PR #82, opened 2025-10-07
Incorporate new Events to the simulator

toslali-ibm:


susiejojo (commented):
This will eventually be updated anyway in #88, hopefully won't cause conflicts.

susiejojo (commented):
experiment_constants.py will be updated anyway in #88, hopefully won't cause conflicts with this.

PR #98, opened 2025-11-07
Throughput calculation correction for memory leak

### LINKED ISSUE #97: Growing memory usage (maybe leak?) with higher betas
Local runs with:

* beta0 = 1e4, beta1 = 1e4, beta2 = 1e4

   - Memory usage: >50%
   - Speed: >20min

* beta0 = 1e-4, beta1 = 1e-4, beta2 = 1e-4

  - Memory usage: 0.7%
  - Speed: 30s

Image

susiejojo:
## Changes Implemented:

* The per-sec binned throughput calculation in `metrics.go` was causing a massive memory leak in the simulation. vLLM itself does not calculate binned throughput, the output throughput (req/s) in vLLM is simply total_requests/total_time. This PR gets rid of the binned throughput calculation and simplifies the throughput calculation similar to vLLM. 
* With very large betas, the KV usage and blocks calculation spills beyond int to int64 (and sometimes even bigger!). This PR typecasts things to higher precision to handle larger betas.

## Issues addressed

Closes #97

### DIFF
--- a/sim/metrics_utils.go
+++ b/sim/metrics_utils.go
@@ -44,38 +44,6 @@ func CalculatePercentile(data []float64, p float64) float64 {
 	}
 }
 
-// Calculate Binned throughput with bins of 1 sec each
-// return mean count of completed requests per second
-func CalculateBinnedThroughput(data []float64) float64 {
-	tempBins := make(map[int]int)
-	maxBinKeyActual := 0
-	totalCompletionCounts := 0
-
-	if len(data) == 0 {
-		return 0
-	}
-
-	for _, val := range data {
-		binKey := int(val)
-		tempBins[binKey]++
-		maxBinKeyActual = max(maxBinKeyActual, binKey)
-	}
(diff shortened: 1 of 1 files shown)


sim/metrics.go
```diff
@@ -83,10 +83,10 @@ func (m *Metrics) Print(horizon int64, totalBlocks int, startTime time.Time) {
 		avgE2E := float64(m.TotalLatency) / float64(m.CompletedRequests)
 		medianE2E := CalculatePercentile(sortedE2Es, 50)
 		p99E2E := CalculatePercentile(sortedE2Es, 99)
-		perSecThroughput := CalculateBinnedThroughput(sortedCompletionTimes)
+		reqThroughput := float64(m.CompletedRequests) / vllmRuntime
```
sriumcp: Is CalculateBinnedThroughput used anywhere? If not, please delete the func.

PR #110, opened 2025-11-21
Enable streaming and TTFT, E2E, ITL metrics

susiejojo:
## Changes Implemented

* Convert all calculations internal in `simulator.go` into ticks (microseconds) to avoid confusion. Note: earlier beta coefficients used to be in seconds/token, now their unit will be ticks/token.
* Streaming calculations:
    * TTFT - `(timestamp of the first output token - arrival timestamp) + alpha2`
    * E2E - `(timestamp of last output token - arrival timestamp) + alpha2 * len(outputTokens)`
    * ITL (calculated on-the-fly during decode) - `(timestamp of current output token + alpha2) - (timestamp of last generated output token + alpha2)`
* Incorporate mean,P90,P95,P99 metrics
* Metrics format matching RH GuideLLM data for easy parsing during optimization

### DIFF
--- a/cmd/coefficients_config.go
+++ b/cmd/coefficients_config.go
@@ -6,8 +6,6 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const TRAINED_COEFFICIENTS_FILEPATH string = "coefficients.yaml"
-
 // Define struct for YAML
 type Config struct {
 	Models  []Model `yaml:"models"`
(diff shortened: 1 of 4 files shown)


sriumcp:
Streaming calculations:

> TTFT - (timestamp of the first output token) + alpha2

Shouldn't this be `(timestamp of the first output token  - arrival timestamp) + alpha2`


`ITL (calculated on-the-fly during decode) = (timestamp of current output token - timestamp of last generated output token)`

Makes sense.

***

I think it would be nice for the sim to always output these numbers. In other words, the sim doesn't need to have a "streaming" vs "non-streaming" distinction at all. The way we use it during training/inference will of course differ for different scenarios.

PR #121, opened 2025-12-16
Expose use-cases of BLIS

### LINKED ISSUE #120: Expose more use cases from BLIS
Support different uses such as 

- `estimate-saturation` where given a workload, BLIS checks a bunch of request arrival rates, and detects saturation using output RPS throughput. It returns the both the max input arrival rate and output RPS throughput such that the system doesn't get saturated.

- `replay request` Identify the maximum sustainable throughput or request rate before saturation
- `vllm config search` Search for an optimal vLLM configuration across key knobs

toslali-ibm:
Resolves #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ BLIS supports several practical and commonly needed inference-planning workflows
 
 
 **2. Saturation analysis (capacity limits)**
-Identify the maximum sustainable throughput or request rate before saturation:
+- Identify the maximum sustainable throughput or request rate before saturation:
 
 ```bash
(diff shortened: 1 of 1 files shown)


susiejojo:
Usecase (saturation detection): https://github.com/inference-sim/inference-sim/issues/120

susiejojo:
Regarding requests replay:

- We should probably expect tokenized requests, since HF tokenizers are not compatible with Golang directly, so we cannot tokenize prompts natively in Go.
- User should also provide a request-rate or use some defaults to replay requests at.

PR #431, opened 2026-02-25
fix(skill): redesign convergence-review loop with two-phase execution

### LINKED ISSUE #430: convergence-review skill: loop breaks after fixes — structural redesign
## The Core Problem

The convergence-review skill describes a mandatory re-run loop (dispatch → tally → fix → re-dispatch), but in practice Claude often skips the re-run after fixing issues. The skill relies entirely on prose instructions to enforce a programmatic loop — and LLMs are unreliable at self-enforcing multi-turn iterative protocols from prose alone.

### 1. No Structural Loop Enforcement

The "algorithm" is written as a pseudocode block inside a markdown document. Once the skill text is loaded into context, the LLM treats it as guidance, not as a control flow it must execute. After dispatching Round 1 agents, tallying findings, and fixing issues, the LLM has completed a satisfying narrative arc (identify problems → fix problems). The instruction to "go back to step 1" competes with the LLM's natural tendency to conclude after resolution.

**Key insight:** The skill is loaded once at invocation. There is no mechanism that re-injects the loop instruction after fixes are applied. By the time fixes are done, the loop instruction is buried far back in the context window under layers of agent outputs, fix diffs, and user interactions.

### 2. Context Window Dilution

A single round of convergence review generates enormous context:
- 5-10 background agent outputs (each potentially 1000+ tokens)
- Reading and tallying each output
- Reporting findings to the user
- Fixing each CRITICAL/IMPORTANT item (reading files, editing files, running tests)

By the time all fixes are applied, the original skill instructions (including the loop mandate) are thousands of tokens back in context. The "re-run is mandatory" instruction has been diluted by pages of operational content.

### 3. The "After fixes" Section Suggests Manual Re-invocation

The skill's "Integration with Other Skills" section says:

```
### After fixes
Re-invoke with the same arguments to start the next round:
/convergence-review pr-code  (Round 2 after fixes)
```

This implies the user must manually re-invoke the skill for Round 2. This contradicts the algorithm section which presents it as an automatic loop. The skill is ambiguous about whether the loop is self-driven or requires re-invocation.

### 4. No Persistent State Between Rounds

The skill says "Track rounds across invocations" but provides no mechanism for this. There is:
- No round counter file written to disk
- No state file tracking which round we're on
- No way to detect "this is a continuation, not a fresh run"

Each invocation starts fresh with no memory of previous rounds unless the conversation context happens to still contain that history.

### 5. Completion Bias After Fixes

After fixing 3 CRITICAL and 5 IMPORTANT issues, the LLM has done substantial work. It has a strong implicit bias toward reporting success ("I've fixed all the issues") rather than acknowledging that fixes must be verified by re-running all reviewers. The fix itself feels like completion.

---

### S7: Structural Skill Redesign — Two-Phase Execution (Primary Fix)

Redesign the skill into two distinct phases. This is the primary enforcement mechanism — it makes the loop structural rather than aspirational.

This is implemented as a single skill with state-file-driven phase detection. The user always runs the same command (`/convergence-review pr-code`). On invocation, the skill checks the state file: no file or `status: "converged"/"stalled"` → Phase A (fresh review); `status: "awaiting-rerun"` → Phase B (fix, then re-enter Phase A). No second skill or mode argument is needed — the state file determines which instruction block applies.

**Phase A: Review** — Dispatch all perspectives, collect results, tally findings. Ends with exactly one of three outcomes:

- **Converged** (`total_critical == 0 AND total_important == 0`): Update state file to `"converged"`, delete it (cleanup), report convergence, proceed to next workflow step.
- **Not converged, rounds remaining** (`round < max_rounds`): Update state file to `"awaiting-rerun"`, emit the status banner, then enter Phase B.
- **Not converged, round limit reached** (`round >= max_rounds`): Update state file to `"stalled"`, emit the status banner with remaining issues, use `AskUserQuestion` to hand control back to the user with options: (a) increase limit and continue, (b) accept remaining issues and proceed, (c) abort.

Default `max_rounds`: 5. The round limit is a safety net (R19 — circuit breaker for unbounded loops), not an expected outcome — if reviews routinely hit it, the fixes are oscillating or introducing new issues, which requires human judgment.

Phase A never applies fixes. Its only job is to produce a verdict.

**Phase B: Fix-and-Rerun** — Only entered when Phase A reports non-convergence with rounds remaining. Applies all CRITICAL and IMPORTANT fixes, then immediately re-enters Phase A (next round). Phase B has no exit path other than invoking Phase A.

**Why this works:** Each phase is short enough that its exit instruction is never far from the current position in context. Phase A can't skip re-dispatch because it never fixes anything. Phase B can't skip re-review because its only exit is invoking Phase A. The state file is the structural enforcement — a concrete artifact on disk that determines behavior, not a prose instruction buried in context.

### S1: Convergence State Directory

Write per-review state files to `.claude/convergence-state/` keyed by gate type and artifact path. This supports concurrent sessions (different terminals reviewing different artifacts) and sequential sessions (finishing one review, starting another) without collision.

**State directory layout:**

```
.claude/convergence-state/
  pr-code--feature-scorer-framework.json
  design--docs-plans-archive-kv-tiered-design.md.json
  pr-plan--docs-plans-pr19-scorer-plan.md.json
```

Filename: `{gate}--{artifact-id-with-slashes-replaced-by-dashes}.json`

**Artifact identity by gate type:**

| Gate type | Artifact ID | Example |
|-----------|------------|---------|
| `pr-code` | Branch name | `feature/scorer-framework` |
| `pr-plan` | Micro-plan file path | `docs/plans/pr19-scorer-plan.md` |
| `macro-plan` | Macro-plan file path | `docs/plans/weighted-scoring-macro-plan.md` |
| `design` | Design doc file path | `docs/plans/archive/kv-tiered-design.md` |
| `h-design` | Hypothesis doc file path | `hypotheses/h-cross-model/HYPOTHESIS.md` |
| `h-code` | Branch name | `hypothesis/h-cross-model` |
| `h-findings` | FINDINGS.md file path | `hypotheses/h-cross-model/FINDINGS.md` |

For diff-based gates (`pr-code`, `h-code`), the branch name is used — the diff content changes between rounds but the branch is stable.

**State file schema:**

```json
{
  "gate": "pr-code",
  "artifact_id": "feature/scorer-framework",
  "round": 2,
  "max_rounds": 5,
  "history": [
    {"round": 1, "critical": 2, "important": 5, "suggestions": 3, "status": "fixed"}
  ],
  "status": "awaiting-rerun",
  "updated_at": "2026-02-25T14:30:00Z"
}
```

### S4: Remove "Re-invoke" Escape Hatch

Remove the "After fixes / Re-invoke" section that implies manual re-invocation. Replace with language stating the loop is automatic.

### S8: Round-Boundary Status Banners

After every Phase A verdict, emit a visible status banner so the user sees the round status and expects the next round.

---

## Dropped Alternatives

| Suggestion | Reason dropped |
|---|---|
| S2: Explicit STOP gate after tally | Subsumed by S7 — Phase A structurally enforces the binary decision |
| S3: Re-inject loop instructions | Subsumed by S7 — each phase is short enough that instructions aren't diluted |
| S5: Self-check before completion | Subsumed by S7 — Phase A can only report convergence if tally is 0/0 |
| S6: AskUserQuestion every round | User friction — slows down every non-converged round unnecessarily |

## Priority

| # | Suggestion | Impact | Effort | Role |
|---|-----------|--------|--------|------|
| S7 | Two-phase skill redesign | High | High | Primary fix — structural loop enforcement |
| S1 | Convergence state directory | High | Low | Persistence across rounds/sessions |
| S4 | Remove "re-invoke" escape hatch | Medium | Trivial | Remove contradictory instruction |
| S8 | Visible round-boundary logging | Low | Trivial | User accountability |

kalantar:
## Summary

Fixes #430

The convergence-review skill relied on prose instructions to enforce a re-run loop after fixes, but LLMs reliably skipped the re-run once fixes felt "complete" (context window dilution, completion bias, ambiguous re-invocation instructions). This PR makes the loop structural rather than aspirational.

- **S7 — Two-phase protocol**: Phase A (review-only, produces one of three outcomes: converged / not-converged / stalled) and Phase B (fix-and-rerun, only exit is re-entering Phase A). Each phase is short enough that its exit instruction is never far from current context position.
- **S1 — Convergence state directory**: `.claude/convergence-state/` with JSON state files keyed by gate type + artifact identity. State-file-driven phase detection on invocation (no file → Phase A; `awaiting-rerun` → Phase B).
- **S4 — Remove re-invoke escape hatch**: The "After fixes / Re-invoke with same arguments" section that implied manual re-invocation is removed. The loop is now automatic and self-driven.
- **S8 — Round-boundary status banners**: Three banner formats (converged, not-converged, stalled) emitted after every Phase A verdict so the user sees round status.
- **Max rounds 10→5**: Updated canonical source (hypothesis.md) and pr-workflow.md. Stall protocol uses AskUserQuestion with three options (increase limit / accept / abort) instead of silently suspending.

## Test plan
- [ ] Invoke `/convergence-review pr-code` on a branch with known issues — verify Phase A dispatches, tallies, and transitions to Phase B
- [ ] After Phase B fixes, verify automatic re-entry into Phase A (Round 2) without manual re-invocation
- [ ] Verify state file created in `.claude/convergence-state/` with correct schema
- [ ] Verify convergence (0 CRITICAL + 0 IMPORTANT) produces converged banner and state file cleanup
- [ ] Verify `*.json` gitignore pattern covers state files
- [ ] Verify no stale "10 rounds" or "re-invoke" references remain

### DIFF
--- a/.claude/skills/convergence-review/SKILL.md
+++ b/.claude/skills/convergence-review/SKILL.md
@@ -73,22 +73,32 @@ For diff-based gates (`pr-code`, `h-code`), the branch name is used — the diff
 
 ### State File Operations
 
-On invocation, determine the artifact ID from the gate type:
-- `pr-code` / `h-code`: Run `git branch --show-current` to get the branch name
-- All other gates: Use the `$1` argument (artifact file path)
+On invocation:
 
-Then derive the state file path:
-```
-.claude/convergence-state/{gate}--{artifact_id with / replaced by -}.json
-```
+1. **Ensure the state directory exists:** `mkdir -p .claude/convergence-state`
 
-Read the state file. If it does not exist or `status` is `"converged"` or `"stalled"`, enter **Phase A**. If `status` is `"awaiting-rerun"`, enter **Phase B**.
+2. **Determine the artifact ID** from the gate type:
+   - `pr-code` / `h-code`: Run `git branch --show-current` to get the branch name
+   - All other gates: Use the `$1` argument (artifact file path)
+
+3. **Derive the state file path:**
+   ```
+   .claude/convergence-state/{gate}--{artifact_id with / replaced by -}.json
+   ```
+
+4. **Read the state file and determine phase:**
+   - File does not exist → **Phase A** (fresh review, round 1)
+   - `status` is `"converged"` or `"stalled"` → **Phase A** (fresh review, round 1)
+   - `status` is `"awaiting-rerun"` **and** `updated_at` is less than 24 hours old → **Phase B** (continue review)
+   - `status` is `"awaiting-rerun"` **but** `updated_at` is more than 24 hours old → **stale state file.** Use AskUserQuestion: *(a) Resume the previous review at round N+1, (b) Start fresh (delete state file and begin round 1).* This prevents a crashed session from silently hijacking the next invocation.
 
 ---
 
 ## Two-Phase Convergence Protocol (non-negotiable)
 
 > **Canonical source:** [docs/process/hypothesis.md — Universal Convergence Protocol](../../../docs/process/hypothesis.md#universal-convergence-protocol). The rules below are a self-contained copy for skill execution; hypothesis.md is authoritative if they diverge.
+>
+> **Intentional skill-level override:** The stall protocol below uses `AskUserQuestion` with three interactive options (increase limit / accept / abort). This differs from hypothesis.md's "suspend the experiment" directive. Rationale: interactive stall handling gives the user control over the outcome rather than silently suspending, which is better UX in a Claude Code session where the user is present. If hypothesis.md is updated to match, this note can be removed.
 
 The convergence loop is **automatic and self-driven**. The user invokes `/convergence-review <gate> [artifact]` once. The skill manages the loop internally via state files. There is no manual re-invocation between rounds.
 
@@ -131,21 +141,38 @@ Phase A dispatches all perspectives, collects results, tallies findings, and pro
 
 ### Phase B: Fix-and-Rerun
 
-Phase B is only entered from Phase A outcome (b). It applies all CRITICAL and IMPORTANT fixes, then immediately re-enters Phase A. **Phase B has no exit path other than invoking Phase A.**
+Phase B is only entered from Phase A outcome (b). It applies all CRITICAL and IMPORTANT fixes, then re-invokes the skill to start Phase A. **Phase B's only exit is a Skill tool re-invocation.**
 
 ```
-1. Fix all CRITICAL items first, then all IMPORTANT items
-2. Run tests/verification as appropriate for the gate type
-3. Update state file: mark current round's history entry status = "fixed"
-4. Re-enter Phase A (next round)
+1. Let M = total number of CRITICAL + IMPORTANT findings.
+
+2. For each finding (CRITICAL items first, then IMPORTANT), in order:
+   a. Fix the issue
+   b. Emit progress: "Fixed N/M — K remaining"
+
+3. After ALL fixes applied:
+   a. Run tests/verification as appropriate for the gate type
+   b. Update state file: mark current round's history entry status = "fixed"
+   c. Emit the re-entry banner:
+
+   ╔═══════════════════════════════════════════════════════════╗
+   ║  ALL FIXES APPLIED — RE-INVOKING PHASE A (Round N+1)     ║
+   ║  /convergence-review <gate> [artifact]                    ║
+   ╚═══════════════════════════════════════════════════════════╝
+
+   d. Invoke the Skill tool with the SAME arguments used to start this review:
+      /convergence-review $0 $1
+      The state file (status: "awaiting-rerun") will route this invocation
+      to Phase A at the correct round. Do NOT proceed to any other step.
+      The ONLY thing you do after fixing is invoke the Skill tool.
 ```
 
-**Why this works:** Each phase is short enough that its exit instruction is never far from the current position in context. Phase A cannot skip re-dispatch because it never fixes anything. Phase B cannot skip re-review because its only exit is invoking Phase A. The state file is the structural enforcement — a concrete artifact on disk that determines behavior, not a prose instruction buried in context.
+**Why this works:** The Skill tool re-invocation reloads the entire skill text fresh, solving context dilution — the Phase A instructions are proximate to the decision point, not buried under pages of fix diffs and test output. The state file routes the fresh invocation to Phase A at the correct round. The countdown progress ("Fixed 3/8, 5 remaining") keeps the LLM in task-oriented mode rather than "wrapping up" mode. This makes skipping the re-review unlikely, though not structurally impossible — see the manual recovery fallback below.
 
 ### Hard Rules
 
 1. **Zero means zero.** One CRITICAL from one reviewer = not converged.
-2. **Re-run is mandatory and automatic.** After fixing issues, Phase B MUST re-enter Phase A. There is no manual re-invocation. There is no "fixes were trivial, skip re-review."
+2. **Re-run is mandatory and automatic.** After fixing issues, Phase B MUST re-invoke the Skill tool to re-enter Phase A. There is no "fixes were trivial, skip re-review." If the automatic re-invocation fails, the user can manually re-invoke (see recovery fallback).
 3. **SUGGESTION items do not block.** Only CRITICAL and IMPORTANT count.
 4. **Independent tallying.** Read each agent's output file. Count findings yourself. Agents have fabricated "0 CRITICAL, 0 IMPORTANT" when actual output contained 3 CRITICAL + 18 IMPORTANT (#390).
 5. **No partial re-runs.** Re-run ALL perspectives, not just the ones that found issues. Fixes can introduce new issues in other perspectives.
@@ -318,4 +345,11 @@ The hypothesis-experiment skill calls this skill at Steps 2, 5, and 8:
 /convergence-review h-design
 /convergence-review h-code hypotheses/h-<name>/
 /convergence-review h-findings hypotheses/h-<name>/FINDINGS.md
+```
+
+### Manual recovery fallback
+If the automatic loop is interrupted (session crash, terminal closed, context limit reached), re-invoke with the same arguments to resume from the current round. The state file tracks progress and will route the invocation to the correct phase automatically:
+```
+/convergence-review pr-code                    (resumes from state file)
+/convergence-review pr-plan docs/plans/...     (resumes from state file)
 ```
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,9 @@ go.work.sum
 __pycache__/*
 */__pycache__/*
 
+# convergence review state files
+.claude/convergence-state/
+
 # data files
 *.json
 !testdata/goldendataset.json


kalantar:
## Enforcement Options: What the "Deeper Fix" Looks Like

Following up on the review's recommendation. Three concrete options, lightest to heaviest — they're complementary.

### Option 1: Self-re-invocation via Skill tool (lightest, strongest single fix)

The key insight: "Re-enter Phase A" is abstract. A Skill tool call is concrete. LLMs are far more reliable at executing a specific tool call than following an abstract instruction to "go back."

Phase B step 4 changes from:

```
4. Re-enter Phase A (next round)
```

to:

```
4. Invoke the Skill tool with the SAME arguments used to start this review:
   /convergence-review <gate> [artifact]
   The state file (status: "awaiting-rerun") will route this invocation
   to Phase A at the correct round. Do NOT proceed to any other step.
   The ONLY thing you do after fixing is invoke the Skill tool.
```

Why this works:
- The state file already handles routing — `awaiting-rerun` → Phase A at round N+1
- A concrete tool call is a single action, not a multi-step re-execution of a section the LLM has to re-read
- The skill text gets **re-loaded on re-invocation**, solving context dilution entirely — fresh instructions, no buried context

Every problem from the issue (context dilution, completion bias, ambiguous re-invocation) is addressed in one change.

### Option 2: Phase B completion banner + countdown (medium, complements Option 1)

Add a progress counter during Phase B and a visible re-entry banner at the end:

```
Phase B step-by-step:

1. For each finding (in order: CRITICAL first, then IMPORTANT):
   a. Fix the issue
   b. Emit progress: "Fixed N/M — K remaining"

2. After ALL fixes:
   a. Run tests/verification
   b. Update state file: mark history entry status = "fixed"
   c. Emit re-entry banner:

   ╔═══════════════════════════════════════════════════════════╗
   ║  ALL FIXES APPLIED — RE-INVOKING PHASE A (Round N)       ║
   ║  /convergence-review <gate> [artifact]                    ║
   ╚═══════════════════════════════════════════════════════════╝

   d. Invoke Skill tool: /convergence-review <gate> [artifact]
      Do NOT proceed to any other step.
```

The countdown ("Fixed 3/8, 5 remaining") keeps the LLM in task-oriented mode rather than "wrapping up" mode. The banner sits immediately before the tool call, so even with long context, the last thing the LLM sees before its next action is the re-invocation instruction.

### Option 3: Claude Code hook as a safety net (heaviest, most robust)

A `PostToolUse` hook that checks for state files after every tool call. When it detects Phase B is complete (history entry marked `"fixed"` but status still `"awaiting-rerun"`), it injects a system-level reminder.

In `.claude/settings.json`:
```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write|Bash",
        "command": ".claude/hooks/convergence-check.sh"
      }
    ]
  }
}
```

```bash
#!/bin/bash
# .claude/hooks/convergence-check.sh
STATE_DIR=".claude/convergence-state"
[ -d "$STATE_DIR" ] || exit 0

for f in "$STATE_DIR"/*.json; do
  [ -f "$f" ] || continue
  status=$(jq -r '.status' "$f" 2>/dev/null)
  [ "$status" = "awaiting-rerun" ] || continue

  last_status=$(jq -r '.history[-1].status' "$f" 2>/dev/null)
  [ "$last_status" = "fixed" ] || continue

  gate=$(jq -r '.gate' "$f" 2>/dev/null)
  round=$(jq -r '.round' "$f" 2>/dev/null)
  echo "CONVERGENCE REVIEW: Round $round fixes complete for gate '$gate'. You MUST now re-invoke: /convergence-review $gate"
done
```

This is the only option that provides enforcement **independent of the skill text**. Even if the LLM ignores Phase B instructions entirely, the hook fires after every Edit/Write/Bash and nags until re-invocation happens. The LLM sees the reminder as a system message it can't dismiss.

Downside: fires on every qualifying tool call during Phase B, which is noisy. Could refine timing to only fire after the state file update marks fixes complete.

---

### Recommendation

**Option 1 is the real fix.** Making the loop a concrete Skill tool call + re-loading the skill fresh solves context dilution, completion bias, and ambiguous re-invocation in one move. Option 2 makes it more reliable. Option 3 is insurance.

Combined with restoring the manual re-invocation fallback (from the review), the recovery story is complete:
- **Happy path:** Phase B ends → Skill re-invocation → fresh Phase A
- **LLM drops the loop:** User sees "After fixes" section → manually re-invokes → state file routes to correct round
- **With Option 3:** Hook nags the LLM before the user even notices

kalantar:
## Critical Review — Does this PR solve #430?

### Summary Assessment

This PR makes **meaningful improvements** to the convergence-review loop — state persistence, phase separation, and status banners are all valuable. However, it does **not fully deliver the "structural" enforcement** it claims. The loop closure (Phase B → Phase A) still depends on LLM compliance with prose instructions, which is the same class of enforcement the issue identifies as the root cause.

---

### What Works Well

1. **State file persistence** is a genuine improvement. Round tracking, history, and phase detection via on-disk state files solve Root Cause #4 (no persistent state between rounds) completely.

2. **Phase separation** makes each phase's job clearer. Phase A's "never fix, only tally" contract and Phase B's "fix, then re-invoke" contract are easier for the LLM to follow than the old monolithic loop.

3. **"Manual recovery fallback" reframing** (replacing the "After fixes / Re-invoke" section) solves Root Cause #3 directly. The old language was confusing about whether the loop was self-driven or manual.

4. **Status banners** improve user visibility into round progress.

5. **Stall protocol with `AskUserQuestion`** is genuinely better UX than silently suspending.

---

### Issues

#### 1. The core enforcement gap is NOT solved (fundamental)

The PR's central mechanism is: Phase B ends by invoking the Skill tool to re-enter Phase A. But the instruction to invoke the Skill tool is loaded once at skill startup, and by the time Phase B completes all fixes (reading files, editing code, running tests for 5-10 findings), that instruction is subject to the **same context window dilution** the issue identifies as Root Cause #2.

The PR itself acknowledges this:

> "This makes skipping the re-review **unlikely, though not structurally impossible**"

This is an honest admission that contradicts the PR's framing as a "structural redesign." The state file tracks state but doesn't *force* the LLM to read it or act on it. The actual loop closure is still behavioral enforcement via prose — the same mechanism class that failed in the original design.

**The improvement is real but probabilistic**: Phase B is shorter than the old monolithic loop (it only fixes, doesn't dispatch/wait/tally), so the re-invocation instruction stays closer to the decision point. For rounds with few findings (1-3), this is likely sufficient. For rounds with many findings (8+), the context dilution is comparable to the old design.

**Calling this "structural" is overstated.** Structural enforcement means the loop cannot be broken by LLM behavior (like a `while` loop in code). This design makes it *less likely* to break, not *impossible* to break. A more honest framing: "makes the loop more reliable through shorter phases and persistent state."

#### 2. `max_rounds` inconsistency between PR description and implementation

The PR description claims:

> **Max rounds 10→5**: Updated canonical source (hypothesis.md) and pr-workflow.md.

But:
- The diff only changes `SKILL.md` and `.gitignore` — **neither `hypothesis.md` nor `pr-workflow.md` are modified**
- The state file schema example uses `"max_rounds": 10`
- The algorithm says `create state file with max_rounds = 10`
- Hard Rule 7 says `Max 10 rounds per gate (default).`
- `hypothesis.md` still says "Max 10 rounds per gate"
- `pr-workflow.md` still says "Max 10 rounds per gate"

The PR description is inaccurate about what was actually changed.

#### 3. 24-hour stale check is underspecified

The state file operations say:

> `status` is `"awaiting-rerun"` **but** `updated_at` is more than 24 hours old → **stale state file.**

This requires datetime arithmetic (parsing `updated_at` ISO timestamp, comparing against current time). The skill doesn't specify how the LLM should perform this comparison. In practice, Claude can approximate this, but edge cases (timezone handling, parsing errors) are unaddressed. A simpler mechanism (e.g., a `stale_after` epoch timestamp computed at write time) would be more robust.

#### 4. `.gitignore` entry is redundant

The existing `.gitignore` already has:
```
# data files
*.json
!testdata/goldendataset.json
```

Since state files are `.json`, they're already excluded. The `.claude/convergence-state/` entry is belt-and-suspenders — not wrong, but worth noting.

---

### The Honest Question

Issue #430 correctly identifies that **LLMs are unreliable at self-enforcing multi-turn iterative protocols from prose alone.** The fix applies... more prose instructions, arranged differently. The state file is a genuine structural addition, but it only solves state *tracking*, not state *enforcement*.

Within the constraints of the Claude Code skill system (skills are loaded as text, no mechanism to force specific tool call sequences), this is probably the best available approach. But the PR should be honest about what it achieves:

- **Solved**: State persistence across rounds (Root Cause #4)
- **Solved**: Ambiguous re-invocation language (Root Cause #3)
- **Improved but not solved**: Context dilution (Root Cause #2) — shortened phases help but don't eliminate
- **Improved but not solved**: Completion bias (Root Cause #5) — progress banners help but don't eliminate
- **Not solved**: No structural loop enforcement (Root Cause #1) — Skill re-invocation is still prose-enforced

If the acceptance bar is "make the loop significantly more reliable," this PR meets it. If the acceptance bar is "make the loop structural rather than aspirational" (the issue's own language), this PR falls short.

kalantar:
❯ what will solve the structural enforcement issue? 

⏺ The fundamental constraint is: skills are text loaded into LLM context, and
  nothing in Claude Code can force the LLM to make a specific tool call. Any
  enforcement within the skill is behavioral, no matter how it's arranged.

  The only way to make it truly structural is to move the loop out of the
  LLM's context and into actual code.

  **The fix: an external orchestrator**

  A bash script (or small Go program) that invokes separate Claude Code
  sessions for each phase:

```
  #!/bin/bash
  gate=$1; artifact=$2; max_rounds=10
  state_file=".claude/convergence-state/${gate}--$(echo $artifact | tr '/'
  '-').json"

  for round in $(seq 1 $max_rounds); do
      # Phase A: fresh Claude session, short context, review only
      claude -p "You are in Phase A of convergence review (round $round).
        Dispatch all perspectives for gate '$gate', tally findings.
        Write results to $state_file. Do NOT fix anything." \
        --allowedTools "Task,Read,Bash,Glob,Grep,Write,Skill"

      status=$(jq -r '.status' "$state_file" 2>/dev/null)
      [ "$status" = "converged" ] && echo "CONVERGED in round $round" && exit
  0

      # Phase B: fresh Claude session, short context, fix only
      claude -p "You are in Phase B of convergence review (round $round).
        Read findings from $state_file. Fix all CRITICAL and IMPORTANT items.
        Run tests. Update $state_file when done." \
        --allowedTools "Read,Edit,Write,Bash,Glob,Grep"
  done


  echo "STALLED after $max_rounds rounds"
```

  Each Claude session handles one short, focused task. The for loop is
  structural — it can't be "convinced" to skip an iteration. The state file is
   the communication channel between sessions (which the PR already designs
  well).

  **What this means for the PR**

  The PR's state file design (S1) and phase separation (S7) are still valuable
   — they're exactly the building blocks the orchestrator needs. The changes
  I'd suggest:

  1. Keep the state file design as-is — it's the inter-session contract
  2. Keep the phase definitions — they define what each session does
  3. Replace the "Skill tool re-invocation" mechanism with documentation that
  the loop is driven by an external script
  4. Ship the orchestrator script as a companion to the skill (e.g.,
  .claude/scripts/convergence-loop.sh)

  The skill itself becomes the phase definition (what Phase A and Phase B do),
   not the loop driver. The script is the loop driver.

  **The tradeoff**

  This requires Claude Code's claude -p (or claude --print) non-interactive
  mode to be reliable enough for multi-agent dispatch. If that's a limitation,
   a middle-ground is a hook-based reminder: a PostToolUse hook on Edit that
  checks the state file and injects "REMINDER: You must re-invoke
  /convergence-review after all fixes" into every turn during Phase B. That's
  not fully structural, but it combats context dilution by re-injecting the
  instruction continuously rather than relying on the one-time skill load.

kalantar:
@claude summarize this pr

sriumcp:
@kalantar Let me absorb this work as part of the larger tracking issue: https://github.com/inference-sim/inference-sim/issues/665

PR #441, opened 2026-02-26
feat(latency): add GPU memory capacity to hardware calibration data

susiejojo:
## Summary

- Add `MemoryGiB float64` field to `HardwareCalib` struct for KV capacity auto-calculation
- Add GPU memory data (80 GiB) for H100 and A100-SXM in `hardware_config.json`
- Add `A100-80` alias entry reconciling `defaults.yaml` naming with `hardware_config.json` (BC-14)
- Update all non-zero test fixtures in `sim/latency/` (6 sites per R4 audit)
- Add `TestGetHWConfig_MemoryGiB_ParsedFromRealConfig` table-driven test (H100, A100-SXM, A100-80)

**Part 1 of 2** for #432 (auto-calculate `total_kv_blocks` in roofline mode).

**Design doc:** `docs/plans/2026-02-25-kv-capacity-auto-calculation-design.md`
**Macro plan:** `docs/plans/2026-02-25-kv-capacity-auto-calculation-macro-plan.md`

## Key decisions

- `MemoryGiB` is plain `float64` (not `*float64`) — zero is never a valid GPU memory value (R9)
- `ValidateRooflineConfig()` is NOT modified — `MemoryGiB` validation belongs in the KV capacity function (PR 2, Decision 4)
- A100-80 alias is a full entry copy (not a programmatic alias) for simplicity
- Zero-value `HardwareCalib{}` construction sites (~40+ test files) need no changes — zero `MemoryGiB` means no auto-calculation

## Test plan

- [x] `TestGetHWConfig_MemoryGiB_ParsedFromRealConfig` verifies all 3 GPU entries parse correctly
- [x] All existing tests pass unchanged (`go test ./...`)
- [x] R4 construction site audit: 6 non-zero sites updated, ~40+ zero-value sites verified safe

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #474, opened 2026-03-02
feat: auto-calculate KV cache blocks from model/hardware in roofline mode

### LINKED ISSUE #432: Auto-calculate total_kv_blocks from model and hardware config in roofline mode
## Problem

When running BLIS in roofline mode, `total_kv_blocks` must either be:

1. **Supplied explicitly** by the user via `--total-kv-blocks`, which requires domain knowledge (model architecture, GPU memory, dtype, TP degree) to set correctly.
2. **Fallen back to the hardcoded CLI default of 1,000,000 blocks**, which is unrealistically large for most configurations and can silently produce misleading simulation results (e.g., no KV evictions ever fire, no memory pressure effects observed).

This is particularly problematic in roofline mode because the user has *already* provided all the information needed to derive the correct value — a HuggingFace `config.json` (model architecture) and `hardware_config.json` (GPU specs) — yet BLIS still requires a separate manual input for KV capacity.

## Desired Behavior

When running in roofline mode, BLIS should **automatically calculate `total_kv_blocks`** from the model and hardware configuration that is already provided, unless the user explicitly overrides it via `--total-kv-blocks`.

**The implementation MUST replicate the calculation from [`llm-d-benchmark capacity_planner.py:total_kv_cache_blocks()`](https://github.com/llm-d/llm-d-benchmark/blob/e724935c533d468b7ab02c542ed2047790f91d6b/config_explorer/src/config_explorer/capacity_planner.py#L596).** This is the canonical reference — BLIS must produce the same block counts for the same inputs to stay consistent with the llm-d ecosystem.

### Reference calculation (from capacity_planner.py)

The reference function `total_kv_cache_blocks()` follows these steps:

**Step 1 — Per-token KV memory:**
For standard attention (non-MLA): `num_layers × 2 × head_dim × num_kv_heads × precision_bytes`
(The factor of 2 accounts for separate K and V caches.)

**Step 2 — Account for tensor parallelism:**
`per_token_kv_bytes = per_token_memory / (tp × pp)`
(BLIS does not model pipeline parallelism, so `pp = 1`.)

**Step 3 — Per-block memory:**
`per_block_bytes = per_token_kv_bytes × block_size`
(vLLM default block size: 16 tokens.)

**Step 4 — Allocatable KV cache memory:**
The reference calls `allocatable_kv_cache_memory()` which subtracts three components from the total GPU memory budget:
- **Model weights** — total parameter bytes, scaled by data parallelism
- **Activation memory** — **5.5 GiB for dense models, 8.0 GiB for MoE models** (fixed constants in the reference)
- **Non-torch system overhead** — **0.15 GiB for TP=1, 0.6 GiB for TP≥2** (NCCL buffers, CUDA context)

The total GPU budget uses `gpu_memory × gpu_mem_util` where `gpu_mem_util = 0.9`.

**Step 5 — Total blocks:**
`total_blocks = floor(allocatable_kv_memory / per_block_bytes)`

### Inputs that need to be added

- `hardware_config.json` needs a **GPU memory field** (e.g., `"MemoryGiB": 80` for H100 SXM).
- The **activation memory** and **non-torch overhead** constants from the reference must be replicated. These could be hardcoded constants (matching the reference) or optionally configurable.

### Override semantics

- If the user passes `--total-kv-blocks` explicitly, that value takes precedence (consistent with R18 — CLI flag precedence).
- If running in blackbox mode (no `--model-config-folder`), behavior is unchanged — lookup from `defaults.yaml` or CLI default.
- In roofline mode, if auto-calculation is possible, it replaces the 1,000,000 fallback. A log message should report the derived value.

### Validation

The auto-calculated values should be compared against the empirical `total_kv_blocks` values in `defaults.yaml` (which come from real vLLM deployments). For example, Llama-3.1-8B on H100 with TP=2 should produce a value close to 132,139. Significant deviations indicate a formula bug.

## Why This Matters

- **Accuracy**: The 1M-block default means simulations in roofline mode never experience memory pressure, producing optimistic and unrealistic results.
- **Usability**: Users adopting roofline mode shouldn't need to manually compute KV block counts — the information is already available.
- **Consistency with llm-d ecosystem**: The [llm-d-benchmark capacity planner](https://github.com/llm-d/llm-d-benchmark/blob/e724935c533d468b7ab02c542ed2047790f91d6b/config_explorer/src/config_explorer/capacity_planner.py#L596) already does this calculation; BLIS must produce matching results.

susiejojo:
## Summary

Closes #432 (PR2 of 2).

- **Core function**: `CalculateKVBlocks()` in `sim/latency/kv_capacity.go` derives the maximum KV cache block count from model architecture (HuggingFace config.json) and GPU memory capacity, matching the llm-d-benchmark `capacity_planner.py` reference formula. Supports dense and MoE models, tied embeddings, SwiGLU-family activations.
- **CLI integration**: When `--roofline` is active and `--total-kv-blocks` is not explicitly set, the block count is auto-calculated. Graceful fallback with warnings on any error (R18 precedence, R1 no silent data loss).
- **Empirical fidelity**: Llama-3.1-8B/H100/TP=2 within 5.2% of empirical baseline, TP=4 within 2.9%, Mixtral-8x7B/TP=2 within 16.1% (all within 10%/20% tolerance bands).

## Changes

| File | Lines | Description |
|------|-------|-------------|
| `sim/latency/kv_capacity.go` | +278 | Core `CalculateKVBlocks`, `computeModelWeightBytes`, `ExtractKVCapacityParams` |
| `sim/latency/kv_capacity_test.go` | +646 | 21 tests: validation, fidelity, monotonicity, purity, MoE, tied embeddings, extraction |
| `cmd/root.go` | +25 | CLI integration after roofline config loading |
| `CLAUDE.md` | +4/-1 | Document new file and auto-calculation feature |

## Design

Follows the [design doc](docs/plans/2026-02-25-kv-capacity-auto-calculation-design.md) and [macro plan](docs/plans/2026-02-25-kv-capacity-auto-calculation-macro-plan.md). PR1 (#441, merged) added `MemoryGiB` to `HardwareCalib`. This PR2 implements the calculation and CLI wiring.

**Memory budget formula** (total across all TP GPUs):
```
available = gpu_mem × 0.9 × tp
overhead  = model_weights + activation (5.5 dense / 8.0 MoE) + non_torch × tp
blocks    = floor((available - overhead) × GiB / per_block_bytes)
```

## Test plan

- [x] `go test ./sim/latency/... -run TestCalculateKVBlocks` — 21 tests pass
- [x] `go test ./...` — full suite passes, no regressions
- [x] `go vet ./...` — clean
- [x] Spec compliance review — all BCs satisfied
- [x] Code quality review — approved

### DIFF
--- a/sim/latency/config.go
+++ b/sim/latency/config.go
@@ -17,6 +17,52 @@ type HFConfig struct {
 	Raw map[string]any
 }
 
+// GetString returns a string value for a key if present and of the right type.
+func (c *HFConfig) GetString(key string) (string, bool) {
+	if v, ok := c.Raw[key]; ok {
+		if s, ok := v.(string); ok {
+			return s, true
+		}
+	}
+	return "", false
+}
+
+// GetInt tries to coerce a JSON number to int.
+func (c *HFConfig) GetInt(key string) (int, bool) {
+	if v, ok := c.Raw[key]; ok {
+		if f, ok := v.(float64); ok {
+			return int(f), true
+		}
+	}
+	return 0, false
+}
+
+// GetBool returns a bool for a key.
+func (c *HFConfig) GetBool(key string) (bool, bool) {
+	if v, ok := c.Raw[key]; ok {
+		if b, ok := v.(bool); ok {
+			return b, true
+		}
+	}
+	return false, false
+}
+
+// MustGetString returns the string or a default.
+func (c *HFConfig) MustGetString(key, def string) string {
+	if s, ok := c.GetString(key); ok {
+		return s
+	}
+	return def
+}
+
+// MustGetInt returns the int or a default.
+func (c *HFConfig) MustGetInt(key string, def int) int {
+	if i, ok := c.GetInt(key); ok {
+		return i
+	}
+	return def
+}
+
 func parseHWConfig(HWConfigFilePath string) (map[string]sim.HardwareCalib, error) {
 	data, err := os.ReadFile(HWConfigFilePath)
 	if err != nil {


sriumcp (commented):
_Test review — please ignore._

sriumcp (commented):
_Test review — please ignore._

PR #623, opened 2026-03-11
feat(latency): switch default latency model from blackbox to roofline

### LINKED ISSUE #622: Switch default latency model from blackbox to roofline
## Summary

The current default latency model backend is `blackbox` (alpha/beta regression from trained coefficients). Now that the roofline, crossmodel, and trained-roofline backends are mature and well-tested, the default should be switched to `roofline` for better out-of-the-box accuracy without requiring pre-trained coefficients.

## Motivation

- **Blackbox requires pre-trained coefficients**: New models need alpha/beta values in `defaults.yaml` before they work, creating a chicken-and-egg problem for users trying new models.
- **Roofline works from first principles**: Only needs `config.json` (auto-fetched from HuggingFace) and `hardware_config.json` (bundled) — no training data required.
- **Better default experience**: Users running `./blis run --model <name>` should get reasonable results immediately, which roofline provides via analytical FLOPs/bandwidth estimation.

### Code changes
- [ ] Update `--latency-model` flag default from `""` (blackbox) to `"roofline"` in `cmd/root.go`
- [ ] Update any fallback/guard logic that uses `backend == ""` to mean blackbox
- [ ] Ensure roofline auto-fetch (`--latency-model roofline` behavior) works when it becomes the default path

### Documentation updates
All docs that reference "blackbox" as the default need updating:

- [ ] `CLAUDE.md` — Latency Estimation section ("Blackbox mode (default)")
- [ ] `docs/guide/latency-models.md` — primary latency model guide
- [ ] `docs/getting-started/quickstart.md` — first-run examples
- [ ] `docs/getting-started/tutorial.md` — capacity planning walkthrough
- [ ] `docs/reference/configuration.md` — configuration reference
- [ ] `docs/reference/models.md` — supported models catalog
- [ ] `docs/concepts/core-engine.md` — core engine description
- [ ] `docs/concepts/roofline.md` — roofline concept page
- [ ] `docs/concepts/glossary.md` — glossary entries
- [ ] `docs/contributing/extension-recipes.md` — extension guides
- [ ] `docs/contributing/templates/design-guidelines.md` — design guidelines
- [ ] `cmd/root.go` flag help string — currently says "blackbox (default)"

### HF_TOKEN documentation for gated models

Since roofline auto-fetches `config.json` from HuggingFace, gated models (e.g., LLaMA, Granite) will fail with HTTP 401 unless `HF_TOKEN` is set. With roofline becoming the default, users will hit this on their very first run if they pick a gated model. Currently `HF_TOKEN` is only mentioned in passing in `docs/guide/latency-models.md` and `docs/reference/configuration.md` — it needs to be much more prominent:

- [ ] **`docs/getting-started/installation.md`** — Add a "Prerequisites" or "Environment Setup" section explaining `HF_TOKEN`: what it is, how to get one from [huggingface.co/settings/tokens](https://huggingface.co/settings/tokens), and how to set it (`export HF_TOKEN=hf_...`). This is the natural place since users read installation before quickstart.
- [ ] **`docs/getting-started/quickstart.md`** — Add a callout/admonition near the first `blis run` example noting that gated models require `HF_TOKEN`, linking back to the installation page for details.
- [ ] **`docs/guide/latency-models.md`** — Expand the existing one-liner into a proper subsection with troubleshooting (what the 401 error looks like, how to fix it).
- [ ] **`cmd/root.go`** — Improve the error message when HF fetch fails with 401 to suggest `export HF_TOKEN=...` and point to the docs.
- [ ] **`README.md` or project root** — Consider a brief mention in any "Getting Started" section if one exists.

### Test updates
- [ ] Update any tests that assert or depend on the default being blackbox
- [ ] Verify golden dataset compatibility or regenerate if needed

## Notes

- The `defaults.yaml` alpha/beta coefficients should remain available for users who explicitly choose `--latency-model blackbox`
- Consider whether `trained-roofline` might be an even better default (7% MAPE), but roofline is simpler and has no learned coefficients dependency
- The HF_TOKEN error path already exists in `cmd/hfconfig.go:178` — it returns a clear message, but we should also guide users in docs before they hit it

susiejojo:
## Summary

- Switch the CLI default `--latency-model` from `""` (blackbox) to `"roofline"`, giving users a working simulation out-of-the-box for any HuggingFace model without needing pre-trained coefficients
- Update factory dispatch so `""` (unspecified backend) maps to roofline instead of blackbox
- Extract hardware/TP/vllm defaults loading into an early block in `cmd/root.go` so roofline can auto-resolve `--hardware` and `--tp` for models in `defaults.yaml`
- Update all 72 test `SimConfig` constructions across 16 test files to explicitly pass `Backend: "blackbox"` (no golden dataset regeneration needed)
- Update 14 documentation files for consistency (README, quickstart, tutorial, latency-models guide, configuration reference, models reference, concepts, glossary, CLAUDE.md)

Fixes #622

### Code (3 files)
- **`cmd/root.go`**: Flag default `""` → `"roofline"`, extracted early defaults resolution block, updated error messages for analytical backends
- **`sim/latency/latency.go`**: Factory dispatch `case "", "roofline":` (was `case "roofline":`), `case "blackbox":` (was `case "", "blackbox":`)
- **`sim/config.go`**: Backend field comment updated

### Tests (16 files)
- Added explicit `Backend: "blackbox"` to all test `SimConfig` constructions that rely on blackbox coefficients — ensures tests continue using blackbox mode now that `""` maps to roofline

### Documentation (14 files)
- Reordered latency model listings (roofline first as default)
- Simplified README/quickstart output tables for first-time users
- Verified metrics definitions against `sim/metrics.go` code
- Added HF_TOKEN environment setup section to installation guide
- Added DP/EP/quantization limitation notes to advanced docs

## Test plan

- [x] `go build -o blis main.go` — builds successfully
- [x] `go test ./... -count=1` — all packages pass (0 failures)
- [x] `go vet ./...` — no issues
- [x] Verify `./blis run --model qwen/qwen3-14b` uses roofline mode by default
- [x] Verify `./blis run --model qwen/qwen3-14b --latency-model blackbox` still works
- [x] CI: `golangci-lint run ./...`

### DIFF
--- a/sim/simulator_test.go
+++ b/sim/simulator_test.go
@@ -2039,7 +2039,7 @@ func TestSimulator_ProactiveCap_EliminatesOvershoot(t *testing.T) {
 		KVCacheConfig:       NewKVCacheConfig(1000, 16, 0, 0, 0, 0),
 		BatchConfig:         NewBatchConfig(256, 2048, 0),
 		LatencyCoeffs:       NewLatencyCoeffs([]float64{6910, 17.67, 2.84}, []float64{0, 0, 0}),
-		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "test", "H100", 1, "", 100),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "test", "H100", 1, "blackbox", 100),
 	}
 	sim := mustNewSimulator(t, cfg)
 
@@ -2081,7 +2081,7 @@ func TestSimulator_ProactiveCap_MaxModelLen2_ZeroOutput(t *testing.T) {
 		KVCacheConfig:       NewKVCacheConfig(100, 16, 0, 0, 0, 0),
 		BatchConfig:         NewBatchConfig(256, 2048, 0),
 		LatencyCoeffs:       NewLatencyCoeffs([]float64{1000, 1, 1}, []float64{0, 0, 0}),
-		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "test", "H100", 1, "", 2),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "test", "H100", 1, "blackbox", 2),
 	}
 	sim := mustNewSimulator(t, cfg)
 
@@ -2118,7 +2118,7 @@ func TestProcessCompletions_LengthCapped_MetricsRefreshed(t *testing.T) {
 		KVCacheConfig:       NewKVCacheConfig(1000, 16, 0, 0, 0, 0),
 		BatchConfig:         NewBatchConfig(256, 2048, 0),
 		LatencyCoeffs:       NewLatencyCoeffs([]float64{1000, 1, 1}, []float64{0, 0, 0}),
-		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "", 100),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "blackbox", 100),
 	}
 	sim := mustNewSimulator(t, cfg)
 
@@ -2154,7 +2154,7 @@ func TestRecordRequestCompletion_LengthCapped_ITL(t *testing.T) {
 		KVCacheConfig:       NewKVCacheConfig(1000, 16, 0, 0, 0, 0),
 		BatchConfig:         NewBatchConfig(256, 2048, 0),
 		LatencyCoeffs:       NewLatencyCoeffs([]float64{1000, 1, 1}, []float64{0, 0, 0}),
-		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "", 100),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "blackbox", 100),
 	}
 	sim := mustNewSimulator(t, cfg)
 
@@ -2189,7 +2189,7 @@ func TestRecordRequestCompletion_NormalRequest_ITL(t *testing.T) {
 		KVCacheConfig:       NewKVCacheConfig(1000, 16, 0, 0, 0, 0),
 		BatchConfig:         NewBatchConfig(256, 2048, 0),
 		LatencyCoeffs:       NewLatencyCoeffs([]float64{1000, 1, 1}, []float64{0, 0, 0}),
-		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "", 0),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "", "", 0, "blackbox", 0),
 	}
 	sim := mustNewSimulator(t, cfg)
 


sriumcp (commented):
## Convergence Review — Round 1 (Gate: pr-code, 10 perspectives)

Dispatched 10 parallel review perspectives (Substance & Design, Code Quality, Test Quality, Getting-Started, Automated Reviewer, DES Expert, vLLM/SGLang Expert, Distributed Inference, Performance, Security). Independently tallied and deduplicated findings below.

---

### CRITICAL (1)

**F1. `case "", "roofline"` in the factory breaks the library API contract (`sim/latency/latency.go:128`)**

Reported by 7 of 10 perspectives. The change from `case "", "blackbox"` to `case "", "roofline"` means any library caller, YAML policy bundle, or test that constructs `ModelHardwareConfig{Backend: ""}` now attempts roofline instantiation — which requires non-zero `ModelConfig`, `HWConfig`, and `TP > 0`. The `validLatencyBackends` map in `sim/bundle.go:65` still accepts `""`, so YAML bundles pass validation but fail at factory time.

**Recommended fix (endorsed by 7/10 perspectives):** Keep `case "", "blackbox"` in the factory. Only change the CLI flag default to `"roofline"`. The CLI already passes the explicit backend string to the factory, so this achieves the UX goal without breaking library callers.

---

### IMPORTANT (7)

**F2. Dead code: implicit roofline detection block (`cmd/root.go:488-501`)**
With flag default `"roofline"`, `backend` is never `""` in the CLI path. This block — including its `logrus.Fatalf` calls — is unreachable. Remove it.

**F3. No test coverage for `Backend=""` → roofline factory path**
All 66 existing `""` test sites changed to `"blackbox"`, but no new test verifies the new `""` dispatch behavior. If F1 is fixed (keep `""` → blackbox), this is moot. Otherwise, add a test.

**F4. `validLatencyBackends` map accepts `""` but semantics changed silently**
YAML bundles omitting `backend` now silently get roofline instead of blackbox — no validation error, just different behavior. If F1 is fixed, no change needed.

**F8. Error messages don't explain roofline is the new default**
A user running `./blis run --model unknown-model` gets `"--latency-model roofline requires --hardware"` with no context that roofline is the default or that `--latency-model blackbox` is an offline alternative. Suggest:
```
Default latency model is roofline, which requires --hardware and --tp.
For offline coefficient-based estimation, use --latency-model blackbox.
```

**F9. Air-gapped/offline environments break with new default**
Default mode now requires HuggingFace network access for `config.json` auto-fetch. README and quickstart should document `--latency-model blackbox` as the offline alternative.

**F14. KV capacity auto-calculation silently differs from defaults.yaml**
When roofline auto-calculates KV blocks via `CalculateKVBlocks`, the result may differ from the trained `total_kv_blocks` in `defaults.yaml`. This silently changes batch formation, preemption thresholds, and routing scorer behavior for existing users. Log the difference when the auto-calculated value overrides a defaults.yaml value.

**F16. Verify all golden dataset test paths are among the 66 changed sites**
The 3 golden paths (`TestSimulator_GoldenDataset`, `TestClusterSimulator_SingleInstance_GoldenEquivalence`, `TestInstanceSimulator_GoldenDataset_Equivalence`) must pass explicit `"blackbox"`. If any passes `Backend: ""`, it will fail roofline validation with zero-valued `ModelConfig`.

---

### SUGGESTION (8, non-blocking)

- **F5.** Redundant `GetDefaultSpecs` calls (2-3 per invocation) — init-only, minor.
- **F6.** Blackbox guard drops `modelConfigFolder`/`hwConfigPath` checks — document the semantic change.
- **F7.** Early defaults uses value checks not `Flags().Changed()` — pre-existing pattern, not a regression.
- **F10.** Roofline overestimates MoE step time ~4x (pre-existing, documented in H-Cross-Model) — note in release notes that MoE users should prefer `--latency-model crossmodel` or `trained-roofline`.
- **F11.** Roofline `StepTime` allocates 4B+4 heap objects per call vs 0 for blackbox (pre-existing) — consider map→struct refactor as follow-up.
- **F12.** Stale docstrings — appears addressed in PR diff.
- **F13.** Documentation pages listing "Blackbox (default)" — appears addressed but verify completeness.
- **F15.** Zero-alpha with roofline only warns, doesn't error — reasonable for default mode.

---

### Convergence Status

| Severity | Count |
|----------|-------|
| CRITICAL | 1 |
| IMPORTANT | 7 |
| SUGGESTION | 8 |

**NOT CONVERGED.** 1 CRITICAL + 7 IMPORTANT remaining.

The core recommendation from 7 of 10 perspectives: **keep `""` → blackbox in the factory, only change the CLI flag default.** This resolves F1, F3, F4, and F16 simultaneously. The remaining items (F2, F8, F9, F14) are straightforward fixes.

PR #834, opened 2026-03-25
fix(observe): calibrate prefix token ratio for BPE tokenizer fidelity

### LINKED ISSUE #832: Observe prompt fidelity: ~2x token inflation from word-based prompt generation
**Describe the bug**

`blis observe` generates synthetic prompts by repeating words (e.g., `"hello hello hello ..."`) at a ratio of ~1 word per intended input token. However, the server's tokenizer typically counts significantly more tokens than intended due to:

1. **Chat template overhead**: system/user/assistant role markers, BOS/EOS tokens, and special formatting tokens added by the chat completions API
2. **Tokenizer divergence**: the 1-word ≈ 1-token heuristic is an approximation that varies by model and tokenizer

The result is that `server_input_tokens` (ground truth from the server's `usage.prompt_tokens`) is often ~1.5–2x larger than `input_tokens` (the workload spec target). This means:

- `blis run` simulates latency for X input tokens
- `blis observe` sends a prompt that the server processes as ~2X input tokens
- `blis calibrate` compares these mismatched workloads

**CLI command to reproduce**

```bash
# Generate a workload with 512 input tokens and observe
./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
  --api-format chat --rate 5 --num-requests 50 --prompt-tokens 512 \
  --trace-header trace.yaml --trace-data trace.csv

# Inspect trace — server_input_tokens will be ~800-1000 for input_tokens=512
```

**Expected behavior**

`server_input_tokens` should be close to `input_tokens` for calibration to produce meaningful sim-vs-real comparisons. The observe pipeline should either:
- Generate prompts that tokenize to the target count, OR
- Adjust calibration to account for the known inflation

**Actual behavior**

For chat completions format, `server_input_tokens` is typically 1.5–2x `input_tokens`. The `calibrate` command's token mismatch counter flags this, but latency comparisons proceed using the mismatched token counts.

Confirmed in `baseline-trace_data.csv`: rows with `input_tokens=512` show `server_input_tokens` in the 800-1000 range.

- [ ] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [ ] Work-conserving
- [x] None / unknown

No numbered invariant is violated, but **R23 (parallel code path parity)** applies: `blis run` and `blis observe` should produce comparable token-level workloads for the same spec, but the prompt generation layer creates a systematic gap.

**Impact**

- **Calibration accuracy**: Sim predicts latency for X tokens; observed latency is for ~2X tokens. TTFT and E2E comparisons are biased.
- **Capacity planning**: Users trusting calibration reports will under-estimate real serving costs because the server processes more tokens than the sim assumes.
- **Replay fidelity**: `blis replay` uses `input_tokens` from the trace (the spec value), not `server_input_tokens`, so replayed latency predictions don't match what the server actually experienced.

**Environment**

- OS: macOS 25.3.0 (darwin)
- Go version: go1.24.x darwin/arm64
- Commit hash: 91aa8ad

**Discussion: tokenizer integration tradeoffs**

The root fix is to generate prompts that tokenize to the target count. Several approaches with different tradeoff profiles:

| Approach | Accuracy | Complexity | Dependencies | Latency |
|----------|----------|------------|--------------|---------|
| **A. Offline calibration factor** — measure word-to-token ratio once per model/format, apply as a divisor to word count | ~±10% | Low | None | None |
| **B. Tokenizer library** — integrate a Go tokenizer (e.g., tiktoken port) to count tokens before sending | Exact | Medium | New dependency | Startup cost |
| **C. Server-side tokenize endpoint** — call `/tokenize` or equivalent to calibrate prompt length | Exact | Medium | Server must support endpoint | Extra HTTP call |
| **D. Iterative binary search** — send a test prompt, read `prompt_tokens`, adjust, repeat | Exact | High | None | Multiple round trips |
| **E. Use `server_input_tokens` in calibrate** — instead of fixing observe, have calibrate compare sim predictions at `server_input_tokens` level | Compensates downstream | Low | None | None |

**Considerations:**
- **Approach A** is simplest but model-dependent. Different tokenizers (BPE, SentencePiece, Unigram) produce different ratios for the same text. Chat template overhead varies by model family.
- **Approach B** gives exact control but adds a non-trivial dependency. Go tokenizer libraries exist but may not support all models. BLIS is currently dependency-light by design.
- **Approach C** is accurate but assumes server endpoint availability. vLLM exposes `/tokenize`, but not all backends do.
- **Approach D** is accurate but slow — impractical for high-rate workloads.
- **Approach E** is a pragmatic short-term fix: adjust calibration to compare apples-to-apples by using the actual token count the server processed. This doesn't fix the workload mismatch itself but makes calibration reports accurate. Could be combined with A or B as a longer-term fix.

A reasonable path: **E (short-term)** to make calibration accurate now, then **A or B (longer-term)** to reduce the prompt inflation itself. The chat template overhead (~50-100 tokens) is likely a constant additive factor per request, so even approach A could be quite effective if the base word-to-token ratio is calibrated per model.

**Additional context**

Discovered during investigation of issue #828. The `server_input_tokens` field was designed to capture this ground truth (see `requestToPending` comment: "Actual token count varies by tokenizer; ServerInputTokens provides ground truth"), but no downstream consumer currently uses it to adjust comparisons.

workload file and trace.csv

[workload_glia_prefix_heavy.yaml](https://github.com/user-attachments/files/26254863/workload_glia_prefix_heavy.yaml)
[baseline-trace_data.csv](https://github.com/user-attachments/files/26254864/baseline-trace_data.csv)

toslali-ibm:
## Summary

- Adds a startup calibration request that measures the server's tokens-per-word ratio by sending 100 vocabulary words and reading back `prompt_tokens`
- Scales prefix word counts in `buildPrefixStrings` by `round(N / tokensPerWord)` so the server tokenizes them to approximately the target token count
- Falls back to 1:1 ratio (current behavior) if calibration fails or ratio is outside [1.0, 3.0]

Fixes #832

## Test plan

- [x] `TestCalibratePrefixTokenRatio_ReturnsRatio` — verifies 167/100 = 1.67 ratio from mock server
- [x] `TestCalibratePrefixTokenRatio_FallbackOnError` — server 500 → ratio 1.0
- [x] `TestCalibratePrefixTokenRatio_FallbackOnOutOfBounds` — pathological ratio → 1.0
- [x] `TestBuildPrefixStrings_ScalesWordCount` — ratio 1.0 → 1000 words; ratio 1.67 → 599 words; prefixLengths stores target tokens in both cases
- [x] Existing `TestBuildPrefixStrings_DeterministicAndDistinct` updated (pass `1.0`) — no behavioral change
- [x] Existing `TestRequestToPending_PrependsPrefixString` updated (pass `1.0`) — no behavioral change
- [x] Full `go test ./cmd/... -count=1` passes

### DIFF
--- a/docs/guide/observe-replay-calibrate.md
+++ b/docs/guide/observe-replay-calibrate.md
@@ -136,6 +136,14 @@ Used when `--rate` mode is active (ignored when `--workload-spec` is provided):
 !!! info "Prefix sharing"
     When the workload spec defines prefix groups, observe builds deterministic prefix strings from a fixed vocabulary, seeded by the RNG seed and group name. This activates the server's prefix cache for realistic KV cache hit rates.
 
+    Before dispatching requests, observe sends a single calibration request to measure the server's tokens-per-word ratio (typically 1.5–1.7 for BPE tokenizers). Prefix word counts are then scaled so the server tokenizes them to approximately the target `prefix_length` in the spec — matching what `blis run` simulates. The calibration result is logged at startup:
+
+    ```
+    INFO Prefix token calibration: 100 words → 167 server tokens (1.670 tokens/word)
+    ```
+
+    If calibration fails (server unreachable, timeout, or abnormal ratio), observe falls back to 1:1 word-to-token mapping with a warning.
+
 !!! info "Session support"
     If the workload spec contains session clients, observe runs in closed-loop mode: each completed request may trigger follow-up requests from the session manager, interleaved with pre-generated arrivals by arrival time.
 


toslali-ibm:
@claude /pr-review-toolkit:review-pr Look at the issue this closes and ensure this PR correctly
  handles the intent of that issue. Check thoroughly to avoid correctness bugs or performance critical  
  aspects, and most importantly, adherence to @docs/contributing/standards. Is this PR ready to merge?

sriumcp:
@claude /pr-review-toolkit:review-pr also check for documentation correctness and completeness.

PR #856, opened 2026-03-27
feat(workload): add normalized exponential arrival sampler (#845)

### LINKED ISSUE #845: feat(workload): add normalized exponential arrival sampler for inference-perf fidelity
## Problem Statement

BLIS cannot replicate inference-perf's "constant" load generation behavior, causing systematic bias in training data. Ground-truth experiments collected with kubernetes-sigs/inference-perf use **normalized exponential inter-arrival times** that span the exact configured duration, while BLIS uses **truncated Poisson arrivals** that end earlier than the target duration.

### Real Example
From `training/trainval_data/20260217-155451-llama-2-7b-tp1-codegen/profile.yaml`:

```json
{
  "load": {
    "stages": [
      {"duration": 600, "rate": 5},
      {"duration": 600, "rate": 10}
    ],
    "type": "constant"
  }
}
```

**Expected behavior:**
- Stage 1: 3000 requests (5 × 600) spanning 0-600s
- Stage 2: 6000 requests (10 × 600) spanning 600-1200s
- **Total: 9000 requests spanning exactly 1200.0 seconds**

### How inference-perf Implements This

From [`inference_perf/loadgen/load_timer.py`](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/loadgen/load_timer.py#L48-L61):

```python
class ConstantLoadTimer(LoadTimer):
    def start_timer(self, initial: Optional[float] = None) -> Generator[float, None, None]:
        num_requests = int(self._rate * self._duration)
        
        # Generate random intervals using EXPONENTIAL distribution
        intervals = self._rand.exponential(1 / self._rate, num_requests)
        
        # Normalize intervals to sum EXACTLY to duration
        total_interval_time = np.sum(intervals)
        scale_factor = self._duration / total_interval_time
        normalized_intervals = intervals * scale_factor
        
        # Yield request times
        next_time = time.monotonic() if initial is None else initial
        for interval in normalized_intervals:
            next_time += interval
            yield next_time
```

**Key properties:**
1. Pre-generates **all** intervals before dispatch
2. Uses exponential distribution (same as Poisson) 
3. **Normalizes** intervals to sum exactly to duration
4. Result: N requests with exponential spacing spanning exactly T seconds
5. Coefficient of variation: **CV < 1.0** (constrained variance)

### Code Flow

1. **`training/run_blis_and_compute_loss.py`** computes expected count:
   ```python
   total_requests = sum(round(s["rate"] * s["duration"]) for s in stages_config)
   spec = {
       "num_requests": total_requests,  # 9000
       "inference_perf": {...}
   }
   ```

2. **`sim/workload/inference_perf.go:124,168`** hardcodes Poisson:
   ```go
   clients = append(clients, ClientSpec{
       Arrival: ArrivalSpec{Process: "poisson"},  // ← Hardcoded
       // ...
   })
   ```

3. **`sim/workload/generator.go:247-256`** generates incrementally:
   ```go
   for currentTime < horizon {
       iat := arrivalSampler.SampleIAT(rng)  // Exponential, unconstrained
       currentTime += iat
       if currentTime >= horizon {
           break  // ← Stops at horizon boundary
       }
       // Generate request...
   }
   ```

4. **`sim/workload/generator.go:322-325`** truncates to exact count:
   ```go
   // Truncate to maxRequests after merge-sort
   if maxRequests > 0 && int64(len(allRequests)) > maxRequests {
       allRequests = allRequests[:maxRequests]
   }
   ```

### What Happens

For the example above (rate=5, duration=600s, target=3000 requests):

1. Generator samples exponential IATs until `currentTime >= 600s`
2. Due to Poisson variance, generates ~3054 requests by 600s (√3000 ≈ 55 stdev)
3. Truncates to exactly 3000 requests
4. **Last request arrives at ~588s** (12 seconds lost)
5. Coefficient of variation: **CV = 1.0** (unconstrained Poisson)

### Missing: Pre-Generation with Normalization

BLIS's architecture generates requests **incrementally** (streaming), controlled by:
1. **Horizon** (time boundary): `currentTime < horizon`
2. **maxRequests** (count cap): post-generation truncation

There is **no mechanism** to:
- Pre-generate N intervals
- Normalize them to sum to exactly T seconds
- Preserve the exponential distribution shape while constraining the sum

### Existing Arrival Samplers

From `sim/workload/arrival.go`:

| Sampler | Inter-Arrival Time | CV | Request Count | Duration |
|---------|-------------------|-----|---------------|----------|
| `PoissonSampler` | `Exponential(1/λ)` | 1.0 | Approximate (N ± √N) | Exact (T) |
| `ConstantArrivalSampler` | `1/λ` | 0.0 | Approximate | Exact (T) |
| `GammaSampler` | `Gamma(α, β)` | Configurable | Approximate | Exact (T) |
| `WeibullSampler` | `Weibull(k, λ)` | Configurable | Approximate | Exact (T) |

**None support:** Exponential IAT + exact count + exact duration simultaneously.

### Systematic Bias

Ground-truth data:
- ✅ Requests span full duration (0-600s)
- ✅ Late-stage load observable (590-600s has requests)
- ✅ Stage transitions at exact boundaries (600.0s)

BLIS simulation:
- ❌ Requests span shortened duration (0-588s)
- ❌ Late-stage load missing (590-600s has no requests)
- ❌ Stage transitions blurred (truncation artifacts)

### Quantitative Difference

For a typical stage (rate=10, duration=600s, target=6000 requests):

**inference-perf normalized exponential:**
- Mean IAT: 100.0ms
- Std IAT: ~99.9ms (normalized CV ≈ 0.999)
- Duration: exactly 600.000s
- Count: exactly 6000

**BLIS truncated Poisson:**
- Mean IAT: 100.0ms  
- Std IAT: 100.0ms (CV = 1.0)
- Duration: ~588s (2% shorter)
- Count: exactly 6000 (post-truncation)

**Lost coverage:** ~12s per 600s stage = **2% of measurement window**

### Training Implications

The training pipeline (`run_blis_and_compute_loss.py`) compares:
- Ground truth: requests distributed 0-600s
- BLIS predictions: requests distributed 0-588s

The loss metric (RMSE of APE) includes error from:
1. ✅ Latency model accuracy (what we want to measure)
2. ❌ **Workload generation mismatch** (confounding variable)

Late-arriving requests in ground truth have higher latency (queue buildup). BLIS never predicts these because it doesn't generate them.

## Proposed Solution

Add a `NormalizedExponentialSampler` to `sim/workload/arrival.go`:

### High-Level Design

```go
type NormalizedExponentialSampler struct {
    intervals []int64  // Pre-generated normalized intervals
    index     int      // Current position
}

func NewNormalizedExponentialSampler(rate float64, duration int64, count int) *NormalizedExponentialSampler {
    // 1. Generate count exponential samples
    raw := make([]float64, count)
    for i := range raw {
        raw[i] = rng.ExpFloat64() / rate
    }
    
    // 2. Normalize to sum exactly to duration
    sum := 0.0
    for _, v := range raw {
        sum += v
    }
    scale := float64(duration) / sum
    
    // 3. Convert to microseconds
    intervals := make([]int64, count)
    for i, v := range raw {
        intervals[i] = int64(v * scale)
    }
    
    return &NormalizedExponentialSampler{intervals: intervals}
}

func (s *NormalizedExponentialSampler) SampleIAT(_ *rand.Rand) int64 {
    if s.index >= len(s.intervals) {
        return 1  // Defensive: should never happen
    }
    iat := s.intervals[s.index]
    s.index++
    return iat
}
```

### Integration Points

1. **`sim/workload/inference_perf.go`**: Pass duration and request count to arrival spec
   ```go
   Arrival: ArrivalSpec{
       Process: "normalized_exponential",
       Duration: &stageDurationUs,
       Count: &stageRequestCount,
   }
   ```

2. **`sim/workload/generator.go`**: Check for pre-generated mode
   ```go
   // Skip horizon loop for pre-generated samplers
   if sampler.IsPreGenerated() {
       for i := 0; i < sampler.Count(); i++ {
           iat := sampler.SampleIAT(clientRNG)
           currentTime += iat
           // Generate request...
       }
   }
   ```

### Alternative: Extend Existing Samplers

Could add a `Normalized` flag to `ArrivalSpec` that wraps any sampler:
```go
type ArrivalSpec struct {
    Process    string
    Normalized bool    // Pre-generate and normalize to exact duration
    Count      *int64  // Required when Normalized=true
}
```

## Related Work

- Issue #481: Inference-perf multi-turn mapping (fixed request count, didn't address timing)
- H30-H32 experiments: Discovered 2% throughput variance attributed to "Poisson noise"
- `docs/plans/h30-h32-replay-plan.md:63`: "±2.5% measures Poisson noise, not step-time accuracy"

## Acceptance Criteria

1. BLIS can generate workloads matching inference-perf's normalized exponential timing
2. For ground-truth profile with rate=R, duration=D:
   - Request count: exactly R×D
   - Duration span: exactly D seconds (last arrival ≈ D)
   - Inter-arrival distribution: exponential with normalized sum
3. Training loss metric no longer includes workload mismatch confound
4. Backward compatible: existing Poisson specs unchanged

## References

- inference-perf source: https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/loadgen/load_timer.py
- Ground-truth data: `training/trainval_data/*/profile.yaml`
- BLIS arrival samplers: `sim/workload/arrival.go`

susiejojo:
## Summary

Adds support for inference-perf's "normalized exponential" arrival pattern, eliminating systematic 2% duration bias in training data.

**Problem:** inference-perf generates exactly N requests spanning exactly T seconds by normalizing exponential samples. BLIS uses Poisson which stops at horizon, generating ~2% fewer requests.

**Solution:** Direct port of inference-perf's algorithm (~15 lines of Python → ~25 lines of Go):
1. Add `CustomSampler` field to `ClientSpec` (programmatic use only)  
2. Implement `NormalizedExponentialSampler` with zero-IAT exhaustion signal
3. Construct directly in `ExpandInferencePerfSpec` for single-stage workloads

## Implementation

- **Modified files:** 6 (spec.go, arrival.go, inference_perf.go, generator.go, arrival_test.go, generator_test.go)
- **Lines changed:** ~280 LOC (production + tests)
- **Tests:** All passing (`go test ./... -count=1`)

### Key Changes

1. **Interface contract update:** `ArrivalSampler.SampleIAT()` now allows 0 return value to signal exhaustion for stateful samplers
2. **Validation guards:** count limits, degenerate distribution checks, division-by-zero guards
3. **Zero-IAT guards:** Both reasoning paths in generator.go handle exhaustion signal
4. **Comprehensive tests:** Edge cases (minimal duration, large count), panic validation, zero-IAT exhaustion, reasoning integration

## Commits

1. `bc4d814` - docs(plan): ultra-simplified normalized exponential approach
2. `ed14f1c` - feat(workload): add CustomSampler field and zero IAT stop signal (BC-2)
3. `e74c893` - feat(workload): add NormalizedExponentialSampler (T2)
4. `d9215ca` - feat(workload): integrate NormalizedExponentialSampler into inference-perf (T3)
5. `b71e04e` - fix(workload): address Round 1 convergence review findings

## Review Status

**Convergence review in progress:** Round 2 active (2 CRITICAL + 36 IMPORTANT findings being addressed)

## References

- Closes #845
- Plan: `docs/plans/pr845-normalized-exponential-simple-plan.md`
- Parent branch: `main` (excludes training branch commits as requested)

### DIFF
--- a/docs/plans/pr845-normalized-exponential-simple-plan.md
+++ b/docs/plans/pr845-normalized-exponential-simple-plan.md
@@ -39,18 +39,20 @@ THEN sum(intervals) ≈ 600,000,000 µs (within count microseconds for rounding)
 **BC-2: Request Generation Guarantees**
 ```gherkin
 GIVEN a client with CustomSampler=NormalizedExponentialSampler(count=3000, duration=600s)
-WHEN workload is generated
+WHEN workload is generated with horizon >= duration
 THEN exactly 3000 requests are generated for that client
   AND all 3000 requests arrive within the duration window
+NOTE: If horizon < duration, some requests may be dropped by the horizon guard
 ```
 
-**BC-3: Inference-Perf Auto-Application**
+**BC-3: Inference-Perf Auto-Application (Single-Stage)**
 ```gherkin
-GIVEN an InferencePerfSpec with any stages
+GIVEN an InferencePerfSpec with a single stage
 WHEN expanded to WorkloadSpec
 THEN all clients use NormalizedExponentialSampler via CustomSampler field
-  AND count is set from (stage.Rate * stage.Duration / numClients)
+  AND count is set from ceil(stage.Rate * stage.Duration / numClients)
   AND duration is set from stage.Duration in microseconds
+NOTE: Multi-stage workloads continue to use Poisson arrival for now
 ```
 
 ### C. Component Interaction
--- a/docs/plans/pr845-normalized-exponential-simple-plan.md
+++ b/docs/plans/pr845-normalized-exponential-simple-plan.md
@@ -45,14 +45,44 @@ THEN exactly 3000 requests are generated for that client
 NOTE: If horizon < duration, some requests may be dropped by the horizon guard
 ```
 
-**BC-3: Inference-Perf Auto-Application (Single-Stage)**
+**BC-3: Inference-Perf Auto-Application (Single-Stage Only)**
 ```gherkin
-GIVEN an InferencePerfSpec with a single stage
+GIVEN an InferencePerfSpec with a single stage AND no multi-turn
 WHEN expanded to WorkloadSpec
-THEN all clients use NormalizedExponentialSampler via CustomSampler field
+THEN all clients use NormalizedExponentialSampler via CustomSamplerFactory
   AND count is set from ceil(stage.Rate * stage.Duration / numClients)
   AND duration is set from stage.Duration in microseconds
-NOTE: Multi-stage workloads continue to use Poisson arrival for now
+NOTE: Multi-stage workloads continue to use Poisson arrival (see BC-4)
+NOTE: Multi-turn workloads continue to use Poisson arrival (see BC-5)
+```
+
+**BC-4: Multi-Stage Asymmetry (Architectural Constraint)**
+```gherkin
+GIVEN an InferencePerfSpec with multiple stages
+WHEN expanded to WorkloadSpec
+THEN all clients use Poisson arrival (not NormalizedExponentialSampler)
+RATIONALE: Multi-stage workloads use per-stage client cohorts with lifecycle windows.
+  Each cohort is active only during its stage's window. NormalizedExponentialSampler
+  pre-generates N intervals spanning the FULL duration, but multi-stage clients are
+  only active for a SUBSET of that duration (their stage's window). This mismatch
+  would waste intervals or require complex windowing logic. Poisson generates
+  intervals incrementally during the active window, matching the architecture.
+FUTURE WORK: Per-stage NormalizedExponentialSampler construction (each stage gets
+  its own sampler with count/duration scoped to that stage).
+```
+
+**BC-5: Multi-Turn Asymmetry (Session Start Time Only)**
+```gherkin
+GIVEN an InferencePerfSpec with multi-turn enabled
+WHEN expanded to WorkloadSpec
+THEN all clients use Poisson arrival for session start times
+RATIONALE: Multi-turn workloads with SingleSession=true use ONE arrival sample
+  for the session start time. The rounds within the session are spaced by
+  ThinkTimeUs (not sampled IATs). NormalizedExponentialSampler pre-generates N
+  intervals, but only the first would be used — wasting N-1 intervals.
+  Poisson generates the session start time on-demand without waste.
+NOTE: The rounds within each session still occur at exact ThinkTimeUs intervals,
+  maintaining inference-perf's round-robin cycling semantics.
 ```
 
 ### C. Component Interaction
--- a/docs/plans/pr845-normalized-exponential-simple-plan.md
+++ b/docs/plans/pr845-normalized-exponential-simple-plan.md
@@ -87,12 +87,17 @@ NOTE: The rounds within each session still occur at exact ThinkTimeUs intervals,
 
 ### C. Component Interaction
 
-**Generator Loop (add CustomSampler check):**
+**Generator Loop (CustomSamplerFactory pattern with sub-RNG isolation):**
 ```go
-// generator.go:~115 (new code):
+// generator.go:~114 (post-refactor):
 var arrivalSampler ArrivalSampler
-if client.CustomSampler != nil {
-    arrivalSampler = client.CustomSampler
+if client.CustomSamplerFactory != nil {
+    // Derive sub-RNG for factory with single entropy draw from clientRNG.
+    // This isolates the sampler's N-draw RNG consumption from downstream
+    // content sampling, keeping input/output distributions stable.
+    subSeed := clientRNG.Int63()
+    subRNG := newRandFromSeed(subSeed)
+    arrivalSampler = client.CustomSamplerFactory(subRNG)
 } else {
     arrivalSampler = NewArrivalSampler(client.Arrival, ratePerMicrosecond)
 }
@@ -100,7 +105,7 @@ if client.CustomSampler != nil {
 // Rest of loop unchanged:
 for currentTime < horizon {
     iat := arrivalSampler.SampleIAT(clientRNG)
-    if iat == 0 { break }  // Exhausted sampler signals stop
+    if iat == 0 { break }  // Stateful sampler exhausted (e.g., NormalizedExponentialSampler)
     currentTime += iat
     if currentTime >= horizon { break }
     // generate request...
--- a/docs/plans/pr845-normalized-exponential-simple-plan.md
+++ b/docs/plans/pr845-normalized-exponential-simple-plan.md
@@ -1,0 +1,302 @@
+# Implementation Plan: Normalized Exponential Arrival Sampler (Ultra-Simplified)
+
+**PR Number:** #845
+**Goal:** Add normalized exponential arrival sampler for inference-perf fidelity
+**Source:** GitHub issue #845
+**Closes:** Fixes #845
+**Tier:** Small (3 files, ~80 LOC total)
+
+---
+
+## Part 1: Design Validation
+
+### A. Executive Summary
+
+This PR adds support for inference-perf's "normalized exponential" arrival pattern, eliminating systematic 2% duration bias in training data.
+
+**Problem:** inference-perf generates exactly N requests spanning exactly T seconds by normalizing exponential samples. BLIS uses Poisson which stops at horizon, generating ~2% fewer requests.
+
+**Solution:** Direct port of inference-perf's algorithm (~15 lines of Python → ~25 lines of Go), bypassing the factory pattern entirely:
+1. Add `CustomSampler` field to `ClientSpec` (programmatic use only)
+2. Implement `NormalizedExponentialSampler` (no factory registration)
+3. Construct directly in `ExpandInferencePerfSpec`
+
+**Files modified:** 3 files (spec.go, arrival.go, inference_perf.go)
+**Lines changed:** ~50 LOC production + ~30 LOC tests = ~80 LOC total
+
+**Key simplification:** No factory changes, no YAML changes, no sentinel overflow, no signature changes.
+
+### B. Behavioral Contracts
+
+**BC-1: Normalized Distribution**
+```gherkin
+GIVEN a sampler with count=3000, duration=600s (600M µs), rate=5/s
+WHEN all 3000 intervals are generated
+THEN sum(intervals) ≈ 600,000,000 µs (within count microseconds for rounding)
+  AND exactly 3000 intervals are returned
+```
+
+**BC-2: Request Generation Guarantees**
+```gherkin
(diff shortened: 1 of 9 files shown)


sriumcp (changes requested):
## Review: Workload Non-Reusability + Related Fixes

The normalized exponential sampler is the right approach for inference-perf parity, and the multi-turn/closed-loop separation is correct. Five things to fix before merge.

---

### 1. IMPORTANT — `spec.go` line 128: Replace pre-built instance with factory function

**Problem:** `CustomSampler ArrivalSampler` stores a stateful `NormalizedExponentialSampler` with an internal `index` counter. Once `GenerateRequests` consumes all its intervals, the sampler is permanently exhausted. Any second call to `GenerateRequests` on the same `WorkloadSpec` silently produces zero requests for those clients. `WorkloadSpec` was previously idempotent; this breaks that contract without any warning.

**Fix:** Store a factory function instead of an instance. The factory is invoked once per `GenerateRequests` call, constructing a fresh sampler each time. `WorkloadSpec` is reusable again.

```go
// REMOVE:
CustomSampler ArrivalSampler `yaml:"-"`

// REPLACE WITH:
// CustomSamplerFactory allows programmatic injection of arrival samplers,
// bypassing the factory-based construction from Arrival.Process.
//
// When set, GenerateRequests calls it once per invocation with the client's RNG
// to obtain a fresh sampler — making WorkloadSpec reusable (INV-6 safe).
// Not exposed in YAML (yaml:"-" tag).
CustomSamplerFactory func(*rand.Rand) ArrivalSampler `yaml:"-"`
```

---

### 2. IMPORTANT — `spec.go` ~line 268: Update `validateClient` guard to new field name

Rename the guard condition to match the renamed field:

```go
// CHANGE:
if c.CustomSampler == nil {

// TO:
if c.CustomSamplerFactory == nil {
```

---

### 3. IMPORTANT — `generator.go` ~line 115: Use factory + sub-RNG

**Problem 1:** The `client.CustomSampler` reference must become `client.CustomSamplerFactory`.

**Problem 2:** `NormalizedExponentialSampler` draws `N` random values during construction. If `clientRNG` is passed directly to the factory, it consumes a large, variable amount of entropy before `inputSampler`/`outputSampler` are called downstream. This makes the generated request content (input/output lengths) unpredictably sensitive to request count. Fix: derive a sub-RNG with a single draw, so `clientRNG`'s downstream behaviour is stable.

```go
// REMOVE:
var arrivalSampler ArrivalSampler
if client.CustomSampler != nil {
    arrivalSampler = client.CustomSampler
} else {
    arrivalSampler = NewArrivalSampler(client.Arrival, clientRate)
}

// REPLACE WITH:
var arrivalSampler ArrivalSampler
if client.CustomSamplerFactory != nil {
    // Derive an independent sub-RNG for sampler construction.
    // Consumes exactly one value from clientRNG, keeping the
    // content-sampler (input/output lengths) stream clean and predictable.
    samplerRNG := rand.New(rand.NewSource(clientRNG.Int63()))
    arrivalSampler = client.CustomSamplerFactory(samplerRNG)
} else {
    arrivalSampler = NewArrivalSampler(client.Arrival, clientRate)
}
```

Add `"math/rand"` to the import block if not already present.

---

### 4. IMPORTANT — `inference_perf.go` ~line 155: Replace instance construction with factory closure; remove `baseRNG`

With the factory approach, interval generation is deferred to `GenerateRequests`, so the separate `baseRNG` / `clientSeed` / `clientRNG` derivation inside `ExpandInferencePerfSpec` is no longer needed. `requestsPerClient` and `durationUs` are captured in the closure at expansion time (correct — these are fixed per spec).

```go
// REMOVE the entire baseRNG block:
totalClients := int64(sp.NumUniqueSystemPrompts) * int64(sp.NumUsersPerSystemPrompt)
if totalClients > 1_000_000 {
    return nil, fmt.Errorf("total client count %d exceeds safety limit (1M)", totalClients)
}
baseRNG := rand.New(rand.NewSource(seed))

// AND inside the inner loop, REMOVE:
clientSeed := baseRNG.Int63()
clientRNG := rand.New(rand.NewSource(clientSeed))
sampler := NewNormalizedExponentialSampler(clientRNG, requestsPerClient, durationUs)

// AND CHANGE:
CustomSampler: sampler,

// TO (capturing loop variables explicitly to avoid closure capture bugs):
rpc := requestsPerClient
dur := durationUs
// ... and in the ClientSpec literal:
CustomSamplerFactory: func(rng *rand.Rand) ArrivalSampler {
    return NewNormalizedExponentialSampler(rng, rpc, dur)
},
Arrival: ArrivalSpec{Process: "poisson"}, // validation placeholder; CustomSamplerFactory is used for generation
```

Also remove the `import "math/rand"` from `inference_perf.go` if nothing else in that file uses it after this change.

---

### 5. IMPORTANT — `inference_perf.go` `else` branch (multi-stage): Document the asymmetry

Single-stage language workloads now use `NormalizedExponentialSampler` (exact count). Multi-stage language workloads still use Poisson (random count). This is invisible to users; document it:

```go
} else {
    // Multi-stage workloads use Poisson arrival (not NormalizedExponentialSampler).
    // NormalizedExponentialSampler requires a fixed duration per client, which
    // conflicts with lifecycle windows (clients start/stop at different times
    // across stages). Multi-stage language workloads therefore retain Poisson,
    // meaning the request count is a random variable rather than exact.
    // Single-stage language workloads use NormalizedExponentialSampler for
    // inference-perf parity (exact count within a fixed duration).
```

---

### Minor fixes (no behaviour change)

**`cohort_test.go` — update the field exclusion key:**
```go
// CHANGE:
"CustomSampler": true,
// TO:
"CustomSamplerFactory": true,
```

**`generator.go` Guard 1 comment — minor clarification:**
```go
// CHANGE:
continue // Sampler exhausted; skip this client
// TO:
continue // Sampler exhausted (stateful sampler returned 0); no intervals remain for this client
```

**`arrival.go` — update interface doc comment** to reflect that 0 is only possible from `NormalizedExponentialSampler`, not standard samplers:
```go
// SampleIAT returns the next inter-arrival time in microseconds.
// Standard samplers (Poisson, Gamma, Weibull, Constant) always return >= 1.
// NormalizedExponentialSampler returns 0 after all pre-generated intervals
// are exhausted; callers must treat 0 as a stop signal (see generator.go guards).
SampleIAT(rng *rand.Rand) int64
```

---

### What's correct and should not change

- Multi-turn path (`EnableMultiTurnChat == true`) correctly keeps Poisson and no `CustomSamplerFactory`. This is right — session rounds are spaced by `ThinkTimeUs`, not arrival IATs.
- Closed-loop session generation in `GenerateWorkload` is unaffected — the zero-IAT guards are no-ops for all standard samplers.
- The `NormalizedExponentialSampler` algorithm itself is correct.
- Guards 2 and 3 (`break` on `iat == 0`) are correctly placed.
- YAML round-trip safety is fine — `yaml:"-"` on `CustomSamplerFactory` means it is invisible to serialization.

PR #914, opened 2026-04-01
feat(sim): add precise-prefix-cache and no-hit-lru scorers

### LINKED ISSUE #883: feat: Add precise prefix cache scoring to match llm-d production implementation

BLIS's current prefix routing uses a **router-side approximate** approach (`PrefixCacheIndex` with hierarchical SHA256 block hashing and LRU eviction). The production llm-d system uses a fundamentally different **precise prefix cache scorer** backed by real KV cache events via `llm-d-kv-cache`.

When translating BLIS routing to llm-d, prefix affinity was dropped entirely (hardcoded to `0.0`) because the sim's approximate approach has no production equivalent. This means:
- Simulation routing decisions diverge from production — the sim guesses what's cached, production knows
- The BLIS weighted scoring algorithm cannot be validated against llm-d since they use different prefix mechanisms
- The BLIS-to-llm-d translation gap forces prefix affinity to be ignored, reducing the translated scorer to load-balance-only

**Proposed solution**

Add a precise prefix cache scorer to the sim that mirrors the llm-d `LongestPrefixMatch` scoring, using the sim's existing KV cache state instead of router-side approximation.

### What llm-d does (reference implementation)

The production [`PrecisePrefixCacheScorer`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/precise_prefix_cache.go):

- Uses `kvcache.Indexer` with real KV block events from the engine
- Tokenizes prompts, computes block hashes via [`llm-d-kv-cache`](https://github.com/llm-d/llm-d-kv-cache)
- Scores using [`LongestPrefixMatch`](https://github.com/llm-d/llm-d-kv-cache/blob/main/pkg/kvcache/kvblock_scorer.go) strategy: longest consecutive block match from block 0
- **Normalization:** Raw scores are **min-max normalized** across all candidate endpoints via [`indexedScoresToNormalizedScoredPods`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/utils.go) — highest raw score → 1.0, lowest → 0.0, all-equal → 1.0. This is NOT proportional (`matched/total`); it uses the relative range across candidates.
- Optional speculative indexing: after routing, proactively adds predicted cache entries

Supporting scorers that compose with it:
- [`NoHitLRU`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/no_hit_lru.go): distributes cold requests (cache misses) to least-recently-used endpoints.
- Production config ([`epp-precise-prefix-cache-config.yaml`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/deploy/config/epp-precise-prefix-cache-config.yaml)): `precise-prefix-cache:2, kv-cache-utilization:1, queue-scorer:1` — notably does **not** include `no-hit-lru`.

### What BLIS already has (building blocks)

The sim already has the key building block for precise prefix scoring:

- **`KVStore.GetCachedBlocks(tokens []int) []int64`** ([`kv_store.go:7`](https://github.com/inference-sim/inference-sim/blob/main/sim/kv_store.go)) — returns the longest consecutive prefix block match against actual instance KV cache state. Already used by batch formation ([`batch_formation.go:161`](https://github.com/inference-sim/inference-sim/blob/main/sim/batch_formation.go)). This is the sim-side equivalent of llm-d's `LongestPrefixMatch`.
- **`InstanceSimulator`** already exposes KV metrics (`KVUtilization`, `CacheHitRate`, `FreeKVBlocks`) via public methods that delegate to the unexported `sim.KVCache` field — precedent for cluster-layer reading instance KV state. The precise scorer needs a new public method on `InstanceSimulator` following this same pattern (see below).
- **Existing approximate scorer** ([`routing_prefix_scorer.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/routing_prefix_scorer.go)) — uses proportional formula (`matchedBlocks / totalBlocks`) against a router-side LRU.
- **Existing min-max normalization** — `scoreQueueDepth` in [`routing_scorers.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/routing_scorers.go) already implements min-max normalization (all-equal → 1.0, otherwise `(max - value) / (max - min)`). The precise scorer should follow this same pattern for internal consistency and llm-d parity.

### What needs to happen

**1. New scorer: `precise-prefix-cache`**

Queries actual instance KV cache state and scores using **min-max normalization** across candidates (matching llm-d's `indexedScoresToNormalizedScoredPods` and BLIS's own `scoreQueueDepth`):
- Raw score per instance: `len(GetCachedBlocks(req.InputTokens))` (matched block count)
- Normalized: highest raw → 1.0, lowest raw → 0.0, all-equal → 1.0
- Requests shorter than one block: raw score 0 for all instances → all-equal → all score 1.0

No router-side index, no approximation. Observer is nil (state is ground truth from the instance).

**New public method on `InstanceSimulator`:** Since `inst.sim.KVCache` is unexported, add a public accessor following the existing `KVUtilization()` / `FreeKVBlocks()` / `CacheHitRate()` pattern:
```go
func (i *InstanceSimulator) GetCachedBlockCount(tokens []int) int {
    return len(i.sim.KVCache.GetCachedBlocks(tokens))
}
```
The cluster layer wraps this into the `cacheQueryFn` map passed to `NewRoutingPolicy`.

**Factory signature change required:** The current scorer factory chain is:
```
NewRoutingPolicy(name, scorerConfigs, blockSize, rng)
  → newScorerWithObserver(name string, blockSize int) → (scorerFunc, observerFunc)
```

`scorerFunc` signature is `func(req *Request, snapshots []RoutingSnapshot) map[string]float64` — it has no access to per-instance `KVStore`. The precise scorer needs to call `GetCachedBlocks` on each instance's KV cache.

**Proposed approach:** Add a per-instance cache query function to the factory chain. Concretely:
- Add a new parameter to `newScorerWithObserver`: `cacheQueryFn map[string]func([]int) int` — keyed by instance ID, returns matched block count for given tokens. Nil for scorers that don't need it.
- `NewRoutingPolicy` receives this map from the cluster layer (where `InstanceSimulator` references are available) and passes it through.
- The precise scorer calls `cacheQueryFn[instanceID](req.InputTokens)` for each snapshot.
- Stateless scorers ignore the parameter.

**2. New scorer: `no-hit-lru`**

For cold requests (no prefix cache hits across any instance), distributes to least-recently-used endpoints. Matching the llm-d [`no-hit-lru-scorer`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/no_hit_lru.go) algorithm:

**Scoring logic:**
- **Warm requests** (any instance has cached blocks): return **0.5** (neutral) for all instances
- **Cold requests** (no cached blocks on any instance): score by LRU position:
  - **Never-used endpoints rank above all used endpoints** — they get the highest scores
  - **Positional scoring:** `score = 1.0 - rank / (totalEndpoints - 1)` where rank 0 = never-used (or oldest used)
  - **Single endpoint:** score 1.0 directly (avoids division by zero)
- **LRU is only updated on cold request routing** — warm requests do not touch the LRU order

**Cache hit detection:** `no-hit-lru` also receives `cacheQueryFn` and independently checks `cacheQueryFn[id](req.InputTokens) > 0` for each instance. This is self-contained — no cross-scorer coupling, no ordering dependency with `precise-prefix-cache`. `GetCachedBlocks` is an in-memory hash lookup, so the redundant call is negligible.

**LRU state:** Uses a dedicated LRU data structure (keyed by instance ID). Updated via `observerFunc` — but only for cold requests. The observer checks the same `cacheQueryFn` to determine if the routed request was cold before updating. Never-used endpoints are those absent from the LRU.

**3. Integration with `WeightedScoring`**

Register both in `newScorerWithObserver` switch and `validScorerNames` map (2 touch points per scorer, matching existing pattern).

**4. New named scorer preset**

Add a preset matching the llm-d production config ([`epp-precise-prefix-cache-config.yaml`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/deploy/config/epp-precise-prefix-cache-config.yaml)). This is a **new named preset**, not a replacement for `DefaultScorerConfigs()`:
```
precise-prefix-cache:2.0
kv-utilization:1.0
queue-depth:1.0
```
`no-hit-lru` is available as an opt-in scorer but is **not** in the default preset (matching production).

### CLI example

```bash
# Production-matching preset
./blis run --model qwen/qwen3-14b --num-instances 4 \
  --routing-policy weighted \
  --scorer-config "precise-prefix-cache:2.0,kv-utilization:1.0,queue-depth:1.0"

# With optional no-hit-lru for cache growth distribution
./blis run --model qwen/qwen3-14b --num-instances 4 \
  --routing-policy weighted \
  --scorer-config "precise-prefix-cache:2.0,no-hit-lru:1.0,kv-utilization:1.0"
```

- [x] Core simulator (`sim/`) — new scorer implementations, `newScorerWithObserver` signature change
- [x] Cluster simulation (`sim/cluster/`) — new `GetCachedBlockCount` method on `InstanceSimulator`; pass per-instance cache query refs to `NewRoutingPolicy`
- [ ] KV cache (`sim/kv/`) — `GetCachedBlocks` already exists, no changes needed
- [x] CLI (`cmd/`) — register new scorer names
- [ ] New package needed

**Extension friction check**

- ~4-5 files modified, ~1-2 new files (precise scorer, no-hit-lru scorer)
- `newScorerWithObserver` gains one parameter (`cacheQueryFn`); existing callers pass nil
- `InstanceSimulator` gains one public method (`GetCachedBlockCount`)
- Registration in `newScorerWithObserver` switch + `validScorerNames` map (2 touch points per scorer)
- Does not affect invariants — routing policy is a pure scoring concern. KV cache read is read-only.

**Alternatives considered**

1. **Port BLIS's approximate approach to llm-d.** Rejected — llm-d already has the precise scorer, and approximation is strictly worse (drifts without eviction feedback). The sim should match production, not the other way around.

2. **Make the translated BLIS scorer compose with `PrecisePrefixCacheScorer` via CycleState.** Rejected for the sim — CycleState is an llm-d-inference-scheduler concept, not a sim concept. The sim should have its own precise scorer that reads instance state directly.

3. **Replace the existing `prefix-affinity` scorer.** Rejected — keep it as an alternative for researchers who want to study approximate vs. precise routing. The new scorer is additive.

4. **`no-hit-lru` reads scores from `precise-prefix-cache` via shared state.** Rejected — creates implicit coupling and ordering dependency between scorers. Instead, `no-hit-lru` independently queries `cacheQueryFn` to detect cache hits. `GetCachedBlocks` is an in-memory hash lookup, so the redundant call is negligible.

5. **Proportional normalization (`matched/total`) instead of min-max.** Rejected — llm-d uses min-max normalization, and BLIS's own `scoreQueueDepth` uses min-max. Proportional scoring compresses prefix scores into a narrow band near zero (e.g., 10/100 = 0.1), causing them to be dominated by other scorers. Min-max uses the full [0, 1] range.

**Known fidelity gaps (non-blocking)**

- **Device-tier weighting:** llm-d's `LongestPrefixScorer` weights matched blocks by device tier (GPU=1.0, CPU=0.8). BLIS's `GetCachedBlocks` queries the GPU tier only (`tiered.go:324-326`), so there's no way to distinguish tiers currently. Acceptable for first pass; can be addressed later if tiered prefix scoring becomes important.

**Non-goals**

- Do NOT add speculative indexing — the sim is synchronous (no routing-to-KV blind spot)
- Do NOT add P/D disaggregation logic (separate concern)
- Do NOT change `DefaultScorerConfigs()` — the new preset is additive

**References**

llm-d:
- [`precise_prefix_cache.go`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/precise_prefix_cache.go) — production scorer
- [`kvblock_scorer.go`](https://github.com/llm-d/llm-d-kv-cache/blob/main/pkg/kvcache/kvblock_scorer.go) — `LongestPrefixMatch` strategy
- [`utils.go`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/utils.go) — `indexedScoresToNormalizedScoredPods` (min-max normalization)
- [`no_hit_lru.go`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/no_hit_lru.go) — cold request distribution
- [`epp-precise-prefix-cache-config.yaml`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/deploy/config/epp-precise-prefix-cache-config.yaml) — production config

BLIS:
- [`kv_store.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/kv_store.go) — `GetCachedBlocks` interface (the key building block)
- [`batch_formation.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/batch_formation.go) — existing `GetCachedBlocks` usage
- [`routing_prefix_scorer.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/routing_prefix_scorer.go) — current approximate scorer
- [`routing_scorers.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/routing_scorers.go) — scorer factory, registration, and `scoreQueueDepth` (min-max normalization reference)
- [`routing.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/routing.go) — `WeightedScoring`, `scorerFunc`, `observerFunc`, `NewRoutingPolicy`

mtoslalibu:
## Summary

- Add `precise-prefix-cache` scorer: queries actual instance KV cache state via `cacheQueryFn`, applies min-max normalization (llm-d production parity)
- Add `no-hit-lru` scorer: distributes cold requests to least-recently-used instances; warm requests score neutral 0.5
- Add `NewRoutingPolicyWithCache` constructor (dual-constructor pattern: `NewRoutingPolicy` unchanged, zero test churn)
- Add `GetCachedBlockCount` accessor on `InstanceSimulator`
- Wire `cacheQueryFn` in cluster layer including deferred NodePool instance registration

## Behavioral Contracts

- **BC-1**: GIVEN instances with varying cached prefix blocks, WHEN routed with precise-prefix-cache, THEN min-max normalization scores highest→1.0, lowest→0.0
- **BC-2a**: GIVEN all instances have the same **nonzero** cached block count, THEN all score 1.0
- **BC-2b**: GIVEN all instances have **zero** cached blocks (no cache data to differentiate), THEN all score 0.5 (neutral)
- **BC-3**: GIVEN a cold request (zero cache hits everywhere), WHEN routed with no-hit-lru, THEN never-used instances score highest, used instances by LRU position
- **BC-4**: GIVEN a warm request (cache hit on any instance), WHEN routed with no-hit-lru, THEN all instances score 0.5 (neutral)
- **BC-5**: GIVEN a warm request routed via observer, THEN LRU state is not updated
- **BC-6**: GIVEN `NewRoutingPolicyWithCache` called with non-nil `cacheQueryFn`, WHEN precise-prefix-cache or no-hit-lru configured, THEN scorers use provided functions
- **BC-7**: GIVEN nil `cacheQueryFn` (all existing call sites via `NewRoutingPolicy`), THEN behavior identical to before this PR
- **BC-8**: GIVEN precise-prefix-cache scorer, THEN observer is nil (stateless ground truth)

## Test Plan

- [x] `TestPrecisePrefixCache_MinMaxNormalization` — BC-1 with 3 instances (5,3,0 cached blocks)
- [x] `TestPrecisePrefixCache_AllEqual` — BC-2a/BC-2b (zero→0.5 and nonzero→1.0 subtests)
- [x] `TestPrecisePrefixCache_MissingInstanceInCacheQueryFn` — missing instance treated as 0 cached blocks
- [x] `TestPrecisePrefixCache_ObserverIsNil` — BC-8
- [x] `TestPrecisePrefixCache_SingleInstance` — single instance edge case
- [x] `TestPrecisePrefixCache_NilCacheQueryFn` — BC-7 for precise-prefix
- [x] `TestNoHitLRU_ColdRequestDistribution` — BC-3 multi-step sequence
- [x] `TestNoHitLRU_WarmRequestNeutral` — BC-4
- [x] `TestNoHitLRU_ObserverColdOnly` — BC-5 (mutable closure verifies original scorer)
- [x] `TestNoHitLRU_LRURepromotion` — re-routing same instance re-promotes to MRU
- [x] `TestNoHitLRU_ObserverMismatchedReqID` — mismatched req.ID is observer no-op
- [x] `TestNoHitLRU_SingleInstanceCold` — single instance edge case
- [x] `TestNoHitLRU_NilCacheQueryFn` — BC-7 for no-hit-lru
- [x] `TestNoHitLRU_Determinism` — INV-6
- [x] `TestAllScorers_ReturnScoreForEveryInstance` — all 5 scorers return [0,1] scores (+ nil-request subtest for stateless scorers)
- [x] `TestNewScorerFactory_PrecisePrefixAndNoHitLRU` — BC-6 factory wiring via `NewRoutingPolicyWithCache`
- [x] All existing tests pass with unchanged `NewRoutingPolicy` signature (zero test churn)

Fixes #883

### DIFF
--- a/sim/metrics.go
+++ b/sim/metrics.go
@@ -125,15 +125,16 @@ func (m *Metrics) SaveResults(instanceID string, horizon int64, totalBlocks int6
 			output.ResponsesPerSec = float64(m.CompletedRequests) / vllmRuntime
 			output.TokensPerSec = float64(m.TotalOutputTokens) / vllmRuntime
 		}
-
-		// Print to stdout (results are primary output, not log messages)
-		fmt.Println("=== Simulation Metrics ===")
-		data, err := json.MarshalIndent(output, "", "  ")
-		if err != nil {
-			return fmt.Errorf("error marshalling metrics: %w", err)
-		}
-		fmt.Println(string(data))
 	}
+
+	// Always emit the metrics section so callers can reliably parse output,
+	// even when CompletedRequests == 0 (e.g., all requests dropped as unservable).
+	fmt.Println("=== Simulation Metrics ===")
+	data, err := json.MarshalIndent(output, "", "  ")
+	if err != nil {
+		return fmt.Errorf("error marshalling metrics: %w", err)
+	}
+	fmt.Println(string(data))
 
 	// --- Write to JSON File ---
 	if outputFilePath != "" {
--- a/sim/metrics_test.go
+++ b/sim/metrics_test.go
@@ -3,8 +3,10 @@ package sim
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -380,6 +382,48 @@ func TestSaveResults_DroppedUnservable_InJSON(t *testing.T) {
 	if output.InjectedRequests != 2 {
 		t.Errorf("InjectedRequests = %d, want 2 (should include dropped)", output.InjectedRequests)
 	}
+}
+
+// TestSaveResults_AlwaysEmitsHeader_ZeroCompletions verifies that the
+// === Simulation Metrics === header is written to stdout even when
+// CompletedRequests == 0 (issue #926).
+//
+// GIVEN metrics with zero completions (all requests dropped as unservable)
+// WHEN SaveResults is called
+// THEN stdout contains "=== Simulation Metrics ===" and a valid JSON block
+func TestSaveResults_AlwaysEmitsHeader_ZeroCompletions(t *testing.T) {
+	m := NewMetrics()
+	m.DroppedUnservable = 100
+	m.SimEndedTime = 1_000_000
+
+	// Capture stdout
+	origStdout := os.Stdout
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+	os.Stdout = w
+
+	saveErr := m.SaveResults("test", 10_000_000, 100, "")
+
+	require.NoError(t, w.Close())
+	os.Stdout = origStdout
+
+	var buf []byte
+	buf, err = io.ReadAll(r)
+	require.NoError(t, err)
+	stdout := string(buf)
+
+	require.NoError(t, saveErr)
+	assert.Contains(t, stdout, "=== Simulation Metrics ===")
+
+	// Extract and parse the JSON block after the header
+	parts := strings.SplitN(stdout, "=== Simulation Metrics ===\n", 2)
+	require.Len(t, parts, 2, "expected header + JSON")
+
+	var out MetricsOutput
+	require.NoError(t, json.Unmarshal([]byte(strings.TrimSpace(parts[1])), &out))
+	assert.Equal(t, 0, out.CompletedRequests)
+	assert.Equal(t, 100, out.DroppedUnservable)
+	assert.Equal(t, 100, out.InjectedRequests)
 }
 
 // BC-3: LengthCappedRequests appears in JSON output


sim/routing_precise_prefix_scorer.go
```diff
@@ -0,0 +1,49 @@
...
+				scores[snap.ID] = 1.0
+			}
+			return scores
+		}
+		// Pass 1: compute raw scores and find min/max
+		raw := make(map[string]int, len(snapshots))
+		minRaw, maxRaw := math.MaxInt, 0
+		for _, snap := range snapshots {
+			count := 0
+			if fn, ok := cacheQueryFn[snap.ID]; ok {
```
sriumcp: **S-1 (Suggestion): Add test for missing `cacheQueryFn` entry**

**Problem:** When `snap.ID` is not in `cacheQueryFn`, `count` silently defaults to 0. No test covers this. A wiring bug (instance in snapshot but not in `cacheQueryFn`) would be indistinguishable from an instance with zero cached blocks.

**How to add test** (in `routing_precise_prefix_scorer_test.go`):

```go
func TestPrecisePrefixCache_MissingInstanceInCacheQueryFn(t *testing.T) {
    // cacheQueryFn has "a" and "b" but NOT "c"
    cacheQueryFn := CacheQueryFn{
        "a": func(tokens []int) int { return 5 },
        "b": func(tokens []int) int { return 0 },
    }
    scorer, _ := newPrecisePrefixCacheScorer(cacheQueryFn)
    scores := scorer(
        &Request{ID: "r1", InputTokens: []int{1}},
        []RoutingSnapshot{{ID: "a"}, {ID: "b"}, {ID: "c"}},
    )
    // "c" missing from cacheQueryFn -> treated as 0 cached blocks (same as "b")
    if scores["a"] \!= 1.0 {
        t.Errorf("a: got %.3f, want 1.0", scores["a"])
    }
    if scores["c"] \!= 0.0 {
        t.Errorf("c (missing): got %.3f, want 0.0 (treated as zero cache)", scores["c"])
    }
}
```

sim/routing_nohit_lru_scorer.go
```diff
@@ -0,0 +1,121 @@
...
+
+		if lastWarm {
+			// BC-4: warm request → neutral 0.5 for all
+			for _, snap := range snapshots {
+				scores[snap.ID] = 0.5
+			}
+			return scores
+		}
+		// BC-3: cold request → LRU positional scoring
+		total := len(snapshots)
```
sriumcp: **S-2 (Suggestion): Add guard for empty snapshots slice**

**Problem:** When `total == 0`, the division `float64(total-1)` on lines 73 and 85 produces `-1` in the denominator. Currently unreachable because `WeightedScoring.Route` panics on empty snapshots first, but this is fragile.

**How to fix:** Add after line 59:

```go
if total == 0 {
    return scores
}
```

sim/routing_nohit_lru_scorer.go
```diff
@@ -0,0 +1,121 @@
...
+		if req == nil {
+			return
+		}
+		// BC-5: use scorer's warm/cold determination (not re-derived).
+		// This avoids disagreement between scorer (checks all instances)
+		// and observer (would only check target instance).
+		if lastWarm || req.ID != lastReqID {
+			return
+		}
+		// Move targetInstance to front of LRU (most-recently-used)
```
sriumcp: **S-3 (Suggestion): Add test for LRU re-promotion (routing same instance twice)**

**Problem:** The observer's `lruSet[targetInstance]` branch (remove-then-reinsert at front) is only indirectly tested via `TestNoHitLRU_Determinism` (which checks two runs match, not actual score values after re-promotion).

**How to add test** (in `routing_nohit_lru_scorer_test.go`):

```go
func TestNoHitLRU_LRURepromotion(t *testing.T) {
    cacheQueryFn := CacheQueryFn{
        "a": func(tokens []int) int { return 0 },
        "b": func(tokens []int) int { return 0 },
        "c": func(tokens []int) int { return 0 },
    }
    scorer, obs := newNoHitLRUScorer(cacheQueryFn)
    snapshots := []RoutingSnapshot{{ID: "a"}, {ID: "b"}, {ID: "c"}}

    // Route: a, b, a (re-promote a to most-recently-used)
    for _, target := range []string{"a", "b", "a"} {
        req := &Request{ID: "r-" + target, InputTokens: []int{1}}
        scorer(req, snapshots)
        obs(req, target)
    }

    // After [a, b, a]: lruOrder = [a, b]. c is never-used.
    // c=1.0 (never-used, rank 0), b=0.5 (oldest-used, rank 1), a=0.0 (newest-used, rank 2)
    req := &Request{ID: "r-check", InputTokens: []int{1}}
    scores := scorer(req, snapshots)
    if scores["c"] \!= 1.0 {
        t.Errorf("c: got %.3f, want 1.0 (never-used)", scores["c"])
    }
    if scores["b"] \!= 0.5 {
        t.Errorf("b: got %.3f, want 0.5 (oldest-used)", scores["b"])
    }
    if scores["a"] \!= 0.0 {
        t.Errorf("a: got %.3f, want 0.0 (newest-used, re-promoted)", scores["a"])
    }
}
```

sim/routing_precise_prefix_scorer_test.go
```diff
@@ -0,0 +1,102 @@
...
+				count := count // capture
+				cacheQueryFn[id] = func(tokens []int) int { return count }
+			}
+			scorer, _ := newPrecisePrefixCacheScorer(cacheQueryFn)
+			req := &Request{ID: "r1", InputTokens: []int{1}}
+			var snapshots []RoutingSnapshot
+			for id := range tt.counts {
+				snapshots = append(snapshots, RoutingSnapshot{ID: id})
+			}
+			scores := scorer(req, snapshots)
```
sriumcp: **S-4 (Suggestion): Non-deterministic snapshot order in `TestPrecisePrefixCache_AllEqual`**

**Problem:** `for id := range tt.counts` iterates a Go map with non-deterministic order, producing a non-deterministic `snapshots` slice. Since the test asserts all scores are 1.0 (all-equal case), this doesn't affect correctness — but non-deterministic snapshot ordering could mask order-dependent bugs.

**How to fix:** Replace the map iteration with a fixed slice:

```go
ids := []string{"a", "b", "c"}
var snapshots []RoutingSnapshot
for _, id := range ids {
    snapshots = append(snapshots, RoutingSnapshot{ID: id})
}
```

sim/routing_precise_prefix_scorer.go
```diff
@@ -0,0 +1,49 @@
...
+			raw[snap.ID] = count
+			if count < minRaw {
+				minRaw = count
+			}
+			if count > maxRaw {
+				maxRaw = count
+			}
+		}
+		// Pass 2: min-max normalize (higher cached → higher score)
+		for _, snap := range snapshots {
```
sriumcp: **S-5 (Suggestion): All-equal returns 1.0 even when all instances have zero cache hits**

**Problem:** When `maxRaw == minRaw == 0`, the scorer returns 1.0 for all — semantically "perfect cache affinity" despite zero hits. This inflates the cache scorer's contribution to the composite score when no instance has any cached blocks.

This is consistent with `scoreQueueDepth` (same all-equal-returns-1.0 pattern), so not a correctness bug. But a more semantically accurate approach would return 0.5 (neutral) when `maxRaw == 0`:

```go
if maxRaw == minRaw {
    neutral := 1.0
    if maxRaw == 0 {
        neutral = 0.5 // no cache affinity detected
    }
    for _, snap := range snapshots {
        scores[snap.ID] = neutral
    }
}
```

**Note:** This would diverge from the existing `scoreQueueDepth` pattern, so only adopt if the team agrees on the semantic distinction.

sim/routing_scorers_test.go
```diff
@@ -224,17 +224,27 @@ func TestAllScorers_ReturnScoreForEveryInstance(t *testing.T) {
...
 		{"load-balance", scoreLoadBalance},
+		{"precise-prefix-cache", precisePrefixScorer},
+		{"no-hit-lru", noHitLRUScorer},
 	}
+	req := &Request{ID: "r1", InputTokens: []int{1, 2, 3}}
 	for _, sf := range scorerFns {
 		t.Run(sf.name, func(t *testing.T) {
-			scores := sf.fn(nil, snapshots)
+			scores := sf.fn(req, snapshots)
 			// INV-2: score for every instance
```
sriumcp: **S-6 (Suggestion): `TestAllScorers` nil-request coverage regression**

**Problem:** Changed from `sf.fn(nil, snapshots)` to `sf.fn(req, snapshots)` to accommodate new scorers. This means `scoreQueueDepth`, `scoreKVUtilization`, and `scoreLoadBalance` lost nil-request path coverage.

**How to fix:** Add a separate subtest for the nil-request case for the original scorers:

```go
// After the existing loop, add:
for _, sf := range scorerFns[:3] { // original stateless scorers only
    t.Run(sf.name+"/nil-request", func(t *testing.T) {
        scores := sf.fn(nil, snapshots)
        assert.Len(t, scores, len(snapshots))
    })
}
```

sim/routing_nohit_lru_scorer.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	observer := func(req *Request, targetInstance string) {
+		if req == nil {
+			return
+		}
+		// BC-5: use scorer's warm/cold determination (not re-derived).
+		// This avoids disagreement between scorer (checks all instances)
+		// and observer (would only check target instance).
+		if lastWarm || req.ID != lastReqID {
```
sriumcp: **S-7 (Suggestion): Add test for observer with mismatched `req.ID`**

**Problem:** The `req.ID \!= lastReqID` guard on line 103 prevents stale LRU updates. No dedicated test exercises this safety net.

**How to add test** (in `routing_nohit_lru_scorer_test.go`):

```go
func TestNoHitLRU_ObserverMismatchedReqID(t *testing.T) {
    cacheQueryFn := CacheQueryFn{
        "a": func(tokens []int) int { return 0 },
        "b": func(tokens []int) int { return 0 },
    }
    scorer, obs := newNoHitLRUScorer(cacheQueryFn)
    snapshots := []RoutingSnapshot{{ID: "a"}, {ID: "b"}}

    // Score req1 (cold) — sets lastReqID = "r1"
    req1 := &Request{ID: "r1", InputTokens: []int{1}}
    scorer(req1, snapshots)

    // Call observer with DIFFERENT request — should NOT update LRU
    req2 := &Request{ID: "r2", InputTokens: []int{1}}
    obs(req2, "a")

    // Score req3 (cold) — "a" should still be never-used (observer was no-op)
    req3 := &Request{ID: "r3", InputTokens: []int{1}}
    scores := scorer(req3, snapshots)
    if scores["a"] \!= 1.0 {
        t.Errorf("a: got %.3f, want 1.0 (never-used — mismatched observer should be no-op)", scores["a"])
    }
}
```

sim/routing_precise_prefix_scorer.go
```diff
@@ -0,0 +1,56 @@
...
+				// All-equal: 1.0 when there is actual cache affinity, 0.5 (neutral)
+				// when no instance has any cached blocks — avoids inflating the cache
+				// scorer's contribution when there is no cache data to differentiate.
+				if maxRaw == 0 {
+					scores[snap.ID] = 0.5
+				} else {
+					scores[snap.ID] = 1.0
+				}
+			} else {
+				scores[snap.ID] = float64(raw[snap.ID]-minRaw) / float64(maxRaw-minRaw)
```
sriumcp: **BC-2 contract mismatch (doc fix needed)**

The PR description states BC-2 as: *"GIVEN all-equal cached blocks... THEN all instances score 1.0 (all-equal case)"*, but the implementation splits this into two sub-cases:
- all-equal **nonzero** → 1.0
- all-equal **zero** → **0.5** (neutral)

The code behavior is correct and intentional (avoiding inflating scorer contribution when no cache data exists). The test at `TestPrecisePrefixCache_AllEqual` correctly asserts 0.5 for the zero case. But the stated BC-2 in the PR description is wrong — please update it to reflect the split:
> BC-2a: GIVEN all instances have the same **nonzero** cached block count, THEN all score 1.0
> BC-2b: GIVEN all instances have **zero** cached blocks (no cache data to differentiate), THEN all score 0.5 (neutral)

sriumcp: **BC-2 contract mismatch (doc fix needed)**

The PR description states BC-2 as: *"GIVEN all-equal cached blocks... THEN all instances score 1.0 (all-equal case)"*, but the implementation splits this into two sub-cases:
- all-equal **nonzero** → 1.0
- all-equal **zero** → **0.5** (neutral)

The code behavior is correct and intentional (avoiding inflating scorer contribution when no cache data exists). The test at `TestPrecisePrefixCache_AllEqual` correctly asserts 0.5 for the zero case. But the stated BC-2 in the PR description is wrong — please update it to reflect the split:
> BC-2a: GIVEN all instances have the same **nonzero** cached block count, THEN all score 1.0
> BC-2b: GIVEN all instances have **zero** cached blocks (no cache data to differentiate), THEN all score 0.5 (neutral)

PR #931, opened 2026-04-02
fix(cluster): PD TTFT includes KV transfer and first decode step

### LINKED ISSUE #930: PD disaggregated TTFT understates user-visible latency (missing KV transfer + first decode step)
## Problem

In PD disaggregated mode, BLIS reports TTFT as the prefill sub-request's completion time relative to the original arrival time:

```
TTFT_current = prefill_completion_time - arrival_time
```

This captures prefill queuing + prefill compute, but **misses two components** that are on the critical path to the user receiving the first token:

1. **KV cache transfer time** (prefill pod → decode pod)
2. **First decode step time** (decode pod recomputes last prompt token to derive logits, then samples the first output token)

The correct user-visible TTFT, matching llm-d behavior, is:

```
TTFT_correct = prefill_time + KV_transfer_time + first_decode_step_time
             = TransferCompleteTime + first_decode_step - ArrivalTime
```

## How llm-d actually works

Investigation of the [llm-d inference scheduler](https://github.com/llm-d/llm-d-inference-scheduler) and the vLLM NIXL connector reveals the following flow:

1. **Sidecar** sends the prefill request to the prefill pod with `max_tokens=1`
2. **Prefill pod** runs full prefill, samples one token, returns `kv_transfer_params` — the generated token text is **discarded** by the sidecar
3. **Sidecar** sends the original request (full `max_tokens`) to the decode pod with the `kv_transfer_params`
4. **Decode pod** initiates async RDMA KV transfer via NIXL connector
5. After transfer completes, the vLLM scheduler **backs off by 1 token** (`num_computed_tokens = num_tokens - 1` in `scheduler.py:_update_waiting_for_remote_kv`) so the last prompt token is recomputed — this produces the logits needed to sample the first output token
6. **First token** streams to the user from the decode pod

The first token the user sees comes from the **decode pod**, not the prefill pod. The prefill pod's generated token is thrown away.

## Current behavior in BLIS

In `projectPDMetrics()` (`sim/cluster/cluster.go:1181-1193`), the prefill sub-request's TTFT is rekeyed to the parent request ID with the value unchanged:

```go
// TTFT: rekey from prefill sub-request ID to parent ID (value unchanged).
if completed {
    if ttft, ok := m.RequestTTFTs[pfx]; ok {
        m.RequestTTFTs[pid] = ttft
    }
}
```

The decode sub-request never records a TTFT because `AllocateTransferredKV` sets `ProgressIndex = inputLen` before any step runs, so the TTFT check at `simulator.go:621` (`ProgressIndex == len(InputTokens)`) never fires during decode execution.

## Recommended fix

All the necessary data is already tracked. `ParentRequest` has `TransferCompleteTime` and `ArrivalTime`. The decode sub-request's first ITL entry (`decodeSubReq.ITL[0]`) captures the first decode step duration. The fix is contained to `projectPDMetrics`:

```go
// Replace the current TTFT rekeying with:
if completed {
    decodeSubReq := parent.DecodeSubReq
    firstDecodeStep := int64(0)
    if decodeSubReq != nil && len(decodeSubReq.ITL) > 0 {
        firstDecodeStep = decodeSubReq.ITL[0]
    }
    ttft := float64(parent.TransferCompleteTime + firstDecodeStep - parent.ArrivalTime)
    m.RequestTTFTs[pid] = ttft
}
```

This computes: `prefill_queuing + prefill_compute + KV_transfer + first_decode_step`, matching the user-visible TTFT in llm-d.

### What this does NOT include

- **Decode-side queuing delay** (time from decode injection to first scheduled step). This is likely negligible under normal load but could matter at saturation. A more complete fix would track `FirstDecodeStepTime` on `ParentRequest`, but that requires additional plumbing through the completion detection path.

## Impact

- PD TTFT is currently understated, particularly at higher request rates where KV transfer time and decode queuing become significant
- Comparisons between colocated and disaggregated TTFT (e.g., replicating DistServe Figure 1) are misleading because the PD curve is missing real latency components
- `TTFTSum` in aggregated metrics is also affected and should be updated consistently

namasl:
## Summary

- PD disaggregated TTFT was understated: it only captured prefill sub-request completion time (arrival → prefill done)
- In llm-d, the first token reaches the user from the **decode pod**, not prefill: prefill completes → KV transfers → decode pod recomputes last prompt token and samples first output token
- Updated `projectPDMetrics()` to compute user-visible TTFT as `prefillTTFT + transferDuration + firstDecodeStep`
- `TTFTSum` adjusted by the delta to maintain consistency (BC-3)
- Defensive fallback to prefill-only TTFT if decode data is unavailable

Closes #930

## Behavioral Contracts

- **BC-1**: `RequestTTFTs[parentID] = prefillTTFT + (TransferCompleteTime - TransferStartTime) + DecodeSubReq.ITL[0]`
- **BC-2**: New TTFT > prefill-only TTFT (transfer + decode add positive time)
- **BC-3**: `TTFTSum` adjusted by delta to stay consistent with `RequestTTFTs`
- **BC-4**: `TTFT <= E2E` (causality preserved)
- **BC-5**: Non-PD clusters unaffected (`projectPDMetrics` is a no-op when `parentRequests` is empty)

## Test plan

- [x] `TestDisaggregation_TTFT_IncludesTransferAndDecode` — verifies BC-1 (exact formula), BC-2 (exceeds prefill-only), BC-4 (causality)
- [x] Full `go test ./...` passes
- [x] `go vet ./...` clean

### DIFF
--- a/sim/cluster/pd_metrics.go
+++ b/sim/cluster/pd_metrics.go
@@ -111,9 +111,9 @@ func CollectPDMetrics(
 		}
 
 		// BC-1: parent TTFT from the aggregated TTFT map, keyed by parent ID.
-		// After projectPDMetrics(), the prefill sub-request TTFT has been rekeyed
-		// from PrefillSubReqID to the parent ID. The value is unchanged: first_token_time -
-		// arrival_time, recorded when ProgressIndex == inputLen (prefill completion).
+		// After projectPDMetrics(), the user-visible TTFT is stored under the parent ID.
+		// Value = prefillTTFT + KV transfer duration + first decode step — the full
+		// time from arrival to the first token reaching the user (issue #930).
 		// Missing key returns 0.0 in Go maps; exclude 0.0 values (BC-11).
 		if ttft := aggregated.RequestTTFTs[p.ID]; ttft > 0 {
 			ttftValues = append(ttftValues, ttft)


namasl:
@claude /pr-review-toolkit:review-pr does this pr tackle the parent issue correctly? Validate correctness, ease of use, separation of concerns, adherence to our @docs/contributing/standards, invariants, behavioral discipline, behavioral tests and non-structural tests, modularity, clean api boundaries and contracts, performance implications, documentation correctness and completeness.

Issue #1062, opened 2026-04-16
bug(sim): TTFT recording lacks TTFTSet guard - possible double-counting


sriumcp:
In sim/simulator.go:621-626 (executeBatchStep), TTFT metrics are recorded when req.ProgressIndex == len(req.InputTokens). The TTFTSet field is set to true but never used as a guard. If a request lingers at the prefill-complete boundary with NumNewTokens=0 across multiple steps, TTFT could be re-recorded.

Discovered during convergence review of #963 (PC-6 DES Expert perspective). Pre-existing issue.

sriumcp:
## Assessment: SUPERSEDED

This bug was real at the time of filing, but has already been fixed — twice.

### What was wrong

Before PR #1099, `sim/simulator.go` had:

\`\`\`go
if req.ProgressIndex == util.Len64(req.InputTokens) { // prefill complete
    req.TTFTSet = true
    req.FirstTokenTime = now + ...
    sim.Metrics.RequestTTFTs[req.ID] = ...
}
\`\`\`

\`TTFTSet\` was written but never read as a guard. The diagnosis here was exactly correct: if a request in the prefill branch had \`NumNewTokens=0\`, \`ProgressIndex\` is assigned from \`reqNumComputedTokens[req.ID]\`. If that value equals \`InputLen\`, the TTFT block fires. On the next step the same assignment repeats (still \`InputLen\`), and the block fires again — overwriting \`FirstTokenTime\` and re-inserting into \`RequestTTFTs\`.

### What fixed it

| PR | What it fixed |
|----|---------------|
| #1099 | Added \`&& !req.TTFTSet\` to the TTFT condition — directly addresses this issue |
| #1138 | Added \`TTFTSet = false\` on preemption in \`batch_formation.go\` — fixes a secondary issue where preemption left \`TTFTSet=true\`, blocking re-prefill from updating \`FirstTokenTime\` |

Current code at \`sim/simulator.go:680\`:

\`\`\`go
if req.ProgressIndex == util.Len64(req.InputTokens) && !req.TTFTSet {
    req.TTFTSet = true
    req.FirstTokenTime = ...
    sim.Metrics.RequestTTFTs[req.ID] = ...
}
\`\`\`

The guard is present and correct. No remaining work.

Closing as superseded by #1099 and #1138.

Issue #1061, opened 2026-04-16
bug(batch): Phase 2 head-of-line deadlock when front WaitQ request cannot allocate KV blocks


sriumcp:
### Description

When Phase 2 of `FormBatch` (`batch_formation.go:177-179`) fails to allocate KV blocks for the front WaitQ request, it `break`s, leaving the request permanently stuck at the front of the queue. On subsequent steps, the same request is `Peek()`d, the same allocation fails, and the cycle repeats indefinitely.

### Root Cause

```go
// batch_formation.go:177-179
if ok := ctx.KVCache.AllocateKVBlocks(next, startIndex, endIndex, cachedBlocks); \!ok {
    break  // request stays at front of WaitQ, retried every step
}
```

The request is `Peek()`d (line 139) but only `DequeueBatch()`d on success (line 181). On failure, `break` exits Phase 2 but the request remains at the front.

### Evidence

Investigation by @susiejojo in [#963 comment](https://github.com/inference-sim/inference-sim/issues/963#issuecomment-4261128669):
- Experiment 70: 173,622 consecutive allocation failures for request_2759
- Running batch is empty, yet `countFreeBlocks() = 0`
- Only 2,759/19,800 requests complete before deadlock

### vLLM Comparison

vLLM also `break`s in Phase 2 on allocation failure (`scheduler.py:1091-1098`), but **never reaches this state** because:
- Real vLLM traces show preemptions are rare: 7.4% of requests, max 4 per request, largest cascade = 2 requests ([#963 comment](https://github.com/inference-sim/inference-sim/issues/963#issuecomment-4261334853))
- The running batch is always active (GPU execution provides forward progress between scheduler steps)
- Blocks are freed by completions before Phase 2 runs out of capacity

### Two sub-problems

1. **Head-of-line blocking**: When the front WaitQ request can't fit, no other request gets a chance. Fix: skip the stuck request and try the next one (move to back of queue), or implement admission-aware ordering.

2. **Possible block accounting leak**: `countFreeBlocks() = 0` with an empty running batch implies `UsedBlockCnt = TotalBlocks` but no request holds those blocks. This needs investigation — is there a path where `ReleaseKVBlocks` doesn't fully decrement `UsedBlockCnt`?

### Relationship to #963

#963 addresses the I/O amplifier (log storm) and StepTime inflation. This issue tracks the underlying deadlock mechanism that produces the stuck request in the first place.

### Reproduction

[...]

PR #1063, opened 2026-04-16
fix(sim): preemption log storm and StepTime idle request inflation (#963)

### LINKED ISSUE #963: Preemption log storm and near-livelock under KV exhaustion with stale queue-depth signal

When an instance reaches KV cache exhaustion, the simulator enters a near-livelock state where `FormBatch` → `preemptForTokens` → `AllocateKVBlocks` fails repeatedly across consecutive step events. Each failure produces a `logrus.Warnf("KV cache full: cannot allocate %d new blocks for req %s")` log line, generating millions of identical log messages while the simulation clock advances only marginally.

**Mechanism:**

1. Instance KV cache fills completely (requests arrive faster than they complete)
2. Each step event calls `FormBatch` → Phase 1 tries to allocate blocks for a request
3. `preemptForTokens` (`batch_formation.go:201`) calls `AllocateKVBlocks` → fails → evicts tail requests
4. Evicted requests re-enter the wait queue and immediately get re-scheduled
5. Next step event: same request still needs blocks, same failure, same log line at `kv/cache.go:221`
6. Cycle repeats millions of times; simulation technically progresses but takes hours instead of seconds

**This is not a correctness bug** — no invariant is violated, and the simulation eventually terminates. It is a performance pathology: the sim spends wall-clock time logging warnings rather than making progress.

## Reproduction

Single instance, no routing, no custom YAML files needed:

```bash
./blis run \
  --model qwen/qwen3-14b \
  --hardware H100 \
  --tp 1 \
  --num-instances 1 \
  --total-kv-blocks 200 \
  --num-requests 100 \
  --rate 20 \
  --prompt-tokens 512 \
  --output-tokens 128 \
  --seed 42 \
  --log warn
```

**Setup:** 200 KV blocks × 16 tokens/block = 3,200 tokens of KV capacity. Each request needs ~32 blocks (512 prompt tokens), so ~6 concurrent requests saturate the cache. At 20 req/s, requests pile up faster than they can drain.

**Result (on `main` @ fcc3a7a):**
- **14 seconds** of wall time → **1.29M** "KV cache full" warning lines, **144 MB** of stderr output
- Zero simulation progress (no stdout produced)
- Process still running and producing identical log lines when killed

In multi-instance setups, stale routing signals (e.g., `--snapshot-refresh-interval 5000000`) make this easier to trigger by concentrating load onto one instance, but the root cause is entirely within single-instance batch formation.

## Comparison with vLLM

The same preemption retry cycle exists in vLLM's v1 scheduler ([`vllm/v1/core/sched/scheduler.py`](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/sched/scheduler.py)). Under sustained KV exhaustion, vLLM also preempts → re-queues to front of waiting (`prepend_request`) → re-schedules next step → preempts again. However, three differences prevent it from becoming a visible pathology in vLLM:

1. **No log on allocation failure.** vLLM's KV cache manager (`allocate_slots`) returns `None` silently — no warning log. Preemption is recorded as a lightweight event (`record_event`), not a per-attempt log line. BLIS logs at `kv/cache.go:221` on every single failed `AllocateKVBlocks` call, which is the primary cause of the 144 MB I/O storm.

2. **One preemption pass per step.** vLLM gates the waiting-queue phase on `if not preempted_reqs` — if any request was preempted during the running-request phase, the waiting queue is skipped entirely for that step. Preempted requests cannot re-enter until the next scheduling step. In BLIS, `FormBatch` can repeatedly trigger `preemptForTokens` for multiple requests within a single step call.

3. **No swap in v1 either.** vLLM v1 removed the GPU→CPU swap that v0 had, so there is no escape hatch for exhausted GPU KV cache — same as BLIS.

**Conclusion:** The near-livelock in BLIS is ~95% caused by the I/O cost of millions of `logrus.Warn` calls. Fix (1) below (rate-limit the log) would bring BLIS in line with vLLM's behavior under the same conditions.

## Invariants

- [ ] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [x] Work-conserving
- [ ] None (pure refactoring)

INV-8 (work-conserving) is technically maintained (step events keep firing), but the spirit is violated — the simulator spends wall-clock time on futile allocation retries rather than making meaningful progress.

## Proposed fix

Two complementary mitigations:

1. **Rate-limit the "KV cache full" log** in `kv/cache.go:221`. Use `logrus.Warn` once per request per N failures (e.g., log first failure, then every 1000th), or downgrade to `Debug` after first occurrence. This eliminates the I/O bottleneck and matches vLLM's silent-return behavior.

2. **Back-off on futile preemption cycles.** When `preemptForTokens` evicts all batch members for the same request multiple consecutive times (across step events), the request is effectively stuck. Consider: (a) marking the request as deferred/preempt-limited after N consecutive failures, (b) adding a per-request preemption counter that triggers a cooling-off period, or (c) skipping allocation for requests that failed in the previous step and have not seen any new block frees.

**Files affected**

- `sim/kv/cache.go` — log rate limiting for "KV cache full" warning
- `sim/batch_formation.go` — potential preemption back-off logic in `preemptForTokens`
- `sim/simulator.go` — step event scheduling may need awareness of stuck instances

**Design document**

None exists. The fix for (1) is straightforward; (2) may warrant a design doc if it changes preemption semantics.

sriumcp:
## Summary

- Downgrade `AllocateKVBlocks` failure log from `Warn` to `Debug`, matching vLLM's silent `return None` (`kv_cache_manager.py:334-336`)
- Filter idle requests (`NumNewTokens=0`) from `StepTime` computation, matching vLLM's `scheduled_running_reqs` pattern (`scheduler.py:870,1236-1242`)

## Behavioral Contracts

**BC-1: Silent allocation failure**
- GIVEN `AllocateKVBlocks` is called when the KV cache has insufficient free blocks
- WHEN the allocation check fails
- THEN the function returns `false` without producing any `Warn`-level log output

**BC-2: StepTime receives only active requests**
- GIVEN a running batch with mixed active (`NumNewTokens > 0`) and idle (`NumNewTokens = 0`) requests
- WHEN `executeBatchStep` computes step duration
- THEN `StepTime` is called with only the active requests

**BC-3: Idle requests persist in RunningBatch**
- GIVEN idle requests exist in `RunningBatch` after `FormBatch`
- WHEN `executeBatchStep` and `processCompletions` run
- THEN idle requests remain in `RunningBatch` for the next step

**BC-4: Determinism preserved**
- GIVEN any simulation configuration and seed
- WHEN the simulation runs before and after this change
- THEN completion metrics are identical (INV-6)

## Testing

- `TestAllocateKVBlocks_Failure_NoWarnOutput` — captures logrus output, verifies no Warn on allocation failure
- `TestExecuteBatchStep_FiltersIdleRequests_BC2` — spy model verifies only active requests reach `StepTime`
- `TestExecuteBatchStep_IdleRequestsPersist_BC3` — verifies idle requests are not dropped from batch
- Full suite: `go test ./... -count=1` all pass, `golangci-lint run ./...` 0 issues

## Known Limitation

This PR does NOT fix the Phase 2 head-of-line deadlock identified by @susiejojo ([comment](https://github.com/inference-sim/inference-sim/issues/963#issuecomment-4261128669)). That is tracked in #1061.

## Discovered Issues

- #1061 — Phase 2 head-of-line deadlock when front WaitQ request cannot allocate KV blocks
- #1062 — TTFT recording lacks TTFTSet guard (pre-existing, found during convergence review)

Fixes #963

### DIFF
--- a/sim/kv/cache.go
+++ b/sim/kv/cache.go
@@ -218,7 +218,7 @@ func (kvc *KVCacheState) AllocateKVBlocks(req *sim.Request, startIndex int64, en
 
 		// Cannot allocate enough KV cache blocks
 		if numNewBlocks > kvc.countFreeBlocks() {
-			logrus.Warnf("KV cache full: cannot allocate %d new blocks for req %s", numNewBlocks, req.ID)
+			logrus.Debugf("KV cache full: cannot allocate %d new blocks for req %s", numNewBlocks, req.ID)
 			return false
 		}
 	} else {
--- a/sim/kv/cache_test.go
+++ b/sim/kv/cache_test.go
@@ -1,9 +1,12 @@
 package kv
 
 import (
+	"bytes"
 	"fmt"
+	"os"
 	"testing"
 
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -714,6 +717,39 @@ func TestAllocateKVBlocks_ChunkedPrefill_NoPhantomBlocks(t *testing.T) {
 			t.Errorf("block %d has empty Tokens (phantom block)", blk.ID)
 		}
 	}
+}
+
+// BC-1 (#963): AllocateKVBlocks failure must not produce Warn-level output.
+// vLLM proof: kv_cache_manager.py:334-336 returns None silently.
+func TestAllocateKVBlocks_Failure_NoWarnOutput(t *testing.T) {
+	// GIVEN a KV cache with 1 block (16 tokens) that is fully occupied
+	kvc := NewKVCacheState(1, 16)
+	filler := &sim.Request{
+		ID:          "filler",
+		InputTokens: make([]int, 16),
+	}
+	ok := kvc.AllocateKVBlocks(filler, 0, 16, nil)
+	require.True(t, ok, "setup: filler allocation must succeed")
+
+	// Capture log output at Warn level
+	var buf bytes.Buffer
+	logrus.SetOutput(&buf)
+	logrus.SetLevel(logrus.WarnLevel)
+	defer func() {
+		logrus.SetOutput(os.Stderr)
+		logrus.SetLevel(logrus.InfoLevel)
+	}()
+
+	// WHEN a second request tries to allocate and fails
+	victim := &sim.Request{
+		ID:          "victim",
+		InputTokens: make([]int, 16),
+	}
+	ok = kvc.AllocateKVBlocks(victim, 0, 16, nil)
+
+	// THEN allocation fails but no Warn-level output is produced
+	assert.False(t, ok, "allocation must fail (cache full)")
+	assert.Empty(t, buf.String(), "no Warn-level log output expected (BC-1: vLLM returns None silently)")
 }
 
 func TestKVCacheState_SnapshotCachedBlocksFn_FrozenView(t *testing.T) {
--- a/sim/simulator.go
+++ b/sim/simulator.go
@@ -587,8 +587,22 @@ func (sim *Simulator) scheduleBatch(now int64) {
 // executeBatchStep handles Phase 2: model execution (prefill + decode) for all requests
 // in the running batch. Returns the step time advance in ticks.
 func (sim *Simulator) executeBatchStep(now int64) int64 {
-	// Estimate step time via LatencyModel (blackbox or roofline, selected at construction)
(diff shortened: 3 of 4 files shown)


sriumcp:

PR #1092, opened 2026-04-20
fix(sim): prevent TotalOutputTokens double-count after preemption

namasl:
## Summary

Closes #1091

- **Root cause**: `TotalOutputTokens++` was incremented inline during every decode step in `executeBatchStep`. When a request was preempted its `ProgressIndex` was reset to 0, causing all decode tokens from the aborted run to be counted again when the request re-ran.
- **Fix**: Remove the inline `TotalOutputTokens++` and instead compute decode tokens once at completion time in `recordRequestCompletion` using `ProgressIndex - len(InputTokens)`, which equals the exact number of decode tokens generated regardless of preemption history.
- **Test**: `TestSimulator_TotalOutputTokens_NoDoubleCountAfterPreemption` — two requests with distinct token content (to prevent prefix-cache sharing), 4 KV blocks tight enough to force A's eviction after one decode step. Asserts `PreemptionCount > 0` (precondition) and `TotalOutputTokens == 8` (4 per request). Fails with the old code (returns 9), passes with the fix.

## Test plan

- [x] New failing test `TestSimulator_TotalOutputTokens_NoDoubleCountAfterPreemption` added and confirmed to fail before fix (returns 9, wants 8)
- [x] Fix applied; test now passes
- [x] `go test ./...` passes with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1091, opened 2026-04-20
Bug: TotalOutputTokens double-counts decode tokens after preemption


namasl:
**Describe the bug**
`TotalOutputTokens` is incremented on every decode step (`sim/simulator.go:647`), but when a request is preempted (`sim/batch_formation.go:232`), its `ProgressIndex` is reset to 0 and all KV blocks are released. When the request is rescheduled, it re-enters prefill and then re-enters decode from scratch — and every re-generated output token is counted again. This inflates the throughput numerator (`tokens_per_sec = TotalOutputTokens / SimEndedTime`).

**CLI command to reproduce**
```bash
# Run a workload that triggers preemption (high concurrency + long output)
./blis run --model meta-llama/Llama-3.1-8B-Instruct --hardware H100 --tp 1 \
  --latency-model trained-physics --concurrency 256 --num-requests 1024 \
  --prompt-tokens 4096 --output-tokens 2048 --seed 42 \
  --max-num-seqs 1024 --max-num-scheduled-tokens 2048 \
  --long-prefill-token-threshold 2048
```

Observe `total_output_tokens` in the JSON metrics output. Compare against the theoretical maximum: `completed_requests * output_tokens`. The reported value exceeds the maximum.

**Expected behavior**
`TotalOutputTokens` should count each request's output tokens exactly once, regardless of how many times the request was preempted and rescheduled. For a workload with fixed OSL=2048 and N completed requests, `TotalOutputTokens` should equal exactly `N * 2048`.

This is a violation of request conservation (INV-1) in the token accounting dimension: the system reports more output tokens than were logically produced.

**Actual behavior**
At concurrency=256, long-long (ISL=4096, OSL=2048): `TotalOutputTokens` = 2,402,782 but the theoretical maximum (completed × OSL) = 2,097,152. The 305,630 excess tokens (14.6%) are ghost tokens from preemption re-counting.

At concurrency=512: `TotalOutputTokens` = 4,336,751 vs theoretical max = 4,194,304 (3.4% excess).

The effect compounds with `SimEndedTime` compression at high concurrency, producing wildly inflated `tokens_per_sec` (e.g., 14,087 tok/s simulated vs 2,497 tok/s real for long-long at c=512 — a 5.6x overestimate).

**Which invariant is violated?**
- [x] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
[...]

PR #1098, opened 2026-04-21
fix(sim): prevent ITL and TTFTSum inflation after preemption

### LINKED ISSUE #1095: Bug: ITLSum and TTFTSum double-count metrics after preemption (same class as #1091)
**Describe the bug**

Two aggregate metrics in the simulator accumulate phantom values when a request is preempted and re-runs — the same root cause as #1091 (`TotalOutputTokens` double-count, fixed in #1092). Both follow the inline-accumulation pattern: a metric is incremented during execution rather than computed once at completion, so preemption-driven re-runs count the aborted work a second time.

**Bug 1: `req.ITL` not cleared on preemption**

`req.ITL` is appended to at every decode step (`sim/simulator.go:660`) but is never reset in `preemptForTokens` (`sim/batch_formation.go:231-234`). After a preempted request re-prefills and re-decodes, its ITL slice contains entries from both the aborted run and the successful re-run. At completion, `recordRequestCompletion` sums the entire slice:

```go
// sim/simulator.go:504-507
var itlSum int64
for _, v := range req.ITL {
    itlSum += v
}
lat := req.FirstTokenTime + itlSum + postDecodeOverhead
```

This inflates:
- `RequestE2Es[req.ID]` and `RequestCompletionTimes[req.ID]` — via `lat`, which includes phantom pre-eviction timings
- `RequestITLs[req.ID]` (TPOT) — numerator (`itlSum`) inflated; denominator (`len(req.OutputTokens)-1`) correct for non-length-capped requests, so TPOT is over-estimated
- `AllITLs` — phantom ITL entries from the aborted run are appended at line 536

**Bug 2: `TTFTSum` accumulated inline**

`TTFTSum` is incremented each time `ProgressIndex == len(InputTokens)` (`sim/simulator.go:663-666`):

```go
if req.ProgressIndex == util.Len64(req.InputTokens) {
    req.TTFTSet = true
    req.FirstTokenTime = now + currStepAdvance + ... - req.ArrivalTime
    sim.Metrics.TTFTSum += req.FirstTokenTime  // fires again after re-prefill
    sim.Metrics.RequestTTFTs[req.ID] = float64(req.FirstTokenTime)
}
```

After preemption, `ProgressIndex` resets to 0. When the request re-prefills and again reaches `ProgressIndex == len(InputTokens)`, this block fires a second time — adding the re-run's TTFT to `TTFTSum` without removing the pre-eviction TTFT that was already added. `RequestTTFTs[req.ID]` is correct (it overwrites), but `TTFTSum` double-counts.

Note: `req.TTFTSet` is also not reset on preemption, so it cannot be used as a re-entry guard without a companion reset in `preemptForTokens`.

**CLI command to reproduce**

```bash
# High concurrency + long output triggers preemptions; compare TTFTSum/CompletedRequests
# against p50 TTFT and inspect AllITLs distribution
./blis run --model meta-llama/Llama-3.1-8B-Instruct --hardware H100 --tp 1 \
  --latency-model trained-physics --concurrency 256 --num-requests 1024 \
  --prompt-tokens 4096 --output-tokens 2048 --seed 42 \
  --max-num-seqs 1024 --max-num-scheduled-tokens 2048 \
  --long-prefill-token-threshold 2048
```

**Expected behavior**

- `TTFTSum` should equal `sum(RequestTTFTs.values())` for all completed requests — each request's TTFT counted exactly once.
- `AllITLs` should contain exactly `sum(len(req.OutputTokens) - 1)` entries across all completed requests.
- `RequestE2Es` should not include timing from decode steps that were discarded by preemption.

**Actual behavior**

For any workload with preemptions:
- `TTFTSum / CompletedRequests` (mean TTFT) is inflated relative to the median of `RequestTTFTs`.
- `AllITLs` contains extra entries; the ITL distribution is skewed by phantom decode step timings.
- Per-request E2E latencies for preempted requests are inflated by the sum of their pre-eviction decode step times.

- [ ] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [ ] Work-conserving
- [x] None / unknown (metric correctness — analogous to the token conservation dimension of INV-1)

**Possible fixes**

**Bug 1 (`req.ITL`):** Add `preemptedRequest.ITL = nil` in `preemptForTokens` alongside the existing `ProgressIndex = 0` reset (line 232). This mirrors the `TotalOutputTokens` fix: discard any work from the aborted run.

**Bug 2 (`TTFTSum`):** Move accumulation to `recordRequestCompletion` using the already-correct `req.FirstTokenTime`:
```go
sim.Metrics.TTFTSum += req.FirstTokenTime
```
Remove the inline `TTFTSum +=` from `executeBatchStep`. This is the same pattern as moving `TotalOutputTokens` to completion time.

**Environment**
- OS: Linux (Fedora 42)
- Go version: 1.22
- Commit hash: dac52ce

**Additional context**
Discovered during code review of #1092 (fix for `TotalOutputTokens` double-count after preemption). All three bugs — `TotalOutputTokens`, `req.ITL`, and `TTFTSum` — share the same inline-accumulation pattern and are only observable at high concurrency where KV cache pressure causes preemptions.

namasl:
## Summary

Fixes two metric accumulation bugs that inflate `AllITLs`, `E2E` latency, `TPOT`, and `TTFTSum` when requests are preempted and re-run. Both share the same root cause as #1091 (`TotalOutputTokens` double-count, fixed in #1092): metrics were incremented inline during execution rather than computed once at completion.

- **Bug 1 (`req.ITL`):** The ITL slice retained entries from aborted runs. After re-prefill and re-decode, new entries appended to the stale slice, inflating `AllITLs`, per-request E2E, and TPOT. Fix: `preemptedRequest.ITL = nil` in the preemption reset block alongside the existing `ProgressIndex = 0` reset.

- **Bug 2 (`TTFTSum`):** Accumulated inline at prefill completion. After preemption resets `ProgressIndex` to 0, re-prefill triggered the accumulation a second time. Fix: move `TTFTSum +=` to `recordRequestCompletion` (same pattern as `TotalOutputTokens` in d8271ee).

Closes #1095

## Test plan

- [ ] `TestSimulator_ITL_NoDuplicateEntriesAfterPreemption` — verifies `len(AllITLs) == sum(OutputLen - 1)` across completed requests; fails without the ITL fix
- [ ] `TestSimulator_TTFTSum_NoDoubleCountAfterPreemption` — verifies `TTFTSum == sum(RequestTTFTs.values())`; fails without the TTFTSum fix
- [ ] `TestSimulator_TotalOutputTokens_NoDoubleCountAfterPreemption` — prior regression test still passes
- [ ] `go test ./...` — full suite green
- [ ] `go vet ./...` — clean


namasl:

sriumcp:
## Code-proof review: PR is correct

Checked the three changes in the diff against the codebase. Here is the reasoning for each, line by line.

---

### Fix 1 — `preemptedRequest.ITL = nil` (`batch_formation.go`)

**The diff adds one line after `ProgressIndex = 0`:**

```go
preemptedRequest.State = StateQueued
preemptedRequest.ProgressIndex = 0
preemptedRequest.ITL = nil          // ← new
ctx.KVCache.ReleaseKVBlocks(preemptedRequest)
```

**Why this is the right fix:** At completion, `recordRequestCompletion` (`simulator.go:504`) sums `req.ITL` and appends it to `AllITLs`. With the nil reset, the re-run starts from an empty slice, so only re-run decode steps are counted. Without it, stale entries from the aborted run were included. The placement — alongside the existing `ProgressIndex = 0` reset — is exactly right. ✅

---

### Fix 2 — `TTFTSum` moved from `executeBatchStep` to `recordRequestCompletion` (`simulator.go`)

**The diff removes one line:**
```go
// executeBatchStep (line ~666) — REMOVED:
- sim.Metrics.TTFTSum += req.FirstTokenTime
```

**And adds it to the top of `recordRequestCompletion`:**
```go
// recordRequestCompletion — ADDED:
sim.Metrics.CompletedRequests++
+ sim.Metrics.TTFTSum += req.FirstTokenTime
```

**Why this is correct:** `req.FirstTokenTime` and `RequestTTFTs[req.ID]` are still set inline in `executeBatchStep` (scalar/overwrite, safe through preemption). Only `TTFTSum` — the running aggregate — is moved. At the point `recordRequestCompletion` reads `req.FirstTokenTime`, the value is already final: `executeBatchStep` runs first, `processCompletions` → `recordRequestCompletion` runs second (BC-5 ordering documented in the `processCompletions` comment). So `FirstTokenTime` is always written before it is read. ✅

**Zero-output-token requests are safe:** For prefill-only requests, `ProgressIndex` reaches `len(InputTokens)` during the same step as completion. `executeBatchStep` sets `FirstTokenTime`, then `processCompletions` calls `recordRequestCompletion` in the same `Step()` call — correct ordering preserved. ✅

**`TTFTSet` is unaffected:** The field is still set to `true` inline (only `TTFTSum +=` was removed), so `tracev2.go:394` and other consumers of `TTFTSet` are not disturbed. ✅

---

### Tests

Both new tests follow the same structure as `TestSimulator_TotalOutputTokens_NoDoubleCountAfterPreemption` (the #1092 regression test) and include a critical precondition guard:

```go
if s.Metrics.PreemptionCount == 0 {
    t.Fatal("precondition violated: expected at least one preemption — scenario does not test the bug")
}
```

This ensures the test fails loudly if the scenario configuration ever changes and preemption stops occurring, rather than silently passing without testing anything.

`PreemptionCount` is a real field (`sim/metrics.go:25`, incremented at `simulator.go:598`). `sort` is already imported in `simulator_test.go`. Both compile cleanly. ✅

**`TestSimulator_ITL_NoDuplicateEntriesAfterPreemption`** asserts:
```go
wantITLCount := (len(bOutput) - 1) + (len(aOutput) - 1)  // exactly one entry per decode step
gotITLCount  := len(s.Metrics.AllITLs)
```
This is the right invariant: each request contributes exactly `OutputLen - 1` ITL entries (TPOT convention: first generated token excluded). ✅

**`TestSimulator_TTFTSum_NoDoubleCountAfterPreemption`** asserts:
```go
TTFTSum == sum(RequestTTFTs.values())
```
It correctly uses sorted map iteration before summing to avoid non-deterministic float accumulation. ✅

---

### Verdict

Both fixes are **correct and complete** for the reported bugs. No other fields need resetting in `preemptForTokens` to fix these two issues. The tests will fail on the old code and pass on the new code.

sim/batch_formation.go
```diff
@@ -230,6 +230,7 @@ func (v *VLLMBatchFormation) preemptForTokens(req *Request, numNewTokens int64,
 
 			preemptedRequest.State = StateQueued
 			preemptedRequest.ProgressIndex = 0
+			preemptedRequest.ITL = nil
```
susiejojo, line 233: QQ: If I understand correctly, since TTFTSum is only reported in metrics and not used for downstream calculations elsewhere, only the reported ITL metrics were incorrect earlier in the event of preemptions, and this fixes that, right?

susiejojo (changes requested):
Can we add the TTFTSum metric in all the golden datasets too to avoid unwanted changes to this calculation in the future? Calculation and logic lgtm, verified manually.

susiejojo:
I think the current golden dataset tests include experiments that deliberately don't trigger any preemptions at all. It might also be very useful to add at least 1 test case with preemptions that can catch these bugs.

sim/batch_formation.go
namasl: Not quite — Bug 1 (ITL not cleared) inflated **E2E latency** too, not just the ITL metrics. At completion, `lat = FirstTokenTime + itlSum + postDecodeOverhead`, and `itlSum` was computed from the stale ITL slice including pre-preemption entries. So `RequestE2Es` and `RequestCompletionTimes` were both inflated for any preempted request.

You're right that `TTFTSum` (Bug 2) was reporting-only — it feeds the mean TTFT output but isn't used for per-request E2E or downstream routing decisions.

Summary:
- **Bug 1 (ITL slice):** affected AllITLs, TPOT, and **E2E / completion times**
- **Bug 2 (TTFTSum):** affected only the aggregate TTFTSum (mean TTFT output)

namasl:
Done — added `ttft_sum_us` (int64, in µs) to `goldenExpected` and regenerated both `roofline_goldendataset.json` and `trained_physics_iter29.json` with the new field (44d9536). Any future change to TTFTSum accumulation will now fail the golden tests.

Stored as int64 µs (not float64 ms) for exact comparison — consistent with the other integer golden fields (`completed_requests`, `total_output_tokens`).

sim/batch_formation.go
susiejojo: Thanks, I think we need test cases in the golden dataset with moderate/heavy preemptions to avoid future changes to these metrics (E2E, TTFTSum, ITL etc).

susiejojo (approved):
LGTM, I opened an issue for extending golden datasets with preemption cases: https://github.com/inference-sim/inference-sim/issues/1101

sriumcp (approved):
lgtm

PR #1100, opened 2026-04-21
fix(sim): prevent orphaned StepEvent cascade from TimeoutEvent batch-empty path

### LINKED ISSUE #1096: Bug: TimeoutEvent cascade creates orphaned StepEvents, inflating throughput at high concurrency
**Describe the bug**

When many requests share the same timeout deadline (e.g., all concurrency-mode seeds arriving at tick 0 with default 300s timeout), a cascade interaction between `TimeoutEvent.Execute` and the INV-8 work-conserving guard causes hundreds of orphaned `StepEvent`s to accumulate permanently in the event heap. These orphaned events fire at stale timestamps, causing the simulation clock to advance in tiny increments (~10-100µs) instead of the computed step time (~39ms), which collapses `SimEndedTime` and inflates `tokens_per_sec` by up to ~33x.

**CLI command to reproduce**

```bash
# long-long workload (ISL=4096, OSL=2048) at c=512 triggers the bug
./blis run \
  --model meta-llama/Llama-3.1-8B-Instruct \
  --model-config-folder model_configs/meta-llama-3.1-8b \
  --hardware H100 --tp 1 \
  --max-num-running-reqs 1024 \
  --max-num-scheduled-tokens 2048 \
  --long-prefill-token-threshold 2048 \
  --prompt-tokens 4096 --prompt-tokens-stdev 0 \
  --output-tokens 2048 --output-tokens-stdev 0 \
  --concurrency 512 --num-requests 10240 \
  --seed 42

# Compare with short-short (ISL=128, OSL=128) which does NOT trigger:
./blis run \
  --model meta-llama/Llama-3.1-8B-Instruct \
  --model-config-folder model_configs/meta-llama-3.1-8b \
  --hardware H100 --tp 1 \
  --max-num-running-reqs 1024 \
  --max-num-scheduled-tokens 2048 \
  --long-prefill-token-threshold 2048 \
  --prompt-tokens 128 --prompt-tokens-stdev 0 \
  --output-tokens 128 --output-tokens-stdev 0 \
  --concurrency 512 --num-requests 10240 \
  --seed 42
```

**Expected behavior**

- Step events should fire at intervals matching the computed step time (~39ms for a batch of ~70-80 decode requests with Llama-3.1-8B on H100)
- `SimEndedTime` for the long-long c=512 case should be ~11,600s (292K steps × 39.6ms)
- `tokens_per_sec` should be ~1,700 tok/s (consistent with the c=256 measurement)

**Actual behavior**

- 226 orphaned `StepEvent`s accumulate in the event heap starting at the 300s timeout mark
- Steps fire at ~10-100µs intervals instead of ~39ms — **73% of step deltas are under 100µs**
- `SimEndedTime` collapses to 350s (vs expected ~11,600s)
- `tokens_per_sec` = 58,146 (vs expected ~1,700) — a **33x inflation**
- Step count (292,798) and per-step latency (~39ms) are computed correctly — the bug is in event scheduling, not the latency model

Metrics comparison across concurrency levels (long-long workload):

| Concurrency | SimEndedTime | Avg step time | tokens_per_sec | Bug? |
|---|---|---|---|---|
| 64 | 2,084s | 31.7ms | 1,964 | No |
| 256 | 5,910s | 39.6ms | 1,720 | No |
| 512 | **350s** | **1.2ms** | **58,146** | **Yes** |

**Root cause**

The bug is a cascade interaction between `TimeoutEvent.Execute` (`event.go:138-192`) and StepEvent scheduling at the same tick. Here is the exact mechanism:

**Setup**: At c=512 with ISL=4096/OSL=2048, only ~68 requests fit in KV cache. The remaining ~444 sit in WaitQ. All 512 seed requests share the same 300s deadline (tick 300,000,000).

**Cascade** (repeats ~434 times, once per queued seed):

1. A `TimeoutEvent` for a running request fires, removing it from `RunningBatch`. When this empties the batch, `event.go:169-172` sets `sim.RunningBatch = nil` and `sim.stepEvent = nil` — **orphaning** the legitimate StepEvent already scheduled in the heap at tick ~300,012,075.

2. The INV-8 guard (`event.go:179-184`) sees `sim.stepEvent == nil && WaitQ.Len() > 0` and creates a new `StepEvent` at tick 300,000,000 (the current tick).

3. **Priority inversion**: `StepEvent` has priority 2, `TimeoutEvent` has priority 5. At the same tick, StepEvents fire first. So this new StepEvent fires immediately, before the remaining TimeoutEvents.

4. The Step runs `FormBatch`, pulling one queued seed from WaitQ into RunningBatch. `scheduleNextStep` creates a new StepEvent at 300,012,075. The step sets `sim.stepEvent` to this new event.

5. The next `TimeoutEvent` fires for the **same request that was just admitted** (it's an original seed with deadline 300,000,000). This times it out, empties RunningBatch again, sets `sim.stepEvent = nil` (orphaning the StepEvent at 300,012,075), and the cycle repeats from step 2.

After all timeouts are processed, **~226 orphaned StepEvents** remain at tick ~300,012,075 in the heap. These fire before the legitimate "next step" StepEvents (which are scheduled ~39ms in the future), causing the simulation to process the running batch hundreds of times more frequently than intended.

**Instrumentation evidence:**

```
# Step 7623: last normal step (heap clean)
Step(299969022): entering step 7623, RunningBatch=78, WaitQ=434, StepEventsInHeap=0

# Step 7624: first timeout-triggered step (cascade begins)
StepEvent firing at tick 300000000, creator=TimeoutINV8
Step(300000000): entering step 7624, RunningBatch=0, WaitQ=434, StepEventsInHeap=1

# Steps 7625-7850: cascade continues, heap grows by 1 per step
StepEvent firing at tick 300000000, creator=TimeoutINV8
Step(300000000): entering step 7625, RunningBatch=0, WaitQ=433, StepEventsInHeap=2
...
Step(300000000): entering step 7851, RunningBatch=1, WaitQ=208, StepEventsInHeap=226

# Steady state: 226 orphaned StepEvents remain permanently
Step(316120505): entering step 100001, RunningBatch=71, WaitQ=137, StepEventsInHeap=226
```

**Trigger conditions**:
1. Many simultaneous timeouts (concurrency mode where all seeds share the same arrival/deadline)
2. Long-running requests still in-flight at the 300s timeout (ISL+OSL large enough)
3. Non-empty WaitQ at timeout time (more seeds than KV-cache capacity)

Short-short (ISL=128, OSL=128) doesn't trigger because requests complete in ~2.2s, well before the 300s timeout.

**Fix direction**

The root problem is `sim.stepEvent = nil` on `event.go:171` — when a TimeoutEvent empties RunningBatch, it orphans the legitimate pending StepEvent. Three possible fixes:

**Option A — Lazy cancellation flag on StepEvent**: Add a `cancelled bool` field to `StepEvent`. When `scheduleNextStep` or the INV-8 guard overwrites `sim.stepEvent`, set `cancelled = true` on the old event. In `StepEvent.Execute`, check `cancelled` and skip if true. This prevents orphaned events from firing.

**Option B — Strengthen the orphaned-StepEvent guard**: Change `Step()`'s guard from `RunningBatch == nil && WaitQ.Len() == 0` to also check whether the firing StepEvent matches `sim.stepEvent`. If it doesn't match, it's orphaned → skip.

**Option C — Don't nil stepEvent in TimeoutEvent**: Instead of setting `sim.stepEvent = nil` when emptying RunningBatch (line 171), leave it pointing to the already-scheduled StepEvent. The StepEvent will fire at its scheduled time and find an empty RunningBatch, which the existing orphaned-StepEvent guard handles. Remove the INV-8 StepEvent creation from TimeoutEvent entirely, since `scheduleNextStep`'s work-conserving path already handles the "batch drained + WaitQ non-empty" case.

- [x] Work-conserving (INV-8 — the guard intended to preserve this invariant is the trigger)
- [x] Request lifecycle (requests are admitted and immediately timed out in a tight loop)
- [x] Causality (steps fire at stale timestamps, not at the computed future time)

**Environment**
- OS: Linux 6.18.5 (Fedora 42)
- Go version: 1.25.6
- Commit hash: d8271ee (branch: fix/preemption-double-counts-output-tokens)

**Additional context**

Discovered during investigation of throughput deviation at high concurrency values in `sweeps.ipynb` "Real benchmarks" section. The bug only manifests when concurrency-mode workloads have requests that don't complete before their timeout deadline. The shorter workloads (short-short, long-short) showed inflated but not catastrophically wrong throughput because the bug didn't trigger (no timeouts) — their deviation from real benchmarks is a separate latency-model accuracy issue.

Related: #1095 (ITL/TTFTSum preemption bugs, same PR branch but independent issue).

namasl:
## Summary

Fixes a DES event-ordering bug where simultaneous `TimeoutEvent`s for running requests cascade into N orphaned `StepEvent`s, collapsing `SimEndedTime` and inflating `tokens_per_sec` by up to ~33x at high concurrency.

**Root cause:** When `TimeoutEvent.Execute` emptied `RunningBatch`, it nil'd `sim.stepEvent`, orphaning the legitimate `StepEvent` already in the heap. The INV-8 work-conserving guard then created a *new* `StepEvent` at the current tick. Because `StepEvent` (priority 2) fires before `TimeoutEvent` (priority 5) at the same tick, this new event fired before the remaining `TimeoutEvent`s — pulling a queued request that was immediately timed out by its own deadline. Each cascade iteration left one orphaned `StepEvent` at the next legitimate step tick. These orphans fired in lock-step with subsequent legitimate `StepEvent`s, processing the batch N+1 times per tick and collapsing `SimEndedTime`.

**Fix:** Do not nil `sim.stepEvent` when `TimeoutEvent` empties `RunningBatch`. The already-scheduled `StepEvent` fires at the correct future time; the existing `Step()` orphan guard handles the empty-batch case. Also clear `sim.stepEvent` in the orphan guard to prevent a stale non-nil pointer from blocking future `QueuedEvent` INV-8 scheduling when all WaitQ requests time out before the orphaned `StepEvent` fires.

Closes #1096

## Behavioral Contracts

- **BC-1:** GIVEN simultaneous `TimeoutEvent`s at the same deadline tick for one running request and N queued requests, WHEN `TimeoutEvent` empties `RunningBatch`, THEN no new `StepEvent` is created at the current tick (the INV-8 guard does not fire because `sim.stepEvent` remains non-nil).
- **BC-2:** GIVEN an orphaned `StepEvent` that fires and finds `RunningBatch == nil` and `WaitQ.Len() == 0`, WHEN the `Step()` orphan guard returns early, THEN `sim.stepEvent` is cleared to nil so future `QueuedEvent` INV-8 guards can schedule normally.

## Antipattern Checklist

**Critical (correctness):**
- [x] No silent data loss — every error path returns error, panics, or increments counter (R1) — N/A: no new error paths
- [x] Struct construction sites audited for new fields (R4) — N/A: no new struct fields
- [x] Resource allocation loops rollback on mid-loop failure (R5) — N/A: no resource allocation changes
- [x] No `logrus.Fatalf` or `os.Exit` in `sim/` packages (R6) — N/A: no new log calls
- [x] Division by runtime-derived denominators guarded (R11) — N/A: no division
- [x] Unbounded retry/requeue loops have circuit breakers (R19) — N/A: no loops
- [x] No `range` over slices that can shrink during iteration (R21) — N/A: no range changes

**Important (quality):**
- [x] Map keys sorted before float accumulation or ordered output (R2) — N/A
- [x] Every new numeric parameter validated — CLI flags AND library constructors (R3) — N/A: no new parameters
- [x] Invariant tests alongside golden tests (R7) — new test asserts INV-1 conservation and SimEndedTime bounds
- [x] No exported mutable maps (R8) — N/A
- [x] `*float64` for YAML fields where zero is valid (R9) — N/A
- [x] YAML strict parsing with `KnownFields(true)` (R10) — N/A
- [x] New interfaces work for 2+ implementations (R13) — N/A: no interfaces
- [x] No method spans multiple module responsibilities (R14) — N/A
- [x] Routing scorer signals documented for freshness tier (R17) — N/A
- [x] CLI flag values not silently overwritten by defaults.yaml (R18) — N/A
- [x] Detectors and analyzers handle degenerate inputs (R20) — N/A
- [x] Pre-check estimates consistent with actual operation (R22) — N/A
- [x] Parallel code paths apply equivalent transformations (R23) — N/A

**Hygiene (maintenance):**
- [x] Golden dataset regenerated if output changed (R12) — N/A: no output format changes
- [x] Stale PR references resolved (R15) — N/A
- [x] Config params grouped by module (R16) — N/A

## Invariants

- [x] Request conservation: injected == completed + queued + running (INV-1) — verified by new test
- [x] Request lifecycle: queued -> running -> completed (INV-2) — unchanged
- [x] KV block conservation: allocated + free == total (INV-4) — unchanged
- [x] Causality: arrival <= schedule <= completion (INV-5) — unchanged
- [x] Clock monotonicity (INV-3) — unchanged
- [x] Work-conserving (INV-8) — fix prevents false-positive INV-8 triggers that caused the cascade; stale-pointer clear ensures INV-8 remains functional for future arrivals

## Test Plan

New test: `TestTimeout_CascadeDoesNotCreateOrphanedStepEvents` in `sim/timeout_test.go`

Scenario: 6 requests with a shared 50ms deadline + 2 requests with no deadline; KV cache constrained to force queueing. The deadline requests time out and trigger the cascade. The surviving requests are processed by the remaining `StepEvent`s.

Without the fix: `SimEndedTime` collapses to ~100ms (orphans process the batch faster than real step timing allows). With the fix: `SimEndedTime` is ~140ms (correct single-step-per-tick progression).

```bash
go test ./sim/... -run TestTimeout_CascadeDoesNotCreateOrphanedStepEvents -v
go test ./... -count=1
go vet ./...
```

All pass.

## Source

Fixes #1096

### DIFF
--- a/sim/event.go
+++ b/sim/event.go
@@ -163,12 +163,17 @@ func (e *TimeoutEvent) Execute(sim *Simulator) {
 			}
 		}
 		sim.RunningBatch.Requests = remaining
-		// If batch is now empty, clear state so the stale stepEvent doesn't
-		// fire Step() on an empty batch (which could crash in StepTime).
-		// This also enables the INV-8 guard below to reschedule if needed.
+		// If batch is now empty, clear RunningBatch. Do NOT nil sim.stepEvent:
+		// leaving it pointing to the already-scheduled StepEvent prevents the
+		// INV-8 guard below from creating a duplicate StepEvent at the current
+		// tick. At the same tick, StepEvents fire before TimeoutEvents (priority
+		// 2 < 5), so a newly-created StepEvent would pull a queued request and
+		// then its same-deadline TimeoutEvent would immediately time it out —
+		// cascading once per queued seed and leaving N orphaned StepEvents.
+		// The already-scheduled StepEvent fires at the correct future time and
+		// the Step() orphan guard handles the empty-batch case.
 		if len(remaining) == 0 {
 			sim.RunningBatch = nil
-			sim.stepEvent = nil
 		}
 	} else {
 		sim.WaitQ.Remove(e.Request)
--- a/sim/simulator.go
+++ b/sim/simulator.go
@@ -539,13 +539,15 @@ func (sim *Simulator) recordRequestCompletion(req *Request) {
 // Step simulates a single vllm step(): batch scheduling, model execution, mirroring, and completion.
 // Phases: (1) schedule batch, (2) execute prefill/decode, (2.5) mirror to CPU, (3) process completions, (4) schedule next step.
 //
-// Orphaned StepEvent guard: when a TimeoutEvent empties the RunningBatch and nils stepEvent,
-// a previously-scheduled StepEvent may still be in the heap. If the INV-8 guard also scheduled
-// a new StepEvent, two StepEvents fire at the same tick. The second finds RunningBatch nil and
-// WaitQ empty (already processed by the first). This guard prevents the phantom double-step.
+// Orphaned StepEvent guard: when a TimeoutEvent empties the RunningBatch it leaves
+// sim.stepEvent pointing to the already-scheduled StepEvent (preventing the cascade
+// described in #1096). If that StepEvent fires and finds nothing to do, clearing
+// sim.stepEvent here prevents future QueuedEvent INV-8 guards from seeing a stale
+// non-nil pointer and skipping their step-scheduling.
 func (sim *Simulator) Step(now int64) {
 	if sim.RunningBatch == nil && sim.WaitQ.Len() == 0 {
-		return // orphaned StepEvent — nothing to process
+		sim.stepEvent = nil
+		return
 	}
 	sim.scheduleBatch(now)
 	currStepAdvance := sim.executeBatchStep(now)
--- a/sim/timeout_test.go
+++ b/sim/timeout_test.go
@@ -402,3 +402,107 @@ func TestTimeout_OrphanedTimeout_MultipleOrphans_NoneInflateClock(t *testing.T) 
 		t.Errorf("SimEndedTime inflated: got %d µs, want < 500_000 µs", sim.Metrics.SimEndedTime)
 	}
 }
+
+// TestTimeout_CascadeDoesNotCreateOrphanedStepEvents verifies that simultaneous
+// TimeoutEvents for running requests do not trigger the INV-8 guard and cascade
+// into orphaned StepEvents (issue #1096).
+//
+// Without the fix, TimeoutEvent.Execute nil'd sim.stepEvent when emptying
+// RunningBatch. The INV-8 guard then created a new StepEvent at the current tick,
+// which (priority 2 < timeout priority 5) fired before remaining TimeoutEvents,
+// pulling a queued seed that was immediately timed out. Each iteration left an
+// orphaned StepEvent at the scheduled future tick. These orphans fired in lock-step
+// with subsequent legitimate StepEvents, processing the batch N+1 times per tick
+// and collapsing SimEndedTime by ~2x.
+//
+// GIVEN: 8 requests sharing the same arrival time; 6 with a short deadline (50ms),
+//
+//	2 with no deadline. KV cache: 4 blocks × 16 tokens, each request needs 2
+//	blocks (InputLen=16 fills block0; first decode token at PI=16 needs block1).
+//	This fits exactly 2 requests, leaving 6 to queue.
+//
+// WHEN: The simulation runs to completion.
+//
+// THEN: The 6 deadline requests time out, the 2 no-deadline requests complete
+//
+//	normally, and SimEndedTime reflects realistic per-step latency (~130ms),
+//	not collapsed cascade timing (~70ms).
+func TestTimeout_CascadeDoesNotCreateOrphanedStepEvents(t *testing.T) {
+	const (
+		deadlineTicks = int64(50_000) // 50ms — requests will not complete in time
+		stepTimeTicks = int64(10_000) // 10ms per step (alpha[0]=10000)
+		inputLen      = 16            // tokens per request — fills exactly 1 KV block (positions 0-15)
+		outputLen     = 8             // needs 1 decode block (positions 16-23); total 2 blocks
+		numDeadline   = 6             // requests that will time out
+		numSurviving  = 2             // requests with no deadline (will complete)
+		numRequests   = numDeadline + numSurviving
+	)
+	// KV cache: 4 blocks × 16 tokens.
+	// Each request needs 2 blocks (input block + first decode block), so 2 fit simultaneously.
+	// r0–r5 carry deadline=50ms; r6–r7 carry no deadline (Deadline==0 → no TimeoutEvent).
+	cfg := SimConfig{
+		Horizon:             500_000, // 500ms — well beyond the ~130ms expected completion
+		Seed:                42,
+		KVCacheConfig:       NewKVCacheConfig(4, 16, 0, 0, 0, 0),
+		BatchConfig:         NewBatchConfig(10, 10_000, 16),
+		LatencyCoeffs:       NewLatencyCoeffs([]float64{float64(stepTimeTicks), 0, 0}, []float64{0, 0, 0}),
+		ModelHardwareConfig: NewModelHardwareConfig(ModelConfig{}, HardwareCalib{}, "test", "H100", 1, "blackbox", 0),
+	}
+	sim := mustNewSimulator(t, cfg)
+
+	for i := 0; i < numRequests; i++ {
+		deadline := deadlineTicks
+		if i >= numDeadline {
+			deadline = 0
+		}
+		sim.InjectArrival(&Request{
+			ID:           fmt.Sprintf("r%d", i),
+			ArrivalTime:  0,
+			InputTokens:  make([]int, inputLen),
+			OutputTokens: make([]int, outputLen),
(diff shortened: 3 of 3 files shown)


namasl:

susiejojo:
Can we fix the PR description above - Claude seems to have truncated most of the text, and it's hard to figure out what it is solving?

susiejojo:
## Minor: Test Bound Looseness (Non-blocking)

In `TestTimeout_CascadeDoesNotCreateOrphanedStepEvents`:

```go
const (
    simEndedMin = int64(120_000) // 120ms: midpoint between fixed (140ms) and buggy (100ms)
    simEndedMax = int64(250_000) // 250ms: generous upper bound
)
```

The upper bound of 250ms is 1.78x the expected 140ms. This means a performance regression that doubles step timing would silently pass. A tighter bound (e.g., `200_000` = 1.43x expected) would catch more regressions while still allowing for environment variance.

That said, this is **consistent with existing test style** (e.g., `TestTimeout_OrphanedTimeout_*` tests use similar loose upper bounds), so it's fine as-is. Just noting for future test maintenance considerations.

---

**Overall: The fix is correct, the test validates the key behavioral invariant (SimEndedTime not collapsed), and this is ready to merge.** ✅

susiejojo (commented):
lgtm, nit: https://github.com/inference-sim/inference-sim/pull/1100#issuecomment-4292127210

susiejojo (approved):

PR #1146, opened 2026-04-23
fix(workload): per-phase rate normalization for phased workloads

mtoslalibu:
## Summary

- **Per-phase normalization**: `normalizeRateFractions` now normalizes each client's `rate_fraction` by the sum of co-active clients (overlapping lifecycle windows) instead of the global sum. Clients without lifecycle windows are always-on and overlap with every phase.
- **inference-perf multi-stage bypass**: Multi-stage clients now use `CustomSamplerFactory` with Poisson at the exact per-client rate, bypassing fraction normalization entirely.
- Golden datasets regenerated due to RNG sequence shift from `CustomSamplerFactory` sub-RNG allocation.

Fixes #1144

## Test plan

- [x] Unit tests for `normalizeRateFractions` (7 cases: BC-1..4 + edge cases)
- [x] Unit tests for `windowsOverlap` (9 cases)
- [x] Updated `TestExpandInferencePerfSpec_MultiStage_CustomSamplerFactory` — verifies factory produces correct per-stage rate
- [x] Integration test: phased workload produces correct aggregate rate per phase (±15%)
- [x] Integration test: inference-perf multi-stage produces correct rate per stage (±20%)
- [x] Golden datasets updated (`-update-golden`) and verified
- [x] Full `go test ./...` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1147, opened 2026-04-23
fix(workload): peak-overlap normalization for lifecycle-windowed clients (#1144)

### LINKED ISSUE #1144: Bug: normalizeRateFractions halves rate for phased (lifecycle-windowed) workloads
**Describe the bug**

`normalizeRateFractions` in `sim/workload/client.go:15` sums `RateFraction` across ALL clients unconditionally, then divides each client's fraction by that total. For phased workloads where clients have non-overlapping lifecycle windows, the total exceeds 1.0 and every client's rate is incorrectly scaled down.

A two-phase workload specifying `aggregate_rate: 40` and `num_requests: 4000` produces only ~2072 requests at ~20.7 QPS — exactly half the intended rate.

**Root cause**

```go
// client.go:15-29
func normalizeRateFractions(clients []ClientSpec, aggregateRate float64) []float64 {
    totalFraction := 0.0
    for i := range clients {
        totalFraction += clients[i].RateFraction  // sums ALL clients, ignoring lifecycle windows
    }
    // ...
    normalizedFraction := clients[i].RateFraction / totalFraction
    rates[i] = aggregateRate * normalizedFraction / 1e6
}
```

For the repro workload below, the three clients have fractions 0.7 + 0.3 + 1.0 = 2.0. But Phase 1 clients (0.7 + 0.3) and Phase 2 client (1.0) are never active simultaneously. The normalization divides by 2.0, giving:

| Client | Window | Fraction | Expected rate | Actual rate |
|--------|--------|----------|---------------|-------------|
| phase1-A | 0-50s | 0.70 | 40 x 0.70 = 28 QPS | 40 x 0.35 = **14 QPS** |
| phase1-B | 0-50s | 0.30 | 40 x 0.30 = 12 QPS | 40 x 0.15 = **6 QPS** |
| phase2 | 50-100s | 1.00 | 40 x 1.00 = 40 QPS | 40 x 0.50 = **20 QPS** |

Every phase runs at half rate.

**CLI command to reproduce**

Save the workload spec below as `repro_phased.yaml`, then:

```bash
# 1. Start a dummy server (any HTTP server returning SSE completions, e.g. the python snippet below)
python3 dummy_server.py 9999 &

# 2. Run blis observe
./blis observe \
  --server-url http://localhost:9999 \
  --model dummy \
  --workload-spec repro_phased.yaml \
  --trace-header /tmp/repro_header.yaml \
  --trace-data /tmp/repro_trace.csv

# 3. Check results
python3 -c "
import csv
rows = list(csv.DictReader(open('/tmp/repro_trace.csv')))
N = len(rows)
sends = [int(r['send_time_us']) for r in rows]
qps = (N-1) / ((max(sends) - min(sends)) / 1e6)
print(f'Requests: {N} (expected 4000)')
print(f'Actual QPS: {qps:.1f} (expected ~40)')
"
```

**Minimal repro workload** (`repro_phased.yaml`):

```yaml
version: "2"
seed: 42
aggregate_rate: 40
num_requests: 4000
horizon: 100000000  # 100s

clients:
  - id: "phase1-A"
    rate_fraction: 0.70
    streaming: true
    arrival: {process: poisson}
    input_distribution: {type: gaussian, params: {mean: 128, std_dev: 32, min: 32, max: 256}}
    output_distribution: {type: gaussian, params: {mean: 64, std_dev: 16, min: 16, max: 128}}
    lifecycle:
      windows:
        - start_us: 0
          end_us: 50000000

  - id: "phase1-B"
    rate_fraction: 0.30
    streaming: true
    arrival: {process: poisson}
    input_distribution: {type: gaussian, params: {mean: 128, std_dev: 32, min: 32, max: 256}}
    output_distribution: {type: gaussian, params: {mean: 64, std_dev: 16, min: 16, max: 128}}
    lifecycle:
      windows:
        - start_us: 0
          end_us: 50000000

  - id: "phase2"
    rate_fraction: 1.0
    streaming: true
    arrival: {process: poisson}
    input_distribution: {type: gaussian, params: {mean: 128, std_dev: 32, min: 32, max: 256}}
    output_distribution: {type: gaussian, params: {mean: 64, std_dev: 16, min: 16, max: 128}}
    lifecycle:
      windows:
        - start_us: 50000000
          end_us: 100000000
```

**Dummy server** (`dummy_server.py`):

```python
#!/usr/bin/env python3
import json, sys
from http.server import HTTPServer, BaseHTTPRequestHandler

class H(BaseHTTPRequestHandler):
    def do_POST(self):
        body = json.loads(self.rfile.read(int(self.headers.get("Content-Length", 0))))
        n = body.get("max_tokens", 10)
        self.send_response(200)
        self.send_header("Content-Type", "text/event-stream")
        self.end_headers()
        for i in range(n):
            self.wfile.write(f'data: {json.dumps({"choices":[{"text":"x","finish_reason":None}],"usage":None})}\n\n'.encode())
            self.wfile.flush()
        self.wfile.write(f'data: {json.dumps({"choices":[{"text":"","finish_reason":"stop"}],"usage":{"completion_tokens":n}})}\n\n'.encode())
        self.wfile.write(b'data: [DONE]\n\n'); self.wfile.flush()
    def log_message(self, *a): pass

HTTPServer(("", int(sys.argv[1]) if len(sys.argv) > 1 else 9999), H).serve_forever()
```

**Expected behavior**

4000 requests at ~40 QPS over 100s, with ~2000 in Phase 1 (0-50s) and ~2000 in Phase 2 (50-100s). Each phase should run at the intended aggregate rate of 40 QPS.

The user's `aggregate_rate` represents the **instantaneous** target rate during any active phase — not a budget to be divided across all phases.

**Actual behavior**

~2072 requests at ~20.7 QPS. Both phases run at exactly half the intended rate.

- [x] None / unknown

This is a workload generation bug, not a simulation invariant violation. However, it silently produces incorrect experimental results — sim2real deployments using phased workloads ran at half load, making latency comparisons unreliable.

**Impact**

This bug affected the sim2real adaptive-routing deployment. Workloads W1 and W3 (both phased) ran at 20 QPS instead of 40 QPS in production, invalidating latency gain measurements. W2 (stationary, no lifecycle windows, fractions sum to 1.0) was unaffected.

**Suggested fix direction**

The fix should be **phase-aware normalization**, not a simple flag check. `normalizeRateFractions` should:

1. Detect distinct time phases from lifecycle windows across all clients
2. For each phase, identify which clients are active (their lifecycle window covers that phase)
3. Normalize fractions within each phase so that concurrently-active fractions sum to 1.0
4. Assign per-client rates such that `aggregate_rate` is achieved during every phase

This is more robust than "skip normalization when lifecycle windows exist" because it handles edge cases like overlapping windows, partially-active clients, and mixed (windowed + always-on) client sets.

Note: `ExpandInferencePerfSpec` (multi-stage) sidesteps this bug because it pre-computes `aggregateRate = sum(stageRates)`, making the math work out after normalization. Hand-written phased YAMLs don't have this escape hatch.

**Environment**
- OS: macOS 26.3.1
- Go version: 1.24.0
- Commit hash: 045692f

**Additional context**

Discovered during sim2real parity check comparing simulation predictions against real llm-d deployment results. The real deployment used `blis observe` to generate workloads, and the half-rate bug caused W1 and W3 to run at insufficient load, masking the adaptive routing algorithm's gains.

SAY-5:
## Summary

`normalizeRateFractions` summed `RateFraction` across every client and used that sum as the divisor, which halved the effective aggregate rate whenever clients had non-overlapping lifecycle windows. A two-phase workload with phase-1 fractions 0.7+0.3 and phase-2 fraction 1.0 ran at ~20 QPS instead of the specified 40 QPS.

This PR changes the normalization divisor to the maximum `RateFraction` sum over any sub-interval of simulated time, computed by an event-sweep over all lifecycle windows. Clients without a `Lifecycle` spec are treated as always-active, which preserves the existing behaviour when no windows are configured (the peak equals the unconditional sum).

Windows are half-open `[start, end)`: removals are processed before additions at tied instants so that back-to-back phases do not falsely register a moment of overlap.

`inference_perf` multi-stage expansion is updated in lock-step: `aggregateRate` is now the peak stage rate (not the sum), which combined with the peak-overlap normalization keeps per-client and per-stage rates exactly at their configured values. BC-2 tests and the override warning message are updated to match.

Fixes #1144

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./sim/workload/ -count=1\` (new TestNormalizeRateFractions_* cases)
- [x] \`go test ./sim/cluster/ -run TestTrainedPhysics_GoldenDataset -count=1\` (golden dataset still matches)
- [x] \`go test ./... -count=1 -timeout 600s\` (all packages green)

### DIFF
--- a/sim/workload/client.go
+++ b/sim/workload/client.go
@@ -1,7 +1,9 @@
 package workload
 
 import (
+	"math"
 	"math/rand"
+	"sort"
 
 	"github.com/inference-sim/inference-sim/sim"
 )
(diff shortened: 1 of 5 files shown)


sriumcp:
Thank you @SAY-5 for this PR. However the parent issue https://github.com/inference-sim/inference-sim/issues/1144 appears to have been resolved through #1146.

cc: @mtoslalibu

PR #1174, opened 2026-04-27
feat(sim): priority preemption mode — evict least-urgent request under KV pressure

sriumcp:
## Summary

- Adds `selectPriorityVictim()`: `min(SLOPriority)` with `max(ArrivalTime)` tiebreak — analog of vLLM `scheduler.py:827-829` `max(priority, arrival_time)` with inverted BLIS convention (higher=more-urgent)
- Changes `preemptForTokens` signature to `(req, numNewTokens, result, ctx, budget, reqIndex) → (bool, int)` — returns `reqAdjustment` counting evictions at indices below `reqIndex`
- Phase 1 index correction: `reqIndex -= adj` after each `preemptForTokens` call (analog of vLLM `scheduler.py:853` `req_index -= 1`) — prevents element skipping after non-tail eviction
- `NewBatchFormation(preemptionPolicy string, sloMap *SLOPriorityMap)` — 2-arg constructor; `nil` sloMap → `DefaultSLOPriorityMap()` (real map threaded in #1170)
- `NewSimulator` wired: `NewBatchFormation(cfg.PreemptionPolicy, nil)`
- INV-12 (Phase 1 Completeness Under Priority Preemption) documented
- vLLM `adjustment` conditionality divergence documented with rationale

## Behavioral Contracts

- **BC-1**: FCFS tail eviction preserved — `"third"` (tail) evicted under default mode
- **BC-2**: Priority evicts least-urgent regardless of position (table-driven: head/middle/tail)
- **BC-3**: Equal-priority tiebreak by latest arrival (`max(ArrivalTime)`)
- **BC-4**: KV conservation (INV-4) after priority preemption — `used <= capacity`
- **BC-5**: Phase 1 completeness (INV-12) — all non-preempted requests get decode tokens after non-tail eviction
- **BC-6**: Multi-eviction ordering — cascading evictions in non-decreasing urgency order
- **NC-1**: Empty batch no-panic — circuit breaker fires
- **NC-2**: Self-preemption guard — current request evicted → goes to WaitQ

## Test Plan

- [x] `TestPreemption_FCFS_EvictsTail` — BC-1
- [x] `TestPreemption_Priority_EvictsLeastUrgent` (3 subtests) — BC-2
- [x] `TestPreemption_Priority_TiebreakByLatestArrival` — BC-3
- [x] `TestPreemption_Priority_KVConservation` — BC-4 (INV-4)
- [x] `TestPreemption_Priority_Phase1Completeness` — BC-5 (INV-12)
[...]

PR #1171, opened 2026-04-27
feat(sim): PreemptionPolicy type + --preemption-policy flag (zero behavior change)

sriumcp:
## Summary

- Introduces `PreemptionPolicy` string-enum type (`fcfs`, `priority`) co-located with `VLLMBatchFormation` in `sim/batch_formation.go`
- Adds `PreemptionConfig` struct and `Preemption` field to `PolicyBundle` with YAML support and `Validate()` coverage (`sim/bundle.go`)
- Adds `PreemptionPolicy string` field to `PolicyConfig` and updates `NewPolicyConfig()` to 3-arg canonical constructor, updating all 10 call sites (`sim/config.go`)
- Registers `--preemption-policy` CLI flag (default: `fcfs`) shared by `blis run` and `blis replay`, with bundle override and validation (`cmd/root.go`)

**Zero behavior change**: the `"fcfs"` default maps exactly to the existing tail-of-batch eviction path. The `"priority"` value is registered and validated but not yet wired — that happens in #1169.

## Behavioral Contracts

- **BC-1**: `IsValidPreemptionPolicy("")` / `IsValidPreemptionPolicy("fcfs")` / `IsValidPreemptionPolicy("priority")` → `true`; all other strings → `false`
- **BC-2**: `PolicyBundle{Preemption: PreemptionConfig{Policy: "random"}}.Validate()` → non-nil error
- **BC-3**: `NewPolicyConfig("constant", "fcfs", "").PreemptionPolicy == ""`
- **BC-4**: `blis run` without `--preemption-policy` → default `"fcfs"` logged in policy config line
- **BC-5**: `blis run --preemption-policy random` → `logrus.Fatalf` with valid options listed
- **BC-6**: bundle `preemption.policy: priority` overrides CLI default when `--preemption-policy` not explicitly set
- **BC-7**: any run with `--preemption-policy fcfs` produces byte-identical output to the same run without the flag

## Test Plan

- [x] `TestPreemptionPolicy_ValidNames` — BC-1
- [x] `TestPolicyBundle_Validate_InvalidPolicy` (table: `"bad preemption"`) — BC-2
- [x] `TestNewPolicyConfig_DefaultPreemptionPolicy` — BC-3
- [x] `TestNewPolicyConfig_FieldEquivalence` (updated to 3-arg) — BC-3
- [x] `TestResolvePolicies_PolicyFlagsRegisteredInBothCommands` — BC-4/BC-6
[...]

Issue #1169, opened 2026-04-27
feat(sim): implement priority preemption victim selection with phase-1 index fix


sriumcp:
**Subissue of #1086.** Depends on #1168 (config + CLI plumbing). Together with #1168 and #1166, resolves the preemption parity gap.

## What problem does this solve?

With `--preemption-policy priority` (added by #1168), the simulator should evict the **least-urgent** running request instead of unconditionally taking the batch tail. This matches vLLM's `--scheduling-policy priority` preemption behavior.

There is also a correctness bug that only manifests with priority mode: when a non-tail request is evicted, every element after it shifts left by one index. The Phase 1 loop uses an explicit index variable, so without compensating, it skips the element that moved into the vacated slot. This is the exact bug vLLM fixes at `scheduler.py:853` with `req_index -= 1`.

## Background: priority convention difference vs. vLLM

vLLM uses **lower value = more urgent** and calls `max(priority, arrival_time)` to select the least-urgent victim.
BLIS uses **higher value = more urgent** (GAIE-compatible: critical=4, standard=3, batch=-1, sheddable=-2, background=-3) and must call `min(priority)` with `max(ArrivalTime)` tiebreak.

Do **not** port vLLM's `max()` directly — that would evict the most urgent request.

## Eviction order under priority mode

From least urgent (evicted first) to most urgent (evicted last):
`background (-3) → sheddable (-2) → batch (-1) → standard (3) → critical (4)`

Among requests with equal priority, the **most recently arrived** is evicted first (same tiebreak as vLLM).

### `sim/batch_formation.go`

**`VLLMBatchFormation` struct**: add `sloMap *SLOPriorityMap` field (the `NewBatchFormation` factory from #1168 already accepts and stores it).

**`selectPriorityVictim(requests []*Request) int`** — new method:
```go
func (v *VLLMBatchFormation) selectPriorityVictim(requests []*Request) int {
    victimIdx := len(requests) - 1
    victimPri := v.sloMap.Priority(requests[victimIdx].SLOClass)
    victimArrival := requests[victimIdx].ArrivalTime
    for i := len(requests) - 2; i >= 0; i-- {
        pri := v.sloMap.Priority(requests[i].SLOClass)
        if pri < victimPri || (pri == victimPri && requests[i].ArrivalTime > victimArrival) {
            victimIdx = i
            victimPri = pri
            victimArrival = requests[i].ArrivalTime
        }
    }
    return victimIdx
}
```

[...]

Issue #1168, opened 2026-04-27
feat(sim): add PreemptionPolicy config type and --preemption-policy CLI flag


sriumcp:
**Subissue of #1086.** Together with #1086-2 and #1166, resolves the preemption parity gap.

## What problem does this solve?

`VLLMBatchFormation` has no named policy concept for victim selection — the eviction strategy is an implicit constant baked into the code. Before the priority algorithm (subissue #1086-2) can land, the config type, validation, and CLI flag wiring need to exist as a zero-behavior-change foundation. This separation keeps #1086-2 focused purely on correctness.

## Proposed solution

```bash
./blis run   --model qwen/qwen3-14b --preemption-policy fcfs      # explicit default, identical behavior
./blis run   --model qwen/qwen3-14b --preemption-policy priority  # enables #1086-2 algorithm
./blis replay --trace-header t.yaml --trace-data d.csv \
              --model qwen/qwen3-14b --preemption-policy priority
```

Policy can also be set in a policy bundle YAML:
```yaml
preemption:
  policy: priority
```
CLI flag takes precedence over bundle when explicitly set.

### `sim/batch_formation.go`
- Add `PreemptionPolicy` string type with two constants:
  ```go
  type PreemptionPolicy string
  const (
      PreemptionFCFS     PreemptionPolicy = "fcfs"
      PreemptionPriority PreemptionPolicy = "priority"
  )
  ```
- Add `preemptionPolicy PreemptionPolicy` field to `VLLMBatchFormation` struct.
- Change `NewBatchFormation()` signature to `NewBatchFormation(preemptionPolicy string, sloMap *SLOPriorityMap) BatchFormation`. If `preemptionPolicy == ""`, default to `PreemptionFCFS`. If `sloMap == nil`, default to `DefaultSLOPriorityMap()`. **No algorithm branching yet** — `preemptForTokens` is unchanged.
- Update the one existing callsite in `sim/simulator.go:145` from `NewBatchFormation()` to `NewBatchFormation(cfg.PreemptionPolicy, nil)`.

### `sim/config.go`
- Add `PreemptionPolicy string` field to the `PolicyConfig` struct (alongside `PriorityPolicy` and `Scheduler`).
[...]

PR #1175, opened 2026-04-27
feat(sim): thread SLOPriorityOverrides from DeploymentConfig to NewBatchFormation

### LINKED ISSUE #1170: fix(sim): thread SLOPriorityMap overrides from AdmissionConfig into preemption victim selection
**Subissue of #1086.** Depends on #1168 and #1169. Completes the priority preemption feature by making it consistent with the admission layer's priority configuration.

Originally filed as #1166. Rewritten here as a subissue with full implementation detail.

Users can customize SLO tier priorities via `slo_priorities` in a policy bundle YAML:
```yaml
admission:
  slo_priorities:
    batch: 0        # make batch non-sheddable
    critical: 10    # raise critical above default
```

These overrides are respected by admission control and the gateway queue. They are **not** respected by preemption victim selection, which always uses `DefaultSLOPriorityMap()` (critical=4, standard=3, batch=-1, sheddable=-2, background=-3) regardless of what the user configured.

This is a code-path parity violation (R23): the same priority concept is evaluated differently in two control-plane layers. With the example above, `--preemption-policy priority` would still evict `batch` requests before `standard` ones even though the user declared `batch: 0` (making batch higher priority than standard=3 is false, but batch=0 is higher than batch=-1... actually the point is that the user's override is ignored entirely).

## Root cause

The `SLOPriorityOverrides` map lives in `DeploymentConfig` (cluster-level config at `sim/cluster/deployment.go:104`). It is built into a `*SLOPriorityMap` at `sim/cluster/cluster.go:181` and passed to admission and gateway queue constructors. However:

1. `sim.SimConfig` (the per-instance config struct, `sim/config.go`) has **no** `SLOPriorityOverrides` field.
2. `resolveConfigForRole()` returns a `sim.SimConfig` — it cannot carry overrides because the field doesn't exist there.
3. `sim/simulator.go:145` calls `NewBatchFormation(cfg.PreemptionPolicy, nil)` — the `nil` is hardcoded, so `NewBatchFormation` always falls back to `DefaultSLOPriorityMap()`.

### `sim/config.go`

Add `SLOPriorityOverrides map[string]int` field to the `SimConfig` struct. Place it near `PolicyConfig` since it is policy-related:
```go
type SimConfig struct {
    // ... existing fields ...
    SLOPriorityOverrides map[string]int // nil means use GAIE defaults
}
```
Do **not** add it to `PolicyConfig` — it is not part of the policy name selection, it is data used by multiple policies.

### `sim/cluster/cluster.go`

After every call to `config.resolveConfigForRole(role)` that produces a `simCfg`, propagate the overrides before passing `simCfg` to `NewInstanceSimulator`. There are **three** such call sites (search for `NewInstanceSimulator(id, simCfg)` to find all of them):

```go
simCfg := config.resolveConfigForRole(role)
simCfg.SLOPriorityOverrides = config.SLOPriorityOverrides  // add this line
inst := NewInstanceSimulator(id, simCfg)
```

Apply to all three call sites. Missing even one will cause inconsistent behavior depending on how the instance was constructed.

### `sim/simulator.go`

Change line 145 from:
```go
batchFormation := NewBatchFormation(cfg.PreemptionPolicy, nil)
```
to:
```go
batchFormation := NewBatchFormation(cfg.PreemptionPolicy, sim.NewSLOPriorityMap(cfg.SLOPriorityOverrides))
```
(`NewSLOPriorityMap(nil)` is equivalent to `DefaultSLOPriorityMap()` — nil-safe, no behavior change when no overrides are configured.)

### `sim/batch_formation_test.go`

Add one new test to verify the fix:

**P-7 — Custom priority map is respected by victim selection**

Setup: `NewBatchFormation("priority", NewSLOPriorityMap(map[string]int{"background": 10}))` — this overrides background to priority 10, making it more urgent than critical (4). Running batch: [background request, standard request]. Expected victim: standard (priority 3, the minimum under the custom map). Without this fix, the victim would be background (priority -3 under the default map).

- [x] Core simulator (`sim/`) — `config.go`, `simulator.go`, `batch_formation_test.go`
- [x] Cluster simulation (`sim/cluster/`) — `cluster.go`

## Which invariants are at risk?
- [x] None (this is a consistency fix — no conservation or causality impact)

Semantic impact is real but bounded: only affects runs where **both** `--preemption-policy priority` is active **and** custom `slo_priorities` are configured. Default-config users (no custom `slo_priorities`) see no behavior change because `NewSLOPriorityMap(nil)` returns identical values to `DefaultSLOPriorityMap()`.

## Alternatives considered

Passing `*SLOPriorityMap` as a constructor parameter through `PolicyConfig` — rejected because `PolicyConfig` holds policy *names*, not data. Keeping the data in `SimConfig` (as a plain `map[string]int`) and converting it at the point of use is the correct layering.

Subissue of #1086. **Blocked by #1168 and #1169** (the `NewBatchFormation` signature and `selectPriorityVictim` must exist first). Originally tracked as #1166.

sriumcp:
## Summary

- Moves `SLOPriorityOverrides` field from `DeploymentConfig` to `SimConfig`, closing the config propagation gap that caused `slo_priorities` to reach admission but not preemption
- Replaces hardcoded `nil` in `NewSimulator` with `NewSLOPriorityMap(cfg.SLOPriorityOverrides)`, so policy bundle `slo_priorities` now affects `--preemption-policy priority` victim selection
- Field removal from `DeploymentConfig` is transparent via Go struct embedding promotion — `cluster.go:181` access is unchanged
- Updates user-facing docs to document that `slo_priorities` is shared between admission and preemption

## Behavioral Contracts

- **BC-1**: Custom `slo_priorities` reach preemption — `batch: 0` override causes background(-3) to be evicted before batch(0)
- **BC-2**: Default behavior unchanged — nil overrides → GAIE defaults (same as before)
- **BC-3**: Cluster admission still resolves `config.SLOPriorityOverrides` via promoted field — no cluster.go edits needed

## Test Plan

- [x] `TestNewSimulator_CustomSLOPriorityMap_AffectsPreemption` — BC-1: verifies custom priorities propagate to victim selection (9/10 block saturation, distinct input tokens, Phase 1 preemption path)
- [x] All existing tests pass (BC-2, BC-3 covered by existing cluster + admission tests)
- [x] `go build ./...` + `go test ./...` + `golangci-lint run ./...` — all pass

## Dependency chain

- **#1168** (merged as #1171): type + flag scaffolding
- **#1169** (merged as #1174): priority preemption algorithm + INV-12
- **#1170 (this PR)**: thread `SLOPriorityOverrides` to complete the config pipeline

Fixes #1170


sriumcp:

  --- CROSS-PATH PARITY (run / replay / observe) ---
                                                   
  BLIS has three command paths that must maintain behavioral parity:
  - blis run (DES with synthetic workload) 
  - blis replay (DES with trace-driven workload)                                                                                                         
  - blis observe (real HTTP dispatch to live server)
                                                                                                                                                         
  For every feature, flag, or behavioral change in this PR:                                                                                              
  1. Does this feature logically apply to the other two paths?                                                                                           
  2. If yes: does the PR implement it for all applicable paths, or at minimum file a follow-up issue?                                                    
  3. If the PR only covers one path: is there an explicit justification for why parity is not needed?                                                    
                                                                                                                                                       
  Common parity gaps to check:                                                                                                                           
  - CLI flags added to one command but not others (e.g., --timeout, --think-time-dist)                                                                   
  - Workload spec fields consumed by one path but ignored by others                                                                                      
  - Metrics computed differently across paths (e.g., timeout counting)                                                                                   
  - Default values that differ between paths without justification                                                                                       
                                                                                                                                                         
  --- PREEMPTION SAFETY ---                                                                                                                              
                                                                                                                                                         
  Preemption (request evicted from RunningBatch, ProgressIndex reset to 0, re-queued) is the #1 source of metric bugs. For any change that touches:      
  - Per-request metrics (TTFT, ITL, E2E, TotalOutputTokens, TTFTSum)                                                                                     
  - Request state transitions (StateQueued, StateRunning, StateCompleted, StateTimedOut)                                                                 
  - Aggregate counters (CompletedRequests, TimedOutRequests)                                                                                             
                                                                                                                                                         
  Verify: what happens when the request is preempted mid-execution and re-runs? Specifically:                                                            
  - Are inline metrics (recorded during execution) overwrite-safe or accumulate-and-double-count?                                                        
  - Are aggregate sums deferred to completion time (recordRequestCompletion) to avoid double-counting?                                                   
  - Does the TimeoutEvent for a preempted-then-completed request get lazily cancelled?                                                                   
                                                                                                                                                       
  --- TIMEOUT / DEADLINE CONSISTENCY ---                                                                                                                 
                                                                                                                                                         
  BLIS has two distinct timeout mechanisms:                                                                                                              
  - DES TimeoutEvent: fires at req.Deadline (simulation ticks), enforced by the event loop                                                               
  - HTTP client timeout: --timeout flag (wall-clock seconds), enforced by Go http.Client                                                                 
                                                                                                                                                         
  For any change that touches timeout, deadline, or horizon:                                                                                             
  - Are DES deadlines and HTTP timeouts producing equivalent observable behavior?                                                                        
  - Does the trace pipeline preserve deadline semantics end-to-end (observe → trace → replay)?                                                           
  - Is the 300s default consistent across both mechanisms (DefaultTimeoutUs in generator.go:676 vs defaultHTTPTimeoutSeconds in observe.go:22)?          
                                                                                                                                                         
sriumcp:
Addressed all three review observations:

1. **`resolve.go` struct-copy safety comment** — extended lines 53-58 to document that `SLOPriorityOverrides map[string]int` also shares its backing map across pool-resolved instances (same safety reasoning as `LatencyCoeffs` slices: `NewSLOPriorityMap` only reads the map, never mutates it).

2. **Alignment inconsistency** — reverted the extra space on `PolicyConfig:` in `cmd/root.go` and `cmd/replay.go` (was widened to column-align with `SLOPriorityOverrides:`, inconsistent with surrounding fields).

3. **BC-3 cluster-layer test gap** — added `TestResolveConfigForRole_SLOPriorityOverrides_Propagates` in `sim/cluster/resolve_test.go`, verifying that `SLOPriorityOverrides` set on `DeploymentConfig.SimConfig` survives `resolveConfigForRole` for all three roles (default, prefill, decode).

Also rebased on upstream/main to resolve the `CLAUDE.md` merge conflict (ServeGen temporal parity entry from `main` and the #1170 entry are now both present).

susiejojo:
## vLLM Parity Check ✅

Verified against vLLM :

**Core algorithm:** ✅ Matches vLLM's `max(self.running, key=lambda r: (r.priority, r.arrival_time))` at lines 827-829. BLIS uses inverted convention (lower priority = higher urgency), which is correctly documented.

**Index adjustment:** ✅ Matches vLLM's `req_index -= 1` pattern at line 853 for non-tail eviction.

**Key difference:** vLLM uses direct `Request.priority: int` fields. BLIS adds an `SLOPriorityMap` layer (SLO class → priority integer). This PR correctly threads that mapping to preemption. This is a valid extension, not a parity violation.

**Verdict:** PR is in parity with vLLM's priority preemption implementation.

susiejojo:
## vLLM Parity Check ✅

Verified against vLLM v1/core/sched/scheduler.py:

**Core algorithm:** ✅ Matches vLLM's max(self.running, key=lambda r: (r.priority, r.arrival_time)) at lines 827-829. BLIS uses inverted convention (lower priority = higher urgency), which is correctly documented.

**Index adjustment:** ✅ Matches vLLM's req_index -= 1 pattern at line 853 for non-tail eviction.

**Key difference:** vLLM uses direct Request.priority: int fields. BLIS adds an SLOPriorityMap layer (SLO class → priority integer). This PR correctly threads that mapping to preemption. This is a valid extension, not a parity violation.

**Verdict:** PR is in parity with vLLM's priority preemption implementation.

mtoslalibu (approved):

susiejojo (approved):

PR #1187, opened 2026-04-28
fix(observe): handle io.ReadAll error on non-200 HTTP response (#1120)

### LINKED ISSUE #1120: fix(observe): handle io.ReadAll error on non-200 HTTP response
## Description

In `cmd/observe.go`, the non-200 HTTP status code path at line ~199 discards the `io.ReadAll` error:

```go
bodyData, _ := io.ReadAll(resp.Body)
```

If the body read fails (e.g., connection reset mid-error-response), the error message included in `record.ErrorMessage` will be empty or truncated, losing diagnostic information.

## Current Behavior

The error from `io.ReadAll` is silently discarded. The request still gets `status=error` (correct), but `record.ErrorMessage` may contain an empty or partial server error body without any indication that the read itself failed.

## Expected Behavior

Log a warning if `io.ReadAll` fails on the error-body path, so operators know the error message may be incomplete. The status should remain `error` (the HTTP status code is the authoritative signal, not the body).

## Impact

Low — this only affects the diagnostic quality of error messages for non-200 responses where the body read also fails. The status is correct either way.

## Context

Identified during PR review of #1118 (streaming timeout fix). Pre-existing issue, not introduced by that PR.

mtoslalibu:
## Summary

- Log a warning when `io.ReadAll` fails on a non-200 HTTP response body (e.g., connection reset mid-error-response)
- Annotate `record.ErrorMessage` with "(body read failed: ...)" so operators know the trace error message may be incomplete
- Preserve existing behavior when body reads successfully (no change to success path)

Fixes #1120

## Behavioral Contracts

**BC-1: ReadAll failure warning**
- GIVEN a server returns a non-200 HTTP status code AND the response body read fails
- WHEN `RealClient.Send` processes the response
- THEN a `logrus.Warn` message is emitted containing "failed to read error response body" AND `record.Status` is `"error"` AND `record.ErrorMessage` contains the HTTP status code and a note that the body read failed

**BC-2: ReadAll success preserves existing behavior**
- GIVEN a server returns a non-200 HTTP status code AND the response body reads successfully
- WHEN `RealClient.Send` processes the response
- THEN `record.Status` is `"error"` AND `record.ErrorMessage` contains the HTTP status code and the full body text AND no warning is logged

## Test plan

- [x] `TestRealClient_ServerError_BodyReadFailure` — uses `http.Hijacker` to simulate partial response with Content-Length mismatch; verifies warning logged and error message annotated (BC-1)
- [x] `TestRealClient_ServerError_BodyReadSuccess` — normal non-200 response; verifies full body in error message and no spurious warning (BC-2)
- [x] Existing `TestRealClient_ServerError_RecordsError` still passes (regression guard)
- [x] Full test suite passes: `go test ./...`

### DIFF
--- a/cmd/observe.go
+++ b/cmd/observe.go
@@ -197,11 +197,17 @@ func (c *RealClient) Send(ctx context.Context, req *PendingRequest) (*RequestRec
 
 	if resp.StatusCode != http.StatusOK {
 		bodyData, readErr := io.ReadAll(resp.Body)
-		record.Status = "error"
 		if readErr != nil {
+			if isTimeoutError(readErr) {
+				record.Status = "timeout"
+				record.ErrorMessage = fmt.Sprintf("HTTP %d: (body read timed out: %v)", resp.StatusCode, readErr)
+			} else {
+				record.Status = "error"
+				record.ErrorMessage = fmt.Sprintf("HTTP %d: (body read failed: %v)", resp.StatusCode, readErr)
+			}
 			logrus.Warnf("observe: request %d: failed to read error response body: %v", record.RequestID, readErr)
-			record.ErrorMessage = fmt.Sprintf("HTTP %d: (body read failed: %v)", resp.StatusCode, readErr)
 		} else {
+			record.Status = "error"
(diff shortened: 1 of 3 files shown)


sriumcp (approved):
lgtm

PR #1202, opened 2026-04-28
fix(cluster): criticality-based gateway queue shedding + tenant budget decorator

mtoslalibu:
## Summary
- **Gateway queue criticality filter**: `GatewayQueue.Enqueue()` now only evicts sheddable requests (`priority < 0`) on overflow. Non-sheddable requests (priority >= 0) are protected. When no sheddable victim exists, incoming request is rejected via new `Rejected` outcome.
- **TenantBudgetAdmission decorator**: Inline tenant budget check removed from `cluster_event.go` and replaced with `TenantBudgetAdmission` wrapping any inner `AdmissionPolicy` through the standard interface. Uses `TenantBudgetTracker` interface in `sim/` to avoid import cycle.
- **INV-1 updated**: Conservation formula extended with `+ gateway_queue_rejected` across CLAUDE.md, invariants.md, conservation test, and CLI output.

Closes #1190

### Deviation from #1190 spec
The issue calls for removing the `tenantTracker` field from `ClusterSimulator`. This PR retains it because it has two roles: (1) admission via `IsOverBudget()` — moved to `TenantBudgetAdmission` policy as specified, and (2) in-flight tracking via `OnStart()`/`OnComplete()` — still needed at 6+ call sites in `cluster_event.go`. Removing the field entirely requires reworking the in-flight tracking path, which is out of scope for Phase 1.

## Test plan
- [x] `TestGatewayQueue_CriticalityProtection_NonSheddableNeverEvicted` — non-sheddable entries are never displaced
- [x] `TestGatewayQueue_CriticalityProtection_SheddableEvicted` — sheddable entries can be displaced by higher-priority incoming
- [x] `TestGatewayQueue_CriticalityProtection_LowerSheddableDoesNotEvictHigherSheddable` — lower sheddable cannot displace higher sheddable
- [x] `TestGatewayQueue_CriticalityProtection_EqualPrioritySheddableTieBreak` — equal-priority FIFO tie-break verified
- [x] `TestTenantBudgetAdmission_DelegatesToInnerPolicy` — inner policy rejection propagates
- [x] `TestTenantBudgetAdmission_SheddableOverBudgetRejected` — sheddable + over-budget = rejected
- [x] `TestTenantBudgetAdmission_NonSheddableOverBudgetAdmitted` — non-sheddable always passes budget check
- [x] `TestTenantBudgetAdmission_SheddableUnderBudgetAdmitted` — sheddable + under-budget = admitted (happy path)
[...]

PR #1214, opened 2026-04-30
fix(cluster): restore trace-gap comment + pass through admission reason in TenantBudgetAdmission

atantawi:
## Summary

Follow-up to #1202 addressing two findings from post-merge review:

- **`cluster_event.go`**: The `Rejected` case now carries the trace-gap warning that existed on the old `shed == true` branch. `RecordAdmission(Admitted:true)` is already written before `Enqueue` is called, so trace consumers must not assume all `Admitted:true` requests have routing records when flow control is enabled. The comment was dropped in #1202 and this reinstates the equivalent for the new `Rejected` path.

- **`admission.go`**: `TenantBudgetAdmission.Admit` returns `(true, reason)` on success instead of `(true, "")`, preserving the inner policy's admission reason and following the standard decorator pass-through contract.

## Test plan

- [x] `go test ./sim/... ./sim/cluster/...` passes — no behavioral change, comment + one-liner fix only

### DIFF
--- a/sim/admission.go
+++ b/sim/admission.go
@@ -348,5 +348,5 @@ func (t *TenantBudgetAdmission) Admit(req *Request, state *RouterState) (bool, s
 	if t.tracker.IsOverBudget(req.TenantID) && t.priorityMap.IsSheddable(req.SLOClass) {
 		return false, "tenant-budget-shed"
 	}
-	return true, ""
+	return true, reason
 }
--- a/sim/admission_test.go
+++ b/sim/admission_test.go
@@ -349,3 +349,19 @@ func TestSLOPriorityMap_InvertForVLLM_PreservesUrgencyOrder(t *testing.T) {
 			critical, standard, batch, sheddable, background)
 	}
 }
+
+type stubAdmitWithReason struct{ reason string }
+
+func (s *stubAdmitWithReason) Admit(*Request, *RouterState) (bool, string) { return true, s.reason }
+
+func TestTenantBudgetAdmission_PassesThroughInnerAdmissionReason(t *testing.T) {
+	inner := &stubAdmitWithReason{reason: "inner-policy-reason"}
+	policy := NewTenantBudgetAdmission(inner, &stubTracker{overBudget: false}, DefaultSLOPriorityMap())
+	admitted, reason := policy.Admit(&Request{ID: "r1", TenantID: "t1", SLOClass: "standard"}, &RouterState{})
+	if !admitted {
+		t.Fatal("expected request to be admitted")
+	}
+	if reason != "inner-policy-reason" {
+		t.Errorf("expected reason %q from inner policy, got %q", "inner-policy-reason", reason)
+	}
+}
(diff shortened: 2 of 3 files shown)


mtoslalibu:
## Review

The two code changes are correct. Three things to address before merging.

---

### 1. Missing test (required)

`admission.go` changes `return true, ""` to `return true, reason` — a real behavioral change — but the existing tests don't catch it. Both admitted-path tests use `AlwaysAdmit` (which returns `""`) and discard the reason with `_`. Reverting the fix would leave all tests green.

Add a test that wraps a stub inner policy returning `(true, "some-reason")` and asserts the decorator passes it through unchanged.

---

### 2. Comment uses an undefined term (required)

`cluster_event.go` — "No correction record is emitted here" — there is no type called a "correction record" in the trace package (`AdmissionRecord`, `RoutingRecord`, `DisaggregationRecord`). The original comment on the old `shed == true` branch was more precise.

Suggest replacing with something like:
> No `trace.RoutingRecord` is emitted for this request. Requests here are counted in `GatewayQueueRejected()` (gw_rejected, INV-1), not in `rejectedRequests`.

Also: the trace-gap warning on `Rejected` ("Trace consumers must not assume all `Admitted:true` requests have routing records...") applies equally to the `ShedVictim` case just below. Consider moving it above the whole `switch` block.

---

### 3. Design docs look like they landed in the wrong PR

The two new files under `docs/plans/` (windowed collector and queueing model analyzer designs) describe future unimplemented features and aren't related to the trace-gap comment or the admission reason fix. They don't appear in the PR description either. Were these intentional? If not, they belong in a separate PR.

---

The `gateway_queue.go` nit from the #1202 review is fine to defer.

atantawi:
Addressed all three items from the review:

**1. Missing test** — Added `TestTenantBudgetAdmission_PassesThroughInnerAdmissionReason` in `sim/admission_test.go`. Uses a `stubAdmitWithReason` inner policy returning `"inner-policy-reason"` and asserts the decorator returns it unchanged when under budget. Reverting `return true, reason` now breaks this test.

**2. Undefined term in comment** — In `cluster_event.go`, replaced "No correction record is emitted here" with "No `trace.RoutingRecord` is emitted for this request. Counted in `GatewayQueueRejected()` (gw_rejected, INV-1), not in `rejectedRequests`." Also moved the trace-gap warning ("Trace consumers must not assume all `Admitted:true` requests have routing records…") above the `switch outcome` block so it covers both `Rejected` and `ShedVictim`.

**3. Unrelated design docs** — Removed `docs/plans/2026-04-23-windowed-collector-design.md` and `docs/plans/2026-04-23-queueing-model-analyzer-design.md` from this branch. They can go in a separate PR.

atantawi:
Rebased onto upstream/main to resolve the merge conflict.

The conflict was in `sim/admission_test.go`: upstream added three `InvertForVLLM` tests while our commit added the `PassesThroughInnerAdmissionReason` test at the same location. Kept both.

Also dropped the two design doc commits (`windowed-collector-design`, `queueing-model-analyzer-design`) from the branch history entirely — they were only on the fork's main, not on upstream/main, and were showing up in the PR diff. The branch now has exactly two commits touching three files: `sim/admission.go`, `sim/admission_test.go`, `sim/cluster/cluster_event.go`.

mtoslalibu (approved):

Issue #899, opened 2026-03-31
feat: GIE flow control parity — per-band queues, TTL, criticality-based shedding, tenant fairness


mtoslalibu:
## Goal

Make BLIS model GIE's flow control accurately so we can use it to discover admission control algorithms for llm-d — maximizing SLO attainment, penalizing bad actors, and prioritizing urgent or well-behaved users.

**Design principle:** Match GIE's interface contracts so algorithms developed in BLIS transfer cleanly to GIE/llm-d. Internals can differ (DES events vs goroutines, count-based vs byte-based capacity) but the pluggable boundaries must align.

---

## Interface Parity: BLIS ↔ GIE

Each row maps a GIE interface to what BLIS has today and what this issue delivers. The goal: same interface shapes, swappable implementations underneath.

### Already Matching

| Boundary | GIE Interface | BLIS Interface | Status |
|---|---|---|---|
| **Routing** | `Filter → Scorer → Picker` pipeline: request in, endpoint out | `RoutingPolicy.Route(req, state) → RoutingDecision` | **Parity** |
| **Scoring** | `Scorer.Score(req, pods) → map[Endpoint]float64` | `scorerFunc(req, snapshots) → map[string]float64` | **Parity** |
| **Saturation detection** | `SaturationDetector.Saturation(ctx, endpoints) → float64` | `SaturationDetector.Saturation(state) → float64` | **Parity** |

### Gaps This Issue Closes

| Boundary | GIE Interface | BLIS Today | Target |
|---|---|---|---|
| **Flow controller entry** | `FlowController.EnqueueAndWait(ctx, req) → QueueOutcome` — single call: request in, outcome out (dispatched / rejected / expired) | Split across `AdmissionPolicy.Admit()` + `GatewayQueue.Enqueue()/Dequeue()` — no unified contract | Unified flow control interface with `QueueOutcome` semantics (Gap 3) |
| **Fairness policy** | `FairnessPolicy.Pick(ctx, band) → flow` — pluggable, stateful per-band | None — hardcoded FIFO within priority level | Pluggable `FairnessPolicy` interface with global-strict and round-robin implementations (Gap 4) |
| **Ordering policy** | `OrderingPolicy.Less(a, b) → bool` — pluggable comparator (FCFS, SLO-deadline, EDF) | `"fifo"` / `"priority"` string switch baked into heap | Pluggable `OrderingPolicy` interface (Gap 5) |
[...]

PR #1311, opened 2026-05-08
feat(cluster): add encode pool and E/P/D disaggregation for multimodal (GAP-4)

namasl:
## Summary

Adds an encode pool role, an `EncodeDecider` interface (`never` / `always` / `multimodal`), and a synchronous encode routing stage inside `executeDisaggregatedRouting`. Purely additive — with `--encode-instances 0` (default), behavior is byte-identical to the pre-PR simulator.

Closes #1264. Parent tracker: #1260.

## Design doc

See `docs/plans/2026-05-08-epd-disaggregation-design.md` for the full design (module contract, key decisions, INV-1 update, follow-ups). The micro-plan is in `docs/plans/2026-05-08-epd-disaggregation-plan.md`.

### Key design decisions

- **D1. Synchronous inline stage, not a new event.** The issue's own "Extension friction check" states: *"If GAP-1 (routing path unification) is resolved first, add the encode stage to `executeDisaggregatedRouting` rather than to `DisaggregationDecisionEvent.Execute`."* GAP-1 is merged (#1266). Under option A (zero-duration encode), a dedicated `EncodeRoutingEvent` adds bookkeeping without modeling fidelity. Refactors cleanly to an event when non-zero encode latency is introduced.
- **D2. Encode latency = 0** (option A from the issue). Follow-up will add `--encode-latency-ms`.
- **D3. `IsMultimodal` is derived**, not a new field. `Request.ImageTokenCount + AudioTokenCount + VideoTokenCount > 0` already flows through spec → generator → TraceV2 → replay (present since PR #420). No TraceV2 schema change.
- **D4. Encode requires a decode-capable pool** (validated at CLI + construction). Encode-only clusters are rejected.
- **D5. Default decider is `never`** and `encodeDecider` is `nil` when the pool is disabled → no runtime path touches encode.

## Behavioral contracts

| Contract | Guarantee |
|----------|-----------|
| BC-EPD-1 | `--encode-instances 0` is byte-identical to pre-PR (INV-6). |
| BC-EPD-2 | `MultimodalEncodeDecider` fires iff any per-modality token count > 0. |
| BC-EPD-3 | With disagg + encode: `EncodeRoutingRecord.Clock ≤ PrefillRoutingRecord.Clock`; `ParentRequest.EncodeInstanceID` populated. |
| BC-EPD-4 | With non-disagg + encode: `EncodeRoutingRecord` emitted; no `PrefillRoutingRecord`; request injected to pre-selected decode pod. |
[...]

PR #1318, opened 2026-05-11
feat(admission): optional request TTL in gateway queue (#1193)

### LINKED ISSUE #1193: feat(admission): Phase 4 — optional request TTL in gateway queue
**What external behavior are we replicating?**

GIE's optional per-request TTL in the flow controller queue. When enabled, requests that sit in the queue longer than their TTL are expired (removed) rather than waiting indefinitely. Default off — same as llm-d.

**Reference implementation (MANDATORY)**

llm-d FlowControl config — `DefaultRequestTTL` field, defaults to zero (disabled):
https://github.com/llm-d/llm-d-inference-scheduler/blob/70183816/pkg/epp/flowcontrol/controller/config.go#L36-L56

GIE TTL context creation — only applies when TTL > 0:
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/dc908434/pkg/epp/flowcontrol/controller/controller.go#L384-L400

GIE QueueOutcomeEvictedTTL — terminal state for TTL expiry:
https://github.com/llm-d/llm-d-inference-scheduler/blob/70183816/pkg/epp/flowcontrol/controller/controller.go#L384-L400

llm-d flowControlRequest.InitialEffectiveTTL — returns 0 (use controller default):
https://github.com/llm-d/llm-d-inference-scheduler/blob/70183816/pkg/epp/requestcontrol/admission.go#L195

**Key behaviors to preserve**

- [x] Behavior 1: TTL defaults to 0 (disabled) — when disabled, requests wait indefinitely (reference: config.go DefaultRequestTTL)
- [x] Behavior 2: When TTL > 0, requests that exceed their TTL while queued are removed and counted as expired (reference: TTL context creation)
- [x] Behavior 3: If a request is dispatched before TTL fires, the TTL event is a no-op (reference: context cancellation pattern)

**Known intentional deviations**

- DES event-based TTL instead of Go context deadlines — schedule a `GatewayQueueTTLEvent` at `enqueue_time + TTL`
- Request-ID index for O(1) removal instead of GIE's handle-based `*list.Element` removal — simpler since DES is single-threaded
- Note: this is separate from instance-level `Request.Deadline`/`TimeoutEvent` which is a client-side timeout at the instance, not gateway queue TTL

**Target parity version**

- GIE: `kubernetes-sigs/gateway-api-inference-extension` @ [`dc908434`](https://github.com/kubernetes-sigs/gateway-api-inference-extension/commit/dc908434a7d5a4a151902130f00031580003b6f9)
- llm-d: `llm-d/llm-d-inference-scheduler` @ [`70183816`](https://github.com/llm-d/llm-d-inference-scheduler/commit/70183816964138a0facd3cc22f87b2865018f837)

- [x] Core simulator (`sim/`) — TTL event type
- [x] Cluster simulation (`sim/cluster/`) — `gateway_queue.go` (request-ID index, Remove method), `cluster_event.go` (TTLEvent), `deployment.go` (config field)
- [ ] KV cache (`sim/kv/`)
- [x] CLI (`cmd/`) — new flag: `--request-ttl`
- [ ] New package needed

**Extension friction check**

- ~100 lines: TTL event + request-ID index + removal logic + config
- Extends existing `GatewayQueue` with `Remove(requestID)` method (added in Phase 2)
- Affects INV-1 (request conservation) — new bucket: `gw_expired`

**What to build**

1. `FlowControlRequestTTL` config field on `DeploymentConfig` (microseconds, 0 = disabled)
2. On enqueue: schedule `GatewayQueueTTLEvent` at `enqueue_time + TTL`
3. Add request-ID index to `GatewayQueue` for O(1) lookup + removal
4. TTL event fires: look up request in index → if found, remove from queue, count as expired. If not found (already dispatched/shed), no-op.
5. `expiredCount` counter, extend INV-1 with `gw_expired` bucket

### Behavioral contract

```
GIVEN FlowControlRequestTTL = 5000 (5ms) and a queued request enqueued at tick 100
WHEN tick reaches 5100 and request is still queued
THEN request removed, gw_expired incremented, INV-1 balanced

GIVEN FlowControlRequestTTL = 5000 and a queued request enqueued at tick 100
WHEN request is dispatched at tick 200 (before TTL)
THEN TTL event at tick 5100 is no-op (request no longer in queue)

GIVEN FlowControlRequestTTL = 0 (default)
WHEN any request is enqueued
THEN no TTL event scheduled, requests wait indefinitely
```

### CLI flag

```
--request-ttl int   Gateway queue request TTL in microseconds (0=disabled, default 0)
```

- Part of #899 (GIE flow control parity — Phase 4)
- Low priority — TTL is default off in GIE and llm-d doesn't enable flow control in production today
- Deliver when needed for experiments

mtoslalibu:
## Summary

- Add `--request-ttl` CLI flag (microseconds, default 0=disabled) for gateway queue request expiry
- When enabled with `--flow-control`, requests that sit in the queue longer than the TTL are expired and removed
- Matches GIE's `DefaultRequestTTL` behavior — Phase 6 of #899 (GIE flow control parity)

## Behavioral Contracts

**BC-1:** `--request-ttl` not set (default 0) → no TTL events scheduled, requests wait indefinitely
**BC-2:** TTL expiry removes queued request, increments `gatewayExpired`, INV-1 holds
**BC-3:** Dispatched before TTL → TTL event is a no-op (request no longer in queue)
**BC-4:** TTL expiry counts in `ShedByTier` per SLO class
**BC-5:** TTL expiry triggers `tryDispatchFromGatewayQueue()` (freed slot may allow dispatch)
**BC-6:** INV-1 conservation: `+ gateway_expired` bucket added to equation

## Implementation

- `requestIndex map[string]requestLocation` on `GatewayQueue` for O(1) TTL lookup
- `RemoveByRequestID()` with arbitrary-position slice removal (does NOT use `removeEntryByIndex` which panics on non-tail)
- `GatewayQueueTTLEvent` (priority 6, after eviction at 5) fires at `enqueue_time + TTL`
- `gatewayExpired` counter wired into `ProgressSnapshot`, `RawMetrics`, CLI output
- INV-1, invariants.md, CLAUDE.md updated with `gateway_expired` bucket

## Test plan

- [x] `TestGatewayQueue_RemoveByRequestID` — middle, head, and post-shed removal
- [x] `TestGatewayQueueTTL_ExpiresQueuedRequest` — TTL fires, request removed, shedByTier updated
- [x] `TestGatewayQueueTTL_NoOpWhenDispatched` — all dispatched before TTL, gatewayExpired=0
- [x] `TestGatewayQueueTTL_DisabledByDefault` — TTL=0, no expirations
- [x] ShedByTier conservation invariant updated to include `GatewayExpired`
- [x] Full cluster test suite passes (150s)

Fixes #1193


mtoslalibu:
## Post-Implementation Review Analysis

Ran comprehensive PR review with code-reviewer, test-analyzer, and silent-failure-hunter agents. Key findings and resolutions:

### Fixed (commit a95a2ab)

**1. CRITICAL: `replay.go` missing `FlowControlRequestTTL` wiring (INV-13 violation)**
- `blis replay --flow-control --request-ttl 5000` silently ignored TTL because `DeploymentConfig` in `replay.go` omitted the field
- Fixed: wired `FlowControlRequestTTL`, `rawMetrics.GatewayExpired`, anomaly condition, and output print — matching `cmd/root.go` exactly

**2. MEDIUM: TTL value not logged in flow control startup message**
- Other flow control params (detector, dispatch, fairness, maxDepth, perBandCapacity) were logged but TTL was missing
- Fixed: added `requestTTL=%d` to the log format

**3. LOW: Test assertions weakened (`>= 1` instead of `== 1`)**
- With exactly 2 requests (r1 completes, r2 expires), the deterministic DES guarantees exactly 1 expiration
- Fixed: tightened to exact `== 1` assertions

### Evaluated and dismissed

- **"Tenant tracker should call OnComplete for expired requests"** — WRONG. Request was never dispatched, `OnStart()` never called, calling `OnComplete()` would corrupt the tracker. Correct behavior.
- **"Nil dereference without --flow-control"** — WRONG. `requestTTL` is set inside `if config.FlowControlEnabled`, so stays 0 when flow control is disabled. TTL scheduling code is inside `if cs.flowControlAdmission != nil`. No nil dereference possible.
- **"TTL event scheduled for routed requests"** — WRONG. If `tryDispatchFromGatewayQueue()` dispatches immediately, request is removed from `requestIndex`, TTL event fires later as a no-op. This IS BC-3 by design.
- **"O(n) scan in RemoveByRequestID"** — Known tradeoff. Flows are 1-10 entries per tenant-priority. Documented in plan.
- **"TTL without --flow-control should be fatal"** — Warn is appropriate. Follows existing pattern. CLI help text says "Requires --flow-control."

### Cross-path parity verification

| Path | TTL Support | Status |
|------|------------|--------|
| `blis run` | ✅ Full | Implemented in initial commit |
| `blis replay` | ✅ Full | Fixed in this review commit |
| `blis observe` | N/A | Talks to real servers — gateway TTL is a DES concept |

### Preemption safety
TTL expiry removes requests from the gateway queue BEFORE they reach any instance. No interaction with preemption, no instance-level state changes, no metric double-counting.

### Timeout/deadline consistency
Gateway TTL (enqueue-time based, microseconds, `--request-ttl`) is distinct from instance-level `Request.Deadline` (arrival-time based). No confusion — different mechanisms at different layers.

mtoslalibu:
## Final Review Round (post-merge, 4 agents)

Ran code-reviewer, test-analyzer, silent-failure-hunter, and type-design-analyzer on the post-merge state (after resolving conflicts with upstream EPD feature #1311).

### Fixed (commit ea66e16)

**1. IMPORTANT: Negative TTL validation inside `if flowControlEnabled` block (code-reviewer)**
- `--request-ttl -5` without `--flow-control` would skip the `< 0` check entirely
- Fixed: moved validation outside the block so it fires regardless of `--flow-control`

**2. IMPORTANT: R3 factory validation at library boundary (type-design-analyzer)**
- `NewClusterSimulator` accepted negative TTL from programmatic callers (tests, YAML)
- Fixed: added `panic` guard in constructor, consistent with other factory validations

**3. IMPORTANT: Existing INV-1 conservation tests missing `gwExpired` term (test-analyzer)**
- `TestClusterSimulator_FlowControl_Conservation` and `_Eviction_Conservation` had stale formulas
- Fixed: added `+ gwExpired` to both conservation equations, keeping formulas in sync with documented INV-1

### Evaluated and dismissed

- **Duplicate-ID assertion in Enqueue** (type-design-analyzer) — request IDs are always globally unique; adding assertion is overengineering for this PR
- **Full standalone INV-1 conservation test with gateway_expired > 0** (test-analyzer) — the ShedByTier conservation test already covers this path indirectly, and the two existing conservation tests now include the term. Sufficient coverage.
- **Priority dispatch mode test** (test-analyzer) — `RemoveByRequestID` uses `requestIndex` for lookup, doesn't depend on dispatch order. Same Enqueue/index code path for both modes.
- **Stale TTL events in heap** (type-design-analyzer) — known DES tradeoff. At BLIS scale (hundreds to thousands of requests), negligible overhead.

### Agent verdicts

| Agent | Verdict |
|---|---|
| code-reviewer | CLEAN (1 IMPORTANT fixed) |
| test-analyzer | 2 gaps fixed, 4 improvements noted as non-blocking |
| silent-failure-hunter | CLEAN — all 5 index paths verified, merge parity perfect |
| type-design-analyzer | Well-designed types, 1 R3 gap fixed |

All CI checks passing.

namasl:
## Review: scope + docs accuracy

**Blocking issues:** None — PR faithfully implements #1193 without scope creep. Default off, TTL event + O(1) request-ID index, `gateway_expired` INV-1 bucket, INV-13 parity preserved in `replay.go`. Non-tail removal correctly uses slice-shift rather than `removeEntryByIndex` (which panics on non-tail).

**Nits (non-blocking):**

1. **`docs/plans/phase6-request-ttl-plan.md` (+299 lines)** — largest single file in the PR. If recent convention is to archive plans after merge rather than keep them in `docs/plans/`, worth confirming before merge.

2. **`sim/cluster/metrics.go:105` comment** — the updated INV-1 comment now includes both `gw_expired` *and* `encode_routing_rejections`. The previous version was already missing `encode_routing_rejections`, so this is a drive-by fix; worth noting in the commit/PR description so reviewers don't read it as part of the Phase 6 addition.

3. **`GatewayQueueTTLEvent.Priority() = 6`** — the inline comment claims "after eviction (5), before scaling (8)". The event ordering is determinism-safe regardless via `(timestamp, priority, seqID)`, but a quick sanity check that no other priority-6 event exists on the cluster event queue would lock the comment's claim.

4. **Minor naming** — issue #1193 uses `expiredCount`; implementation uses `gatewayExpired`. The latter is more consistent with neighboring counters (`gatewayEvicted`, `gatewayQueueShed`) so this is the right call — flagging only because the issue text will look slightly out of sync with the code.

Docs are otherwise consistently updated across `CLAUDE.md`, `invariants.md`, `metrics.go`, `progress_hook.go`, both conservation tests, the startup log string, and the new CLI example.

namasl (approved):

PR #1362, opened 2026-05-13
fix: make blis-pr-review strictly read-only, move fix loop to impleme…

jgchn:
Make blis-pr-review strictly read-only on human PRs


susiejojo (approved):
lgtm

PR #1407, opened 2026-06-03
feat(workload): per-member unique prefix groups in cohort expansion

### LINKED ISSUE #1406: feat(workload): per-member unique prefix groups in cohort expansion

Cohort expansion (`ExpandCohorts` in `sim/workload/cohort.go`) assigns the same `PrefixGroup` string to every expanded member. Since `generatePrefixTokens` generates one token sequence per unique group name, all members of a cohort share identical prefix tokens and therefore hit the same KV cache entry. This mismodels workloads where each user/session has a distinct initial context — e.g. code-generation (per-session repository map), deep-research (per-session document context), interactive-chat (per-conversation dynamic system prompt). These are exactly the workloads where the catalog tags prefix-aware routing as a key optimization.

The result is that any BLIS experiment using cohort-based workloads to evaluate prefix routing or KV cache hit rates gets artificially inflated results: the simulation treats each user's unique context as a globally shared prefix, when in reality only the fixed `shared_system_prompt_len` portion would be shared.

The manual workaround — defining N explicit clients each with a unique `prefix_group` name — works but is impractical at population scale (e.g. 40 clients × multi-turn spec is hundreds of YAML lines).

**Proposed solution**

Change one line in `ExpandCohorts` (`sim/workload/cohort.go:40`):

```go
// Before: all members share one cache entry
PrefixGroup: cohort.PrefixGroup,

// After: each member gets its own cache entry
PrefixGroup: fmt.Sprintf("%s-%d", cohort.PrefixGroup, j),
```

When `cohort.PrefixGroup` is set, each expanded member `j` gets group name `"<prefix_group>-<j>"`. `generatePrefixTokens` then produces one unique token sequence per member, giving independent KV cache entries across the population while still sharing within a session's own turns (correct behavior).

This matches inference-perf's `num_conversations` model: each conversation has its own unique context that is not shared with other conversations.

- [x] Workload generation (`sim/workload/`)

**Extension friction check**
- 1 file, 1 line changed (`cohort.go`)
- No new interface needed; extends existing cohort expansion logic
- No effect on invariants (INV-1 through INV-13): prefix token generation is deterministic from seed — changing the group-name formula changes which tokens are generated but preserves determinism (INV-6). Conservation, causality, and KV block accounting are unaffected.
- `ExpandCohorts` is a pure function; the change is fully covered by existing cohort expansion tests plus a new test asserting that expanded members have distinct prefix groups and distinct token sequences.

**Alternatives considered**

1. **Schema addition: `per_session_prefix: true` flag on CohortSpec** — adds API surface for a behavior that should arguably be the default for any cohort with a prefix. Rejected in favour of the simpler always-unique-per-member approach.
2. **New `prefix_distribution` field** — a sampled-length-per-session prefix. More expressive but significantly more complex; can be a follow-on. This issue covers the naming fix that unblocks correct KV cache modelling immediately.
3. **Manual N-client workaround** — works today but scales poorly and defeats the purpose of cohorts.

Discovered while auditing inference-perf workload catalog conversions in `workloads/`. The shared-prefix problem affects `inference-perf-code-generation.yaml`, `inference-perf-deep-research.yaml`, and `inference-perf-interactive-chat.yaml` — all workloads the catalog tags with `"Prefix aware routing"` as a suitable optimization.

namasl:
## Summary

- Fix cohort expansion to assign each member a distinct `PrefixGroup` (`<prefix_group>-<j>`) instead of sharing one group name across all members
- Previously, all cohort members shared one KV cache entry, artificially inflating prefix-cache hit rates in experiments using cohort workloads
- Empty `prefix_group` is preserved as empty — no unintended groups created

## Issue

Closes #1406

## Cross-Path Parity

| `blis run` | yes | yes | `ExpandCohorts` feeds `GenerateRequests` which calls `generatePrefixTokens` |
| `blis replay` | no | N/A | Replay uses trace-recorded `PrefixGroup` values directly, not cohort expansion |
| `blis observe` | no | N/A | Observe dispatches to real server; workload generation path does not use `ExpandCohorts` |

## Invariants Affected

- **INV-6 (Determinism):** Preserved. The per-member group name is a deterministic function of `cohort.PrefixGroup` and `j` — same inputs produce identical output.
- **INV-1 (Request conservation):** Unaffected. No change to request counts or routing.

## Test Plan

- [x] Unit tests pass (`go test ./...`)
- [x] `go vet ./sim/workload/...` clean
- [x] `TestExpandCohorts_PrefixGroup_EachMemberGetsUniqueGroup` — each expanded member has a distinct, non-empty PrefixGroup
- [x] `TestExpandCohorts_PrefixGroup_NamePattern` — groups follow `<prefix_group>-<j>` pattern exactly
- [x] `TestExpandCohorts_NoPrefixGroup_RemainsEmpty` — cohorts without `prefix_group` produce empty strings (no regression)
- [x] `TestExpandCohorts_PrefixGroup_DistinctTokenSequences` — `generatePrefixTokens` produces one distinct token sequence per member

---

### DIFF
--- a/docs/guide/workloads.md
+++ b/docs/guide/workloads.md
@@ -414,6 +414,7 @@ Cohorts support the same advanced client features as explicit clients. The field
 | `closed_loop` | \*bool | `null` (omitted): `true` for multi-turn, `false` for all others. `true`: each round waits for the previous reply. `false`: all rounds pre-generated at open-loop arrival times |
 | `timeout` | \*int64 | Per-request timeout in µs. `null`: 300 s default when closed-loop; no deadline when open-loop. `0` = no timeout |
 | `prefix_length` | int | Shared prefix token count prepended to every request |
+| `prefix_sharing` | string | `"shared"` (default, omit to use): all members share one prefix group → one KV cache entry, modelling a global system prompt. `"per_member"`: each member gets an independent prefix group → independent KV cache entries, modelling per-session contexts (code-gen repo map, per-user chat history). Requires `prefix_group`. |
 | `network` | object | Client-side network RTT and bandwidth simulation |
 | `multimodal` | object | Mixed-modality token generation (text + image/audio/video) |
 
--- a/docs/reference/workload-spec.md
+++ b/docs/reference/workload-spec.md
@@ -104,6 +104,7 @@ Each entry in the `cohorts` list defines a population with lifecycle dynamics. C
 | `input_distribution` | object | **Yes** | Input token length distribution |
 | `output_distribution` | object | **Yes** | Output token length distribution |
 | `prefix_group` | string | No | Prefix group name |
+| `prefix_sharing` | string | No | `"shared"` (default): all members share one prefix. `"per_member"`: each member gets its own prefix. Requires `prefix_group`. |
 | `streaming` | bool | No | Whether to simulate streaming output |
 | `rate_fraction` | float64 | **Yes** | Fraction of `aggregate_rate` for each client in this cohort |
 | `diurnal` | object | No | Sinusoidal rate modulation (see below) |
--- a/sim/workload/cohort.go
+++ b/sim/workload/cohort.go
@@ -28,6 +28,14 @@ func ExpandCohorts(cohorts []CohortSpec, seed int64) []ClientSpec {
 		for j := 0; j < cohort.Population; j++ {
 			clientID := fmt.Sprintf("%s-%d", cohort.ID, j)
 
+			// Each member gets its own prefix group so generatePrefixTokens produces
+			// one unique token sequence per member. Without this, all members share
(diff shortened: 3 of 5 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

sriumcp:
@claude Will this PR enable users to create workloads, where within a cohort, there is flexibility to specify prefix-sharing (either, enabled or disabled; and if enabled, then specify the shared prefix length across the clients). Also, is the observe/replay/run parity achieved by this PR?

sriumcp (approved):
lgtm

PR #1408, opened 2026-06-04
fix(saturation): relativize time domain in backlog-drift detector

### LINKED ISSUE #1405: Bug: backlog-drift detector fails on replay with absolute-timestamp observe traces (INV-13 violation)
**Describe the bug**

The backlog-drift saturation detector (`--saturation-report`) produces empty results when run via `blis replay` on traces captured by `blis observe`. The detector assumes the simulation clock starts at 0, but replay preserves the absolute Unix timestamps from the observe trace (e.g., `arrival_time_us ≈ 1.78e15`). This causes `computeWindowMetrics` to compute a `totalDurationUs` of ~113 years, triggering the 7-day safety guard and returning zero windows.

The same bug affects the `blis observe --saturation-report` backlog-drift path (when `--post-hoc-detector` is `none`, the default).

**CLI commands to reproduce**

```bash
# 1. Observe a real server (captures absolute Unix timestamps)
./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
  --workload-spec workload.yaml --trace-header trace.yaml --trace-data trace.csv

# 2. Replay with saturation report — produces empty classification
./blis replay --trace-header trace.yaml --trace-data trace.csv \
  --model qwen/qwen3-14b --num-instances 4 \
  --saturation-report saturation.json

# 3. Observe with saturation-report (no --post-hoc-detector) — same failure
./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
  --workload-spec workload.yaml --trace-header trace.yaml --trace-data trace.csv \
  --saturation-report saturation.json
```

**Expected behavior**

`--saturation-report` should produce a valid `BacklogDriftReport` with windows covering the actual observation duration (~120s), identical in structure to what `blis run --saturation-report` produces. Per **INV-13 (Run/Replay parity)**: for any configuration supported by both commands, results must be equivalent.

**Actual behavior**

```json
{
  "classification": "UNSATURATED",
  "slope": 0,
  "windows": [],
  "note": "Observation too short: 0 windows < MinWindows=5. Cannot reliably classify saturation. Conservatively marking as UNSATURATED.",
  "recommendation": "Run longer observations (at least 300 seconds) for reliable classification."
}
```

Stderr warning:
```
level=warning msg="Saturation analysis: totalDurationUs (3560929781687290) exceeds reasonable limit (604800000000 us = 7 days), skipping window metrics"
```

**Root cause**

`sim/workload/saturation.go`:

1. `ComputeSimEndUs()` (line 18) returns `req.ArrivalTime + FirstTokenTime + Σ(ITL)` — an absolute Unix µs timestamp when the trace comes from observe.
2. `computeWindowMetrics()` (line 238) creates windows starting at `startUs = 0`, expecting `totalDurationUs` to be a relative duration from 0.
3. All intervals have `ArrivalUs ≈ 1.78e15` but windows start at 0, so even without the 7-day guard the data would fall into the last of ~29.7 billion windows.

**Affected code paths**

| Path | How `simEndUs` is computed | Bug triggered? |
|------|---------------------------|----------------|
| `blis run --saturation-report` | `ComputeSimEndUs()` — DES clock starts at 0 | **No** (always relative) |
| `blis replay --saturation-report` | `ComputeSimEndUs()` at `cmd/replay.go:717` — absolute Unix µs from observe trace | **Yes** |
| `blis observe --saturation-report` (backlog-drift) | `max(rec.LastChunkTimeUs)` at `cmd/observe_cmd.go:558-561` — absolute Unix µs | **Yes** |

**Proposed fix**

Relativize the time domain inside `AnalyzeBacklogDrift` (`sim/workload/saturation.go:386`) before calling `computeWindowMetrics`. This single fix location covers all three callers (run, replay, observe):

```go
// AnalyzeBacklogDrift (line 386) — after RequestsToIntervals, before computeWindowMetrics:

// Relativize time domain: shift intervals and simEndUs so windowing starts at 0.
// This handles absolute Unix timestamps from observe-derived traces while being
// a no-op when the DES clock already starts at 0 (blis run).
if len(intervals) > 0 {
    minArrival := intervals[0].ArrivalUs
    for _, iv := range intervals {
        if iv.ArrivalUs < minArrival {
            minArrival = iv.ArrivalUs
        }
    }
    if minArrival > 0 {
        for i := range intervals {
            intervals[i].ArrivalUs -= minArrival
            intervals[i].CompletionUs -= minArrival
        }
        simEndUs -= minArrival
    }
}
```

**Why this is safe for `blis run`**: When the DES clock starts at 0, `minArrival` will be 0 (or very close), so the subtraction is a no-op.

**Edge cases handled**:
- `len(intervals) == 0`: early-exit at line 391 fires before this code
- Still-in-flight requests: `CompletionUs = simEndUs + 1` → after fix: `(simEndUs + 1) - minArrival` = relative end + 1 ✓
- `minArrival > simEndUs`: impossible since completion ≥ arrival and `simEndUs ≥ max(completion_times)`

- [x] INV-13 (Run/Replay parity)

The backlog-drift detector works correctly on `blis run` (clock starts at 0) but silently produces empty/invalid results on `blis replay` and `blis observe` with absolute timestamps. This is a cross-command parity failure.

**Acceptance criteria**

1. `blis replay --saturation-report` on an observe-captured trace produces a non-empty window list and valid classification
2. `blis observe --saturation-report` (no `--post-hoc-detector`) produces the same
3. `blis run --saturation-report` behavior is unchanged (regression test)
4. New test case in `saturation_test.go` with absolute-timestamp intervals (e.g., `ArrivalUs ≈ 1.78e15`) verifies `len(windows) > 0`

**Environment**

- OS: macOS 15 (Darwin 24.6.0, arm64)
- Go version: 1.24.4
- Commit hash: 4e88b200

**Additional context**

Discovered during calibration workflow: ran `blis replay --saturation-report` on a trace from a 4-instance Qwen3-14B cluster observed via `blis observe`. The 7-day guard (added as safety for "MaxInt64 when no workload specified") masks the real issue. `blis run` never triggers this because the DES clock starts at 0. `blis observe` does its own saturation analysis with the composite detector on client-side data when `--post-hoc-detector` is set, so only the backlog-drift fallback path (the default) is affected.

Affected code paths:
- `cmd/replay.go:717` → `workload.ComputeSimEndUs()` → `workload.AnalyzeBacklogDrift()` → `computeWindowMetrics()`
- `cmd/observe_cmd.go:555-575` → `max(rec.LastChunkTimeUs)` → `workload.AnalyzeBacklogDrift()` → `computeWindowMetrics()`

susiejojo:
## Summary

- Fix backlog-drift saturation detector failing on absolute-timestamp traces from `blis observe`
- Relativize time domain in `AnalyzeBacklogDriftWithClassifier` before windowing — subtracts `minArrivalUs` from all intervals and `simEndUs`
- No-op for `blis run` (clock starts at 0), fixes both `blis replay` and `blis observe --saturation-report` paths

## Root Cause

`computeWindowMetrics` assumes timestamps start at 0, but observe-derived traces have absolute Unix µs timestamps (~1.78e15). This triggered the 7-day safety guard, returning zero windows and a false UNSATURATED classification.

## Behavioral Contracts

**BC-1:** Absolute-timestamp traces produce non-empty windows covering actual observation duration
**BC-2:** Zero-origin timestamps (blis run) produce identical results to before the fix
**BC-3:** Classification is preserved regardless of timestamp origin

## Test plan

- [x] New test: `TestAnalyzeBacklogDrift_AbsoluteTimestamps` — verifies non-empty windows with Unix epoch timestamps
- [x] New test: `TestAnalyzeBacklogDrift_AbsoluteTimestamps_ClassificationPreserved` — verifies classification matches relative-timestamp equivalent
- [x] All existing saturation tests pass unchanged (regression guard)
- [x] Full `go test ./...` passes

Fixes #1405

### DIFF
--- a/sim/workload/saturation.go
+++ b/sim/workload/saturation.go
@@ -686,6 +686,23 @@ func AnalyzeBacklogDriftWithClassifier(requests []*sim.Request, simEndUs int64, 
 		}
 	}
 
+	// Relativize time domain: shift intervals and simEndUs so windowing starts at 0.
+	// This handles absolute Unix timestamps from observe-derived traces (#1405) while
+	// being a no-op when the DES clock already starts at 0 (blis run).
+	minArrival := intervals[0].ArrivalUs
+	for _, iv := range intervals {
+		if iv.ArrivalUs < minArrival {
+			minArrival = iv.ArrivalUs
+		}
+	}
+	if minArrival > 0 {
+		for i := range intervals {
+			intervals[i].ArrivalUs -= minArrival
+			intervals[i].CompletionUs -= minArrival
+		}
+		simEndUs -= minArrival
+	}
+
 	// Step 2: Compute per-window metrics (BC-1)
 	windowSizeUs := int64(cfg.WindowSize / time.Microsecond)
 	windows := computeWindowMetrics(intervals, windowSizeUs, simEndUs)
--- a/sim/workload/saturation_test.go
+++ b/sim/workload/saturation_test.go
@@ -1235,3 +1235,97 @@ func TestSaturationProgression_TransitionBoundaries(t *testing.T) {
 		})
 	}
 }
+
+func TestAnalyzeBacklogDrift_AbsoluteTimestamps(t *testing.T) {
+	// GIVEN requests with absolute Unix µs timestamps (as produced by blis observe)
+	// WHEN AnalyzeBacklogDrift is called
+	// THEN the report produces non-empty windows covering the actual observation duration
+	//
+	// This tests the fix for issue #1405: the backlog-drift detector assumed clock starts
+	// at 0, but observe traces preserve absolute Unix timestamps (~1.78e15 µs).
+	cfg := NewBacklogDriftConfig(10*time.Second, 4, 2.0, 0.2, 0.95, 2, 1, 0.95, 0.98)
+
+	// Simulate an observe trace: absolute Unix timestamp base (~June 2026)
+	// with requests spanning 120 seconds of real observation
+	const baseTimeUs = int64(1_780_000_000_000_000) // ~June 2026 in µs
+	var requests []*sim.Request
+	for i := int64(0); i < 50; i++ {
+		requests = append(requests, &sim.Request{
+			ArrivalTime:    baseTimeUs + i*2_400_000, // Every 2.4 seconds over 120s
+			FirstTokenTime: 50_000,                   // 50ms TTFT
+			ITL:            []int64{10_000},           // 10ms per token
+			TTFTSet:        true,
+			State:          sim.StateCompleted,
+		})
+	}
+	// simEndUs is also absolute (as ComputeSimEndUs would return)
+	simEndUs := baseTimeUs + 120_000_000 // base + 120 seconds
+
+	report := AnalyzeBacklogDrift(requests, simEndUs, cfg)
+
+	// BC-1: Must produce non-empty windows (not trigger the 7-day guard)
+	if len(report.Windows) == 0 {
+		t.Fatalf("Expected non-empty windows for absolute-timestamp trace, got 0 windows.\nNote: %s\nRecommendation: %s",
+			report.Note, report.Recommendation)
+	}
+
+	// The actual observation is 120s with 10s windows → should have ~12 windows
+	if len(report.Windows) < 4 {
+		t.Errorf("Expected at least 4 windows for 120s observation with 10s window, got %d", len(report.Windows))
+	}
+
+	// Windows should cover relative time domain [0, ~120s], not absolute Unix timestamps
+	firstWindow := report.Windows[0]
+	if firstWindow.StartUs > 10_000_000 { // 10 seconds — should be near 0
+		t.Errorf("First window StartUs should be near 0 (relative), got %d", firstWindow.StartUs)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


susiejojo:
@claude /blis-pr-review

sriumcp:
@claude /blis-pr-review Does this PR achieve the issue resolution correctly? Will it lead to observe/replay/run parity, and correct calibration? Also, will it resolve the following comment correctly?

```shell
The issue is real, reproducible, well-evidenced, and correctly analyzes the root cause. The proposed fix is correct. The only gap:

The observe --saturation-report path is also affected (when --post-hoc-detector is none, the default). The backlog-drift branch at observe_cmd.go:555-590 passes absolute simEndUs and absolute-ArrivalTime intervals to AnalyzeBacklogDriftWithClassifier. The 7-day guard will fire on that path too for traces captured against a real server. The fix should document (or handle) this path as well.

The fix placement note says "insert in AnalyzeBacklogDrift" but that's the deprecated shim. The correct placement is AnalyzeBacklogDriftWithClassifier (which is what all callers now use).

Neither gap blocks implementation — the fix approach is sound and an implementer will encounter the observe path when verifying parity.
```

sriumcp (approved):
lgtm

Issue #1422, opened 2026-06-10
Hardening: staticcheck SA5011 nil-pointer-deref warnings in cmd/convert_test.go and sim/latency/config_test.go


sriumcp:
## Summary
`golangci-lint run ./...` reports 6 staticcheck **SA5011** (possible nil pointer dereference) issues in two test files. These are pre-existing on main (confirmed byte-identical to commit a185454f) and were surfaced during the #1417 PR review.

## Locations
- `cmd/convert_test.go`: lines ~14, 17, 22, 25 — `windowFlag`/`drainFlag` checked for `== nil` then dereferenced unconditionally on the next line.
- `sim/latency/config_test.go`: lines 89, 92 — `cfg` from `GetModelConfig` checked `== nil` (with `t.Fatal`) but staticcheck still flags the subsequent `.BytesPerParam` deref pattern.

## Reproduction
```bash
golangci-lint run ./cmd/ ./sim/latency/
# SA5011: possible nil pointer dereference (staticcheck) x6
```

## Fix direction
Restructure the nil checks so staticcheck can prove the deref is guarded — e.g. use `require.NotNil`/early `t.Fatalf` return that the analyzer recognizes, or guard the deref inside the `!= nil` branch. Test-only; no production impact.

## Notes
Out of scope for #1417 (does not touch the flagged lines). Filed per the PR-workflow 'Filing Pre-Existing Issues' rule.

Issue #1421, opened 2026-06-10
Bug: run/replay completed_requests mismatch — replay of an exported trace completes far fewer requests


sriumcp:
## Summary
A trace exported by `blis run --trace-output` and replayed with identical flags completes **far fewer requests** than the original run, and replay emits no per-request `requests` array in its metrics. This violates INV-13 (run/replay parity).

## Reproduction (stock dense model, NO DP/EP — fully on current main behavior)
```bash
MODELDIR=model_configs/qwen3-14b
blis run --model qwen/qwen3-14b --model-config-folder $MODELDIR \
  --hardware-config hardware_config.json --hardware H100 --tp 1 \
  --latency-model trained-physics --num-requests 20 --rate 5 --seed 123 \
  --total-kv-blocks 100000 --trace-output /tmp/t --metrics-path /tmp/run.json
blis replay --trace-header /tmp/t.yaml --trace-data /tmp/t.csv \
  --model qwen/qwen3-14b --model-config-folder $MODELDIR \
  --hardware-config hardware_config.json --hardware H100 --tp 1 \
  --latency-model trained-physics --total-kv-blocks 100000
```

## Observed
- run: `completed_requests: 20`, `injected_requests: 20`, per-request `requests` array present (20 entries).
- replay: `completed_requests: 8` (also seen 10 on an MoE config), `injected_requests: 20`, **no `requests` array** in the metrics block.

## Expected (INV-13)
Replaying an exported trace with identical flags should reproduce identical per-request metrics, hence identical `completed_requests`.

## Scope / notes
- Reproduces with **no DP/EP flags** on a dense model, so it is independent of the DP/EP work (#1417) — discovered during that PR's run→replay parity smoke.
- Likely cause to investigate: `mode: generated` trace-replay timing semantics vs the live run, or a horizon/warmup interaction (replay may apply a default horizon that truncates, or the recorded arrival/last-chunk timestamps interact with completion detection).

## Impact
Replay cannot currently be used to faithfully reproduce a run's per-request metrics for these configs — affects calibrate/replay-parity workflows.

Issue #1416, opened 2026-06-10
Tracking: DP + EP support in trained-physics latency backend (#A–#D)


sriumcp:
## Tracking issue: DP + EP support in the trained-physics latency backend (#A–#D)

**Parent umbrella:** #1348 (broad EP/DP vision). This issue is the **concrete, scoped deliverable** for the latency-modeling machinery — sub-issues #A–#D. Calibration (#E/#F) is deferred.

**Source of truth:** `docs/plans/2026-06-10-dp-ep-latency-scoped-design.md` (scoped decision-bearing design; overrides the proposal where they differ). Background proposal: `docs/plans/2026-06-09-dp-ep-support-design.md`. Discussion: [#1415](https://github.com/inference-sim/inference-sim/discussions/1415).

**vLLM parity target:** `vllm-project/vllm` @ `f6ec81c766c63e59216c0845b5cb3b1d05a19cfb` (v0.15.1; vendored in this repo's `vllm/` submodule at the same commit — line numbers below resolve in both). Every behavioral claim is backed by a commit-pinned permalink so a newcomer can verify it without trusting prose.

---

### The problem (3 bugs)

The trained-physics latency model (`sim/latency/trained_physics_model.go`) models **only tensor parallelism (TP)** — it divides every cost by TP and has no notion of **data parallelism (DP)** or **expert parallelism (EP)**. For MoE deployments this produces three errors:

- **B1 — routed-expert weight/compute not scoped to the vLLM MoE group.** vLLM shards routed experts over the flattened `TP·DP` group (EP-off: tensor-sharded; EP-on: expert-sharded). BLIS uses a batch-dependent `nEff/tp` that ignores DP entirely.
- **B2 — MoE dispatch/combine communication is unmodeled.** When `DP>1`, vLLM dispatches/combines hidden states across the MoE group. BLIS charges nothing for it.
- **B3 — shared experts are parsed but ignored** by `StepTime` (DeepSeek/Qwen run a shared expert for every token; the FFN dim is already parsed into `SharedExpertFFNDim` but never used).

### The vLLM facts the whole design rests on

All verified against the pinned commit. **These are the load-bearing facts — if you doubt the design, check these first.**

**Fact 1 — MoE group is flattened to `TP·DP·PCP` (PCP=1 here), in BOTH EP modes.** vLLM always computes a flattened group size first:
```python
# vllm/model_executor/layers/fused_moe/config.py:907
[...]

Issue #1348, opened 2026-05-12
Add Expert Parallelism (EP) and Data Parallelism (DP) support for MoE models


susiejojo:
## What external behavior are we replicating?

**System:** vLLM
**Component:** Parallelism configuration and MoE layer execution
**Specific behavior:** Expert Parallelism (EP) and Data Parallelism (DP) for Mixture-of-Experts models

When EP is enabled in vLLM:
- Expert layers are sharded across `EP_SIZE = TP_SIZE × DP_SIZE` ranks using All-to-All communication
- Attention layers use Tensor Parallelism (TP) if TP>1, or are replicated if TP=1
- DP creates model replicas that process independent batches (for MoE, requires synchronization)

### Permalink 1: ParallelConfig with EP/DP fields
https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/config/parallel.py#L98-L164

Key fields:
- `data_parallel_size` (line 107-109): "Number of data parallel groups. MoE layers will be sharded according to the product of the tensor parallel size and data parallel size."
- `enable_expert_parallel` (line 139-140): "Use expert parallelism instead of tensor parallelism for MoE layers."
- `expert_placement_strategy` (line 145-153): "linear" or "round_robin" expert distribution
- `all2all_backend` (line 154-163): Communication backend selection

### Permalink 2: Expert distribution logic (determine_expert_map)
https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/layer.py#L78-L164

Implements two placement strategies:
- **Linear placement** (lines 126-130): Experts distributed contiguously
  - Example: 16 experts, EP=4 → Rank 0: [0-3], Rank 1: [4-7], Rank 2: [8-11], Rank 3: [12-15]
- **Round-robin placement** (lines 131-138): Experts distributed cyclically
  - Example: 16 experts, EP=4 → Rank 0: [0,4,8,12], Rank 1: [1,5,9,13], Rank 2: [2,6,10,14], Rank 3: [3,7,11,15]

Returns `expert_map` tensor: global expert ID → local expert ID (or -1 if not on this rank)

### Permalink 3: Layer-specific parallelism behavior
[...]

PR #1424, opened 2026-06-10
feat(config): DP/EP config plumbing for trained-physics MoE (#1417)

### LINKED ISSUE #1417: #A — DP/EP config plumbing (ModelHardwareConfig + CLI flags)
## #A — Config plumbing for DP + EP

**Tracker:** #1416 · **Umbrella:** #1348 · **Source of truth:** `docs/plans/2026-06-10-dp-ep-latency-scoped-design.md` §3, §6.5 · **vLLM parity:** `vllm-project/vllm@f6ec81c7`

This is the foundation sub-issue: it adds the parallelism degrees to the config and exposes them on the CLI. No latency math changes here (that's #C) — but the helpers and validation defined here are consumed by #B/#C/#D.

### Why these specific semantics (vLLM proofs)

- **`EffectiveMoEGroupSize() = TP·DP` for MoE, `TP` for dense.** vLLM flattens the MoE group to `dp·pcp·tp` in *both* EP modes (only the partition axis differs): `flatten_tp_size = dp_size * pcp_size * tp_size` → [config.py:901-907](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/config.py#L901-L907). EP-off keeps it as `tp_size` ([config.py:1005-1018](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/config.py#L1005-L1018)); EP-on moves it to `ep_size` ([config.py:1020-1037](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/config.py#L1020-L1037)).
- **`EffectiveEP()` is an EP-mode predicate only** — `use_ep = (dp·pcp·tp > 1) and enable_expert_parallel` ([config.py:992-995](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/config.py#L992-L995)). It is **not** the EP-off sharding divisor (that's `MoEGroupSize`).
- **Reject dense `DP>1`.** vLLM treats non-MoE DP as N fully independent engines, not one instance ([core.py — non-MoE DP becomes independent processes](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/v1/engine/core.py#L1243-L1276) asserts `DPEngineCoreProc` is MoE-only). Dense DP is already expressible via BLIS's router-replica mechanism; routing it through the latency divisors would be a redundant second path.

### Tasks (TDD — write the test first)

**1. `ModelHardwareConfig` (`sim/config.go:98`).** Add two fields after `TP` (`sim/config.go:103`):
```go
DP                  int  // data-parallel degree (default 1)
EnableExpertParallel bool // mirrors vLLM --enable-expert-parallel
```
Update the canonical constructor `NewModelHardwareConfig` (`sim/config.go:111`) — **R4: single construction site**. The positional signature change forces every call site to recompile (intended). Inventory:
- production: `cmd/root.go:1632` (run), `cmd/replay.go:465` (replay), `cmd/blis-kvtime/main.go:913`.
- ~30 test sites (mechanical: pass `1, false`).

**2. Helpers** on `ModelHardwareConfig` (pure query, R13/R14):
```go
func (c ModelHardwareConfig) EffectiveDP() int           // max(1, DP)
func (c ModelHardwareConfig) EffectiveMoEGroupSize() int // MoE: TP·DP; dense: TP
func (c ModelHardwareConfig) EffectiveEP() int           // TP·DP iff EnableExpertParallel && isMoE; else 1
```
`isMoE` ≡ `ModelConfig.NumLocalExperts > 1` (consistent with the existing threshold at `sim/latency/config.go:243`).

**3. Validation (construction-time):**
- `DP >= 1` (else `panic`/`error` per library boundary).
- Allow `DP > 1` for MoE with either EP setting.
- **Reject `DP > 1` for dense models** (see proof above).

**4. CLI flags (`run` + `replay` only).** Package-level vars in `cmd/root.go` next to `tensorParallelism`:
- `--dp int` (default `1`)
- `--enable-expert-parallel` bool (default `false`)

Validate there (`DP >= 1`; reject dense `DP>1`; **reject `DP>1`/EP when backend ≠ `trained-physics`** — see INV BC-ROOFLINE in #C/tracker), then thread into `NewModelHardwareConfig`. Both commands call the constructor identically, so wiring the flags reaches both. Add the flags to **replay's** flag set + validation in lockstep (INV-13 parity). A replay omitting them defaults to `DP=1`, EP off.

- **`observe` gains no flags** — verified it builds **no** `ModelHardwareConfig` and **no** latency model (grep `cmd/observe*.go`). DP/EP have no meaning for observe (the real server's deployment is whatever it is).

### Behavioral contract / acceptance test
Table-driven over the four real `config.json` fixtures (Mixtral-8x7B, DeepSeek-V2-Lite, Qwen3-30B-A3B, Llama-4 Scout) asserting `(NumLocalExperts, NumExpertsPerTok, MoEExpertFFNDim, SharedExpertFFNDim, EffectiveMoEGroupSize, EffectiveEP)` for several `(TP, DP, EnableExpertParallel)` triples. The table **also records known dense-layer-placement gaps** (`numMoELayers` BLIS-today vs vLLM, e.g. DeepSeek's `first_k_dense_replace`) so the fixture does not over-claim parity.

CLI test: `run --latency-model roofline --dp 2 <moe>` must `Fatalf` (INV BC-ROOFLINE); `run --dp 2 <dense>` must `Fatalf` (dense rejection).

### Friction / invariants
- Files touched: `sim/config.go`, `cmd/root.go`, `cmd/replay.go` (+ recompile-only at the ~32 other call sites).
- No new interface. No new events. Affects no runtime invariant except the two new ones (BC-DP1 lands in #C; BC-ROOFLINE validation lands here).

**Depends on:** nothing. **Blocks:** #B, #C, #D.

sriumcp:
## Summary

Sub-issue **#A** of the DP+EP latency tracker (#1416, umbrella #1348). Adds the data-parallel (**DP**) and expert-parallel (**EP**) degrees to `ModelHardwareConfig` and exposes them on the `run`/`replay` CLI. **Config plumbing only** — the latency-math consumer (the `StepTime` divisor / shared-expert / dispatch refactor) is deferred to **#C**. The behavioral contracts and helper laws are captured in the test files and in the "Behavioral contracts" section below (a separate scoped design doc is not part of this repo).

Default `--dp 1`, EP off ⇒ behavior is **byte-identical to today** (INV-6 verified on a real run).

## What changed

- **`sim/config.go`** — `ModelHardwareConfig` gains `DP int` + `EnableExpertParallel bool`. The canonical constructor `NewModelHardwareConfig` (R4: single construction site) validates `DP >= 1` and rejects `DP > 1` for dense models. The positional signature change forces every call site to recompile (intended); ~150 mechanical test sites + 2 production sites updated.
- **Three pure-query helpers** (R13/R14) encoding vLLM's flattened-MoE-group parity:
  - `EffectiveDP() = max(1, DP)`
  - `EffectiveMoEGroupSize() = TP·DP` for MoE, `TP` for dense (PCP assumed 1)
  - `EffectiveEP() = TP·DP` iff `EnableExpertParallel && isMoE`, else `1`
  - `isMoE ≡ NumLocalExperts > 1` (matches the parsing layer / `ExtractKVCapacityParams`)
- **CLI flags** `--dp` / `--enable-expert-parallel` registered once in the **shared** `registerSimConfigFlags` and validated once in the **shared** `resolveLatencyConfig` — so `run` and `replay` get them through a **single code path** (INV-13 parity by construction, not by keeping two sites in sync).
- **Validation:** INV BC-ROOFLINE rejects `DP>1`/EP on non-trained-physics backends. Dense + `--dp > 1` fatals with a message explaining the vLLM online-serving equivalence (`--data-parallel-size` on a dense model = N independent engines = BLIS `--num-instances`). Dense + `--enable-expert-parallel` fatals (matches vLLM's `_verify_with_expert_parallelism` ValueError — the process cannot start).
- **Three real upstream MoE `config.json` fixtures** added (Mixtral-8x7B, DeepSeek-V2-Lite, Qwen3-30B-A3B) so the acceptance test parses all four fixtures through the production `GetModelConfig` path.

## Behavioral contracts

- **GIVEN** an MoE model **WHEN** built with `(TP, DP)` **THEN** `EffectiveMoEGroupSize = TP·DP`, and `= TP` for a dense model regardless of DP/EP.
- **GIVEN** `EnableExpertParallel` **WHEN** the model is MoE **THEN** `EffectiveEP = EffectiveMoEGroupSize`; otherwise `EffectiveEP = 1`. The flattened group is **EP-mode-independent**.
- **GIVEN** `--dp > 1` with a dense model **WHEN** resolving config **THEN** the CLI fatals with a message explaining the vLLM online-serving equivalence (N independent engines → `--num-instances`). Also fatals with a non-trained-physics backend.
- **GIVEN** `--enable-expert-parallel` on a dense model **THEN** the CLI fatals, matching vLLM's fatal rejection at `_verify_with_expert_parallelism` (the vLLM process would crash-loop before serving).
- **GIVEN** `--dp 1`, EP off **THEN** simulation output is unchanged (INV-6 byte-identical).
- **GIVEN** a trace exported by `run --dp D --enable-expert-parallel` **WHEN** replayed with identical flags **THEN** the config is reconstructed through the same constructor (INV-13).

## Testing

- Helper-law + constructor-validation unit tests (incl. DP=0/negative panic, dense-DP rejection, EP-mode-independence law, DP-clamp on zero-value struct).
- 4-fixture acceptance table over the real configs, asserting `(NumLocalExperts, NumExpertsPerTok, MoEExpertFFNDim, SharedExpertFFNDim, EffectiveMoEGroupSize, EffectiveEP)`; known dense-layer-placement gaps (DeepSeek `first_k_dense_replace`, Scout shared-expert) recorded as **non-asserted notes** so the fixture doesn't over-claim parity.
- CLI fatal-path tests (subprocess): roofline+EP rejection, dense-DP rejection, dense-EP rejection (vLLM parity), `--dp 0` rejection, defaults-accepted regression guard.
- Run/replay flag-parity + constructor-threading source guards.

Verification: `go build ./...` clean · `go test ./...` all pass · `golangci-lint run` clean on all changed files. Run→replay parity smoke: DP/EP flags thread identically through both paths; DP/EP produce no step-time change yet (correct — consumer is #C).

## Scope note
It is **correct** that DP/EP have zero step-time effect and zero non-test consumers today — `StepTime` consumes the helpers in #C. This PR delivers and tests the plumbing + validation only.

- #1421 (bug) — `run`/`replay` `completed_requests` mismatch; reproduces on a stock dense model with no DP/EP flags (pre-existing, orthogonal to this PR).
- #1422 (hardening) — pre-existing staticcheck SA5011 in `cmd/convert_test.go` and `sim/latency/config_test.go`.

Closes #1417


sriumcp:
## 🤖 BLIS PR Self-Review — Verdict: READY TO MERGE

Two BLIS-specific review passes (correctness/invariants/parity + adversarial silent-failure/gate-ordering) plus the standard multi-perspective code review (substance, tests, comments). **Zero CRITICAL, zero IMPORTANT.**

### Confirmed
- **All #1417 tasks addressed.** Helper semantics match vLLM (MoEGroup=TP·DP MoE / TP dense; EP=group iff enabled+MoE; isMoE=NumLocalExperts>1).
- **R4** single construction site (only the constructor builds `ModelHardwareConfig{}`; `go build` proves all ~150 call sites updated). **R13/R14** pure single-module query helpers. Library uses `panic`, CLI uses `logrus.Fatalf` — boundary respected.
- **INV-13 parity by construction** — flags + validation + constructor call on the single shared path (`registerSimConfigFlags` → `resolveLatencyConfig` → identical `NewModelHardwareConfig`); `observe` correctly excluded.
- **INV BC-ROOFLINE** enforced; **INV BC-DP1** preserved (no step-time math changed). **INV-6** byte-identical stdout verified on a real run.
- **Gate ordering proven safe** by adversarial truth-table trace: `DP>1` can only reach the dense-rejection gate under `trained-physics`, where `modelConfig` is guaranteed populated — no wrong-reject / wrong-pass possible.
- **Preemption & timeout/deadline:** untouched (config plumbing only).
- **Tests behavioral, acceptance values independently grounded** in real upstream configs.

### Reviewer suggestion addressed
The adversarial pass flagged a latent `TP=0 → 0` divisor hazard in the helpers. Investigated: `NewLatencyModel` (trained-physics) **already** rejects `TP <= 0`, and every helper consumer goes through that factory — so the hazard is already prevented at the model-build boundary, and 3 existing tests rely on the TP=0 factory-validation seam. Resolution: **documented** in the constructor why TP is intentionally not guarded there (turning the concern into an explicit, defensible decision) rather than adding a redundant guard that would degrade those tests.

### Scope note
DP/EP having no step-time effect and no non-test consumers today is **correct** — the `StepTime` consumer is sub-issue #C. This PR is plumbing + validation only.

Verification gate: `go build ./...` clean · `go test ./...` all pass · `golangci-lint run` clean on all changed files.

sriumcp:
@claude /blis-pr-review

cmd/root.go
```diff
@@ -679,6 +681,36 @@ func resolveLatencyConfig(cmd *cobra.Command) latencyResolution {
...
+			"Dense data parallelism is expressed via router replicas (--num-instances), not the latency model.",
+			dataParallelism)
+	}
+	if enableExpertParallel && modelConfig.NumLocalExperts <= 1 {
+		// EP is a no-op for dense models (EffectiveEP stays 1); warn so a likely
+		// misconfiguration is visible rather than silently ignored. Not fatal:
+		// the flag is harmless here and rejecting it would over-constrain. Warning
+		// goes to stderr, so deterministic stdout (INV-6) is unaffected.
+		logrus.Warnf("--enable-expert-parallel has no effect on dense model %q (no MoE experts); ignoring.", model)
+	}
```
susiejojo, line 716: ### Gap 1: EP on dense → Warning vs Fatal (parity gap, BLIS is looser)

vLLM **fatally rejects** `--enable-expert-parallel` on dense models:

```python
# vllm/config/model.py:1017-1022
def _verify_with_expert_parallelism(self) -> None:
    if not self.is_moe:
        raise ValueError(
            "Number of experts in the model must be greater than 0 "
            "when expert parallelism is enabled."
        )
```

Called from [`verify_with_parallel_config`](https://github.com/inference-sim/vllm/blob/f6ec81c76/vllm/config/model.py#L1077-L1078) during [`VllmConfig.__post_init__`](https://github.com/inference-sim/vllm/blob/f6ec81c76/vllm/config/vllm.py#L540) — the process dies before serving.

BLIS only warns here. This means BLIS will simulate a deployment that **cannot exist in production** (the vLLM pod would crash-loop). Consider upgrading to `logrus.Fatalf` for parity — a user whose config would fail on vLLM should get the same signal from the simulator.

cmd/root.go
```diff
@@ -679,6 +681,36 @@ func resolveLatencyConfig(cmd *cobra.Command) latencyResolution {
...
+			"(got --dp=%d, --enable-expert-parallel=%t, --latency-model %s). The roofline backend is "+
+			"DP/EP-blind for step time.",
+			dataParallelism, enableExpertParallel, backend)
+	}
+	if dataParallelism > 1 && modelConfig.NumLocalExperts <= 1 {
+		// Dense DP is the router-replica mechanism, not a latency divisor (vLLM
+		// treats non-MoE DP as N independent engines).
+		logrus.Fatalf("--dp > 1 is only supported for MoE models (got --dp=%d for a dense model). "+
+			"Dense data parallelism is expressed via router replicas (--num-instances), not the latency model.",
+			dataParallelism)
```
susiejojo, line 707: ### Gap 2: DP > 1 on dense → Fatal always (parity gap, BLIS is stricter)

vLLM **allows** `--data-parallel-size > 1` on dense models in online/serving mode — it creates N independent engines with an internal LB. The dense-model rejection at [`parallel.py:590-594`](https://github.com/inference-sim/vllm/blob/f6ec81c76/vllm/config/parallel.py#L590-L594) only fires in the `else` branch (env-var fallback for offline/SPMD mode), not when `data_parallel_size` is set explicitly via CLI.

BLIS fatally rejects this universally. The error message correctly explains the mapping (`--num-instances`), but a user directly translating their vLLM serving config (`--data-parallel-size 4` on a dense model) would hit a fatal on what is a valid production configuration.

**Suggestion:** Keep the fatal (the `--num-instances` mapping is the right BLIS abstraction), but consider noting in docs/error message that vLLM online DP on dense = N independent engines = BLIS `--num-instances`.

sim/config.go
```diff
@@ -101,27 +101,101 @@ type ModelHardwareConfig struct {
...
+		Backend:              backend,
+		MaxModelLen:          maxModelLen,
+	}
+}
+
+// isMoE reports whether the model is a mixture-of-experts model. The threshold
+// (NumLocalExperts > 1) matches the parsing layer (sim/latency/config.go) and
+// ExtractKVCapacityParams: single-expert configs are dense-equivalent.
+func (c ModelHardwareConfig) isMoE() bool {
+	return c.ModelConfig.NumLocalExperts > 1
```
susiejojo: ### Gap 3: MoE detection threshold — `> 1` vs `> 0` (semantic divergence)

vLLM uses `num_experts > 0` as its MoE predicate:

```python
# vllm/config/model.py:1685-1686
@property
def is_moe(self) -> bool:
    return self.get_num_experts() > 0
```

BLIS uses `NumLocalExperts > 1`.

For a model with `num_local_experts=1`:
- vLLM: `is_moe = True` → EP validation passes, DP > 1 enters MoE path
- BLIS: `isMoE() = False` → EP triggers warning, DP > 1 fatals as dense

In practice no real model has exactly 1 expert, so this is low-risk. But the threshold difference is worth documenting in the `isMoE()` docstring as a **known divergence** from vLLM's `is_moe` property.

sim/config.go
```diff
@@ -101,27 +101,101 @@ type ModelHardwareConfig struct {
...
+	}
+	return c.DP
+}
+
+// EffectiveMoEGroupSize returns the size of the flattened MoE tensor-parallel
+// group. For MoE models this is TP·DP (mirroring vLLM's flattened dp·pcp·tp MoE
+// group; PCP is not modeled here and is assumed 1), used by both the EP-off and
+// EP-on MoE paths. For dense models it is just TP. This is the sharding divisor
+// for routed-expert weights/compute.
+func (c ModelHardwareConfig) EffectiveMoEGroupSize() int {
```
susiejojo, line 183: ### Note: PCP=1 assumption is correct for now

vLLM formula: `flatten_tp_size = dp_size × pcp_size × tp_size` ([`config.py:901-909`](https://github.com/inference-sim/vllm/blob/f6ec81c76/vllm/model_executor/layers/fused_moe/config.py#L901-L909)).

BLIS correctly documents PCP=1. The llm-d wide-EP deployment manifests do not use PCP > 1. No action needed — just confirming the assumption holds for current production topologies.

susiejojo (commented):
## Parity Review: BLIS DP/EP Validation vs vLLM

Reviewed against vLLM at [`inference-sim/vllm@f6ec81c`](https://github.com/inference-sim/vllm/tree/f6ec81c76). Found 4 parity gaps worth discussing — see inline comments.

susiejojo:
### Gap 4: Expert field resolution priority (pre-existing, not introduced by this PR)

This PR's `isMoE()` depends on `NumLocalExperts` which is populated by `sim/latency/config.go:245-253`. The resolution order is:

```go
// BLIS: sim/latency/config.go:245-253
numLocalExperts := getInt("num_local_experts")   // ← primary
if numLocalExperts <= 1 {
    for _, key := range []string{"num_routed_experts", "n_routed_experts", "num_experts"} {
        if v := getInt(key); v > 1 { numLocalExperts = v; break }
    }
}
```

vLLM's resolution order is **different** ([`model_arch_config_convertor.py:86-98`](https://github.com/inference-sim/vllm/blob/f6ec81c76/vllm/transformers_utils/model_arch_config_convertor.py#L86-L98)):

```python
# vLLM priority: num_experts → moe_num_experts → n_routed_experts → num_local_experts
```

For a model with **both** `num_experts=256` and `num_local_experts=64` (hypothetically): vLLM sees 256 (global total), BLIS sees 64 (picks `num_local_experts` first). This could produce different `EffectiveMoEGroupSize` calculations.

**Impact on this PR:** Low — the helpers (`EffectiveMoEGroupSize`, `EffectiveEP`) will produce correct results *given* their input. The concern is whether the input (`NumLocalExperts`) itself matches what vLLM would resolve for the same HF config.

Not actionable here (pre-existing in `sim/latency/config.go`), but worth tracking as a potential follow-up if real HF configs surface with conflicting expert-count fields.

cmd/root.go
sriumcp: **Resolved** in d7f74cf2 — upgraded from `logrus.Warnf` to `logrus.Fatalf`.

BLIS now matches vLLM's hard rejection: `--enable-expert-parallel` on a dense model (`!modelConfig.IsMoE()`) fatals, mirroring `_verify_with_expert_parallelism`'s `ValueError`. A user whose config would crash-loop the vLLM pod now gets the same signal from the simulator.

Test: `TestResolveLatencyConfig_DenseEP_Rejected` (subprocess, asserts exit 1 + "vLLM fatally rejects").

sriumcp:
@claude /blis-pr-review Look into the updated PR. Look into the resolution that we came up with and assess merge worthiness.

cmd/root.go
sriumcp: **Resolved** in 91d6097b — kept the fatal (the `--num-instances` mapping is the right BLIS abstraction, as you noted), and clarified the error message to explain the vLLM online-serving equivalence:

> `--dp > 1 is only supported for MoE models (got --dp=N for a dense model). In vLLM online serving, --data-parallel-size on a dense model creates N independent engines; the BLIS equivalent is --num-instances (router replicas), not --dp.`

So a user translating a `--data-parallel-size 4` dense serving config now gets actionable guidance toward the correct BLIS flag rather than an opaque rejection.

sim/config.go
sriumcp: **Resolved** across e9e323ce + 08080ad0, and the divergence is now documented (originally 6a56e195).

We deliberately **kept the `>= 2` threshold** rather than flipping to vLLM's `> 0`, because it's a BLIS-internal correctness guard, not a parity gap:

- BLIS's analytic weight/FLOP formulas read `NumLocalExperts` as a multiplier/divisor (`kv_capacity.go` `3·h·f·N + N·h`, `roofline.go` `(N-k)/N`, `trained_physics` `min(N,…)`). At `N=1` these misestimate. vLLM has no such issue — it builds a real `FusedMoE` kernel and doesn't special-case N=1.
- On **every real model the two thresholds agree** — no real HF config sets a total-count field to 0 or 1 (verified against vLLM's config classes + fixtures). So `>= 2` loses zero classification parity and protects the formulas.

The threshold is now a single named constant `sim.MoEMinExperts = 2`, with `ModelConfig.IsMoE()` as the canonical predicate (replacing the scattered inline `> 1`/`<= 1` checks) and `(*HFConfig).ResolveNumExperts()` as the single resolver shared by both parse paths (R23). The rationale + vLLM divergence is in the `MoEMinExperts` docstring. Keystone test `TestExtractKVCapacityParams_SingleExpert_ClassifiedAsDense` (N=1 ⇒ dense) is mutation-verified to fail if the threshold is ever flipped to `> 0`.

sim/config.go
sriumcp: Acknowledged — thanks for confirming the PCP=1 assumption. No change needed here; `EffectiveMoEGroupSize = TP·DP` matches vLLM's `flatten_tp_size` with PCP=1, and a future PCP>1 extension would slot in as an additional factor. Left as-is.

sriumcp:
### Re: Gap 4 (expert field resolution priority) — **Resolved** in 08080ad0 (+ e9e323ce)

The original note flagged this as pre-existing and "not actionable here," but since we were already unifying the resolution logic, we fixed it properly:

**`(*HFConfig).ResolveNumExperts()` now matches vLLM's `get_num_experts` field order and membership:**

```go
var moeExpertCountFields = []string{
    "num_experts",       // Jamba
    "moe_num_experts",   // Dbrx   ← was MISSING in BLIS before
    "n_routed_experts",  // DeepSeek
    "num_local_experts", // Mixtral
    "num_routed_experts", // BLIS-historical alias (vLLM lacks it; kept for compat)
}
```

This corrects both issues from the comment: (1) BLIS previously checked `num_local_experts` **first** (inverted vs vLLM); now it follows vLLM's order. (2) BLIS previously **omitted** `moe_num_experts` (Dbrx); now included.

**One intentional difference remains, by design:** vLLM's `getattr_iter` is *first-field-that-exists*; BLIS returns *first-field-that-is-`>= MoEMinExperts`*. These differ only when a field is present with value 0 or 1 — which no real config does (verified). The threshold rule additionally protects BLIS's analytic formulas from a degenerate N=1 (see the Gap 3 thread). So on the hypothetical `num_experts=256 + num_local_experts=64` case from your comment, BLIS would now resolve **256** (matching vLLM), where before it resolved 64.

**Unification (R23):** both `GetModelConfigFromHF` (run/replay parse path) and `ExtractKVCapacityParams` (KV-capacity path) now call this single resolver — they can no longer desync. Pinned by `TestResolveNumExperts_ParityAcrossEntryPoints`, which includes synthetic two-field rows locking the order (mutation-verified: reverting either path to the old chain fails the test).

sriumcp:
## Review resolution summary

All review comments addressed. Map of comment → commit:

| Comment | Resolution | Commit(s) |
|---|---|---|
| **Gap 1** — EP on dense should be fatal (vLLM parity) | `Warnf` → `Fatalf`; matches vLLM's `_verify_with_expert_parallelism` | `d7f74cf2` |
| **Gap 2** — DP>1 on dense fatal message unclear | Kept fatal; message now explains vLLM online-DP = N engines = `--num-instances` | `91d6097b` |
| **Gap 3** — MoE threshold `>1` vs `>0` | Kept `>1` (BLIS-internal guard for analytic formulas), documented as intentional vLLM divergence; centralized as `sim.MoEMinExperts` + `ModelConfig.IsMoE()` | `6a56e195`, `e9e323ce` |
| **Gap 4** — expert-field resolution priority | `ResolveNumExperts()` now matches vLLM's `get_num_experts` order + adds `moe_num_experts` (Dbrx) | `08080ad0`, `e9e323ce` |
| **PCP note** | Acknowledged; no change | — |

### Beyond the literal comments — a reuse refactor

Gaps 3 and 4 shared a root cause: the "is this MoE?" question and the expert-count resolution chain were **duplicated** as inline integer comparisons across the codebase (the reason R23 code-path-parity existed at all). Rather than patch each site, this PR unifies them:

- **`sim.MoEMinExperts` + `ModelConfig.IsMoE()`** — one MoE-detection predicate; replaced the scattered inline `> 1`/`<= 1` checks at all detection sites.
- **`(*HFConfig).ResolveNumExperts()`** — one resolver shared by `GetModelConfigFromHF` and `ExtractKVCapacityParams`, so they can't desync (R23 is now structural, not clerical).
- Arithmetic sites that read the literal expert count (weight/FLOP formulas) were deliberately **left reading the raw integer** — only boolean detection sites moved behind the predicate.

### Self-audit

Ran an adversarial end-to-end audit (2 independent bug-hunting passes + the 3 BLIS PR-review agents). Outcome: **no CRITICAL/IMPORTANT findings.** Highlights:
- Every `>1`→`IsMoE()` / `<=1`→`!IsMoE()` substitution proven boundary-exact for all N≥0.
- The threshold guard and R23 parity were **mutation-tested** — flipping `MoEMinExperts` to vLLM's `>0`, or reverting one resolver path, both fail the new tests.
- INV-6 (determinism), INV-13 (run/replay parity) confirmed; verified end-to-end with a byte-identical Mixtral DP=2 run.
- One near-miss caught and corrected: an early pass called the `kv_capacity.go` MoE guard "redundant" — it's actually a defensive guard on a public positional constructor (removing it would allow a degenerate N=0 to silently produce zero MLP weight). Kept and documented.

Only MINOR/LOW comment-accuracy nits surfaced; all fixed (`83993576`, `16609236`). `go build` / `go test ./...` / `golangci-lint` all green.

susiejojo (approved):
lgtm

PR #1431, opened 2026-06-11
feat(observe): add --prewarm-duration flag for system priming (#1430)

### LINKED ISSUE #1430: Add --prewarm-duration flag to blis observe for system priming before measurement

When `blis observe` starts sending workload requests, the target system (vLLM + llm-d) is cold:
- CUDA kernels not yet compiled/cached
- Memory allocators not primed
- vLLM batch scheduler hasn't seen real traffic
- EPP/router connections not established

The existing `--warmup-requests` flag sends the first N requests from the actual workload profile and excludes them from results. This has two problems:

1. At high target rates (e.g., 20 QPS near saturation), warmup at the same rate can itself trigger queue collapse on slow nodes — pre-breaking the system before measurement starts.
2. The warmup behavior is tied to the workload profile. Different workloads need different warmup strategies, and there's no generic solution.

Real experiment data shows all runs have 20-25x worse latency in the first 60s vs steady state, and one run (a3 at 20 QPS) collapsed entirely because it couldn't recover from the cold-start transient.

**Proposed solution**

Add a `--prewarm-duration` flag that runs a fixed, conservative priming phase before the real workload begins. This phase is independent of the workload profile:

```bash
blis observe --server-url http://localhost:8000 --model Qwen/Qwen3-14B \
  --prewarm-duration 60s \
  --workload-spec workload.yaml --trace-header trace.yaml --trace-data trace.csv
```

Prewarm behavior (hardcoded, not tied to workload):
- Concurrency: 4-8 simultaneous requests
- Input tokens: 256 (small, fast prefill)
- Output tokens: 64 (short decode, completes quickly)
- Arrival: constant rate (no Poisson spikes)
- Drain at end: wait for all prewarm requests to complete before starting real workload
- No output: prewarm requests never appear in trace_data

This exercises the full request path (client → gateway → EPP → vLLM → response) enough to warm everything, but can't overload because concurrency and token sizes are small and fixed. Works for any workload — whether 5 QPS or 50 QPS, short or long tokens, flat or phased.

The existing `--warmup-requests` flag remains unchanged for backward compatibility.

- [x] CLI (`cmd/`)
- [ ] Core simulator (`sim/`)
- [ ] KV cache (`sim/kv/`)
- [ ] New package needed

**Extension friction check**
- ~1-2 files change: `cmd/observe_cmd.go` (flag + prewarm loop before `runObserveOrchestrator`)
- No new interface needed — reuses existing `RealClient.Send()`
- No invariants affected — prewarm runs before measurement, nothing recorded

**Alternatives considered**

1. **Increase `--warmup-requests` count:** Simple but dangerous at near-saturation rates. At 20 QPS, 1200 warmup requests might itself cause collapse.
2. **`--warmup-rate` at fraction of target (e.g., 1/4):** Ties warmup to workload config. Not generic — user must calculate the right fraction for each workload.
3. **Two-phase (relaxed + soak at target rate):** Soak at target rate has the same collapse risk as option 1. Over-engineered for what's needed.

The proposed `--prewarm-duration` is preferred because it's generic (one flag, any workload), safe (can't overload), and simple to implement.

Motivated by root cause analysis of `quartic-jc-2instances` experiment where cold-start transients caused one baseline run to collapse entirely (a3 at 20 QPS). See `_run-real-experiments/llmd_bundles/root-cause-analysis/quartic-jc-2instances-debug.md`.

mtoslalibu:
## Summary

Closes #1430.

Adds `--prewarm-duration` to `blis observe` that runs a fixed, conservative priming phase before the real workload begins. This warms CUDA kernels, EPP connections, and memory allocators without risking queue collapse — regardless of the real workload's rate or token sizes.

**Usage:**
```bash
blis observe --server-url http://localhost:8000 --model Qwen/Qwen3-14B \
  --prewarm-duration 60s \
  --workload chatbot --rate 20 --num-requests 500 \
  --trace-header trace.yaml --trace-data trace.csv
```

**Prewarm behavior (hardcoded, not tied to workload):**
- Concurrency: 4 simultaneous requests
- Input tokens: 256 (small, fast prefill)
- Output tokens: 64 (short decode)
- Arrival: constant (no spikes)
- Drain: all prewarm requests complete before real workload starts
- No output: prewarm requests never appear in trace_data.csv

## Behavioral Contracts

- **BC-1:** Requests sent continuously for the specified duration
- **BC-2:** Prewarm requests never appear in trace output
- **BC-3:** All in-flight prewarm requests drain before real workload starts
- **BC-4:** Fixed parameters independent of workload config
- **BC-5:** Disabled by default (0 = no prewarm, backward compatible)
- **BC-6:** Respects context cancellation (Ctrl+C stops immediately)

## Files Changed

| File | What |
|------|------|
| `cmd/observe_cmd.go` | Flag registration, validation, `runPrewarm()` function, wiring |
| `cmd/observe_test.go` | Two behavioral tests (duration, cancellation) |
| `docs/guide/observe-replay-calibrate.md` | Flag table + "System Prewarming" section |
| `CLAUDE.md` | Usage example |

## Design choices

1. **Hardcoded parameters** — no `--prewarm-concurrency` or `--prewarm-tokens`. The point is a single flag that's safe for any workload. Configurability would be overengineering.
2. **Independent of `--warmup-requests`** — both can coexist. `--prewarm-duration` warms the system; `--warmup-requests` excludes initial real-workload requests from trace.
3. **`time.AfterFunc` + `close(done)`** — broadcasts stop signal to all goroutines (unlike `time.After` channel which only fires once to one receiver).

## Test plan

- [x] `go test ./...` — all tests pass
- [x] `go vet ./...` — clean
- [x] `TestRunPrewarm_SendsRequestsForDuration` — verifies requests are sent for ~2s
- [x] `TestRunPrewarm_RespectsContextCancellation` — verifies early stop on cancel

## Invariants

No invariants affected:
- **INV-1 (request conservation):** prewarm requests never enter the recording pipeline
- **INV-6 (determinism):** observe is non-deterministic by nature
- **INV-13 (run/replay parity):** prewarm is observe-only; run/replay unaffected

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,6 +84,13 @@ go build -o blis main.go
   --think-time-dist "lognormal:mu=2.0,sigma=0.6,min=3s,max=30s" \
   --trace-header trace.yaml --trace-data trace.csv
 
+# Observe with system prewarm (recommended for cold systems, #1430)
+# Sends small fixed requests at low concurrency to warm CUDA/EPP/memory before measurement
+./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
+  --prewarm-duration 60s \
+  --workload chatbot --rate 10 --num-requests 100 \
+  --trace-header trace.yaml --trace-data trace.csv
+
 # Observe with ITL (inter-token latency) recording for streaming requests
 # --record-itl forces streaming on non-streaming workloads to capture per-chunk timestamps
 ./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,7 +85,10 @@ go build -o blis main.go
   --trace-header trace.yaml --trace-data trace.csv
 
 # Observe with system prewarm (recommended for cold systems, #1430)
-# Sends small fixed requests at low concurrency to warm CUDA/EPP/memory before measurement
+# --prewarm-duration: warms infrastructure (EPP/gateway connections, TCP pools)
+#   before measurement. Use for cold systems. Shape-independent.
+# --warmup-requests: excludes first N real-workload requests from trace.
+#   Use for statistical trimming. Rarely needed if --prewarm-duration is set.
 ./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
   --prewarm-duration 60s \
   --workload chatbot --rate 10 --num-requests 100 \
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -62,6 +62,7 @@ var (
 	observeRecordITL           bool
 	observeITLOutput           string
 	observeTimeout             int
+	observePrewarmDuration     time.Duration
 	// saturationReport is declared in root.go and shared across run, replay, observe
 )
 
@@ -128,6 +129,7 @@ func init() {
 	observeCmd.Flags().StringVar(&observeServerType, "server-type", "vllm", "Server type (vllm, tgi, etc.)")
 	observeCmd.Flags().IntVar(&observeMaxConcur, "max-concurrency", 256, "Maximum simultaneous in-flight requests")
 	observeCmd.Flags().IntVar(&observeWarmup, "warmup-requests", 0, "Number of initial requests to exclude from trace")
+	observeCmd.Flags().DurationVar(&observePrewarmDuration, "prewarm-duration", 0, "Duration of system priming phase before real workload (e.g., 60s). Sends small fixed requests at low concurrency to warm CUDA/EPP/memory. 0 = disabled.")
 	observeCmd.Flags().BoolVar(&observeNoStreaming, "no-streaming", false, "Disable streaming (use non-streaming HTTP)")
 	observeCmd.Flags().Int64Var(&observeSeed, "seed", 42, "RNG seed for workload generation")
 	observeCmd.Flags().Int64Var(&observeHorizon, "horizon", 0, "Observation horizon in microseconds (0 = from spec or unlimited)")
@@ -293,6 +295,9 @@ func runObserve(cmd *cobra.Command, _ []string) {
 	}
 	if observeWarmup < 0 {
 		logrus.Fatalf("--warmup-requests must be >= 0, got %d", observeWarmup)
+	}
+	if observePrewarmDuration < 0 {
+		logrus.Fatalf("--prewarm-duration must be >= 0, got %v", observePrewarmDuration)
 	}
 	if cmd.Flags().Changed("rate") && (observeRate <= 0 || math.IsNaN(observeRate) || math.IsInf(observeRate, 0)) {
 		logrus.Fatalf("--rate must be a finite value > 0, got %v", observeRate)
@@ -469,6 +474,11 @@ func runObserve(cmd *cobra.Command, _ []string) {
 		logrus.Warn("Received interrupt signal, cancelling observation...")
 		cancel()
 	}()
+
+	// Prewarm if requested (#1430)
+	if observePrewarmDuration > 0 {
+		runPrewarm(ctx, client, observePrewarmDuration)
+	}
 
 	// Run orchestrator
 	startTime := time.Now()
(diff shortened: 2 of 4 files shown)


mtoslalibu:
## Self-Review Summary

Ran full review (code quality, test coverage, silent failures). Verdict: **ready to merge** after one improvement — now addressed in e5e7bba.

### What the review found

**Critical:** 0
**Important:** 0  

**Suggestions addressed (2):**

1. **No user feedback when server is unreachable** — user would see "Prewarm complete" whether 0 or 1000 requests succeeded. Now logs a summary with request count + error count, and warns when all requests fail.

2. **CPU spin on instant failures** — if the server is down, the tight loop burned CPU for the full duration. Now backs off 500ms on errors.

### Cross-path parity

`--prewarm-duration` is correctly observe-only. `run` and `replay` use a DES with no real HTTP — warming CUDA/EPP is meaningless there. No parity gap.

### Preemption safety

Not affected. Prewarm runs entirely before `runObserveOrchestrator` — no trace records are written, no metrics are touched.

### Timeout/deadline consistency

Prewarm reuses the same `RealClient` with its configured HTTP timeout. If a prewarm request times out, the error is counted and the goroutine backs off, then retries. No interference with measurement.

### Invariants

- **INV-1 (request conservation):** prewarm requests never enter the recording pipeline
- **INV-6 (determinism):** observe is non-deterministic by nature
- **INV-13 (run/replay parity):** prewarm is observe-only; run/replay unaffected

### Standards compliance

- **R1:** Errors explicitly discarded (fire-and-forget by design), now with summary diagnostic
- **R3:** Negative duration validated with `logrus.Fatalf`
- **R4:** `PendingRequest` construction site uses zero-value defaults correctly

mtoslalibu:
Thanks for the thoughtful analysis — the CUDA kernel shape-dependency point is well-taken and the ecosystem comparison is useful context. Here's our reasoning for keeping the current design:

### The cold-start problem is mostly shape-independent

The quartic-jc root cause analysis showed 20-25x worse latency for the first 60s. That's far more than CUDA kernel compilation alone. The dominant contributors are:

- EPP/gateway connection establishment — shape-independent
- Memory allocator priming — shape-independent
- vLLM batch scheduler cold state — shape-independent
- TCP connection pooling to backend pods — shape-independent
- CUDA kernel compilation — shape-dependent (your point)

The fixed-shape prewarm addresses the bulk of the problem. The remaining CUDA kernel warmup is a valid follow-up optimization (e.g., a future `--prewarm-shape real` option), but it doesn't invalidate the generic approach.

### The two flags serve different purposes

- `--prewarm-duration`: warms the **infrastructure** before measurement starts. "Turn on the engine before driving."
- `--warmup-requests`: excludes **initial workload transients** from results (Poisson synchronization artifact at t=0). "Discard the first lap while finding your rhythm."

Using both together is legitimate: prewarm the system, then still exclude the first few real requests to absorb the synchronized arrival burst at t=0.

### The alternative is actually more complex

`--warmup-requests 200 --warmup-concurrency 4` requires:
1. Changing existing `--warmup-requests` semantics (it currently sends at full target rate)
2. Users must calculate N ("how many requests at concurrency 4 to warm for 60s with *my* workload's token distribution?")
3. Still subject to duration unpredictability at large token sizes

With `--prewarm-duration 60s`, the answer is always "60s" regardless of workload. The simplicity is the feature.

### Different context from vLLM benchmark tools

vLLM `benchmark_*` and guidellm benchmark a single server they control. blis observe targets any OpenAI-compatible server through a gateway (EPP + Envoy + vLLM). The gateway/routing warmup is the dominant cold-start component and is entirely shape-independent.

### No backward compatibility concern

This is purely additive — the flag defaults to 0 (disabled). Existing users see zero behavior change. It's opt-in for teams like ours who run against cold systems and want stable measurement baselines. Anyone who doesn't need it simply doesn't use it.

---

We'll keep the CUDA kernel shape concern in mind as a potential follow-up if we observe that prewarm doesn't sufficiently warm specific workload code paths. But for the infrastructure-dominated cold-start we're solving, the generic approach is correct.

susiejojo:
## Review: Design feedback on `--prewarm-duration`

Thanks for the solid implementation — the drain semantics, context cancellation, error backoff, and behavioral contracts are all well done. The cold-start problem is real and worth solving.

After digging into this more, I agree with your response that the cold-start is infrastructure-dominated (EPP/gateway connections, TCP pools, Envoy routing) rather than CUDA kernel compilation — vLLM's `compile_or_warm_up_model()` already handles kernel warmup, CUDAGraph capture, and FlashInfer autotune at startup before serving. So the hardcoded request shape is fine for what's actually being warmed. My earlier CUDA shape concern was misplaced.

### Remaining concern: two flags, one decision

My main concern isn't about shape — it's about user experience. With this PR, `blis observe --help` shows:

```
--warmup-requests    Number of initial requests to exclude from trace
--prewarm-duration   System priming phase before real workload
```

These look like competing alternatives to the same question: "how do I handle cold-start?" A user has to understand the distinction between "prewarm" and "warmup" (which are synonyms in plain English) to pick the right one.

More concretely: **would a user ever want both?** If the prewarm works (infrastructure is warm after 60s), request 0 of the real workload hits a warm system. There's no systematic reason to also discard it with `--warmup-requests`. The "Poisson synchronization at t=0" justification is a single-batch, sub-millisecond artifact — not something you'd exclude 50 requests (2.5s at 20 QPS) to avoid.

If you'd never meaningfully use both together, they're competing alternatives for the same lifecycle stage. Having both makes users think they *should* consider using both, adding cognitive load for zero practical benefit.

### Suggestion: unify under one concept

A single mechanism that replaces both:

```bash
# Duration-based (new behavior, what this PR solves):
blis observe --warmup-duration 60s ...

# Count-based (existing behavior, backward compat):
blis observe --warmup-requests 50 ...
```

Both send at throttled concurrency (default 4), both exclude from trace, both use the same code path. The user picks one based on whether they think in time or request count. One concept, one decision, no overlap.

`--warmup-requests` could gain a `--warmup-concurrency` option for the throttling (defaulting to safe value), making the existing flag handle the collapse-at-high-rates problem that motivated this PR.

---

To be clear: the feature is needed and the implementation quality is high. I'm only suggesting a naming/UX consolidation so users face one clear choice rather than two overlapping flags.

susiejojo:
### Minor suggestion: "which flag?" guidance

Since both flags will coexist, it would help to add a short clarification so users don't have to guess. Two small additions:

**1. In `docs/guide/observe-replay-calibrate.md`** (the "System Prewarming" section already added by this PR), append:

> **Which warmup flag?** Use `--prewarm-duration` when targeting a cold system (fresh restart, new connections). Use `--warmup-requests` only if you need to exclude initial requests for other reasons (e.g., rate ramp-up). You typically don't need both.

**2. In `CLAUDE.md`** (usage examples section), add a note near the prewarm example:

```
# --prewarm-duration: warms infrastructure (EPP/gateway connections, TCP pools)
#   before measurement. Use for cold systems. Shape-independent.
# --warmup-requests: excludes first N real-workload requests from trace.
#   Use for statistical trimming. Rarely needed if --prewarm-duration is set.
```

Small cost, prevents the "which one do I pick?" moment.

susiejojo (approved):
lgtm

PR #1432, opened 2026-06-12
feat(sim): ExpertPlacement interface + BalancedPlacement (#1418)

sriumcp:
## Summary

- Adds `sim.ExpertPlacement` — a single-method, pure-query contract (R13/R14) that maps a step's routed-token population onto **busiest-GPU** MoE cost, plus the default `BalancedPlacement` (`imbalanceFactor = 1`). Returning the busiest GPU lets "step = max over GPUs" physics emerge automatically once a future strategy (EPLB / skewed routing / redundant experts) introduces load skew.
- Lives in **`sim` core** (`sim/expert_placement.go`), not `sim/latency/`: expert placement is a model-*deployment* concept (same domain as `ModelConfig` and the parallelism degrees). Future consumers (scheduler, KV-capacity) must reach it without importing `latency/`, which would invert `cmd/ → sim/cluster/ → sim/`. The latency model already imports `sim`, so it gets `sim.ExpertPlacement` for free.
- This is the seam #C (#1419) consumes; no consumer is wired in this PR.

### Formulas (`BalancedPlacement`)

```
PerGPUComputeTokens = globalTokens · kEff / moeGroupSize
PerGPUExpertCount   = numExperts / moeGroupSize
PerGPUCommTokens    = (globalTokens / dp) · kEff · (moeGroupSize-1)/moeGroupSize · 2
```

- `PerGPUExpertCount = numExperts/moeGroupSize` matches **both** vLLM EP modes (`moeGroup = TP·DP`): EP-off tensor-sharded full-expert-equivalent bytes, EP-on whole experts per GPU — the device count is invariant across EP modes, only the partition axis changes.
- `PerGPUCommTokens` is divided by `dp` because it is a **per-rank** latency paid by the busiest source GPU (each DP rank owns ~`globalTokens/dp` disjoint sequence tokens), not an aggregate cluster byte count. The `(moeGroup-1)/moeGroup · 2` factor is the standard dispatch+combine all-to-all volume.

## Issue

Closes #1418 · Tracker #1416 · Source of truth: `docs/plans/2026-06-10-dp-ep-latency-scoped-design.md` §4 · vLLM parity `vllm-project/vllm@f6ec81c7`

## Cross-Path Parity

| Path | Applies | Implemented | Notes |
|------|---------|-------------|-------|
| `blis run` | N/A | N/A | Pure library type; no CLI/event surface. No consumer wired yet (#C wires it via the shared `NewLatencyModel` constructor, preserving INV-13 by construction). |
[...]

Issue #183, opened 2026-02-18
Bug: KV allocation failure path silently drops request from simulation


sriumcp:
## Description

In `sim/simulator.go`, the `Step()` function's completion path has a `continue` statement after KV allocation failure that silently drops the request from the simulation. The request is neither marked as completed nor added to the `remaining` slice, causing it to disappear entirely.

## Location

`sim/simulator.go` — inside `Step()`, in the request completion loop (~line 590):

```go
ok := sim.KVCache.AllocateKVBlocks(req, req.ProgressIndex, req.ProgressIndex+1, []int64{})
if !ok {
    logrus.Errorf("[tick %07d] KV allocation failed for completing request %s — this indicates a cache accounting bug", now, req.ID)
    continue  // ← request silently lost
}
```

## Impact

- The request is not added to `remaining` (so it won't be retried)
- The request is not marked as completed (so no metrics are recorded)
- The request vanishes from the simulation, making throughput/completion metrics inaccurate
- This path should be unreachable (blocks are already allocated for running requests), but if it is reached, the silent loss makes debugging harder

## Possible Fixes

1. **Add to `remaining`** to retry allocation in the next step
2. **Force completion anyway** and track a `KVAllocationFailures` counter
3. **Panic with context** since continuing in an inconsistent state is dangerous for a simulator that needs deterministic, accountable results

## Discovery

Found during code review of #38 (log level recalibration). The pre-existing `logrus.Warnf("[THIS SHOULD NEVER HAPPEN]")` was upgraded to `logrus.Errorf(...)` with context, which prompted closer inspection of the error path.

PR #1433, opened 2026-06-12
feat(latency): StepTime DP/EP refactor + multi-backend MoE comm (#1419)

### LINKED ISSUE #1419: #C — StepTime refactor: DP/EP divisors, shared-expert, MoE comm, β_EP
## #C — `StepTime` refactor (the heart): divisors, shared-expert, MoE comm, β_EP

**Tracker:** #1416 · **Umbrella:** #1348 · **Source of truth:** `docs/plans/2026-06-10-dp-ep-latency-scoped-design.md` §5 · **vLLM parity:** `vllm-project/vllm@f6ec81c7`

The only hot-path change. Fixes B1 (routed-expert scoping), B2 (MoE comm), B3 (shared experts). Keep the single-pass, zero-allocation structure of `TrainedPhysicsModel.StepTime`.

Compute once per step (MoE models): `dp := float64(m.dp)`, `moeGroup := m.EffectiveMoEGroupSize()`, `ep := m.EffectiveEP()`, and `m.placement.Resolve(globalTokens, kEff, numExperts, moeGroup, dp)`.

---

### Part 1 — Divisor map (term by term, NOT a blanket `/tp → /(tp·DP)`)

**Gain a `/dp`** (sequences split across DP ranks — each rank processes `~globalTokens/dp` tokens):
- Attention KV read/write (`bytesPfKv`, `bytesDcKv`): `dKV/tp → dKV/(tp·dp)`.
- Attention/decode projection compute (the `/tp` token-population terms): also `/dp`.
- Dense-FFN compute (interleaved models): `/tp → /(tp·dp)`.
- Shared-expert compute: `/(tp·dp)`.

**Stay `/tp`** (weights replicated across DP groups): attention weights, dense-FFN weight, shared-expert weight.

**vLLM proofs:**
- Attention heads sharded by TP only, DP replicates (`num_heads // tp_size`, `tp_size = get_tensor_model_parallel_world_size()`): [qwen3_moe.py:229-242](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/models/qwen3_moe.py#L229-L242). ⟶ the `hPerGPU` head-sharded attention FLOPs are **not** touched by DP; the DP effect rides the token-count path only.
- Disjoint per-rank token split: [core.py:1243-1276](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/v1/engine/core.py#L1243-L1276).

### Part 2 — Routed-expert terms → `ExpertLoad` (B1 fix)
- Weight bytes use `PerGPUExpertCount` (= `numExperts/moeGroup`) in place of today's batch-dependent `nEff = min(N, max(k, B·k))/tp`. **Applied unconditionally** — all MoE weight loading routes through `ExpertPlacement`, including `DP=1`/EP-off.
- Compute uses `PerGPUComputeTokens`.
- ⚠️ **This intentionally changes existing MoE step-time output** vs today (the saturation-point behavior the model targets). Not covered by byte-identity. See "Behavior-change boundary."

### Part 3 — Shared-expert term (B3 fix)
Gated on `SharedExpertFFNDim > 0` (already parsed at `sim/latency/config.go:306`): compute `/(tp·dp)`, weight `/tp`, scaled by MoE-layer count.
- **vLLM proof:** shared experts run unconditionally per token ([shared_fused_moe.py:59-96](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/shared_fused_moe.py#L59-L96)); shared MLP weight sharded over the **attention TP group** (size `TP`, not `TP·DP`) via `MergedColumn/RowParallelLinear` ([deepseek_v2.py:288-365](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/models/deepseek_v2.py#L288-L365)).
- Documented **no-op for Llama-4 Scout** (no config field exposes its shared-expert dim → `SharedExpertFFNDim = 0`).

### Part 4 — MoE-FFN communication taxonomy (B2 fix)

**Convention (basis vs. coefficient):** every `t*` below is a **raw basis** (µs before its β). The coefficient is applied **once**, in the final step-time sum. Where this writes `tMoEDispatch = …`, the sum contributes `m.Beta[10] · tMoEDispatch`.

vLLM partitions MoE-FFN comm on the **`DP` boundary** (see tracker Fact 3). Helper (unchanged from today): `V(units, group) = units · totalTokens · hidden · 2 · 2 · (group-1)/group / bwHbmUs` (the `·2·2` = BF16 bytes × ring phases). Today's monolithic `tTp` (`allReduceUnits = 2·numDenseLayers + numMoELayers`, `sim/latency/trained_physics_model.go:281-283`) splits into:

**Reduction-class (coefficient `β₄`, `Beta[3]`):**
- `tTpAttention = V(numLayers, tp) / dp` — attention all-reduce, scoped to the local TP group, `1/dp` (DP groups run in parallel).
- `tTpDenseFFN = V(numDenseLayers, tp) / dp` — dense-FFN all-reduce, TP-only.
- `tMoEReduce = V(numMoELayers, tp)` — **regime 1 only** (`DP=1`, `TP>1`):
  ```
  gateReduce = isMoE && numMoELayers > 0 && DP == 1 && tp > 1
  ```
  vLLM reduces over `tp_size or ep_size` (both = TP at DP=1). 0 at any `DP>1` or `TP=1`.
  **Proof:** [layer.py:1561-1569](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/layer.py#L1561-L1569)

**Dispatch/combine (coefficient `β_EP`, `Beta[10]`):**
`PerGPUCommTokens` already carries the per-source-rank top-k and the `(moeGroup-1)/moeGroup · 2` dispatch+combine factors (#B), so do **not** re-multiply by `kEff`:
```
tMoEDispatch = (PerGPUCommTokens · hidden · bpp / bwHbmUs) · numMoELayers
gateDispatch = isMoE && DP > 1
```
The gate is `DP > 1` — **independent of EP** — because the dispatch path is gated on `dp_size > 1`, not `use_ep`:
- naive (EP-off *and* EP-on-non-chunked): `do_naive_dispatch_combine = dp_size > 1 and not modular` — [layer.py:1792-1826](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/layer.py#L1792-L1826). `use_ep` only conditions a post-quant allgather there, **not** the dispatch.
- chunked all2all (EP-on low-latency backends): `use_all2all_kernels = dp_size > 1 and use_ep` — [config.py:867-869](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/config.py#L867-L869).

Both move hidden states over `moeGroup = TP·DP`, so one term covers both. `gateReduce` (`DP==1`) and `gateDispatch` (`DP>1`) **partition** the MoE-FFN comm on the DP boundary — no overlap, no gap → charged exactly once.

> **Mutual-exclusion mechanism (for the skeptic):** in vLLM, `forward_native` always calls `forward_impl` via the `moe_forward` op — they are *not* alternative entry points, both run every pass. What prevents double-comm is the `reduce_results` flag + `must_reduce_shared_expert_outputs()` ([layer.py:1542-1545](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/layers/fused_moe/layer.py#L1538-L1545)), not the `not use_dp_chunking` guard alone. The design's conclusion (comm charged once on the DP boundary) holds regardless of which gate enforces it.

### Part 5 — `β_EP` coefficient wiring (NON-PASSIVE — read carefully)

`β_EP` is the 11th coefficient → `Beta[10]`. Existing index map unchanged (`0=β₁ … 3=β₄ … 7=β₈, 8=β₉, 9=β₁₀`). Today (`sim/latency/trained_physics_model.go:369-370`):
```go
betaSlice := make([]float64, 10)
copy(betaSlice, coeffs.BetaCoeffs[:min(10, len(coeffs.BetaCoeffs))])
```
A naive grow to `make(..., 11)` leaves `Beta[10] = 0` for the common (7–10-coeff) callers, **silently disabling MoE dispatch comm.** The required change is three lines:
```go
betaSlice := make([]float64, 11)
copy(betaSlice, coeffs.BetaCoeffs[:min(11, len(coeffs.BetaCoeffs))])
if len(coeffs.BetaCoeffs) < 11 {
    betaSlice[10] = betaSlice[3] // β_EP defaults to β₄ when not explicitly provided
}
```
Extend `defaults.yaml` with an explicit 11th value (a provided value overrides the β₄ default). `β₈` kept as-is (re-fit deferred to #E).

### Behavior-change boundary
- **Dense at `DP=1`** — byte-identical to today (**INV BC-DP1**, golden test). The split is value-preserving: `tTpAttention + tTpDenseFFN = V(numLayers, tp) + V(numDenseLayers, tp) = V(2·numDenseLayers + numMoELayers, tp)` when `numMoELayers=0` — exactly today's `tTp`; `tMoEReduce=0` (not MoE), `tMoEDispatch=0` (DP=1).
- **MoE** output **intentionally changes** even at `DP=1`/EP-off: (1) B1 expert weight uses `numExperts/moeGroup` not `nEff/tp`; (2) `tMoEReduce` now charges `V(numMoELayers, tp)` which today's code drops (it deferred MoE-FFN comm to `β₈`, which is 0 for uniform MoE). Deliberate fidelity gain. Existing MoE tests assert structural properties + relative comparisons, not absolute golden step-times.

### Tests
- Golden (INV BC-DP1): dense `DP=1`, EP off → byte-identical across the existing matrix.
- `(TP=2, DP=2)` EP off: routed weight/compute use `moeGroup=4`; attention all-reduce stays local `TP=2`; `tMoEDispatch` active (naive); `tMoEReduce` absent.
- `(TP=2, DP=2)` EP on: routed weight/compute use `EP=4`; `tMoEDispatch` active (all2all); `tMoEReduce` absent. **The two `(TP=2,DP=2)` cells produce identical MoE-FFN comm** — the gate is `DP>1`, not EP.
- `(TP=2, DP=1)` EP off: `tMoEReduce` active (over `tp=2`); `tMoEDispatch` absent — the DP boundary.
- Shared-expert term present for DeepSeek-class, absent for Mixtral; a separate test records Scout as a known parser gap.
- `β_EP` default: 10-coeff caller yields `Beta[10] == Beta[3]`; 11-coeff caller yields the provided value.
- One fully-worked `(TP=2, DP=2)` hand-arithmetic example (proposal §6.5) as a golden with the arithmetic in a comment.

### Friction / invariants
- One file: `sim/latency/trained_physics_model.go`. Preserve single-pass / zero-alloc — the term table above is the review checklist.
- Introduces **INV BC-DP1**. INV-3/6 (determinism) preserved (no RNG, no new events).

**Depends on:** #A, #B. **Blocks:** #D (KV) is independent but lands after for a complete model.
**Risk:** medium (only hot-path change).

sriumcp:
## Summary

The heart of the DP/EP trained-physics work (#1416 #C). Refactors `TrainedPhysicsModel.StepTime` to model data parallelism and expert parallelism, consuming the `sim.ExpertPlacement` seam from #B (#1432), and adds an explicit multi-backend MoE communication model.

Fixes the three modeling bugs from the tracker:
- **B1** — routed-expert weight/compute now scoped to the flattened `moeGroup = TP·DP` via `ExpertPlacement` (`PerGPUExpertCount`/`PerGPUComputeTokens`), replacing the batch-dependent `nEff/tp`. EP-mode-agnostic, applied unconditionally for MoE.
- **B2** — MoE dispatch/combine communication (`tMoEDispatch`, β_EP) modeled per comm backend.
- **B3** — shared experts (`SharedExpertFFNDim`) now charged (compute `/(tp·dp)`, weight `/tp`, every token).

Plus the DP divisor map: token-population terms (projection/dense-FFN compute, KV read/write) gain `/dp`; weights stay `/tp`; head-sharded attention FLOPs untouched (DP rides the token-count path). The monolithic TP all-reduce splits into `tTpAttention + tTpDenseFFN + tMoEReduce`.

## Deviation from design §5 (multi-backend comm) — decided with the maintainer

Design §5 specified a **single `kEff`-bearing dispatch term** for all `DP>1`. Verified against `vllm-project/vllm@f6ec81c7`: vLLM's **default** MoE comm backend (`allgather_reducescatter`) and the `naive` path move **dense hidden states with no top_k**; only the opt-in low-latency kernels (pplx, deepep_*, mori, flashinfer) route per-`top_k`. So a single `kEff` term mismodels the default deployment. Instead this PR adds `--moe-comm-backend` (all 7 vLLM names) mapping to two physical volume families:

| Family | Backends | Per-rank dispatch volume |
|--------|----------|--------------------------|
| all-gather | `naive`, `allgather_reducescatter` (default) | `(globalTokens/dp)·(moeGroup-1)/moeGroup·2·hidden·bpp` — **no top_k** |
| modular all-to-all | `pplx`, `deepep_high_throughput`, `deepep_low_latency`, `mori`, `flashinfer_all2allv` | `PerGPUCommTokens·hidden·bpp` — carries `kEff` |

(The scoped design doc is an untracked local working file, not in the repo; this deviation is recorded in code comments, here, and on #1419/#1416.)

## β_EP default = β₄ (derived, not deferred)

β_EP (the new 11th coefficient, `Beta[10]`) defaults to β₄ for **both** families, derived from the NVIDIA NCCL bus-bandwidth model (`nccl-tests/doc/PERFORMANCE.md`): all-gather/reduce-scatter and all-to-all share the same `(n-1)/n` per-phase NVLink efficiency, and ring all-reduce (which β₄ corrects) *is* reduce-scatter+all-gather. The per-family **volume** difference lives in the dispatch basis, so the coefficient must not re-encode it. Corroborated by DeepEP's measured intranode NVLink ≈ 726–740 GB/s (≈0.22× H100 HBM, same band as the ~0.27 β₄ encodes). An explicit 11th coefficient overrides; per-family empirical refit is #E.

## Issue

Closes #1419 · Tracker #1416 · vLLM parity `vllm-project/vllm@f6ec81c7`

## Cross-Path Parity

| `blis run` | yes | yes | `--dp`, `--enable-expert-parallel`, `--moe-comm-backend` |
| `blis replay` | yes | yes | same flags via shared `registerSimConfigFlags` + `resolveLatencyConfig`; INV-13 parity verified (run e2e == replay e2e with `--dp 2 --moe-comm-backend deepep_low_latency`) |
| `blis observe` | N/A | N/A | dispatches to a real server; latency model not used |

## Invariants

- **INV-BC-DP1 (new):** dense `DP=1`/EP-off step time is byte-identical to pre-#1419 (golden across TP matrix). MoE output intentionally changes (B1 + `tMoEReduce`) — a deliberate fidelity gain. Registered in `invariants.md` + `CLAUDE.md`.
- **INV-13:** preserved by construction (run/replay share the constructor); smoke-verified.
- **INV-6:** determinism — pure arithmetic, no RNG/new events; byte-identical stdout confirmed.
- Roofline + (`--dp>1` | EP | `--moe-comm-backend`) rejected at validation.

## Test Plan

- [x] `go build ./...`, `go test ./... -count=1` (all packages), `golangci-lint run ./...` (0 issues)
- [x] INV-BC-DP1 dense golden (byte-identical) + dense-determinism companion
- [x] β_EP default (10-coeff→`Beta[10]==Beta[3]`, 11-coeff→provided); unknown-backend rejected
- [x] DP divisors, B1 batch-independence, MoE comm DP-boundary taxonomy, `tp=1,dp>1` mutex cell
- [x] Per-family dispatch volume **formulas pinned exactly** (all2all/all-gather ratio == kEff)
- [x] Shared-expert present/absent/Scout-no-op; worked `(TP=2,DP=2)` ExpertLoad example
- [x] `--moe-comm-backend` registered on run+replay; rejection paths (unknown name, non-trained-physics)
- [x] Scout MoE golden dataset regenerated (4 experiments, ~2-3% from B1+tMoEReduce; dense unchanged)
- [x] End-to-end smoke: two backends produce different comm cost; run/replay parity exact

## Discovered during self-audit

- Removed a dead `ep` field (EP has no separate step-time divisor; EP-independence is behaviorally tested).

---


sriumcp:
## 🤖 BLIS PR Self-Review — Verdict: CHANGES REQUIRED → addressed (round 1)

A multi-perspective self-review verified the divisor map, the per-family dispatch volumes against vendored `vllm@f6ec81c7`, the tMoEReduce/tMoEDispatch DP-boundary partition, β_EP wiring, INV-BC-DP1, INV-13 parity, R1/R4/R13/R14, determinism, and test quality. **One IMPORTANT correctness bug** found and fixed in 78f9d39.

### IMPORTANT (fixed)

**MoE dispatch comm was sized by the weight dtype instead of the activation dtype.** `moeDispatchBasis` multiplied the hidden-state volume by `weightBPP` (quantized weight precision), but dispatch/combine moves hidden-state **activations**. Verified in `../vllm`: `NaiveAll2AllManager.naive_multicast` allocates the dispatch buffer with `dtype=x.dtype` (the BF16 hidden states); the quantized `post_quant_allgather` path is an explicit opt-in, not the default. For a quantized-weight MoE this under-sized comm by the weight/activation ratio (2× for FP8, 4× for W4A16) — invisible to the current FP16-only tests where `weightBPP == BytesPerParam == 2.0`.

**Fix:** added an `activationBPP` field (`= ModelConfig.BytesPerParam`, the compute dtype) and used it in both `moeDispatchBasis` branches, matching `tpAllReduceBasis` and the KV terms. Added `TestMoEDispatchBasis_UsesActivationDtypeNotWeightDtype` (FP8-weight MoE: weightBPP=1.0, activationBPP=2.0) which fails on the old code. No change to FP16 results — dense INV-BC-DP1 and the Scout goldens are unchanged.

### Verified correct (no action)
- Divisor map (`/tpdp` token terms, `/tp` weights, `hPerGPU` untouched); operator precedence.
- B1: batch-independent `PerGPUExpertCount`; two `Resolve` calls (prefill/decode, no conflation); old `nEff` removed.
- B2 family split faithful to vLLM (all-gather = no top_k; modular all2all = top_k); no double `/dp`, no kEff re-multiply.
- Gate mutual exclusion exhaustive + exclusive over the DP boundary (incl. tp=1,dp>1 dispatch-only; tp=1,dp=1 both off).
- B3 shared experts (every token, `/(tp·dp)` compute, `/tp` weight, Mixtral/Scout no-op).
- β_EP wiring (10→11, defaults to β₄, explicit overrides; defaults.yaml 11th = β₄).
- INV-BC-DP1 dense byte-identity (value-preserving split; real golden + companion law); Scout MoE regen justified (B1+tMoEReduce).
- INV-13 parity (shared registerSimConfigFlags + resolveLatencyConfig; smoke-verified run e2e == replay e2e).
- Determinism/zero-alloc preserved; no preemption/timeout/metric impact (only step-time magnitude).
- Per-family test pins the law `ratio == kEff`, not golden numbers.

### NIT (no action)
- Reviewer flagged the decode-KV "read+write" comment as a new edit; it is unchanged from `main` (verified via `git show main:`), so left as-is.

Verification: `go build ./...` ✅, `go test ./... -count=1` ✅ (all packages), `golangci-lint run ./...` ✅ 0 issues, determinism byte-identical, INV-13 run/replay parity exact.

sriumcp:
## 🤖 Comprehensive PR Review (4 parallel agents) — focus: correctness & accuracy vs vLLM @ f6ec81c7

Ran vLLM-parity correctness, general code, test-quality, and comment-accuracy reviews. Every vLLM claim was re-verified against the vendored tree. **One IMPORTANT correctness issue found and fixed**, plus test-coverage and comment gaps — all addressed in 810c9d7.

### IMPORTANT (fixed) — attention compute not divided by DP

The vLLM-parity audit found an internal inconsistency: **attention compute FLOPs (`flopsAttn`) were divided by `/tp` only, not `/(tp·dp)`**, while projection, dense-FFN, and KV all correctly got `/dp`. Verified against vLLM: each DP rank is an independent `DPEngineCoreProc` processing `num_tokens_across_dp[dp_rank]` — its own ~`tokens/dp` slice — so attention (run per-rank over that slice) must scale `1/(tp·dp)` like the other token-population terms. The head-sharding `/tp` stays; the missing factor was the `/dp` on the token axis.

**Fix:** divided both `flopsAttn` terms by `dp`. It's a **no-op at DP=1**, so dense INV-BC-DP1 byte-identity and the Scout (DP=1) golden are unchanged; only DP>1 MoE prefill is corrected (it had been over-charging attention compute by a factor of `dp`). Confirmed end-to-end: DP=2 Mixtral e2e dropped 5672.86→5672.69ms.

### Test gaps (fixed) — end-to-end witnesses for the new behaviors

The test-quality review correctly flagged that the family split and `tMoEReduce` were pinned only on basis *helpers*, not on `StepTime` itself (the `Beta[10]*tMoEDispatch` wiring could be dropped and tests would pass). Added:
- `TestStepTime_DispatchFamilyAffectsStepTime` — all2all > all-gather at DP>1 in a full step (guards the dispatch wiring).
- `TestStepTime_MoEReduceChargedAtDP1` — `tMoEReduce` reaches the step.
- `TestStepTime_AttentionComputeScalesWithDP` — doubling DP nearly halves compute-bound step time (guards the fix above).

### Comment/doc accuracy (fixed)

- **Scout shared-expert reference was wrong** (would misdirect the #E follow-up): it named `intermediate_size_mlp`, but vLLM's Scout shared expert reuses `config.intermediate_size` (`llama4.py:94,105`); `intermediate_size_mlp` is the *dense* FFN, already mapped to `DenseIntermediateDim`. Corrected in code + guide.
- Refreshed the struct header (`8/9/10 beta` → `8–11` with the β_EP / MoE-comm terms), dropped stale "added in T5/later task" comments, fixed the `7-10 beta` / `10 elements` doc/comment counts, and made `moeCommBackends` a single source of truth for the name list + family lookup (eliminates the slice/switch drift risk).

### Verified correct vs vLLM (no change needed)

B1 routed-expert weight = `numExperts/(TP·DP)` (both EP modes, factor 3 = gate+up+down); B2 family split faithful (default `allgather_reducescatter` + naive = dense hidden, no top_k; modular = top_k); `tMoEReduce` gate & volume; **β_EP=β₄ default is sound and not an over-charge** (default backend dispatch is serial/non-overlapped in vLLM, so additive latency is right; magnitude ~plausible); B3 shared-expert sharding (`/tp` weight, `/(tp·dp)` compute); activation-dtype fix; defaults.yaml 11-beta; INV-BC-DP1 dense byte-identity; INV-13 parity. The DeepSeek/Qwen shared-dim parser handles both field conventions.

### Verification
`go build ./...` ✅ · `go test ./... -count=1` ✅ (all packages) · `golangci-lint run ./...` ✅ 0 issues · INV-6 determinism byte-identical · INV-13 run/replay parity exact (5672.6934 == 5672.6934 with `--dp 2 --moe-comm-backend deepep_low_latency`).

sriumcp:
## Why `testdata/trained_physics_iter29.json` changed (golden dataset)

This PR regenerates 4 entries of the trained-physics golden dataset. Since golden files are where bugs can silently bake in, here is the full justification.

### What changed
Exactly **4 of the 15 experiments** — all of them the **Llama-4 Scout** (MoE) model (IDs 17, 20, 21, 48). The other **11 entries are byte-identical** (Llama-2-7B, Llama-3.1-70B, Mistral-Nemo, Qwen2.5-7B, Yi-34B). The `alpha_coeffs`/`beta_coeffs` block at the top is **unchanged**. The deltas are small (~2-3%, e.g. Scout id=17 e2e 6300→6467 ms).

### Why it changed — intentional MoE fidelity gain, not a regression
The #C refactor's documented behavior-change boundary (INV-BC-DP1) is: **dense models are byte-identical; MoE models intentionally change.** The Scout shifts come entirely from the two MoE fixes:
1. **B1** — routed-expert weight now uses `numExperts/moeGroup` (batch-independent) instead of the old batch-dependent `nEff = min(N, max(k, B·k))/tp`.
2. **tMoEReduce** — the MoE-FFN all-reduce at DP=1/TP>1, which the previous code dropped entirely (it deferred MoE-FFN comm to β₈, which is 0 for uniform MoE), is now charged.

Scout runs at **DP=1**, so the new DP>1 dispatch term, the per-family comm model, and the activation-dtype fix do **not** touch these numbers — the change is purely B1 + tMoEReduce. Scout's shared-expert term is also **not** involved: its config exposes no `shared_expert_intermediate_size`, so `SharedExpertFFNDim=0` (a documented no-op). Values went slightly *up*, consistent with charging a previously-omitted all-reduce.

### How it was regenerated — tool-generated, not hand-edited
Produced with the project's official mechanism (`sim/internal/testutil` `-update-golden`), which runs the real cluster simulation per experiment and writes the simulator's actual output:
```
go test ./sim/cluster/ -run 'TestTrainedPhysics_GoldenDataset' -update-golden
```
No value was typed by hand. Re-running that command against the committed file now yields **zero diff** — i.e. every number is exactly what the DES emits.

### Why the new values are trustworthy (not just "what the code now prints")
Per CLAUDE.md golden-test rules (R7/R12, issue #183), the golden has independent guards that don't depend on the latency numbers:
- The test asserts `completed_requests` / token-conservation per experiment, so a silently-dropped-request bug can't hide behind the regen.
- INV-BC-DP1 is independently pinned by `TestINVBCDP1_DenseStepTimeByteIdentical` + a companion law test in `sim/latency/`, and the dense entries here being byte-identical is the cluster-level confirmation of it.
- The B1/tMoEReduce magnitudes were verified from first principles against `vllm@f6ec81c7` (expert sharding over the flattened TP·DP group; the DP=1 MoE-FFN all-reduce at `fused_moe/layer.py:1561`).

### One convention note for reviewers
The file header says *"Update only when the trained-physics backend is intentionally versioned (rename backend first)."* I did **not** rename the backend, on the rationale that #C is a fidelity *fix* to the existing `trained-physics` backend (the design explicitly frames the MoE change as intended) rather than a new backend version, and the #1419 test plan calls for regenerating this golden. If reviewers prefer the stricter reading of that header (version/rename the backend first), I'm happy to do that instead — flagging it explicitly so it's a conscious decision.

sriumcp:
@claude /blis-pr-review Please review the PR (with its new scope of supporting various vLLM comm backends). Check for soundness and correctness.

sriumcp:
## How `β_EP` should be trained per comm backend (design note for #E)

A reasonable question on this PR: *there are 7 comm backends — should we train a separate `β_EP` for each?* Here is the reasoning, from scratch, for anyone picking this up later.

### Background: what `β_EP` actually is

When `DP > 1`, an MoE step pays a **communication cost** to shuffle token activations between GPUs (the "dispatch/combine" step). BLIS models that cost in two pieces:

1. **The basis** — a physics formula that computes *how many bytes* cross the wire, normalized by HBM bandwidth into a raw time. This is `moeDispatchBasis(...)`.
2. **The coefficient `β_EP`** — a single dimensionless number multiplied onto the basis to turn that idealized byte-time into the *observed* wall-clock time. It absorbs real-world effects the basis doesn't model: the NVLink-vs-HBM speed ratio, collective-algorithm efficiency, kernel overheads.

So: **`step_comm_time = β_EP × basis`**. The basis says "how much data"; `β_EP` says "how efficiently the hardware moves it."

### Key insight: the basis already separates the backends into 2 groups

The 7 vLLM backends fall into **two families** that move *different amounts of data*:
- **all-gather family** (`naive`, `allgather_reducescatter` = the default): every token's activation is shared once — **no top-k factor**.
- **all-to-all family** (`pplx`, `deepep_*`, `mori`, `flashinfer`): each token is sent to its top-k experts — volume carries a **`kEff` (top-k) factor**, so it moves roughly `kEff×` more data.

That volume difference is captured in **the basis, not the coefficient.** This is why, in this PR, all backends ship with the *same* default `β_EP` (= β₄): the families already differ in the basis, so the coefficient shouldn't double-count it. You can see this in practice — at TP=2, DP=2 on Mixtral, the all-gather backends and the all-to-all backends produce two distinct step times, purely from the basis.

### So should #E train 7 values, 2 values, or 1?

**Not 7 (per-backend).** The differences *within* a family (e.g. `deepep_high_throughput` vs `deepep_low_latency`) are kernel-implementation details. The most important one — HT is tuned for prefill, LL for decode — is a **prefill-vs-decode *shape* difference**, and a single scalar fundamentally cannot represent a shape. Fitting 7 scalars would just memorize benchmark noise (and 5 of the 7 need exotic GPU kernels to even measure). If we ever need that fidelity, the right tool is a small per-backend *profile* (a structured cost model), not seven magic numbers.

**Start with 1 (shared), and let the data decide whether to split to 2 (per-family).** The #E experiment is:
1. Fit one `β_EP` on all-gather measurements, one on all-to-all measurements.
2. Test whether they are statistically distinguishable.
3. **If indistinguishable → keep a single shared `β_EP`.** (This is my prior: both are bandwidth-bound collectives over the same NVLink fabric, so after the basis accounts for the volume, the leftover efficiency should be ~the same.)

### What to do if the two families *are* distinguishable

This is the part that matters most, because the obvious move ("just use two coefficients") is often wrong. Before adding a second coefficient, **diagnose *why* they differ** — look at the residuals (the leftover error after the basis), not just the total fit error:

- **Case A — a flat, constant gap** (all-to-all is uniformly ~X% less efficient at every batch size, DP degree, and model). *Then two per-family coefficients is correct* — index `β_EP` by the family (the code already carries a `commFamily` field for exactly this). **This is the only case that justifies "two values."**

- **Case B — the gap changes with the workload** (the error grows with batch size, or differs between prefill and decode). This means **the basis volume formula is wrong for that family** — not that we need a second knob. A second coefficient would fit the *average* and be biased high at one end and low at the other, while *looking* like it improved the fit. The correct fix is to the basis (e.g. DeepEP dispatches in FP8 so its dispatch leg is half the bytes; low-latency kernels may overlap comm with compute, breaking the "add the latency" assumption). After fixing the basis, re-test — you may collapse back to a single coefficient.

- **Case C — the gap only shows up on one model or one GPU and doesn't replicate.** That's overfitting; keep a single coefficient.

**Summary:** fit per-family → test equality → collapse to one if indistinguishable. If distinguishable, check whether the gap is a *flat offset* (→ legitimately use two per-family coefficients) or a *workload-dependent shape* (→ the basis is wrong; fix the volume model, don't add a coefficient). Per-backend scalars are never the answer — a scalar can't represent the prefill/decode shape difference that drives within-family variation; that needs a profile.

The guiding principle throughout: **never let `β_EP` quietly paper over a wrong byte-volume formula.** That's also why #E should first *validate the basis* (does measured all-to-all comm actually scale with `kEff`?) before trusting any fitted coefficient.

(The shipped default in this PR — single `β_EP = β₄`, family volume difference carried by the basis — is the correct *starting point*; this note is guidance for the later empirical refit, tracked as #E.)

susiejojo:
## Review Summary

**vLLM parity**: 6/6 claims verified against `vllm-project/vllm@f6ec81c7` — full match. Backend names, default, volume families, naive path characterization, DP=1 all-reduce, and shared expert independence all confirmed.

**CI**: All green (build, lint, all test shards).

**Overall**: Technically sound and merge-ready after addressing the items below.

---

## Important Issues

### 1. `moeDispatchBasis` default branch silently accepts future families

**File**: `sim/latency/trained_physics_model.go`, `moeDispatchBasis` method

The `switch m.commFamily` uses `default:` for `commFamilyAllGather`. If a 3rd family is added in the future, it silently falls into the all-gather volume model — a potential factor-of-`kEff` error with plausible-looking (not NaN) output.

**Suggestion**: Enumerate both known cases and add `default: panic(...)`:
```go
case commFamilyAllGather:
    return (globalTokens / dpf) * (group-1) / group * 2 * hidden * m.activationBPP / m.bwHbmUs
default:
    panic(fmt.Sprintf("moeDispatchBasis: unhandled commFamily %d", m.commFamily))
```

Currently safe (upstream `moeCommFamilyFor` validates), but documents the contract and catches iota drift.

---

### 2. `activationBPP` silently inherits zero-value `BytesPerParam`

**File**: `sim/latency/trained_physics_model.go`, constructor

If `ModelConfig.BytesPerParam = 0` (parser bug or zero-value struct), `activationBPP` becomes `0.0`, silently zeroing out the MoE dispatch communication term. The step time remains positive (other terms contribute), so no panic fires.

**Suggestion**: Add constructor validation:
```go
if hw.ModelConfig.BytesPerParam <= 0 {
    return nil, fmt.Errorf("trained-physics: BytesPerParam must be > 0 "+
        "(needed for activation dtype sizing in MoE dispatch comm), got %f",
        hw.ModelConfig.BytesPerParam)
}
```

---

### 3. `tpAllReduceBasis` hardcodes `2.0` instead of `activationBPP`

**File**: `sim/latency/trained_physics_model.go`, `tpAllReduceBasis`

```go
return units * tokens * float64(m.hiddenDim) * 2.0 * 2.0 * tpFactor / m.bwHbmUs
//                                              ^^^ hardcoded BF16 bytes
```

Meanwhile `moeDispatchBasis` correctly uses `m.activationBPP`, and `TestMoEDispatchBasis_UsesActivationDtypeNotWeightDtype` explicitly guards that pattern. The same principle applies to TP all-reduce (it also moves activations, not weights). Not a live bug today (all models use BF16), but inconsistent with the moeDispatchBasis fix in commit 10/10. If kept as-is, deserves a comment explaining the BF16 assumption.

---

### 4. β_EP calibration procedure undocumented

The PR introduces β_EP (Beta[10]) with an active default of β₄, justified by the shared NVLink fabric argument. This is physically sound. However, there's no documentation of how to independently train β_EP for deployments where the argument breaks down (e.g., cross-node InfiniBand for EP with intra-node NVLink for TP). A brief note in `docs/guide/latency-models.md` describing the isolation experiment (DP>1 MoE, vary comm backend, fit residual with other 10 coefficients frozen) would make the coefficient actionable for future calibration.

---

### 5. `--moe-comm-backend` at DP=1 or on dense model: silent no-op

**File**: `cmd/root.go`, `resolveLatencyConfig`

The flag help states "MoE + --latency-model trained-physics + --dp > 1", but validation only checks the backend name and latency model — not whether the model is MoE or DP > 1. The flag silently has no effect in those cases.

**Suggestion**: Add a `logrus.Warnf` for the no-op cases:
```go
if moeCommBackend != "" && dataParallelism <= 1 {
    logrus.Warnf("--moe-comm-backend=%s has no effect at --dp=%d; "+
        "MoE dispatch/combine comm is only charged when DP > 1.",
        moeCommBackend, dataParallelism)
}
```

---

## Suggestions (non-blocking)

- **Empty batch edge case**: No test exercises `StepTime(nil)` or `StepTime([]*Request{})` on a MoE model. If `ExpertPlacement.Resolve(0, ...)` doesn't handle zero tokens gracefully, this could panic.
- **Interleaved MoE + DP>1 test**: All DP/EP tests use uniform-MoE (`InterleaveMoELayerStep=0`). A Scout-style interleaved config at DP=2 would exercise `tTpDenseFFN` and `tMoEDispatch` simultaneously.
- **MoE golden pin test**: `TestINVBCDP1_DenseStepTimeByteIdentical` pins dense values but no equivalent MoE golden exists at the unit-test level (the iter29.json dataset covers it implicitly).

sriumcp:
Thanks @susiejojo — thorough review, and I agree with all 5 Important items. Addressed in 2efef8d. Point by point:

**1. `moeDispatchBasis` default branch → explicit cases + panic.** Done. Both families are now explicit `case`s; `default:` panics on a future 3rd family rather than silently inheriting the all-gather (no-`kEff`) volume. Documents the contract and catches iota drift exactly as you suggested.

**2. `BytesPerParam = 0` silently zeroes dispatch comm.** Confirmed real — the HF parser yields `BytesPerParam = 0` for an unrecognized `torch_dtype` (config.go), and it would zero *every* activation-movement term (KV, TP all-reduce, MoE dispatch) while step time stayed positive. Added a constructor guard mirroring the `TFlopsPeak`/`BwPeakTBs` validation, plus `TestNewTrainedPhysicsModel_RejectsZeroBytesPerParam`. (No existing test relied on a zero value — all trained-physics configs already set it.)

**3. `tpAllReduceBasis` hardcodes `2.0`.** Good catch on the inconsistency with the `moeDispatchBasis` activation-dtype fix. Rather than just comment it, I made it correct: it now uses `m.activationBPP` (the compute dtype), matching `moeDispatchBasis` and the KV terms. I verified **every golden model is 16-bit** (`float16`/`bfloat16` → `BytesPerParam = 2.0`), so this is **value-neutral for all goldens** (zero churn) while being correct for future non-BF16 activations. Comments updated to drop the "BF16 bytes" wording.

**4. β_EP calibration undocumented.** Added a "Calibrating β_EP" subsection to `docs/guide/latency-models.md`: the isolation experiment (DP>1 MoE, freeze the other 10 β, fit the dispatch residual), with the note that fitting is per comm-family, not per backend (and a pointer to the per-backend-granularity discussion above on this PR).

**5. `--moe-comm-backend` no-op at DP=1/dense.** Added `logrus.Warnf` (not fatal — the flag is harmless there, just inert) for both the dense-model and DP=1 cases. Verified it fires.

**Suggestions:**
- *Empty batch* — already guarded (`len(batch)==0 → return 1`, so `Resolve(0,…)` is never reached), but added `TestStepTime_EmptyBatch_MoE` as explicit insurance on a MoE model.
- *Interleaved MoE + DP>1* — added `TestStepTime_InterleavedMoE_DP2` (Scout-style, exercises `tTpDenseFFN` and `tMoEDispatch` together). Writing it surfaced a nice property: at top-1 routing (`kEff=1`) the two comm families move identical volume, so they *should* give identical step time — captured as `TestStepTime_CommFamiliesConvergeAtTopK1` (the `kEff=1` boundary of the `ratio == kEff` law).
- *MoE unit golden* — the dense byte-identity golden is the strong INV-BC-DP1 guard; the new end-to-end MoE step-time tests (family-affects-step, reduce-charged, attention-`/dp` scaling) pin the MoE behaviors as laws rather than golden numbers, which I think is the more robust coverage here.

Dense INV-BC-DP1 byte-identity, the Scout golden, INV-6 determinism, and INV-13 run/replay parity are all unchanged. Build + full `go test ./...` + `golangci-lint` green.

susiejojo (approved):
lgtm

Issue #1438, opened 2026-06-24
Lazy request generation to greatly reduce peak memory footprint


namasl:
**What antipattern or issue does this address?**

`blis run` materializes the entire workload — every request with its full input/output token arrays — in memory *before* the simulation event loop starts. For realistically shaped workloads (multi-turn conversations with shared prefixes and accumulating context), this produces a peak resident-memory spike in the tens of gigabytes that drops sharply once the event loop begins consuming requests. On systems without enough headroom the spike can trigger swap thrashing, OOM kills, or full desktop freezes — well before the simulation has done any useful work.

This issue proposes three orthogonal changes. Change A is the structural fix; Changes B and C address remaining factors and compose multiplicatively with it.

All code references below are permalinked to commit [`d059008`](https://github.com/inference-sim/inference-sim/commit/d059008f7abb008d69b2f01ff08b31e4372230ea) (`main` at filing time).

---

## Reproducer (workload shape)

The problem reproduces strongly on workloads with the inference-perf **code-generation** catalog shape:

- Multi-turn conversations with `context_growth: accumulate`
- A large shared prefix per cohort member (system prompt, repository map, retrieved context — typical: 30K tokens)
- A moderate cohort population (e.g., 40 members)
- A moderate `max_rounds` (e.g., 15)
- A modest `num_requests` cap (e.g., 500)

This is not a pathological corner case — it directly mirrors realistic agentic-coding, RAG, and long-context-chat workloads.

---

## Root cause

Three independent multipliers compound during the eager generation phase invoked from [`cmd/root.go#L1455`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/root.go#L1455) (`workload.GenerateWorkload`):

**(1) Eager materialization.** The full request stream is built upfront as a `[]*sim.Request`, with each request carrying fully-realized `InputTokens []int` and `OutputTokens []int` slices. The DES then consumes requests one at a time at their arrival timestamps, so almost all of those token arrays sit in RAM doing nothing until they're needed.

[...]

Issue #1439, opened 2026-06-24
A1 — Introduce `RequestSource` interface + eager adapter; migrate `blis run` and `blis replay` consumers (#1438)


namasl:
## A1 — Introduce `RequestSource` interface + eager adapter; migrate `blis run` and `blis replay` consumers

**Tracker:** #1438 · **Change A — PR 1 of 4**

This is the foundation sub-issue for Change A (lazy streaming request generation) from #1438. It is **deliberately behavior-preserving** — no memory win lands here. It exists to decouple the workload generator from the cluster's arrival-event handler so that a streaming generator can be dropped in later (A3) without touching the cluster code again.

### Why this exists as its own PR

The current call site at [`cmd/root.go#L1455`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/root.go#L1455) eagerly materializes `wl.Requests` and hands the whole slice to `cluster.NewClusterSimulator(config, requests, onRequestDone)` at [`sim/cluster/cluster.go#L152`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/cluster.go#L152). The cluster then iterates that slice in `Run()` ([`sim/cluster/cluster.go#L616-L623`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/cluster.go#L616-L623)). To swap to streaming, every consumer of `[]*sim.Request` has to go through one common abstraction first.

Doing the abstraction extraction in its own PR keeps the streaming-generator PR (A3) reviewable as a single well-scoped change to `sim/workload/generator.go`.

### Scope (what this PR does)

1. **Define a `RequestSource` interface** in `sim/cluster/` (or `sim/workload/` — implementer's call, but the cluster must depend on the interface, not the workload package). Single-method, pure-query, in the spirit of project rule R13 (behavioral contracts) and the BLIS interface-design conventions in `docs/contributing/standards/principles.md`.

   Required signature for A1:
   ```go
   // RequestSource yields requests in non-decreasing ArrivalTime order, exactly once each.
   // Returns (nil, false) when the source is exhausted.
   type RequestSource interface {
       Next() (*sim.Request, bool)
   }
   ```

[...]

PR #1448, opened 2026-06-25
refactor(cluster): per-arrival trace hook (#1440)

### LINKED ISSUE #1440: A2 — Move trace export off the materialized request list onto a per-arrival hook (#1438)
## A2 — Move trace export off the materialized request list onto a per-arrival hook

**Tracker:** #1438 · **Change A — PR 2 of 4** · **Depends on:** A1

This PR migrates the `--trace-output` export path so that records are captured **as requests arrive**, not by iterating a finished list. It is a precondition for A3 (the streaming generator): once requests are produced one at a time, there is no finished list to iterate.

Like A1, this PR is behavior-preserving. The on-disk TraceV2 format does not change. The memory win still doesn't land here — the eager generator from `main` still runs.

Trace export today is at [`cmd/root.go#L1794-L1808`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/root.go#L1794-L1808). It assembles `allRequests = append(preGeneratedRequests, followUpRequests...)`, sorts by `ArrivalTime`, and calls `workload.RequestsToTraceRecords(allRequests)`. Two structural problems:

1. The hook must work in **both eager and lazy modes**. A3 introduces lazy generation as an alpha behind `--lazy-generation` (default false); when the flag is on there is no finished list to iterate. The hook is the only emission point that works uniformly across both modes.
2. Today the trace exporter relies on the assumption that the same eager list feeds both the cluster and the trace, which is what makes INV-13 (run/replay parity) "structural by construction" today. Under lazy mode it becomes test-enforced — capturing trace records at the arrival hook (the same boundary where the cluster sees the request) is what preserves the invariant.

### Scope (what this PR does)

1. **Add an arrival-time trace hook.** When `--trace-output` is set, install a callback that runs every time the cluster receives a request from its `RequestSource` (the boundary defined in A1 — at or near the call into `pushArrival` at [`sim/cluster/cluster.go#L604`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/cluster.go#L604)). The hook converts a `*sim.Request` to a `workload.TraceRecord` and appends it to a per-run trace buffer, or streams it directly to the writer if the writer supports incremental append (implementer's choice — buffer is fine for this PR).

2. **Wire the hook in `blis run`.** Replace the eager `allRequests` assembly + `RequestsToTraceRecords(allRequests)` call at [`cmd/root.go#L1794-L1808`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/root.go#L1794-L1808) with the hook-driven buffer. The buffer is what `workload.ExportTraceV2` consumes at the end of `Run()`.

3. **Keep `followUpRequests` accumulation correct.** The hook must fire for follow-ups too, because closed-loop sessions inject follow-ups via `onRequestDone` at [`cmd/root.go#L1765-L1772`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/root.go#L1765-L1772). Cleanest path: route follow-ups through the same `RequestSource` (or sibling source) and let the hook fire uniformly. A second-best alternative is to fire the hook from the cluster's arrival path regardless of source.

4. **Saturation analysis path stays intact.** The same comment in `cmd/root.go` (lines 1760-1763) notes that follow-up collection is unconditional for saturation correctness. Don't break that — saturation may keep its own buffer if disentangling is cleaner.

### Out of scope

- **Streaming generator.** A3.
- **Removing the `preGeneratedRequests` field on the cluster.** Stays in until A3, since the eager path still produces it.

### Acceptance criteria

- `go test ./...` passes, including the existing trace round-trip tests (e.g., `cmd/observe_test.go#L1688-L1799`).
- Determinism: TraceV2 output (header YAML + data CSV) produced by `blis run --seed 1234 --trace-output X` is byte-identical to `main` at [d059008](https://github.com/inference-sim/inference-sim/commit/d059008f7abb008d69b2f01ff08b31e4372230ea) for at least one chatbot/code-gen workload (INV-6 + INV-13).
- A focused unit test asserting that the per-arrival hook fires exactly once per request (no duplicates from follow-ups, no drops).

### Files expected to change

- `cmd/root.go` (around L1794-L1808 — replace eager assembly with hook installation)
- `sim/cluster/cluster.go` (add the hook-call point in the arrival path; one or two lines plus a setter)
- Possibly `cmd/observe.go` if the recorder shares code with the new hook (probably not in this PR — observe migration is separate).

That's 2-3 files modified.

### Invariants at risk

- **INV-13 (Run/replay parity):** the new hook becomes the sole emission point for trace records on the run side. Replay still loads from disk. Add a parity test (one workload, run with `--trace-output`, replay the produced trace, compare per-request metrics).
- **INV-6 (Determinism):** preserved by writing records in arrival order (which is the order the cluster receives them — already monotonic per INV-3). If the hook ever sees out-of-order arrivals, fail loudly rather than silently sort.

### Notes for the implementer

- Don't introduce a global trace buffer. Pass the hook (or a recorder object) explicitly into the cluster, mirroring how `onRequestDone` is passed today.
- The hook should NOT call expensive logic. Keep it to "convert to TraceRecord, append" — heavy work is on Export at end of run.

### Labels

`hardening`, `refactor`

namasl:
## Summary

- Adds `ClusterSimulator.SetArrivalHook(func(*sim.Request))`, fired exactly once per fresh arrival from `ClusterArrivalEvent.Execute` so trace records emerge in clock-monotonic order (INV-3).
- Replaces the eager `preGeneratedRequests + followUpRequests` append+sort+`RequestsToTraceRecords` block in `cmd/root.go` with a hook-driven buffer that captures `*sim.Request` pointers at the cluster's single arrival boundary.
- `REDIRECT` re-injections are filtered inside the hook so a request never emits two trace records; a non-monotonic timestamp regression panics rather than silently sorting (INV-6 guard).

## Issue

Closes #1440 (Tracker: #1438 · Change A — PR 2 of 4 · depends on A1).

## Why this PR

The hook must work in **both eager and lazy generation modes**. A3 (the next PR in this series) introduces lazy generation as an alpha behind `--lazy-generation` (default false); when the flag is on there is no finished list to iterate. The arrival hook lives at the cluster's single `ClusterArrivalEvent.Execute` boundary, which is independent of how requests are sourced (eager `SliceRequestSource` today, a streaming generator tomorrow) — so it is the only emission point that works uniformly across both modes.

Today the trace exporter also relies on the same eager list feeding both the cluster and the trace, which is what makes INV-13 (run/replay parity) "structural by construction." Under lazy mode that goes away; capturing trace records at the arrival hook (the same boundary the cluster sees) keeps the invariant test-enforced rather than implicit.

This PR is behavior-preserving — the on-disk TraceV2 format does not change, the eager generator from `main` still runs, and the memory win still doesn't land here. The memory win lands in A3.

## Cross-Path Parity

| `blis run` | yes | yes | Hook installed in `cmd/root.go` when `--trace-output` is set; eager list dropped from the trace path. Mode-agnostic: ready for A3's `--lazy-generation`. |
| `blis replay` | no | n/a | Reads requests from disk; no generator-side trace assembly to migrate. |
| `blis observe` | no | n/a | Already streaming-shaped; separate migration per issue notes. |

## Invariants Preserved

- **INV-3 (Clock monotonicity)**: Hook fires from `ClusterArrivalEvent.Execute`, which runs in cluster-clock order. A regression panics in `fireArrivalHook` (loud failure per issue guidance).
- **INV-6 (Determinism)**: Trace YAML+CSV are byte-identical to `main` at d059008 for chatbot and code-gen workloads under `--seed 1234` (verified locally; see Test plan).
- **INV-13 (Run/replay parity)**: A trace produced by the hook replays cleanly through `blis replay` — verified locally on the chatbot trace.

## Behavioral change documented

Beyond-horizon closed-loop follow-ups (scheduled at `clock + think_time` past `config.Horizon`) are now **excluded** from the exported trace. The old eager path emitted records for these requests even though they never executed; the hook fires only when `ClusterArrivalEvent.Execute` runs, which the event loop short-circuits past Horizon. This strengthens INV-13 (replay sees the same arrivals the run executed) and is covered by `TestArrivalHook_BeyondHorizonFollowUpsExcluded`.

## Out of scope

- Streaming generator + `--lazy-generation` flag (A3).
- Removing `preGeneratedRequests` from the cluster (stays until A3).
- `cmd/observe.go` migration (separate PR).

## Test plan

- [x] `go build ./...`
- [x] `go test ./... -count=1`
- [x] Per-arrival hook unit tests (`sim/cluster/arrival_hook_test.go`): fires exactly once per initial request; preserves pointer identity; non-decreasing arrival order; **no duplicate fires for REDIRECT re-injections**; disabled when not set (zero overhead); panics on non-monotonic arrival; panics if `SetArrivalHook` called after `Run`; fires for session follow-ups (exact count); beyond-horizon follow-ups excluded; `SetArrivalHook(nil)` resets the monotonicity floor.
- [x] **End-to-end** `cmd/root_test.go` test (`TestRunCmd_TraceOutput_RecordCountMatchesRequests`) runs `runCmd.Run` with `--trace-output` and asserts trace record count + non-decreasing `ArrivalTimeUs`.
- [x] **REDIRECT integration** (`TestInstanceLifecycle_RedirectDrainPreservesConservation`) installs the hook and asserts `hookCalls == 0` through the live drain path.
- [x] **Byte-identical determinism** (`diff` exit 0 on YAML + CSV) for `blis run --seed 1234 --trace-output … --num-requests 50` vs `main` at d059008:
  - default workload
  - `--workload chatbot` (session follow-ups exercised)
- [x] Replay round-trip: `blis replay --trace-header … --trace-data …` succeeds against the hook-produced chatbot trace.

## Files changed

- `sim/cluster/cluster.go` — `arrivalHook`+`lastArrivalHookTime` fields, `SetArrivalHook`, `fireArrivalHook`.
- `sim/cluster/cluster_event.go` — `ClusterArrivalEvent.Execute` fires the hook before scheduling admission.
- `cmd/root.go` — install hook when `--trace-output` is set; trace-export path consumes `traceArrivals` directly; saturation path keeps its own `allRequests` assembly.
- `sim/cluster/arrival_hook_test.go` — new unit tests.
- `sim/cluster/instance_lifecycle_test.go` — REDIRECT-drain test now installs the hook and asserts no duplicate fires.
- `cmd/root_test.go` — new end-to-end record-count test.

---

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1812,7 +1812,10 @@ var runCmd = &cobra.Command{
 
 		// Assemble allRequests for saturation analysis (BC-12, issue #1298).
 		// Trace export is now driven by the arrival hook above and no longer
-		// shares this slice (issue #1440).
+		// shares this slice (issue #1440). NOTE: allRequests is nil when
+		// --saturation-report is not set — any new downstream consumer must
+		// either trigger this assembly path or source its data elsewhere
+		// (the hook buffer for trace, or aggregateMetrics for per-instance).
 		var allRequests []*sim.Request
 		if saturationReport != "" {
 			allRequests = make([]*sim.Request, 0, len(preGeneratedRequests)+len(followUpRequests))
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1812,10 +1812,8 @@ var runCmd = &cobra.Command{
 
 		// Assemble allRequests for saturation analysis (BC-12, issue #1298).
 		// Trace export is now driven by the arrival hook above and no longer
-		// shares this slice (issue #1440). NOTE: allRequests is nil when
-		// --saturation-report is not set — any new downstream consumer must
-		// either trigger this assembly path or source its data elsewhere
-		// (the hook buffer for trace, or aggregateMetrics for per-instance).
+		// shares this slice (issue #1440). allRequests is nil when
+		// --saturation-report is not set.
 		var allRequests []*sim.Request
 		if saturationReport != "" {
 			allRequests = make([]*sim.Request, 0, len(preGeneratedRequests)+len(followUpRequests))
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1784,8 +1784,15 @@ var runCmd = &cobra.Command{
 		// trace is disabled). Saturation analysis continues to use the
 		// preGeneratedRequests + followUpRequests path below — those slices
 		// remain populated for that purpose only.
+		//
+		// Install/export coupling: traceArrivals is declared nil here and
+		// assigned a non-nil empty slice ONLY inside the install branch.
+		// A nil traceArrivals at the export site below with traceOutput
+		// non-empty means the install branch was dropped — we fail loudly
+		// rather than write a silent empty trace (R1).
 		var traceArrivals []*sim.Request
 		if traceOutput != "" {
+			traceArrivals = make([]*sim.Request, 0)
 			cs.SetArrivalHook(func(req *sim.Request) {
 				traceArrivals = append(traceArrivals, req)
 			})
@@ -1829,6 +1836,13 @@ var runCmd = &cobra.Command{
 		// the arrival hook (issue #1440) — already in clock-monotonic order
 		// per INV-3, so no sort is required.
 		if traceOutput != "" {
+			// Install/export coupling guard (R1): traceArrivals is a non-nil
+			// empty slice when SetArrivalHook ran above. A nil here means
+			// the install branch was dropped or moved without updating this
+			// site — refuse to write a silent empty trace.
+			if traceArrivals == nil {
+				logrus.Fatalf("Trace export: arrival hook was not installed but --trace-output=%q is set — install/export branches diverged (issue #1440)", traceOutput)
+			}
 			records := workload.RequestsToTraceRecords(traceArrivals)
 			header := &workload.TraceHeader{
 				Version:           3,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1771,6 +1771,25 @@ var runCmd = &cobra.Command{
 			}
 		}
 		cs := cluster.NewClusterSimulator(config, cluster.NewSliceRequestSource(preGeneratedRequests), onRequestDone)
+
+		// Arrival hook: capture trace-emission references at the cluster's
+		// single arrival boundary so the trace exporter no longer relies on
+		// the eager preGeneratedRequests + followUpRequests list assembly
+		// (issue #1440). The hook fires once per fresh arrival in
+		// clock-monotonic order — see ClusterArrivalEvent.Execute. We hold
+		// pointers (not copies) so the request's final state (set by the
+		// event loop) is visible at export time.
+		//
+		// Only install when --trace-output is set (BC-1: zero overhead when
+		// trace is disabled). Saturation analysis continues to use the
+		// preGeneratedRequests + followUpRequests path below — those slices
+		// remain populated for that purpose only.
+		var traceArrivals []*sim.Request
+		if traceOutput != "" {
+			cs.SetArrivalHook(func(req *sim.Request) {
+				traceArrivals = append(traceArrivals, req)
+			})
+		}
 		if err := cs.Run(); err != nil {
 			logrus.Fatalf("Simulation failed: %v", err)
 		}
@@ -1791,9 +1810,11 @@ var runCmd = &cobra.Command{
 		}
 		goodputTargets := mergeGoodputTargets(cliTTFT, cliITL, cliE2E, nil, specTargets)
 
-		// Assemble allRequests if trace export or saturation analysis requested (BC-12, issue #1298)
+		// Assemble allRequests for saturation analysis (BC-12, issue #1298).
+		// Trace export is now driven by the arrival hook above and no longer
+		// shares this slice (issue #1440).
 		var allRequests []*sim.Request
-		if traceOutput != "" || saturationReport != "" {
+		if saturationReport != "" {
 			allRequests = make([]*sim.Request, 0, len(preGeneratedRequests)+len(followUpRequests))
 			allRequests = append(allRequests, preGeneratedRequests...)
 			allRequests = append(allRequests, followUpRequests...)
@@ -1803,9 +1824,11 @@ var runCmd = &cobra.Command{
 			})
 		}
 
-		// Export trace if requested (BC-1, BC-7)
+		// Export trace if requested (BC-1, BC-7). Records are sourced from
+		// the arrival hook (issue #1440) — already in clock-monotonic order
+		// per INV-3, so no sort is required.
 		if traceOutput != "" {
-			records := workload.RequestsToTraceRecords(allRequests)
+			records := workload.RequestsToTraceRecords(traceArrivals)
 			header := &workload.TraceHeader{
 				Version:           3,
 				TimeUnit:          "microseconds",
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1144,6 +1144,188 @@ func TestRunCmd_MetricsPath_WritesMetricsOutput(t *testing.T) {
 	}
 }
 
+// TestRunCmd_TraceOutput_RecordCountMatchesRequests verifies the issue
+// #1440 end-to-end contract: when `blis run --trace-output X` is used,
+// the exported `X.csv` contains exactly one TraceV2 record per request
+// produced by the workload generator. The hook installed in cmd/root.go
+// replaced the eager `preGeneratedRequests + followUpRequests + sort +
+// RequestsToTraceRecords` path; this test guards the substitution against
+// silent record-count drift (would corrupt INV-13 replay parity).
+// NOTE: Do NOT use t.Parallel() — mutates package-level vars.
+func TestRunCmd_TraceOutput_RecordCountMatchesRequests(t *testing.T) {
+	tmpDir := t.TempDir()
+	tracePrefix := filepath.Join(tmpDir, "trace")
+
+	mcFolder, hwPath, defaultsPath := setupTrainedPhysicsTestFixturesWithDefaults(t)
+
+	// Save and restore the package-level flag vars runCmd.Run mutates.
+	// Same list as TestRunCmd_MetricsPath_WritesMetricsOutput, with
+	// traceOutput added.
+	origMetrics := metricsPath
+	origModel := model
+	origBackend := latencyModelBackend
+	origBeta := betaCoeffs
+	origAlpha := alphaCoeffs
+	origTotalKV := totalKVBlocks
+	origBlockSize := blockSizeTokens
+	origMaxRunning := maxRunningReqs
+	origMaxSched := maxScheduledTokens
+	origInstances := numInstances
+	origSeed := seed
+	origResults := resultsPath
+	origThreshold := longPrefillTokenThreshold
+	origKVCPU := kvCPUBlocks
+	origOffload := kvOffloadThreshold
+	origBandwidth := kvTransferBandwidth
+	origBaseLatency := kvTransferBaseLatency
+	origSnapRefresh := snapshotRefreshInterval
+	origAdmission := admissionPolicy
+	origRouting := routingPolicy
+	origScheduler := scheduler
+	origPolicyConfig := policyConfigPath
+	origMaxModelLen := maxModelLen
+	origTraceLevel := traceLevel
+	origCounterfactualK := counterfactualK
+	origSimHorizon := simulationHorizon
+	origWorkloadType := workloadType
+	origRate := rate
+	origNumReqs := numRequests
+	origConcurrency := concurrency
+	origThinkTime := thinkTimeMs
+	origPrefix := prefixTokens
+	origPromptMean := promptTokensMean
+	origPromptStdev := promptTokensStdev
+	origPromptMin := promptTokensMin
+	origPromptMax := promptTokensMax
+	origOutputMean := outputTokensMean
+	origOutputStdev := outputTokensStdev
+	origOutputMin := outputTokensMin
+	origOutputMax := outputTokensMax
+	origWorkloadSpec := workloadSpecPath
+	origRequestTimeout := requestTimeoutSecs
+	origTraceOut := traceOutput
+	origLogLevel := logLevel
+	origModelConfigFolder := modelConfigFolder
+	origHwConfigPath := hwConfigPath
+	origGPU := gpu
+	origTP := tensorParallelism
+	defer func() {
+		metricsPath = origMetrics
+		model = origModel
+		latencyModelBackend = origBackend
+		betaCoeffs = origBeta
+		alphaCoeffs = origAlpha
+		totalKVBlocks = origTotalKV
+		blockSizeTokens = origBlockSize
+		maxRunningReqs = origMaxRunning
+		maxScheduledTokens = origMaxSched
+		numInstances = origInstances
+		seed = origSeed
+		resultsPath = origResults
+		longPrefillTokenThreshold = origThreshold
+		kvCPUBlocks = origKVCPU
+		kvOffloadThreshold = origOffload
+		kvTransferBandwidth = origBandwidth
+		kvTransferBaseLatency = origBaseLatency
+		snapshotRefreshInterval = origSnapRefresh
+		admissionPolicy = origAdmission
+		routingPolicy = origRouting
+		scheduler = origScheduler
+		policyConfigPath = origPolicyConfig
+		maxModelLen = origMaxModelLen
+		traceLevel = origTraceLevel
+		counterfactualK = origCounterfactualK
+		simulationHorizon = origSimHorizon
+		workloadType = origWorkloadType
+		rate = origRate
+		numRequests = origNumReqs
+		concurrency = origConcurrency
+		thinkTimeMs = origThinkTime
+		prefixTokens = origPrefix
+		promptTokensMean = origPromptMean
+		promptTokensStdev = origPromptStdev
+		promptTokensMin = origPromptMin
+		promptTokensMax = origPromptMax
+		outputTokensMean = origOutputMean
+		outputTokensStdev = origOutputStdev
+		outputTokensMin = origOutputMin
+		outputTokensMax = origOutputMax
+		workloadSpecPath = origWorkloadSpec
+		requestTimeoutSecs = origRequestTimeout
+		traceOutput = origTraceOut
+		logLevel = origLogLevel
+		modelConfigFolder = origModelConfigFolder
+		hwConfigPath = origHwConfigPath
+		gpu = origGPU
+		tensorParallelism = origTP
+	}()
+
+	// 5 distribution-mode requests — non-session workload, so each
+	// initial request emits exactly one trace record.
+	const wantRecords = 5
+	traceOutput = tracePrefix
+	workloadType = "distribution"
+	rate = 1.0
+	numRequests = wantRecords
+	promptTokensMean = 512
+	promptTokensStdev = 256
+	promptTokensMin = 2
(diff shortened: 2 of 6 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
**BLIS PR Self-Review (Round 5): `refactor(cluster): per-arrival trace hook (#1440)`**

This is a fresh review against `issue-1440-arrival-hook` at `16837825`, looking at the diff vs `main`. CI is green (build/lint/test all SUCCESS); `mergeable` is true.

---

### Addressed since last review

All findings from Rounds 1–4 are resolved:

- Round 1 IMPORTANT (`cluster.go:132` stale `pushArrival` reference) → fixed in `caf1e0a1`.
- Round 2 IMPORTANT #1 (beyond-horizon follow-ups) → documented + `TestArrivalHook_BeyondHorizonFollowUpsExcluded` added in `aab727ba`.
- Round 2 IMPORTANT #2 (no `cmd/root.go` integration test) → `TestRunCmd_TraceOutput_RecordCountMatchesRequests` added.
- Round 2 IMPORTANT #3 (REDIRECT integration gap) → hook + `hookCalls == 0` check added to `TestInstanceLifecycle_RedirectDrainPreservesConservation`.
- Round 2 IMPORTANT #4 (`SetArrivalHook(nil)` watermark) → `lastArrivalHookTime` now resets; `TestSetArrivalHook_NilClearResetsMonotonicityFloor` guards it.
- Round 3 IMPORTANT (`Run() panics` → `fireArrivalHook() panics`) → fixed in `16837825`.

---

### CRITICAL

None.

### IMPORTANT

None.

### MINOR

1. **`cmd/root.go:1783` — `BC-1` tag overloaded twice in the same function.** Used at line 1341 for `--concurrency`/`--rate` mutual exclusion (unrelated) and reused at line 1783 for "zero overhead when trace is disabled." Two unrelated BC labels with the same string in one file is mildly confusing. Non-blocking; either dropping the label or re-numbering inline would be cleaner.

2. **`TestArrivalHook_NoDuplicateFiresForRedirectedRequests` is a white-box unit test.** It calls the unexported `cs.fireArrivalHook(...)` directly to simulate the REDIRECT injection rather than driving it through the real drain path. This was flagged as MINOR in Round 2/3. The complementary integration test in `TestInstanceLifecycle_RedirectDrainPreservesConservation` (now installs a hook and asserts `hookCalls == 0`) provides the resilient behavioral guard, so the white-box unit test is acceptable as a focused unit on the filter predicate. Worth noting that if `fireArrivalHook` is ever renamed or its parameters change, this test will need maintenance.

3. **Tie-break ordering for equal-timestamp arrivals changed (cosmetic).** The old path used `sort.SliceStable(allRequests, ...)` by `ArrivalTime`, preserving source-slice order. The hook stream orders ties by event `seqID` (push order). For initial-only workloads these are identical; for an initial + follow-up at the exact same microsecond, `RequestID` assignment (a sequential index in `RequestsToTraceRecords`) could differ. Since `RequestID` is an opaque trace index with no semantic replay dependence, this is harmless — flagged only for awareness in case any future consumer assumes the previous stable ordering.

4. **Race coverage.** `TestArrivalHook_NoDuplicateFiresForRedirectedRequests` and `TestSetArrivalHook_NilClearResetsMonotonicityFloor` call `fireArrivalHook` directly without going through `Run()`, so the "must call SetArrivalHook before Run" contract is enforced sequentially by `TestSetArrivalHook_AfterRun_Panics`. CI doesn't appear to run `go test -race` on this package. Not a blocker, but the hook is conceptually thread-affine to the cluster goroutine and a `-race` run would help catch a future regression where someone tries to install or fire it concurrently.

---

### Invariant Analysis

| Invariant | Status | Notes |
|-----------|--------|-------|
| INV-1 (Request conservation) | ✅ Preserved | Hook is read-only; REDIRECT filter (`req.Redirected` check at `cluster.go` ~656) prevents double-emission. Confirmed `infra_lifecycle_event.go:283` marks `req.Redirected = true` before re-`pushArrival`. |
| INV-3 (Clock monotonicity) | ✅ Preserved | Hook fires at execute time (`ClusterArrivalEvent.Execute`), which is the cluster-clock-ordered path. Regression panics inside `fireArrivalHook`. |
| INV-6 (Determinism) | ✅ Preserved | No RNG draws, no branching that introduces order ambiguity. Hook append is deterministic. PR description records byte-identical YAML+CSV vs `main` at `d059008` under `--seed 1234`. |
| INV-8 (Work-conserving) | ✅ Unaffected | Hook fires before `AdmissionDecisionEvent` push — event queue shape unchanged. |
| INV-13 (Run/Replay parity) | ✅ Preserved | Trace format unchanged. Hook-driven assembly is now end-to-end tested by `TestRunCmd_TraceOutput_RecordCountMatchesRequests`. Beyond-horizon exclusion strengthens parity (replay reads the same trace the run produced). |

---

### Cross-Path Parity

| Path | Applicable | Status |
|------|-----------|--------|
| `blis run` | Yes | ✅ Hook installed when `--trace-output` is set; eager `allRequests` path removed from trace export. |
| `blis replay` | No | ✅ N/A — `cmd/replay.go:550` reads `requests` directly from the loaded trace and calls `RequestsToTraceRecords(requests)`; no generator-side assembly to migrate. |
| `blis observe` | Deferred | ✅ Already streaming-shaped (`Recorder` accumulates `TraceRecord`s per HTTP completion). The PR explicitly defers migration per the issue notes. |

---

### Preemption Safety / Timeout Consistency

Not applicable. The hook adds a read-only callback before `AdmissionDecisionEvent` and touches no per-request metrics, aggregate counters, timeout/deadline semantics, or horizon enforcement.

---

### API Design

`SetArrivalHook(func(*sim.Request))` is a clean single-method API mirroring `SetProgressHook`. The must-call-before-Run panic guard is correct and tested. The "Pass nil to clear; resets `lastArrivalHookTime`" affordance is documented and tested. R6 (sim/ failure paths use panic, not `logrus.Fatalf`) is satisfied. R4 (single canonical construction site) is unaffected — no new struct fields require updating literals elsewhere.

---

### Test Coverage

12 behavioral tests cover the contract:

- Initial requests: once-per, pointer identity, INV-3 ordering.
- REDIRECT: no duplicate fires (white-box + integration via `instance_lifecycle_test.go`).
- Zero-overhead path: hook nil → no panic, no record.
- Panic on non-monotonic arrival.
- Panic on install-after-Run.
- Session follow-ups: exact-count assertion (not range).
- Beyond-horizon: explicitly excluded.
- Nil-clear: `lastArrivalHookTime` reset.
- cmd-layer: `TestRunCmd_TraceOutput_RecordCountMatchesRequests` (record count + `ArrivalTimeUs` non-decreasing).

Tests are behavioral and refactor-resilient with exact assertions.

---

### Verdict

**READY TO MERGE.**

The PR is correct, complete, and well-tested. All invariants are preserved with positive evidence (panic guards + byte-identity check). Cross-path parity rationale is explicit. The four MINOR items above are style/coverage polish, not correctness issues — none block merge.

namasl:
Issue #1440 was updated to clarify the rationale: A3 introduces lazy generation as an alpha behind `--lazy-generation` (default false), and the hook must work in **both eager and lazy modes** — not "after A3 lands, there is no list."

The implementation here is already mode-agnostic: the hook fires from `ClusterArrivalEvent.Execute`, which is independent of how requests are sourced (today's eager `SliceRequestSource`, tomorrow's streaming generator). The PR description has been updated to reflect the corrected framing. No code change is required — the acceptance criteria are unchanged and continue to hold.

susiejojo:
## Self-review notes — PR #1448 `refactor(cluster): per-arrival trace hook (#1440)`

Thanks for this — it's a clean, well-documented change and the test coverage is genuinely thorough. Sharing some notes from a multi-agent self-review (code / tests / error-handling / comments) below. **Overall this looks ready to merge**; everything in the "suggestions" section is non-blocking and meant to be helpful rather than required.

Reviewed at `16837825` against base `b26f027c` (4 commits, 6 files, +540/−4). `go build ./...` is clean and the new arrival-hook tests pass locally.

---

### Scope clarification (two questions that came up)

**Which command paths does this touch?**

| Path | Affected? | Detail |
|------|-----------|--------|
| `blis run` | **Yes (direct)** | `--trace-output` is now sourced from the arrival hook (`traceArrivals`) rather than the eager `preGeneratedRequests + followUpRequests + sort.SliceStable` assembly. `--saturation-report` still uses the eager slice (unchanged). No change to stdout/metrics. |
| `blis replay` | **No (not migrated)** | `cmd/replay.go:550` still calls `RequestsToTraceRecords(requests)` on the eager slice. The run-side trace-content change *indirectly* affects what a later replay consumes (fewer phantom records), which underpins the INV-13 argument. |
| `blis observe` | **No** | Doesn't construct a `ClusterSimulator`; has its own real-HTTP trace-emission path. |
| `blis calibrate` | **No** | Pure post-processing (trace vs sim results). Untouched. |

**Does this overwrite eager generation, or add lazy generation as an alternate?**

Neither, if I'm reading it right — and that seems intentional. Workload is still eagerly generated into `preGeneratedRequests` and wrapped in `SliceRequestSource` (from #1439). This PR changes only trace *export*, decoupling it from the eager slice via a per-arrival callback. Since `traceArrivals` still accumulates a full `[]*sim.Request`, there's no memory win yet — this reads as pure plumbing (sub-change A2 of #1438) that relocates the trace-emission seam so the future streaming generator (A3) can slot in without touching export. Flagging only so the "lazy generation" framing in #1438 isn't mis-read as landing here.

---

### Behavioral change worth calling out (intentional + well-documented)

Unlike #1439 (a byte-identical refactor), this one does change trace *contents*: firing at `ClusterArrivalEvent.Execute` is clock-monotonic, and the event loop short-circuits past `Horizon` (`cluster.go:760-762`, breaking *before* `Execute`). So beyond-horizon follow-ups — previously emitted as incomplete, zero-timing records — are now excluded, as are REDIRECT re-injections (`req.Redirected`, set at `infra_lifecycle_event.go:283`). Both exclusions look correct and are covered by tests. The godoc explains this nicely.

---

### Two things that might be worth a look before/just-after merge

**1. The INV-13 "strengthens parity" claim isn't end-to-end tested yet.**
The central justification — "replay reads the same trace the run produced" — is architecturally convincing, but I couldn't find a test that actually exports via `run --trace-output`, replays it, and compares per-request TTFT/E2E. `TestRunCmd_TraceOutput_RecordCountMatchesRequests` verifies record *count* + `ArrivalTimeUs` monotonicity, which is great, but stops short of the full loop. A run→replay→compare test would turn the strongest claim into the best-guarded one. Could also be a fast-follow if you'd prefer to keep this PR tight.

**2. Small nil-guard opportunity around `traceArrivals`.**
Hook install (`if traceOutput != ""` → `SetArrivalHook`) and export (`if traceOutput != ""` → `RequestsToTraceRecords(traceArrivals)`) are coupled by a duplicated condition today. If a future refactor ever moves the install, export would run with `traceArrivals == nil` and silently write an empty trace. A `logrus.Fatalf` (or assertion) when `traceOutput != "" && traceArrivals == nil` would keep this aligned with R1 (never fail silently). Very minor, just defensive.

---

### Smaller suggestions (all optional)

- **Filtering is currently silent.** Beyond-horizon + REDIRECT drops produce no diagnostic, so a user who generates 50 requests but sees 37 trace records has nothing in the output to explain the delta. A `logrus.Debugf` in the filter paths and/or a "captured N of M generated" line would help discoverability.
- **Empty-trace visibility.** Consider elevating the 0-records case from `Infof` to `Warnf` so it's visible at `--log-level warn`.
- **A few test additions** that would round things out: (a) field-by-field content equivalence between the hook path and the old eager path (substitution regression guard); (b) an *initial* (non-follow-up) request with `ArrivalTime > Horizon` — only follow-ups are covered today; (c) equal-`ArrivalTime` tie-order determinism (INV-6).
- **Test coupling, very minor.** Three tests call the unexported `cs.fireArrivalHook(...)` directly. `PanicsOnNonMonotonicArrival` and `NilClearResetsMonotonicityFloor` are reasonable (hard to trigger through the min-heap), but `NoDuplicateFiresForRedirectedRequests` largely overlaps the stronger integration test in `instance_lifecycle_test.go` — might be consolidatable.
- **One godoc nuance.** `fireArrivalHook` explains the REDIRECT skip as "the same logical request was observed on its original arrival" — but in the seeded-then-redirected case (the one `instance_lifecycle_test.go` exercises) the hook never saw an original arrival. The broader rationale — "a redirect is a duplicate injection of the same logical request and must not double-emit" — covers both paths.
- **Wording polish.** "clock-monotonic" → "non-decreasing" (the guard uses strict `<`, so equal timestamps are allowed); `TestArrivalHook_DisabledWhenNotSet`'s "verifies zero overhead" is really "verifies no-op safety"; and the INV-13 "strengthens" comment could carry a small "(not end-to-end tested here)" note.
- **API consistency (follow-up).** `SetArrivalHook` adds a nice `hasRun` guard that the older `SetProgressHook` lacks — might be worth aligning `SetProgressHook` separately.
- **Parity note.** Keeping replay's re-export on the eager path is defensible (it re-emits a fixed input trace with simulated timing); a one-line note in the PR body or a follow-up issue would make the asymmetry explicit per the cross-path-parity bar.

---

### Things that are nicely done
- Replacing the silent `sort.SliceStable` with a monotonicity **panic** surfaces the root cause instead of masking ordering bugs (INV-3/INV-6) — good call.
- `SetArrivalHook` panicking if called after `Run()` neatly prevents the "wired up too late, zero records" failure mode.
- The hook is purely observational, and INV-1 conservation is tracked independently via `injectedByClass` (incremented *before* the hook fires), so trace exclusions can't corrupt conservation accounting.
- Zero-overhead-when-disabled path and pointer-identity (not copy) semantics are both correct and documented.
- The `instance_lifecycle_test.go` integration check (REDIRECT → 0 hook fires) drives through the real drain lifecycle — the most robust test in the set.

Thanks again — really solid work. None of the above blocks merge from my side.

*Automated multi-agent self-review; no code was modified.*

namasl:
Thanks for the thorough notes! Investigated both pre-merge items; both warranted fixes. Pushed as `df2c01d`.

**1. INV-13 run→replay parity end-to-end test** — implemented. Added `TestINV13_ArrivalHookRunReplayParity` in `sim/cluster/arrival_hook_test.go`: it runs a workload, captures records via the arrival hook, exports, reloads, replays through a fresh `ClusterSimulator` with the same config, and asserts per-request TTFT and E2E are identical. This is complementary to the existing `TestINV13_RunReplayParity_PD`, which still drives the eager path via `RequestsToTraceRecords(requests)` directly — so we now have parity coverage for both emission paths.

**2. Install/export coupling guard for `traceArrivals`** — implemented. `traceArrivals` is now initialised to a non-nil empty slice inside the `if traceOutput != ""` install branch, and the export site rejects a nil `traceArrivals` with `--trace-output` set via `logrus.Fatalf` (R1: never fail silently). A future refactor that moves the install away from the export's condition will now fail loudly instead of writing an empty trace.

Verification:
- `go test ./... -count=1` — pass
- Byte-identical chatbot trace (`--seed 1234`) vs `main` at d059008 still holds.

The other (smaller) suggestions in your notes are noted as follow-ups — happy to tackle them in a separate PR if you'd like, but per your "non-blocking" framing I'm leaving them out of this one to keep the diff tight.

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo (approved):
lgtm

PR #1452, opened 2026-06-26
hardening(workload): delta-encoded multi-turn rounds (O(R) per session) (#1445)

### LINKED ISSUE #1445: C — Delta-encoded multi-turn rounds (O(R) per session, not O(R²)) (#1438)
## C — Delta-encoded multi-turn rounds (O(R) per session, not O(R²))

**Tracker:** #1438 · **Change C (sole PR)**

In multi-turn workloads with `context_growth: accumulate`, each round currently stores a fully concatenated input — round k's `InputTokens` length is proportional to k, so a session with R rounds uses O(R²) memory. This PR switches to delta encoding: each round stores only its **new** tokens, and the prior context is reconstructed on demand (or, more often, doesn't need to be — the KV-cache prefix hasher operates block-by-block, and round N's first M blocks hash identically to round N-1's first M blocks, so the cache hits before reconstruction would happen).

This change is independent of A and B. It can land in any order.

### Two accumulate paths must be addressed

The same antipattern exists in two places. Both must be covered (or one explicitly deferred to a follow-up before the issue is VALID — see "Out of scope" below):

1. **Open-loop generation** — [`sim/workload/reasoning.go#L53-L57`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/reasoning.go#L53-L57). Eager construction at workload-generation time. Source of the parent issue's measured spike.
2. **Closed-loop replay sessions** — [`sim/workload/session.go#L191-L205`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/session.go#L191-L205). Round N's input is built at runtime as `append(append([]int{}, sess.contextTokens...), newInputTokens...)` — same `O(R²)` bytes per session, just amortized over the run instead of paid up front. This path is hit by `blis run --session-mode closed-loop` and `blis replay --session-mode closed-loop`.

A delta representation that fixes only path (1) leaves runtime closed-loop sessions still quadratic, and INV-13 (run/replay parity) cannot be satisfied without addressing path (2) as well.

#### 1. Pick a single chain representation

Choose one and commit (do not leave to "implementer's choice" — the choice has cascading consequences for trace export and PD sub-request copies):

- **Recommended: shared-tail chain.** A small struct, e.g. `type tokenChain struct { prev *tokenChain; delta []int }` (or a per-session `*[]int` buffer that rounds index into). Round N stores only its delta plus a back-reference. Walking the chain to materialize is O(total input length), traversed only when a consumer cannot work block-by-block.
- **Alternative: previousRound pointer.** Each `*Request` stores `PrevRound *Request` and concatenates on demand. Simpler, but every read of the full sequence walks the chain — and round N → round 0 with R=50 means walking through 50 requests' deltas. Only acceptable if no hot path needs the full sequence (verify under `pprof`).

The contract is the same in either case: a `req.FullInputTokens() []int` accessor returns the same sequence the eager `append` produces today. State the chosen representation in the PR description.

#### 2. Add accessors

On `*sim.Request`:

- `req.FullInputTokens() []int` — materializes the full input sequence (only called by trace export and the rare consumer that needs the flat slice).
- `req.InputLen() int64` — returns the total length without materialization. Replaces `util.Len64(req.InputTokens)` and `len(req.InputTokens)` reads at every call site that needs only length.
- `req.InputTokenSlice(start, end int64) []int` — returns the slice spanning the given absolute index range. Replaces direct `req.InputTokens[start:end]` reads (notably `sim/kv/cache.go:190`).

#### 3. Switch storage in `sim/workload/reasoning.go`

Replace the eager prepend at L53-L57:

```go
if mt.ContextGrowth == "accumulate" && round > 0 {
    inputTokens = append(append([]int{}, contextPrefix...), newInputTokens...)
}
```

With per-round delta storage plus a back-reference (per the chosen representation). The accessor preserves observable behavior.

#### 4. Switch storage in `sim/workload/session.go`

Replace the runtime accumulate at L191-L205:

```go
sess.contextTokens = append(sess.contextTokens, rawConversation[len(sess.contextTokens):]...)
// ...
inputTokens = append(append([]int{}, sess.contextTokens...), newInputTokens...)
```

With the same chain representation. `sess.contextTokens` becomes the head of the chain (or is replaced by the chain handle); each round's `Request.InputTokens` carries only the round's delta.

#### 5. Audit and migrate every direct read of `req.InputTokens`

A grep across `sim/` and `cmd/` (excluding tests) returns ~88 reads. Most are `len(req.InputTokens)` and migrate trivially to `req.InputLen()`. The non-trivial consumers are listed below; the audit must be exhaustive (per project rule R4 — every `*sim.Request{...}` literal must be reviewed when adding chain fields).

**Hot-path consumers needing the full sequence (or a slice into it):**

- [`sim/batch_formation.go#L105`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/batch_formation.go#L105) — `util.Len64(req.InputTokens) - req.ProgressIndex` for chunked-prefill scheduling. Use `req.InputLen()`.
- [`sim/batch_formation.go#L132`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/batch_formation.go#L132) — decode-phase trigger. Use `req.InputLen()`.
- [`sim/batch_formation.go#L182-L183`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/batch_formation.go#L182) — `ctx.KVCache.GetCachedBlocks(next.InputTokens)`. The hasher must accept either a chained input or a flat `[]int`.
- [`sim/kv/cache.go#L182-L190`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/kv/cache.go#L182) — `AllocateKVBlocks` slices `req.InputTokens[startIndex:endIndex]` for prefill block allocation. Replace with `req.InputTokenSlice(startIndex, endIndex)`. This is the hasher API touch point — the issue previously listed `sim/kv_store.go`; the real change is `sim/kv/cache.go` and the `KVStore` interface in `sim/kv_store.go` (`GetCachedBlocks(tokens []int) []int64`) — extend that interface to accept a token-source instead of a flat slice, or have callers materialize.

**Routing/scoring (cluster scope, called per request):**

- `sim/routing_prefix_scorer.go:36,59`, `sim/routing_precise_prefix_scorer.go:33`, `sim/routing_nohit_lru_scorer.go:48` — pass `req.InputTokens` to `ComputeBlockHashes` / `GetCachedBlocks`. Same hasher-interface touch as above.

**PD sub-request construction (cluster scope, decision point):**

- [`sim/cluster/cluster.go#L2005`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/cluster.go#L2005), [`sim/cluster/cluster_event.go#L491`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/cluster_event.go#L491), [`sim/cluster/pd_events.go#L129`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/cluster/pd_events.go#L129) — construct prefill/decode sub-requests by copying `InputTokens` directly (`InputTokens: orig.InputTokens` or `e.request.InputTokens`). The sub-request currently aliases the original slice (cheap copy of a slice header). With the chain representation, the natural choices are:
  1. **Sub-requests share the chain handle** (preferred — keeps the storage win). Requires the chain to be reference-counted or owned by a session-level handle that outlives both sub-requests.
  2. **Flatten on sub-request creation** (defeats the storage win for the duration of the sub-request, but avoids ownership complexity). Acceptable if PD sub-request lifetimes are short relative to session lifetime — measure under a representative PD workload.

  State the choice in the PR description and justify with measurement.

**Length-only reads in the simulator (use `req.InputLen()`):**

- `sim/simulator.go:420-457, 471, 486, 539, 603, 761, 781, 808, 826` — all `len(r.InputTokens)` or `util.Len64(req.InputTokens)`. Mechanical migration.
- `sim/admission.go:55`, `sim/scheduler.go:49`, `sim/metrics_utils.go:39`, `sim/cluster/instance.go:436`, `sim/cluster/parent_request.go:59`, `sim/disaggregation.go:137,148` — same.

#### 6. Trace export flattens

When writing TraceV2, each request's input token sequence is materialized once via `FullInputTokens()` and written out flat (`sim/workload/tracev2.go:209,557` — current call sites). The on-disk format does not change.

#### 7. Free the prior round's context once unreachable

With delta encoding, after round N completes, round N-1's tokens can be GC'd if no live reference remains. With the recommended shared-tail chain, the head buffer is freed when the last referencing round goes out of scope. Verify with a memory test that the working set doesn't accumulate across many sessions.

### Out of scope

- The non-accumulating multi-turn paths (`context_growth: replace` etc.) — they already store O(R) per session.
- The on-disk TraceV2 format.
- Changes B (token-ID compaction) and A (lazy generation) from #1438 — separate PRs.
- ~~Changes B (token-ID compaction) and A (lazy generation)~~ — defer if needed.

### Acceptance criteria

- `go test ./...` passes, including the multi-turn / reasoning tests.
- **INV-6 determinism (open-loop):** byte-identical stdout for a fixed-seed `blis run` on a multi-turn `accumulate` workload before and after.
- **INV-6 determinism (closed-loop):** byte-identical stdout for a fixed-seed `blis run --session-mode closed-loop` on a multi-turn `accumulate` workload before and after.
- **INV-13 run/replay parity:** trace exported by `blis run --trace-output ...` and replayed by `blis replay --trace-header ... --trace-data ...` produces identical per-request metrics. Run the parity check with both `--session-mode open-loop` and `--session-mode closed-loop`.
- **Pinned memory measurement.** PR description includes peak per-session input-token bytes (collected via `pprof` heap profile or instrumented allocation counter) before and after on:
  - `max_rounds=15, prefix_tokens=2000, num_requests=500, context_growth=accumulate, --session-mode open-loop` — target ≥ 5× drop in per-session input-token bytes (R/2 ≈ 7.5).
  - `max_rounds=50, prefix_tokens=2000, num_requests=500, context_growth=accumulate, --session-mode open-loop` — target ≥ 20× drop (R/2 = 25).
  - The same scenario with `--session-mode closed-loop` — same targets.
- **Unit tests:**
  - `req.FullInputTokens()` returns the same sequence the eager `append` produces today (for both open-loop and closed-loop construction paths).
  - `req.InputLen()` returns the same value as `len(req.FullInputTokens())` without materializing.
  - `req.InputTokenSlice(s, e)` returns `req.FullInputTokens()[s:e]`.
  - PD sub-request construction (cluster + cluster_event + pd_events) preserves the chain semantics chosen in step 1 (e.g., assert sub-request `FullInputTokens()` equals original `FullInputTokens()`).

### Files expected to change

- `sim/request.go` — accessors (`FullInputTokens`, `InputLen`, `InputTokenSlice`) + chain struct (or `PrevRound` field, per chosen representation).
- `sim/workload/reasoning.go` — open-loop accumulate (L53-L57).
- `sim/workload/session.go` — closed-loop accumulate (L191-L205).
- `sim/batch_formation.go` — three call sites (L105, L132, L182-183).
- `sim/kv_store.go` and `sim/kv/cache.go` — hasher API to accept chained input or callers materialize via accessor.
- `sim/cluster/cluster.go`, `sim/cluster/cluster_event.go`, `sim/cluster/pd_events.go` — PD sub-request construction.
- `sim/routing_prefix_scorer.go`, `sim/routing_precise_prefix_scorer.go`, `sim/routing_nohit_lru_scorer.go` — pass chained input through hasher.
- `sim/simulator.go`, `sim/admission.go`, `sim/scheduler.go`, `sim/metrics_utils.go`, `sim/cluster/instance.go`, `sim/cluster/parent_request.go`, `sim/disaggregation.go` — `len()` migrations to `req.InputLen()`.
- `sim/workload/tracev2.go` — flatten via `FullInputTokens()` on export.
- Tests.

That is ~12-15 files modified. Larger than the parent tracker's "5–10 call sites" estimate, because the audit must be exhaustive (R4) and the closed-loop path is included.

### Invariants at risk

- **None directly.** Representation change.
- **INV-6 (Determinism)** — verified by tests above. Per-session RNG is unchanged; only the storage representation changes.
- **INV-13 (Run/replay parity)** — verified by parity tests above on both session modes.
- **INV-4 (KV cache conservation)** — unaffected. The KV cache sees the same token sequence; only the storage representation upstream changes.

### Notes for the implementer

- `req.ProgressIndex` semantics still refer to position within the *full* input. Don't change that — only storage changes.
- Tests in [`sim/workload/generator_test.go`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/generator_test.go) and any closed-loop session tests that pin specific input-token contents may need to call the accessor instead of indexing directly.
- Per project rule R4, grep for ALL construction sites of `*sim.Request{...}` before adding chain fields. Each site needs explicit handling (zero-value chain = no previous round).

### Labels

`hardening`, `performance`

namasl:
## Summary

Replaces the O(R²) per-round eager copy in multi-turn `accumulate` workloads with a per-session growable token buffer. Each round's `Request.InputTokens` becomes a flat sub-slice into a shared buffer; total per-session memory drops from O(R²) to O(R). Both the open-loop generator path (`reasoning.go`) and the closed-loop SessionManager path (`session.go`) are fixed.

## Issue

Closes #1445 (parent tracker: #1438, Change C).

## Chosen Representation

**Per-session `*[]int` buffer that rounds index into** (the parenthetical option in the issue's "Recommended: shared-tail chain"). Encapsulated by `SessionTokenBuffer` in `sim/workload/session_buffer.go`. Keeps `Request.InputTokens` a flat `[]int`, so the KV cache hasher, routing scorers, and PD sub-request copies continue to alias slice headers — preserving the storage win across the disaggregation pipeline.

PD sub-request construction (`cluster.go:2021`, `cluster_event.go:491`, `pd_events.go:129`) is **unchanged** — sub-requests share the parent's slice header, which is now a view into the shared session buffer (no flatten).

## What Changed

| Layer | Change |
|------|--------|
| API | New accessors `req.InputLen()`, `req.FullInputTokens()`, `req.InputTokenSlice(s,e)` on `*sim.Request` |
| Buffer | New `SessionTokenBuffer` (sim/workload/session_buffer.go) with `Append`/`Slice`/`Len`/`Cap` |
| Open-loop | `reasoning.go` routes accumulate through the shared buffer; prefix seeded into buffer (passed in as new parameter — `generator.go` no longer prepends per-round) |
| Closed-loop | `session.go` replaces `sess.contextTokens []int` with `*SessionTokenBuffer`; seed logic preserves legacy bounds guard |
| Migrations | ~28 length-only sites in `sim/`, `cmd/` migrated from `len(req.InputTokens)` / `util.Len64(req.InputTokens)` to `req.InputLen()`. KV hasher, routing scorers, and tiered cache use `req.FullInputTokens()` |

## Memory Measurements

Per-session input-token bytes (peak buffer capacity vs legacy O(R²) bound):

| Scenario | Legacy O(R²) | New peak | Win |
|---|---|---|---|
| open-loop R=15, in=100, out=50 | 17,250 | 2,560 | **6.7×** |
| open-loop R=50, in=100, out=50 | 188,750 | 9,216 | **20.5×** |
| open-loop R=15, in=1500, out=425 | 224,625 | 34,816 | **6.5×** (272 KB/sess) |
| open-loop R=50, in=1500, out=425 | 2,433,125 | 110,592 | **22.0×** (864 KB/sess) |
| closed-loop R=20, in=100, out=50 | 30,500 | 3,408 | **8.9×** |
| open-loop R=30, prefix=500, in=100, out=50 | 83,250 | 5,120 | **16.3×** |

All meet or exceed the issue's targets (≥5× at R=15, ≥20× at R=50). Reproducible via:
```
go test ./sim/workload/ -run "TestReasoningAccumulate_PerSessionMemoryIsLinear|TestSessionAccumulate_ClosedLoopMemoryIsLinear|TestReasoningAccumulate_WithPrefix_MemoryIsLinear|TestReportMemoryWinForPR" -v
```

## Cross-Path Parity

| `blis run` | yes | yes | Open-loop reasoning + closed-loop session |
| `blis replay` | yes | yes | Closed-loop session path covered |
| `blis observe` | partial | yes (length sites) | Token storage path not exercised; length-only migrations applied |

## Invariants Affected

- **INV-6 (Determinism):** Preserved. RNG call order unchanged; token VALUES at every position are byte-identical to pre-PR. Verified by new `TestSeedDeterminism_MultiTurnAccumulate` (token-value spot checks) + existing seed tests passing.
- **INV-13 (Run/replay parity):** Preserved. Trace export reads scalar length only; replay regenerates tokens via the same RNG. Existing replay parity test suite passes.
- **INV-4 (KV cache conservation):** Unaffected. Hasher sees the same flat token sequence.

## Sub-request Construction (R4)

The PD sub-request alias pattern (`InputTokens: orig.InputTokens`) at `cluster.go:2021`, `cluster_event.go:491`, `pd_events.go:129` is preserved unchanged — sub-requests are slice-header views into the shared session buffer, no flatten. Verified by all existing `sim/cluster/cluster_test.go` and `sim/cluster/disaggregation_test.go` cases passing.

## Test Plan

- [x] `go build ./...` — exit 0
- [x] `go test ./... -count=1` — all 11 packages OK
- [x] Open-loop memory test: ≥5× at R=15, ≥20× at R=50
- [x] Closed-loop memory test: ≥5× at R=20
- [x] Prefix-bearing workload memory test: ≥10× win confirms prefix is shared in buffer, not copied per-round
- [x] INV-6 multi-turn determinism: token VALUES at multiple positions byte-identical across runs
- [x] Existing session_test.go + reasoning_test.go behavioral tests pass (length invariants, prefix-once, multi-step accumulation, zero-suffix, prefix-only edge cases)
- [x] Buffer aliasing test: consecutive rounds share underlying array

### DIFF
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -1084,7 +1084,7 @@ func adaptForSessionManager(original *sim.Request, record *RequestRecord) *sim.R
 	}
 
 	outputCount := record.OutputTokens
-	adapted.ProgressIndex = int64(len(original.InputTokens) + outputCount)
+	adapted.ProgressIndex = original.InputLen() + int64(outputCount)
 
 	if outputCount > 0 {
 		adapted.OutputTokens = make([]int, outputCount)
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -1127,7 +1127,8 @@ func requestToPending(req *sim.Request, reqIndex int, noStreaming, unconstrained
 	if tokensPerWord <= 0 {
 		tokensPerWord = 1.0
 	}
-	wordCount := int(math.Round(float64(len(req.InputTokens)) / tokensPerWord))
+	inputLen := int(req.InputLen())
+	wordCount := int(math.Round(float64(inputLen) / tokensPerWord))
 	if wordCount <= 0 {
 		wordCount = 1
 	}
@@ -1136,7 +1137,7 @@ func requestToPending(req *sim.Request, reqIndex int, noStreaming, unconstrained
 	if req.PrefixGroup != "" && prefixes != nil {
 		if prefix, ok := prefixes[req.PrefixGroup]; ok {
 			prefixLen := prefixLengths[req.PrefixGroup]
-			suffixTokens := len(req.InputTokens) - prefixLen
+			suffixTokens := inputLen - prefixLen
 			if suffixTokens < 1 {
 				suffixTokens = 1
 			}
@@ -1144,19 +1145,19 @@ func requestToPending(req *sim.Request, reqIndex int, noStreaming, unconstrained
 			if suffixWords < 1 {
 				suffixWords = 1
 			}
-			suffixStart := len(req.InputTokens) - suffixTokens
+			suffixStart := inputLen - suffixTokens
 			if suffixStart < 0 {
 				suffixStart = 0
 			}
-			if suffixStart > len(req.InputTokens) {
-				suffixStart = len(req.InputTokens)
+			if suffixStart > inputLen {
+				suffixStart = inputLen
 			}
-			prompt = prefix + tokensToPrompt(req.InputTokens[suffixStart:], suffixWords)
+			prompt = prefix + tokensToPrompt(req.InputTokenSlice(int64(suffixStart), int64(inputLen)), suffixWords)
 		} else {
-			prompt = tokensToPrompt(req.InputTokens, wordCount)
+			prompt = tokensToPrompt(req.FullInputTokens(), wordCount)
 		}
 	} else {
-		prompt = tokensToPrompt(req.InputTokens, wordCount)
+		prompt = tokensToPrompt(req.FullInputTokens(), wordCount)
 	}
 
 	// Set min_tokens = max_tokens per-request so the server generates exactly MaxOutputLen
@@ -1169,7 +1170,7 @@ func requestToPending(req *sim.Request, reqIndex int, noStreaming, unconstrained
 
 	return &PendingRequest{
 		RequestID:       reqIndex,
-		InputTokens:     len(req.InputTokens),
+		InputTokens:     int(req.InputLen()),
 		MaxOutputTokens: req.MaxOutputLen,
 		Model:           req.Model,
 		Streaming:       req.Streaming && !noStreaming,
--- a/cmd/root_seed_test.go
+++ b/cmd/root_seed_test.go
@@ -125,3 +125,81 @@ func TestSeedOverride_YAMLSeedPreserved_WhenCLINotSpecified(t *testing.T) {
 		}
 	}
 }
+
+// makeMultiTurnAccumulateSpec returns a WorkloadSpec exercising the multi-turn
+// accumulate path that PR #1445 changed (SessionTokenBuffer storage). Drives
+// the open-loop reasoning generator.
+func makeMultiTurnAccumulateSpec(seed int64) *workload.WorkloadSpec {
+	rr := 10.0
+	return &workload.WorkloadSpec{
+		Version: "2", Seed: seed, Category: "language", AggregateRate: rr,
+		Clients: []workload.ClientSpec{{
+			ID: "c1", TenantID: "t1", RateFraction: 1.0, SLOClass: "standard",
+			Arrival:    workload.ArrivalSpec{Process: "poisson"},
+			InputDist:  workload.DistSpec{Type: "gaussian", Params: map[string]float64{"mean": 80, "std_dev": 10, "min": 50, "max": 150}},
+			OutputDist: workload.DistSpec{Type: "exponential", Params: map[string]float64{"mean": 40}},
+			Reasoning: &workload.ReasoningSpec{
+				MultiTurn: &workload.MultiTurnSpec{
+					MaxRounds:     5,
+					ThinkTimeUs:   1000,
+					ContextGrowth: "accumulate",
+				},
+			},
+		}},
+	}
+}
+
+// TestSeedDeterminism_MultiTurnAccumulate verifies INV-6 under the delta-encoded
+// multi-turn path introduced in PR #1445: same seed must produce byte-identical
+// per-request arrival times, input lengths, AND input token VALUES across runs.
+// The token-value check is the load-bearing one — it catches representation
+// changes (e.g. SessionTokenBuffer slicing) that fail to preserve content.
+func TestSeedDeterminism_MultiTurnAccumulate(t *testing.T) {
+	const seed = 7777
+	spec1 := makeMultiTurnAccumulateSpec(seed)
+	spec2 := makeMultiTurnAccumulateSpec(seed)
+
+	horizon := int64(1e7)
+	r1, err := workload.GenerateRequests(spec1, horizon, 100)
+	if err != nil {
+		t.Fatal(err)
+	}
+	r2, err := workload.GenerateRequests(spec2, horizon, 100)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(r1) != len(r2) {
+		t.Fatalf("INV-6: different request counts: %d vs %d", len(r1), len(r2))
+	}
+	if len(r1) == 0 {
+		t.Fatal("expected non-empty request set for multi-turn accumulate workload")
+	}
+	for i := range r1 {
+		if r1[i].ArrivalTime != r2[i].ArrivalTime {
+			t.Errorf("INV-6: request %d arrival diverged: %d vs %d", i, r1[i].ArrivalTime, r2[i].ArrivalTime)
+		}
+		if r1[i].InputLen() != r2[i].InputLen() {
+			t.Errorf("INV-6: request %d input length diverged: %d vs %d", i, r1[i].InputLen(), r2[i].InputLen())
+		}
+		// Spot-check token VALUES at three positions in each request — first,
+		// middle, last. A representation change that breaks determinism would
+		// surface as a value mismatch at one of these.
+		a := r1[i].FullInputTokens()
+		b := r2[i].FullInputTokens()
+		if len(a) != len(b) {
+			t.Errorf("INV-6: request %d FullInputTokens length diverged", i)
+			continue
+		}
+		positions := []int{0, len(a) / 2, len(a) - 1}
+		for _, p := range positions {
+			if p < 0 || p >= len(a) {
+				continue
+			}
+			if a[p] != b[p] {
+				t.Errorf("INV-6: request %d token[%d] diverged: %d vs %d", i, p, a[p], b[p])
+				break
+			}
+		}
+	}
+}
--- a/sim/admission.go
+++ b/sim/admission.go
@@ -52,7 +52,7 @@ func (tb *TokenBucket) Admit(req *Request, state *RouterState) (bool, string) {
 		tb.currentTokens = min(tb.capacity, tb.currentTokens+refill)
 		tb.lastRefill = clock
 	}
-	cost := float64(len(req.InputTokens))
+	cost := float64(req.InputLen())
 	if tb.currentTokens >= cost {
 		tb.currentTokens -= cost
 		return true, ""
--- a/sim/batch_formation.go
+++ b/sim/batch_formation.go
@@ -102,7 +102,7 @@ func (v *VLLMBatchFormation) FormBatch(ctx BatchContext) BatchResult {
 		}
 		req := result.RunningBatch.Requests[reqIndex]
 
-		numNewTokens := util.Len64(req.InputTokens) - req.ProgressIndex
+		numNewTokens := req.InputLen() - req.ProgressIndex
 		// Chunked prefill for running requests
 		if numNewTokens > 0 {
 			if 0 < ctx.PrefillTokenThreshold && ctx.PrefillTokenThreshold < numNewTokens {
@@ -129,7 +129,7 @@ func (v *VLLMBatchFormation) FormBatch(ctx BatchContext) BatchResult {
 			ctx.ComputedTokens[req.ID] += numNewTokens
 		}
 		// Decode phase: allocate 1 token
-		if req.ProgressIndex >= util.Len64(req.InputTokens) && len(req.OutputTokens) > 0 {
+		if req.ProgressIndex >= req.InputLen() && len(req.OutputTokens) > 0 {
 			decodeTokens := int64(1)
 			// Proactive MaxModelLen cap (BC-1): skip decode at boundary.
 			// Equivalent to max(0, maxModelLen-1-PI) < 1, specialized for single-token decode.
@@ -179,8 +179,8 @@ func (v *VLLMBatchFormation) FormBatch(ctx BatchContext) BatchResult {
 			continue
 		}
 
-		cachedBlocks := ctx.KVCache.GetCachedBlocks(next.InputTokens)
-		numNewTokens := util.Len64(next.InputTokens) - util.Len64(cachedBlocks)*ctx.KVCache.BlockSize()
+		cachedBlocks := ctx.KVCache.GetCachedBlocks(next.FullInputTokens())
+		numNewTokens := next.InputLen() - util.Len64(cachedBlocks)*ctx.KVCache.BlockSize()
 
 		if 0 < ctx.PrefillTokenThreshold && ctx.PrefillTokenThreshold < numNewTokens {
 			numNewTokens = ctx.PrefillTokenThreshold
--- a/sim/cluster/cluster.go
+++ b/sim/cluster/cluster.go
@@ -2016,6 +2016,9 @@ func (cs *ClusterSimulator) executeDisaggregatedRouting(req *sim.Request, time i
 	cs.parentRequests[parent.ID] = parent
 
 	// Create prefill sub-request: same input, no output (completes after prefill).
+	// InputTokens is a slice-header alias of req.InputTokens (#1445) — the
+	// sub-request views the same underlying token buffer, no flatten. If
+	// Request.InputTokens ever becomes lazy/chained, this site must update.
 	prefillSubReq := &sim.Request{
 		ID:           parent.PrefillSubReqID,
 		InputTokens:  req.InputTokens,
--- a/sim/cluster/cluster_event.go
+++ b/sim/cluster/cluster_event.go
@@ -486,6 +486,9 @@ func (e *DisaggregationDecisionEvent) Execute(cs *ClusterSimulator) {
 
 	// Create prefill sub-request: same input, no output (completes after prefill).
 	// Output is intentionally nil: zero-output request completes at prefill end.
+	// InputTokens is a slice-header alias of e.request.InputTokens (#1445) — the
+	// sub-request views the same underlying token buffer, no flatten. If
+	// Request.InputTokens ever becomes lazy/chained, this site must update.
 	prefillSubReq := &sim.Request{
 		ID:           parent.PrefillSubReqID,
 		InputTokens:  e.request.InputTokens,
--- a/sim/cluster/instance.go
+++ b/sim/cluster/instance.go
@@ -433,7 +433,7 @@ func (i *InstanceSimulator) TransitionTo(state sim.InstanceState) {
 // the decode instance becomes non-routable between KVTransferStartedEvent
 // and KVTransferCompletedEvent.
 func (i *InstanceSimulator) ReserveTransferredKV(req *sim.Request) bool {
-	inputLen := int64(len(req.InputTokens))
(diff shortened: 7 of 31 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## Self-review: `/blis-pr-review` (delta-encoded multi-turn rounds, #1445)

Thanks for this — it's a clean, well-documented hardening change. I checked it out, rebuilt (`go build ./...` clean), ran `go test ./sim/... ./cmd/...` (all green), reconstructed the authoritative 31-file diff against `b26f027c`, and ran five focused review passes (correctness/standards, tests, error-handling, comments, type-design). Summary below.

### Verdict: **Ready to merge** — no blocking defects found.

The core design is sound and the load-bearing invariants hold. The items below are non-blocking hardening / observability / doc-precision suggestions; a couple are compelling enough that you may want to fold them in before merge, but none gate correctness.

### What I verified holds (the things that could have broken)
- **No mutation hazard.** The whole point of the shared buffer is that each round's `InputTokens` aliases one backing array, so any `append(req.InputTokens, …)` anywhere would corrupt a *later* round. A full grep of `sim/` and `cmd/` confirms **zero** append-through-`InputTokens` sites; the one sub-slicing consumer (`kv/cache.go`) makes a defensive copy. ✅
- **INV-6 determinism preserved.** The RNG-draw region of `session.go` is **unchanged** by the diff (only the storage representation changed, `contextTokens []int` → `*sessionTokenBuffer`). In `reasoning.go`, round-0 sampling is *hoisted* above the loop to size the buffer but is consumed by round 0 in the same order, so the draw sequence `[r0.in, r0.out, r0.tokens, r1.in, …]` is identical. ✅
- **R4 (canonical constructor).** `InputTokens` stays a plain `[]int` field, so no `sim.Request{…}` literal needed changing — the accessors are read-only seams. ✅
- **Continuity guard is sound under preemption.** Preemption resets `ProgressIndex` but never reassigns `InputTokens`, and a closed-loop session has exactly one in-flight request, so `req.InputLen() == buf.Len()` genuinely holds on every subsequent `OnComplete`. ✅
- **Cross-path parity.** `replay` is untouched (zero diff — it loads fixed trace tokens, no multi-turn follow-up generation); `observe`'s closed-loop path goes through the migrated accessors. INV-13 intact. ✅
- **INV-10 / INV-11** (session causality / completeness) preserved; arrival-time path unchanged and covered by `TestSession_AccumulateArrivalTime_INV10`.

### Recommendations (non-blocking)

**1. (Convergent, top pick) Use a three-index slice to enforce the no-mutation contract mechanically.**
`sessionTokenBuffer.Slice()` (session_buffer.go:73) and `Request.InputTokenSlice()` / `FullInputTokens()` (request.go) return `s.b[start:end]`, whose capacity extends to the end of the backing array. The "MUST NOT be modified" contract is therefore comment-enforced only. Returning `s.b[start:end:end]` caps the result at `len`, so a future stray `append` allocates a fresh array instead of silently overwriting the next round's tokens — zero runtime cost, and a natural fit for a hardening PR. (Raised independently by both the correctness and type-design passes.)

**2. Two error-handling paths in `session.go` could be a touch louder / safer:**
- `OnComplete` SessionID-not-found (≈:116) logs at `Warnf` then returns nil. Since this signals a blueprint/trace mismatch (never a normal path), `Errorf` would give it Sentry visibility and better match its "should never happen" nature.
- The over-cap branch (≈:233) logs `Errorf` but then continues with accumulated state the comment itself calls "suspect," which then propagates to every subsequent round. Worth considering cancelling the session instead so the corruption can't silently fan out. (The `max(progressIndex-inputLen, 0)` clamp at :184 is similarly unreachable-but-silent; a guarded `Errorf` would make a future violation observable.)

**3. Test coverage is strong; two gaps worth a follow-up:**
- The closed-loop buffer is created with **zero** capacity (`newSessionTokenBuffer()`), so reallocation *will* occur mid-session — yet the only aliasing test that crosses an append (`TestSessionTokenBufferSliceAliasesAcrossAppends`) pre-allocates to *avoid* realloc. A test that forces a grow and asserts the orphaned prior view is still content-correct would lock in the documented "intentional, bounded" hazard.
- `TestSeedDeterminism_MultiTurnAccumulate` proves same-seed-twice reproducibility but not preservation of the *legacy* output, and it only exercises the open-loop path. (I confirmed by inspection that the RNG order is unchanged, so the risk is low — this is a belt-and-suspenders ask.) A closed-loop determinism case + a small golden snapshot would close it.
- Minor: `TestReportMemoryWinForPR` has no assertions (logging only) so it can't fail — consider giving it a floor like its siblings or marking it clearly as a reporting helper.

**4. Doc-precision nits (since this PR's value is partly its documentation):**
- session_buffer.go:28 "cap ≤ 2× current len" overstates Go's growth guarantee (large slices grow ~1.25×). The 3× test bound is safe; the prose is just over-specific.
- reasoning.go:19 "eliminating the legacy O(R) prefix-copy tax per round" — a fixed prefix is O(P) per round / O(R·P) total, not "O(R) per round." The accumulated-context copy is the quadratic term.
- session_buffer.go:44 "every Slice() result remains valid through all subsequent appends" is true only while capacity isn't exceeded — worth the qualifier.

**5. Trivia:** reasoning.go:74 allocates an empty `newSessionTokenBuffer()` in non-accumulate mode that is never used — harmless, droppable.

Nice work overall — the continuity panic guard and the byte-for-byte malformed-trace fallback are particularly thoughtful, and the memory tests demonstrate the O(R) win convincingly.

<sub>🤖 Automated self-review via `/blis-pr-review` (read-only). Findings, not changes.</sub>

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## Pre-merge status

Re: the self-review above ([#issuecomment-4833864537](https://github.com/inference-sim/inference-sim/pull/1452#issuecomment-4833864537)). Verified against the current branch head — **nothing is required before merge.** All correctness-affecting and documentation findings are addressed in `4dab3eed` and `020b7c85`; `go build`, `go test ./...` (all 11 packages), and `golangci-lint` are clean.

### Addressed (11/12 recommendations)
- ✅ **Three-index slices** on `Request.FullInputTokens` (`[:n:n]`) and `InputTokenSlice` (`[start:end:end]`) — converts the no-mutation contract into a runtime-enforced guarantee. The buffer's internal `Slice()` intentionally keeps full capacity so memory tests can still measure peak via `cap(req.InputTokens)`. (+ `TestRequestInputTokenSliceCappedCapacity`)
- ✅ `OnComplete` SessionID-not-found: `Warnf` → `Errorf`
- ✅ Over-cap branch now **cancels the session** instead of continuing with suspect state (+ `TestSession_AccumulateOverCap_CancelsSession`)
- ✅ Negative `ProgressIndex - InputLen` clamp: guarded `Errorf` so future accounting drift is observable
- ✅ Forced-realloc orphaned-view content-correctness test (`TestSessionTokenBufferForcedReallocOrphanContentCorrect`)
- ✅ `TestReportMemoryWinForPR` now asserts hard floors (≥5× at R=15, ≥20× at R=50) instead of logging only
- ✅ Doc-precision: growth bound (`~2×` small / `~1.25×` large / `3×` ceiling), prefix copy `O(P)` per round / `O(R·P)` total, `WithCapacity` "within the pre-allocated capacity" qualifier
- ✅ Dropped the unused empty buffer allocation in the non-accumulate path

### Optional follow-up (non-blocking — does NOT gate merge)
- **Rec 3b** — a *closed-loop* (`SessionManager.OnComplete`) run-twice determinism test + small golden snapshot. Today `TestSeedDeterminism_MultiTurnAccumulate` covers the **open-loop** path only; the closed-loop path now has strong *correctness* coverage but no *determinism* (run-twice-identical) anchor. The reviewer explicitly flagged this as belt-and-suspenders ("RNG order confirmed unchanged by inspection, risk is low"). Fine to defer to a separate PR.

**Verdict: ready to merge.**

<sub>🤖 Verification summary via `/blis-pr-review` follow-up (read-only).</sub>

susiejojo (dismissed):

namasl:
@claude /blis-pr-review

susiejojo (approved):

PR #1446, opened 2026-06-24
refactor(cluster): RequestSource interface + eager adapter (#1439)

namasl:
## Summary

Foundation PR for #1438 Change A (lazy streaming workload generation). This PR is **deliberately behavior-preserving** — no memory win lands here. It exists to decouple the workload generator from the cluster's arrival-event handler so that a streaming generator can be dropped in later (A3) without touching cluster code or tests again.

- Adds a single-method `RequestSource` interface in `sim/cluster/`: `Next() (*sim.Request, bool)`.
- Adds the trivial eager adapter `SliceRequestSource` wrapping `[]*sim.Request`. Value-identity over the input slice — same elements, same order, no copy, no transformation.
- Migrates `cluster.NewClusterSimulator` signature from `[]*sim.Request` to `RequestSource` (in place — no parallel constructor, per R4).
- Updates all ~270 call sites (`blis run`, `blis replay`, 28 cluster test files, `cmd/replay_test.go`) to wrap via `cluster.NewSliceRequestSource(...)`. Mechanical edits.
- Replaces `Run()`'s slice-range arrival-injection with a `Next()`-driven drain. The empty-source warning is preserved via a local `arrivalCount` counter (since the streaming source can't cheaply expose `Len()`).

## Review-response fixes (commit `c273803`)

Applied the should-fix and important findings from the multi-agent review on the first push:

- **Nil `RequestSource` guard in `NewClusterSimulator`** — matches the existing `NumInstances < 1` panic style; fails fast at the construction site instead of an opaque nil-deref inside `Run()` when A3's second implementation lands.
- **Caller-obligation wording** — the interface's order requirement is now phrased as "Implementations MUST yield..." (caller obligation, not verified) instead of "guaranteed to yield..." which read as an enforcer claim.
- **`(nil, true)` prohibition documented** — `Run()` dereferences the returned request; one doc line covers what the prior slice-range code also assumed implicitly.
- **`TestNewClusterSimulator_PanicsOnNilRequestSource`** — covers the new guard.

The minor items (cluster-level drain spy, aliasing-hazard wording, "future test" comment) the reviewer marked low-priority/acceptable — left for a follow-up.

## Behavioral Contracts

- **BC-1 / BC-2 / BC-3 (adapter):** Order + identity, sticky exhaustion, empty/nil exhausts immediately. Covered by `request_source_test.go`.
- **BC-4 (cluster drains source):** Every existing cluster test (~263 of them) now exercises this path; a counting helper test asserts drain semantics directly.
[...]

PR #1453, opened 2026-06-26
feat(workload): lazy --lazy-generation streaming source (alpha, #1441)

### LINKED ISSUE #1441: A3 — Add lazy request generation behind `--lazy-generation` (alpha, default off) (#1438)
## A3 — Add lazy request generation behind `--lazy-generation` (alpha, default off)

**Tracker:** #1438 · **Change A — PR 3 of 5** · **Depends on:** A1 (#1439, merged), A2 (#1440)

**This is the PR where the lazy generator lands.** A1 and A2 set up the boundaries; this PR introduces a streaming `RequestSource` implementation that produces requests one at a time, with resident memory O(in-flight requests) instead of O(total requests).

Lazy generation is introduced as an **alpha feature** behind a boolean CLI flag. Eager generation remains the default and the on-disk format and observable behavior are unchanged when the flag is off. The flag exists so we can ship the new code path, soak it under tests + experiments, and graduate it once we are confident (see A6/A7/A8 in the parent tracker for the deprecation arc).

The two earlier PRs touch the cluster and the trace path. This PR is concentrated in `sim/workload/generator.go`. Keeping it narrow makes the determinism review tractable: per-client RNG state + heap-merge interleaving is the most subtle part of #1438, and reviewers should not have to chase it across multiple files.

### What the PR does

1. **Add a boolean CLI flag `--lazy-generation`** (default `false`) plumbed through `cmd/root.go` and `cmd/replay.go`. When false, behavior is byte-identical to today. When true, the workload generator emits requests through the new streaming source.

2. **Add a lazy streaming `RequestSource` implementation** in `sim/workload/` (e.g., new `sim/workload/stream.go`). A1 (#1439, merged) placed the `RequestSource` interface and the eager `SliceRequestSource` adapter in [`sim/cluster/request_source.go`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/cluster/request_source.go). A3's lazy source lives in `sim/workload/` and **implements `cluster.RequestSource`** — the interface does not move. Both source implementations live side by side, selected at construction time based on the flag value.

3. **Per-client generator state.** Each cohort/client carries its own RNG (already true today via `clientRNG`) plus enough state to produce its next request on demand. The state is small — a sampler position, the prefix tokens, and the round/session bookkeeping for multi-turn workloads (`reasoning.go`). Two reasoning-client shapes must be supported (mirroring [`generator.go#L155-L209`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator.go#L155) vs the non-`SingleSession` branch):
   - **`SingleSession=true`**: one session per client, all rounds pre-computed by `GenerateReasoningRequests` with monotonically increasing arrival times. Streaming state is `(roundIndex, []*sim.Request)` — yield rounds in order, filtering against horizon and lifecycle windows exactly as the eager loop does.
   - **`SingleSession=false` (default)**: each new IAT starts a fresh session. Streaming state is the IAT sampler plus the current session's `(roundIndex, []*sim.Request)`; when the current session is exhausted, draw the next IAT and start a new session.

4. **A small priority queue keyed by `nextArrivalTime`.** Initially, each client pushes its first request's arrival time. The streaming source's `Next()` (or peek+pop) pops the earliest, asks that client to produce the request, advances that client to its next candidate, and re-pushes the client. The total request stream produced is the same one the eager merge-sort produces today.

5. **In-loop `--num-requests` truncation.** Today, [`sim/workload/generator.go#L99-L107`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator.go#L99) sets `perClientCap = 2 * maxRequests` and [L366-L369](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator.go#L366) truncates after merge. Lazy mode counts as it goes and stops at `maxRequests` — but the eager path retains its current behavior. The `perClientCap = 2 * maxRequests` factor only fully disappears once eager is removed (A8).

6. **Sequential ID assignment.** Today IDs are assigned post-sort (`req.ID = fmt.Sprintf("request_%d", i)` after the merge). Lazy mode assigns IDs in the order requests pop off the priority queue — which is arrival-sorted, so IDs remain arrival-sorted. The eager-vs-lazy ID assignment must match exactly (including ties — the eager path uses `sort.SliceStable` which preserves client order on ties; the lazy priority queue must too). This is enforced by the property tests in A4.

### The central correctness claim

For any seed and workload spec supported by both modes, the request streams produced by eager and lazy generation are **byte-identical** — same count, same IDs, same arrival times, same input/output token slices, in the same order. Downstream metrics, trace exports, and stdout for `blis run` are consequently byte-identical between modes.

This is what makes the alpha flag safe: turning it on cannot change any observable except peak memory. Verifying this claim is the load-bearing job of A4 (property tests).

### Determinism: the load-bearing concern (INV-6)

Same seed must still produce byte-identical stdout in lazy mode. Three correctness conditions:

1. **Per-client RNGs advance in the same sequence as today.** Each client's RNG state is independent; advance order is determined only by what *that client* produces, not by the merge order. This is already how `clientRNG` works today — preserve it.

2. **Per-client seeds are drawn in client-index order before any request is yielded.** The eager path draws `clientSeed := workloadRNG.Int63()` once per client at [`generator.go#L119`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator.go#L119), in the iteration order of `allClients`. The lazy source MUST advance `workloadRNG` in the same client-index order during construction (before pumping any client for its first request), or per-client RNG streams will diverge from the eager path and stdout byte-identity is lost. Concretely: in `NewLazyRequestSource`, iterate `allClients` once, draw `clientSeed` per client, and only then begin building the initial priority queue.

3. **Priority-queue tie-breaking is deterministic.** The queue's ordering key is `(arrival_us, client_index, per_client_seq)` where `client_index` is the position in `allClients` and `per_client_seq` is the production order that client emits (IAT-draw order for non-reasoning clients; round order within a session for reasoning clients). This matches the eager path's `sort.SliceStable` over `allRequests`, which preserves the append order — and `allRequests` is appended in `range allClients` order, then per-client in production order. **Note:** `per_client_seq` is internal state, not an externally observable field on `*sim.Request`; it lives only inside the streaming source.

Add a same-seed determinism test that runs lazy mode twice and asserts byte-identical stdout. A4 strengthens this into a full property-test harness.

### Scope guard: paths NOT covered by lazy mode in this PR

- The **time-varying / per-window-parameter path** at [`sim/workload/generator.go#L92-L94`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator.go#L92) (`generateTimeVaryingRequests`). This path is structurally different (window-based proportional allocation). When `--lazy-generation` is set but `hasPerWindowParameters(allClients)` is true, the workload generator must transparently fall back to the eager path and log a one-line warning (so the alpha flag never silently does the wrong thing). File a follow-on issue to migrate the time-varying path to lazy. The dominant memory cliff is the non-time-varying multi-turn case (per #1438's reproducer), which lazy mode does cover.

- **Replay** path uses [`sim/workload/replay.go`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/replay.go) and is unaffected — it loads requests from a trace file. `cmd/replay.go` **accepts `--lazy-generation` and silently ignores it** (no error, no warning) for CLI symmetry with `blis run`. This matches the pattern already used for other generation-side flags that are accepted-but-ignored on replay (e.g., `--num-requests` is honored by replay only as a slice over the trace, not as a generator control).

### Materialization status after A1 + A2

With A1 (#1439, merged) and A2 (#1440, dependency) in place, **no pre-`Run()` consumer re-materializes the full request list when `--lazy-generation` is on**. Specifically:

- The cluster consumes requests through `cluster.RequestSource.Next()` (A1).
- Trace export is hooked off the per-arrival emission path (A2), not the materialized slice.
- `cmd/root.go` does not iterate the request list before calling `Run()` (verify during implementation; if any pre-`Run()` summary stat still iterates the slice, move it behind a flag-aware path or delete it).

Therefore the memory-savings claim in the acceptance criteria is **unconditional** for non-time-varying workloads when `--lazy-generation` is on. If implementation uncovers a missed pre-`Run()` consumer, call it out in the PR description and either move it or document the residual materialization point.

### Acceptance criteria

- `go test ./...` passes, including the workload generator tests around [`sim/workload/generator_test.go`](https://github.com/inference-sim/inference-sim/blob/b26f027c/sim/workload/generator_test.go). Both modes covered.
- **Default behavior unchanged.** Without `--lazy-generation`, `blis run` stdout and trace output are byte-identical to `main` for at least one chatbot, one code-gen, and one closed-loop session workload at a fixed seed.
- **Eager ≡ lazy byte-identity.** For the same seed/workload/model on at least the same three workloads above, `blis run --lazy-generation` produces byte-identical stdout and byte-identical TraceV2 output (header YAML + data CSV) to `blis run` without the flag. A4 generalizes this with property tests.
- **Memory measurement.** PR description must include peak-RSS measurements (via `/usr/bin/time -v` or equivalent) on a code-generation-shaped workload, comparing eager vs lazy mode. Suggested workload (per #1438 reproducer): N=40 cohort members, R=15 max rounds, 30K-token prefix, 1500-token input, 425-token output, `num_requests=500`. Expected: tens of GB → tens-to-hundreds of MB peak when lazy mode is on.
- **Determinism (INV-6).** Lazy-mode same-seed byte-identical-stdout test passes.
- **Time-varying fallback.** A test that sets `--lazy-generation` on a workload with per-window parameters and asserts the eager path is taken with a warning logged.
- **Replay accepts the flag.** A test that `blis replay --lazy-generation ...` runs without error and produces output byte-identical to `blis replay` without the flag (since replay reads from trace, the flag is a no-op).

### Files expected to change

- `cmd/root.go` and `cmd/replay.go` (add the `--lazy-generation` flag and thread it into the workload generator selection; replay accepts-and-ignores)
- `sim/workload/generator.go` (add the lazy code path; the eager path is **untouched**)
- `sim/workload/reasoning.go` for multi-turn streaming state (keep changes minimal — Change C is the deeper rewrite of this file)
- `sim/workload/cohort.go` if cohort iteration moves into the streaming source
- New file: priority queue + streaming source (`sim/workload/stream.go` or similar), implementing `cluster.RequestSource`
- Tests

That's 4-5 files modified plus 1-2 new files. Larger than A1/A2 by necessity — this is the actual rewrite. The compensating decision is that this PR does not modify any consumer of the workload (the cluster + trace export already consume through abstractions from A1 + A2).

### Invariants at risk

- **INV-6 (Determinism):** central concern. Same-seed test required in this PR; property test in A4.
- **INV-13 (Run/replay parity):** test required in this PR; strengthened in A4.
- **INV-1 (Request conservation), INV-2 (Lifecycle), INV-3 (Clock monotonicity):** preserved trivially — the request stream is identical, just produced lazily.

### Notes for the implementer

- The priority queue is `O(C log C)` total over `C = maxRequests` requests, which is cheaper than today's eager `O(N log N)` merge-sort over `N = sum(perClientCap) >= 2 * C * num_clients`. Performance should improve in addition to memory when lazy is on.
- Don't try to also fix Change B or Change C in this PR. Each is a separate sub-issue (#1444, #1445).
- `Request.ID` assignment must match the eager path's "request_%d" format and arrival-order numbering. Tests in `cmd/root_test.go` and elsewhere may pin specific IDs — run the full suite under both modes.
- The flag is intentionally simple (boolean). If we later need finer-grained source selection, the flag can evolve into an enum — but during alpha, one switch is enough.

### Labels

`hardening`, `refactor`, `architecture`

namasl:
## Summary

- Adds `--lazy-generation` (alpha, default off) on `blis run` so the workload generator emits requests through a streaming `RequestSource` interleaved by a `(arrival, clientIdx, perClientSeq)` priority queue, instead of pre-generating the full `[]*sim.Request` slice. Memory drops from O(total_requests) to O(num_clients + max_session_rounds).
- New file `sim/workload/stream.go` plus a focused unit test file. The lazy source structurally satisfies `cluster.RequestSource` (no new import edge from `sim/workload` to `sim/cluster`). Eager paths (`GenerateRequests`, `GenerateWorkload`, time-varying generator) are untouched.
- Three fallback gates so the flag is safe: time-varying parameters, concurrency clients, and multi-session reasoning (`SingleSession=false`) cause a one-line warning and fall through to the eager `GenerateWorkload`. Single-shot, single-session reasoning, prefix-group sharing, and cohort-expanded workloads are in scope for the alpha.
- `blis replay` accepts `--lazy-generation` and ignores it for CLI symmetry (replay never invokes the generator).

## Issue

Closes #1441 (Tracker: #1438 · Change A — PR 3 of 5 · depends on A1 #1446 + A2 #1448).

## Why this PR

The eager generator builds `[]*sim.Request` with a `perClientCap = 2 * maxRequests` per client and merge-sorts the union before truncating. For multi-turn cohort workloads at the scale called out in #1438 (N=40 cohort members × R=15 max rounds × 30K-token prefix × num_requests=500), that pre-merge slice peaks in the tens of GB — well before the simulator's algorithmic limits. Lazy mode streams each request through the cluster the moment it is produced, releasing token-slice memory as soon as the cluster has it on its event heap.

A1 (#1446) put the `cluster.RequestSource` interface in place; A2 (#1448) moved trace export off the materialized slice onto the per-arrival hook. With both in place, no pre-`Run()` consumer re-materializes the request list when `--lazy-generation` is on, so the memory-savings claim is unconditional for the supported workload shapes.

## Memory measurement

Smoke-tested with a 2-client × 10K-prefix × 200-request workload spec under `--lazy-generation`:

| Mode  | Peak RSS |
|-------|----------|
| Eager | 152.94 MB |
| Lazy  |  63.40 MB |

~58% peak-RSS reduction on this modest workload. The savings ratio grows with cohort size and prefix length — the issue's reproducer (40 cohort members, 30K-token prefix, 500 requests) is where the tens-of-GB → tens-to-hundreds-of-MB claim from #1441 lands.

## Byte-identity verification

Verified byte-identical stdout + trace YAML + trace CSV between `blis run` and `blis run --lazy-generation` at fixed seeds across three workload shapes:

| Workload | Seed | Result |
|----------|------|--------|
| distribution mode (50 reqs, rate=5) | 42 | stdout identical, trace identical |
| chatbot preset (30 reqs, rate=3) | 1234 | stdout identical |
| 2-client + 10K prefix YAML spec (200 reqs) | 7 | stdout identical |

Verified via diff of captured outputs. The cmd/root_test.go end-to-end test `TestRunCmd_LazyVsEager_Distribution_ByteIdentical` covers this in CI.

## Behavioral contracts (BCs)

- **BC-1 / BC-9** — Flag wired through both commands. `blis run --lazy-generation` parses; `blis replay --lazy-generation` is a no-op.
- **BC-2** — Eager mode unchanged (flag off → byte-identical to `main` at upstream head).
- **BC-3 / BC-4** — Lazy stream is byte-identical to the eager stream for every supported workload (same length, IDs, arrival times, token slices, in the same order). Trace YAML + CSV identical.
- **BC-5** — Determinism (INV-6): two lazy runs at the same seed produce byte-identical stdout.
- **BC-6** — In-loop `maxRequests` truncation. Eager's per-client `2 * maxRequests` safety cap is intentionally not applied in lazy mode.
- **BC-7** — Sequential IDs `request_<i>` assigned in arrival-sorted order with `clientIdx` tie-break (matches eager's `sort.SliceStable`).
- **BC-8** — Time-varying / concurrency / multi-session-reasoning fallback: one warning line, eager generator runs.
- **BC-10** — No fields added to any existing exported struct; no changes to `cluster.RequestSource`.

## Determinism (INV-6, the central correctness concern)

1. `workloadRNG.Int63()` per-client seeds are drawn in `allClients` order at construction time, **before any client produces its first request** — matching eager `generator.go:119`.
2. `generatePrefixTokens(allClients, workloadRNG)` runs in the same RNG state as eager.
3. Priority queue ordering key is `(arrivalUs, clientIdx, perClientSeq)` — the `clientIdx` term preserves `sort.SliceStable`'s tie-break behavior.
4. SessionBlueprints are constructed in a pre-pass that replays each closed-loop client's RNG to enumerate session IDs, then drawn from `blueprintRNG` in sorted-session-IDs-per-client order — matching `generator.go:506`.
5. Sequential IDs are assigned in heap-pop yield order, matching `GenerateWorkload`'s post-filter renumbering at `generator.go:619-621`.
6. Intermediate closed-loop rounds (`RoundIndex > 0`) are popped from the heap and **count toward maxRequests**, but are NOT yielded to the cluster — matching eager's "produce all rounds, truncate to maxRequests, then filter round-0" sequencing.

## Scope deviations from #1441

Documented in the plan's deviation log:

1. **Multi-session reasoning (`SingleSession=false`) falls back to eager** with a warning. The eager path's per-client multi-session loop produces all sessions in series and then global-sorts them; a single-entry-per-client streaming source cannot reproduce that interleaving correctly without pre-drawing every session's anchor. Single-session reasoning (the inference-perf shape and the #1441 reproducer shape per its `GenerateReasoningRequests` reference) IS supported.
2. **Concurrency clients (`Concurrency > 0`) fall back to eager** with a warning. Concurrency clients generate seed requests via `concurrencyRNG` in `GenerateWorkload`'s separate path; adding their merge into the heap is out of the alpha scope.
3. **Time-varying parameters fall back to eager** with a warning, as called out in the issue itself.

Each fallback is enforced by a sentinel error (`ErrLazyUnsupportedMultiSession`, `ErrLazyUnsupportedConcurrency`, `ErrLazyUnsupportedTimeVarying`) so the gating cannot drift silently.

## Saturation analysis under lazy mode

In lazy mode `preGeneratedRequests` is nil, so the existing `preGeneratedRequests + followUpRequests + sort` assembly for `--saturation-report` is unavailable. Instead, the per-arrival hook (added by A2 #1448) is installed unconditionally when lazy mode + `--saturation-report` are both set; the hook delivers request pointers in clock-monotonic arrival order (INV-3), so saturation reads them directly with no post-sort. Eager + saturation is unchanged.

## Cross-Path Parity

| `blis run` | yes | yes | Lazy source feeds the cluster via `cluster.RequestSource`; arrival hook handles both trace export (A2) and saturation. |
| `blis replay` | flag accepted (BC-9) | yes (no-op) | Replay reads from a captured trace; the flag is for CLI symmetry only. |
| `blis observe` | no | n/a | Explicitly out of scope for this PR (Change A part 3 of 5). `cmd/observe_cmd.go:386` calls `workload.GenerateWorkload(...)` eagerly and mutates each `wl.Requests[i]` (streaming flags, think-time) — the #1438 memory spike applies equally to `observe`. Migration tracked in follow-up issue #1455. |

## Invariants Preserved

- **INV-6 (Determinism):** central concern — see "Determinism" section. Tested by same-seed determinism test + property byte-identity tests.
- **INV-13 (Run/replay parity):** trace export is unchanged (still flows through the A2 arrival hook); the lazy source produces the same arrival stream eager does, so replay sees an identical trace.
- **INV-1 / INV-2 / INV-3:** preserved trivially — the cluster sees the same request stream, just produced lazily.

## Test plan

- [x] `go build ./...`
- [x] `go test ./... -count=1 -timeout 240s` — all packages pass
- [x] **Streaming unit tests** (`sim/workload/stream_test.go`): per-client production matches eager; priority-queue tie-break by `clientIdx`; sequential IDs in arrival order; `maxRequests` truncation; same-seed determinism; prefix-group sharing; single-session reasoning byte-identity; blueprint pre-pass matches eager; all three fallback paths; zero-horizon and negative-`maxRequests` edge cases.
- [x] **End-to-end byte-identity** (`cmd/root_test.go`): `TestRunCmd_LazyVsEager_Distribution_ByteIdentical` runs blis twice with same seed and asserts YAML + CSV bytes equal; `TestRunCmd_LazyGeneration_SameSeed_Deterministic` asserts INV-6.
- [x] **Replay flag** (`cmd/replay_test.go`): `TestReplayCmd_LazyGenerationFlag_AcceptedAsNoOp` asserts the flag is registered with the expected default.
- [x] **Real binary smoke**: `./blis run --seed 42 --num-requests 50 --rate 5.0 --trace-output …` produces byte-identical stdout/YAML/CSV with vs without `--lazy-generation` (distribution mode); same for chatbot preset.
- [x] **Time-varying fallback smoke**: `./blis run --workload-spec <spec-with-lifecycle-window-overrides> --lazy-generation` logs the expected one-line warning and completes successfully via eager.

## Files changed

- `sim/workload/generator.go` — extracts the spec-mutating prelude into a reusable `validateAndExpandSpec` helper. Behavior-preserving.
- `sim/workload/stream.go` — new. The `lazyRequestSource`, `clientStreamState`, priority queue, and `GenerateWorkloadLazy` factory.
- `sim/workload/stream_test.go` — new. Streaming-source unit tests.
- `cmd/root.go` — adds the `--lazy-generation` flag and wires the streaming source into `cluster.NewClusterSimulator`; routes saturation analysis through the arrival hook in lazy mode.
- `cmd/replay.go` — adds the `--lazy-generation` flag as a no-op.
- `cmd/root_test.go` — adds three end-to-end tests (flag registered, eager ≡ lazy byte-identity, same-seed determinism).
- `cmd/replay_test.go` — adds the no-op flag test.
- `CLAUDE.md` — documents the new flag and its fallback rules.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,6 +114,17 @@ go build -o blis main.go
 ./blis calibrate --trace-header t.yaml --trace-data d.csv --sim-results results.json \
   --slo-ttft "critical=100ms" --slo-e2e "critical=5s" --report calibration.json
 
+# Run with lazy request generation (alpha, #1441). Streams requests from the
+# workload generator into the cluster instead of pre-generating the full
+# slice — reduces peak memory from O(total_requests) to O(in_flight_requests).
+# Falls back to the eager generator with a one-line warning for:
+#   - workloads with per-window parameters (time-varying)
+#   - concurrency clients (Concurrency > 0)
+#   - reasoning clients with SingleSession=false (multi-session)
+# Supported: single-shot, single-session reasoning, prefix-group sharing,
+# multi-client / cohort workloads. Behavior with the flag off is unchanged.
+./blis run --model qwen/qwen3-14b --lazy-generation
+
 # Convert workload formats
 ./blis convert preset --name chatbot --rate 10 --num-requests 100
 ./blis convert servegen --path data/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,7 +116,11 @@ go build -o blis main.go
 
 # Run with lazy request generation (alpha, #1441). Streams requests from the
 # workload generator into the cluster instead of pre-generating the full
-# slice — reduces peak memory from O(total_requests) to O(in_flight_requests).
+# slice — reduces peak generator memory from O(total_requests) to
+# O(num_clients + max_session_rounds): the heap holds one entry per client,
+# and per-client reasoning state holds at most one session's pending rounds.
+# (Cluster-side memory still scales with in-flight cluster requests, which
+# this PR does not change.)
 # Falls back to the eager generator with a one-line warning for:
 #   - workloads with per-window parameters (time-varying)
 #   - concurrency clients (Concurrency > 0)
--- a/cmd/replay.go
+++ b/cmd/replay.go
@@ -784,6 +784,12 @@ func init() {
 	replayCmd.Flags().StringVar(&goodputSLOTTFT, "slo-ttft", "", "Per-class TTFT goodput thresholds (e.g. \"critical=100ms,standard=500ms\"). Precedence: CLI > trace header > workload spec.")
 	replayCmd.Flags().StringVar(&goodputSLOITL, "slo-itl", "", "Per-class mean ITL goodput thresholds (e.g. \"critical=50ms,standard=150ms\").")
 	replayCmd.Flags().StringVar(&goodputSLOE2E, "slo-e2e", "", "Per-class E2E goodput thresholds (e.g. \"critical=5s,standard=30s\").")
+	// --lazy-generation: accepted for CLI symmetry with `blis run` (#1441),
+	// but ignored — replay reads requests from a captured trace and never
+	// invokes the workload generator. The flag binds to a throwaway local
+	// so no global state is mutated. BC-9.
+	var replayLazyGenerationIgnored bool
+	replayCmd.Flags().BoolVar(&replayLazyGenerationIgnored, "lazy-generation", false, "Accepted for symmetry with `blis run` (#1441); has no effect on replay.")
 	rootCmd.AddCommand(replayCmd)
 }
 
--- a/cmd/replay_test.go
+++ b/cmd/replay_test.go
@@ -387,6 +387,26 @@ func TestReplayCmd_TraceOutputFlag_Registered(t *testing.T) {
 	}
 }
 
+// TestReplayCmd_LazyGenerationFlag_AcceptedAsNoOp pins BC-9 of #1441:
+// --lazy-generation is registered on replay for CLI symmetry with
+// `blis run` but has no effect (replay reads requests from a captured
+// trace and never invokes the workload generator).
+func TestReplayCmd_LazyGenerationFlag_AcceptedAsNoOp(t *testing.T) {
+	f := replayCmd.Flags().Lookup("lazy-generation")
+	if f == nil {
+		t.Fatal("replayCmd missing --lazy-generation flag (BC-9, #1441)")
+	}
+	if f.DefValue != "false" {
+		t.Errorf("--lazy-generation default = %q, want %q", f.DefValue, "false")
+	}
+	// Flag value must parse without error.
+	if err := replayCmd.Flags().Set("lazy-generation", "true"); err != nil {
+		t.Errorf("flag parse failed: %v", err)
+	}
+	// Reset for other tests.
+	_ = replayCmd.Flags().Set("lazy-generation", "false")
+}
+
 func TestReplayCmd_TraceHeaderFlag_Registered(t *testing.T) {
 	// GIVEN the replay command
 	// WHEN checking for --trace-header flag
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1447,6 +1447,13 @@ var runCmd = &cobra.Command{
 		// (because expansion inside GenerateWorkloadLazy happens AFTER
 		// applyTimeoutToSpec runs), producing the default 300 s deadline
 		// instead of the user-requested value (PR #1453 self-review).
+		//
+		// REMOVING THIS CALL re-introduces that bug silently. The
+		// regression is covered by
+		// TestGenerateWorkloadLazy_InferencePerf_TimeoutAppliedAfterPreExpand
+		// at the library layer; the cmd-level smoke is covered by the
+		// inference-perf byte-identity check verified during PR review.
+		//
 		// ExpandClientsAndCohorts is idempotent — the generators'
 		// validateAndExpandSpec runs it again with no effect since both
 		// branches guard on len(spec.Clients) == 0. (#1441)
@@ -1899,6 +1906,13 @@ var runCmd = &cobra.Command{
 				// clock-monotonic order — no separate followUpRequests merge
 				// or post-sort required (the cluster delivers them in arrival
 				// order via the hook).
+				//
+				// SAFETY: allRequests aliases the same backing array as
+				// traceArrivals. Both downstream consumers (trace export
+				// below + saturation analysis) MUST be read-only of this
+				// slice — neither appends, reorders, nor mutates element
+				// contents. If a future consumer needs to mutate, copy
+				// first: `allRequests = append([]*sim.Request(nil), traceArrivals...)`.
 				allRequests = traceArrivals
 			} else {
 				allRequests = make([]*sim.Request, 0, len(preGeneratedRequests)+len(followUpRequests))
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1439,6 +1439,20 @@ var runCmd = &cobra.Command{
 		if requestTimeoutSecs == 0 {
 			logrus.Fatalf("--timeout must be positive (seconds) or negative to disable; got 0")
 		}
+		// Pre-expand inference-perf / ServeGen specs so the timeout-
+		// application step below sees every client — including those
+		// populated by expansion. Without this, --lazy-generation +
+		// --workload-spec=inference_perf.yaml + an explicit --timeout
+		// would build streaming states from clients whose Timeout is nil
+		// (because expansion inside GenerateWorkloadLazy happens AFTER
+		// applyTimeoutToSpec runs), producing the default 300 s deadline
+		// instead of the user-requested value (PR #1453 self-review).
+		// ExpandClientsAndCohorts is idempotent — the generators'
+		// validateAndExpandSpec runs it again with no effect since both
+		// branches guard on len(spec.Clients) == 0. (#1441)
+		if err := workload.ExpandClientsAndCohorts(spec); err != nil {
+			logrus.Fatalf("Failed to expand workload spec: %v", err)
+		}
 		if workloadSpecPath == "" || cmd.Flags().Changed("timeout") {
 			applyTimeoutToSpec(spec, requestTimeoutSecs)
 		}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -127,6 +128,7 @@ var (
 
 	// Workload spec config (PR10)
 	workloadSpecPath string // Path to YAML workload specification file
+	lazyGeneration   bool   // --lazy-generation: stream requests from generator (alpha, #1441)
 
 	// Tiered KV cache config (PR12)
 	kvCPUBlocks             int64
@@ -1452,13 +1454,41 @@ var runCmd = &cobra.Command{
 			logrus.Fatalf("Workload requires either num_requests or --horizon to bound generation")
 		}
 
-		wl, err := workload.GenerateWorkload(spec, simulationHorizon, maxRequests)
-		if err != nil {
-			logrus.Fatalf("Failed to generate workload: %v", err)
+		// Lazy generation path (#1441, alpha). Default off. When set, build
+		// a streaming workload source instead of materializing the full
+		// request slice. Falls back to eager (with a warning) for specs
+		// the streaming source cannot handle yet (time-varying parameters,
+		// concurrency clients, multi-session reasoning).
+		var wl *workload.GeneratedWorkload
+		var lazyRequestSource interface{ Next() (*sim.Request, bool) }
+		if lazyGeneration {
+			src, sessions, followUpBudget, lazyErr := workload.GenerateWorkloadLazy(spec, simulationHorizon, maxRequests)
+			switch {
+			case errors.Is(lazyErr, workload.ErrLazyUnsupportedTimeVarying):
+				logrus.Warnf("[workload] --lazy-generation ignored: workload has per-window parameters; using eager generator (issue #1441)")
+			case errors.Is(lazyErr, workload.ErrLazyUnsupportedConcurrency):
+				logrus.Warnf("[workload] --lazy-generation ignored: workload has concurrency clients; using eager generator (issue #1441)")
+			case errors.Is(lazyErr, workload.ErrLazyUnsupportedMultiSession):
+				logrus.Warnf("[workload] --lazy-generation ignored: workload has multi-session reasoning (SingleSession=false); using eager generator (issue #1441)")
+			case lazyErr != nil:
+				logrus.Fatalf("Failed to build lazy workload: %v", lazyErr)
+			default:
+				lazyRequestSource = src
+				wl = &workload.GeneratedWorkload{Sessions: sessions, FollowUpBudget: followUpBudget}
+			}
+		}
+		if wl == nil {
+			var err error
+			wl, err = workload.GenerateWorkload(spec, simulationHorizon, maxRequests)
+			if err != nil {
+				logrus.Fatalf("Failed to generate workload: %v", err)
+			}
 		}
 		// Re-apply timeout to generated requests and session blueprints.
 		// For inference_perf specs, spec.Clients was empty at applyTimeoutToSpec time
 		// and populated inside GenerateWorkload — deadlines need correction here.
+		// In lazy mode wl.Requests is nil (no-op for the request loop); session
+		// blueprint Timeout pointers still need refresh.
 		if workloadSpecPath == "" || cmd.Flags().Changed("timeout") {
 			applyTimeoutToRequests(wl, requestTimeoutSecs)
 		}
@@ -1468,7 +1498,13 @@ var runCmd = &cobra.Command{
 			if wl.FollowUpBudget >= 0 {
 				sessionMgr.SetFollowUpBudget(wl.FollowUpBudget)
 			}
-			logrus.Infof("Generated %d requests + %d session blueprints (closed-loop)", len(wl.Requests), len(wl.Sessions))
+			if lazyRequestSource != nil {
+				logrus.Infof("Generated streaming source + %d session blueprints (closed-loop, lazy)", len(wl.Sessions))
+			} else {
+				logrus.Infof("Generated %d requests + %d session blueprints (closed-loop)", len(wl.Requests), len(wl.Sessions))
+			}
+		} else if lazyRequestSource != nil {
+			logrus.Infof("Generated streaming workload source (lazy, #1441)")
 		} else {
 			logrus.Infof("Generated %d requests via unified workload pipeline", len(wl.Requests))
 		}
@@ -1770,7 +1806,16 @@ var runCmd = &cobra.Command{
 				return followUps
 			}
 		}
-		cs := cluster.NewClusterSimulator(config, cluster.NewSliceRequestSource(preGeneratedRequests), onRequestDone)
+		// RequestSource: streaming in lazy mode, eager-slice otherwise.
+		// The workload package's lazy source satisfies cluster.RequestSource
+		// via structural typing — both define the same Next() method.
+		var clusterRequestSource cluster.RequestSource
+		if lazyRequestSource != nil {
+			clusterRequestSource = lazyRequestSource
+		} else {
+			clusterRequestSource = cluster.NewSliceRequestSource(preGeneratedRequests)
+		}
+		cs := cluster.NewClusterSimulator(config, clusterRequestSource, onRequestDone)
 
 		// Arrival hook: capture trace-emission references at the cluster's
 		// single arrival boundary so the trace exporter no longer relies on
@@ -1790,8 +1835,15 @@ var runCmd = &cobra.Command{
 		// A nil traceArrivals at the export site below with traceOutput
 		// non-empty means the install branch was dropped — we fail loudly
 		// rather than write a silent empty trace (R1).
+		// The arrival hook captures fresh-arrival references at the single
+		// cluster boundary. It powers trace export (#1440) and, in lazy mode
+		// where preGeneratedRequests is nil, also feeds saturation analysis.
+		// In eager mode without --trace-output, the hook stays uninstalled
+		// (BC-1 zero overhead) and saturation falls back to the
+		// preGeneratedRequests + followUpRequests path.
 		var traceArrivals []*sim.Request
-		if traceOutput != "" {
+		arrivalHookNeeded := traceOutput != "" || (lazyRequestSource != nil && saturationReport != "")
+		if arrivalHookNeeded {
 			traceArrivals = make([]*sim.Request, 0)
 			cs.SetArrivalHook(func(req *sim.Request) {
 				traceArrivals = append(traceArrivals, req)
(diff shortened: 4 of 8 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## BLIS PR Self-Review — PR #1453 (`--lazy-generation`)

Independent review focused on the two questions raised: **(1) is the existing eager generation path affected?** and **(2) is run/replay/observe/calibrate parity maintained?** Verified with `go build ./...`, `go test ./sim/workload/... ./cmd/...` (green), and **empirical byte-identity comparison against `main`**.

### Verdict: READY TO MERGE (alpha) — 1 Important item + 1 doc correction recommended

Default-off, fallback-guarded, and byte-identity-verified. No eager regression found. The items below are worth addressing but none block an alpha behind a default-off flag.

---

### 1. Eager path — NOT regressed (empirically verified vs `main`)

Built `main` and branch binaries and diffed **eager** stdout (flag off):

| Scenario | eager `main` vs branch |
|---|---|
| distribution (50 reqs, seed 42) | **IDENTICAL** |
| inference-perf spec (seed 7) | **IDENTICAL** |
| inference-perf + `--timeout 120` | **IDENTICAL** |
| multi-turn reasoning spec (seed 3) | **IDENTICAL** |

The `validateAndExpandSpec` extraction is behavior-preserving.

**Latent coupling (Minor, non-blocking):** `cmd/root.go` now calls `workload.ExpandClientsAndCohorts(spec)` **unconditionally** for every `blis run` (not just lazy). This clears the `InferencePerf`/`ServeGenData` markers *before* `GenerateWorkload`, which re-enables the mixed-`slo_class` client check at `sim/workload/spec.go:374` for expanded inference-perf clients. It is safe **today** only because `inference_perf.go` hardcodes `SLOClass:"standard"` (uniform → check can't reject) and the fix is idempotent. If inference-perf expansion ever emitted empty/mixed `slo_class`, eager specs that previously validated would start failing. Recommend either scoping the pre-expansion to `if lazyGeneration`, or adding a comment + regression guard noting this eager-path side effect.

**Test gap:** no automated test pins eager-on-branch == eager-on-main (verified manually here, but unprotected against future refactors).

### 2. Cross-path parity

| Path | Status | Notes |
|---|---|---|
| `blis run` | ✅ | Lazy source byte-identical to eager (stdout + trace YAML + CSV verified at seed 42). CustomSamplerFactory/inference-perf path also byte-identical. |
| `blis replay` | ✅ | No-op flag correct; trace export byte-identical → replay sees identical input (INV-13 preserved). |
| `blis calibrate` | ✅ (N/A) | Never invokes the generator; consumes trace + sim-results JSON. Traces are byte-identical, so parity is trivially maintained. |
| `blis observe` | ⚠️ | **Justification is inaccurate.** The PR states observe is "already streaming-shaped," but `cmd/observe_cmd.go:386` calls `workload.GenerateWorkload(...)` eagerly and then mutates each `wl.Requests[i]` (streaming flags, think-time). The #1438 memory spike applies equally to observe. Out-of-scope for this PR is fine (PR 3 of 5), but please correct the rationale and file a follow-up for an observe lazy migration. |

Saturation-report + lazy was **empirically verified byte-identical to eager** (`NumEntered:50` confirms the arrival-hook aliasing delivers all arrivals), and the multi-session-reasoning fallback fires the documented warning correctly.

### 3. Important — sampler-error asymmetry (lazy vs eager)

`sim/workload/stream.go:143` and `:277`: when `GenerateMultimodalTokens` / `GenerateReasoningRequests` fails (reachable — `spec.Validate()` does not validate these distribution fields), lazy logs `Errorf` + `exhausted=true` + returns, so the run **continues with reduced traffic and exits 0**. Eager returns the error → `logrus.Fatalf` (aborts). Same invalid spec → divergent outcome; the partial run yields misleading capacity numbers. Recommend propagating the error (e.g. an error field on the source checked after `cs.Run()`, or `Next() (…, error)`) so cmd can `Fatalf` and match eager. If kept as-is for alpha, document it explicitly as an accepted limitation.

### 4. Minor
- **Duplicate error logging:** the blueprint pre-pass and Phase 3 both run `produceNext`, so a sampler error logs twice (`stream.go` pre-pass + Phase 3). Consider a `dryRun` guard.
- **Test comment (still open from prior round):** `sim/workload/stream_test.go` `TestLazyRequestSource_PoppedNotYielded_StopsAtPoppedCap` comment says "NOT SingleSession" but the spec sets `SingleSession:true`.
- **Untested-but-verified paths:** CustomSamplerFactory sub-RNG (verified via inference-perf byte-identity here), non-time-varying lifecycle windows, and cohort-expanded lazy specs lack dedicated lazy tests. Saturation+lazy lacks a regression test (behavior verified manually).

---

**Summary:** Eager generation is unaffected (byte-identical to `main` across four workload shapes). run/replay/calibrate parity is sound; observe is correctly out of scope but its stated rationale is wrong. Recommend addressing the sampler-error asymmetry (Important) and correcting the observe justification before merge; remaining items are Minor follow-ups.

<sub>🤖 Independent review via `/blis-pr-review`. Read-only — no code modified.</sub>

namasl:
@claude /blis-pr-review

susiejojo:
## BLIS PR Self-Review — Verdict: READY TO MERGE

Reviewed against issue #1441 and tracking parent #1438, at head `fddee22d`. All CI green (build, lint, 6 test shards). This verdict is based on independent verification of the determinism-critical core, not just convergence with prior rounds.

### Correctness & invariants (INV-6 / INV-13)
The byte-identity guarantee between the lazy and eager paths holds. The three things that would silently break it all line up with the eager generator:
- Heap ordering key `(arrivalUs, clientIdx, perClientSeq)` (`stream.go:344`) matches eager's `sort.SliceStable` tie-break.
- Per-request RNG draw order matches per client kind — multimodal-vs-standard ordering (`stream.go:143–170`) and the `CustomSamplerFactory` sub-seed draw (`stream.go:678`).
- `clientSeed` is drawn **after** the zero-rate skip (`stream.go:544–548`), mirroring eager's `if clientRate <= 0 { continue }`.
- The `maxRequests`-aware blueprint pre-pass (`enumerateSurvivingSessionsPerClient`) correctly avoids consuming a phantom `blueprintRNG.Int63()` for sessions dropped by a binding cap (`stream.go:572–577`) — a subtle INV-6/INV-13 landmine that was caught and fixed in self-review.
- Determinism rests on `rand.New(rand.NewSource(s))` being a pure function of seed; the clone-and-discard pre-pass + fresh re-seed in Phase 3 is therefore byte-identical to the streamed pass.

### Cross-path parity (run / replay / observe)
Disciplined and explicit:
- **run**: implements `--lazy-generation`.
- **replay**: documented no-op (BC-9) — trace-driven, never invokes the generator; flag binds to a throwaway local (`replay.go`).
- **observe**: real gap, correctly acknowledged (observe eagerly calls `GenerateWorkload` at `observe_cmd.go:386` and the #1438 spike applies equally) and tracked in follow-up issue **#1455**.

### Preemption safety
N/A — this PR touches request **generation**, not per-request metrics, state transitions, or aggregate counters. No preemption surface.

### Timeout / deadline consistency
Preserved. The lazy-scoped `ExpandClientsAndCohorts` pre-expansion (`root.go`) fixes the inference-perf `--timeout` divergence so deadline = arrival + user timeout (not the 300s default), regression-tested by `TestGenerateWorkloadLazy_InferencePerf_TimeoutAppliedAfterPreExpand`. Scoping to lazy mode correctly avoids clearing the `InferencePerf` marker before eager's `Validate` (mixed-slo_class check).

### Error handling (R1)
`lazyRequestSource.Err()` (`stream.go:393`) + the post-Run `Fatalf` (`root.go:175`) make lazy match eager's abort-on-invalid-spec instead of silently exiting 0 with reduced traffic. Library code returns sentinel errors / values (no `logrus.Fatalf` in `sim/` — R6). `dryRun` prevents double-logging across the pre-pass and streaming pass.

### Separation of concerns / API boundaries
`lazyRequestSource` satisfies `cluster.RequestSource` via structural typing (no `sim/workload → sim/cluster` import). The `generator.go` change is a clean extract-method refactor (`validateAndExpandSpec` + idempotent `ExpandClientsAndCohorts`) with the marker-clearing caveat documented.

### Tests & docs
Behavioral coverage is strong: byte-identity vs eager, tight-`maxRequests` blueprint parity, sampler-error surfacing via `Err()`, multimodal parity, fallback gates (concurrency / multi-session / time-varying), empty-horizon, negative-`maxRequests`. Docs (CLAUDE.md + flag help) accurately scope the alpha and its fallbacks.

---

### Non-blocking finding (suggested fast-follow)
The blueprint pre-pass at `stream.go:578` runs **unconditionally**, even for specs with zero closed-loop reasoning clients. For a pure single-shot workload (the primary #1441 target) this dry-generates the full request stream once and then again for real — peak memory stays O(clients) (headline claim holds), but generation **CPU/allocation ~doubles**, and `survivingPerClient` is never read when there are no closed-loop clients (only consulted inside the `isClosedLoop` branch at `stream.go:585–588`). Guarding the pre-pass behind a `hasClosedLoopReasoning(preps)` check would be a pure win (byte-identical output, half the CPU for the common case). Non-blocking given the feature is **alpha, off by default, and memory-focused** — suggest a follow-up issue.

### Verdict
**READY TO MERGE.** No correctness, parity, or standards blockers. The single finding above is a performance optimization suitable for a fast-follow, not a merge gate.

susiejojo (approved):
Approving per BLIS self-review: READY TO MERGE. INV-6/INV-13 byte-identity verified, cross-path parity disciplined (observe deferred to #1455), timeout consistency preserved, error surfacing matches eager. One non-blocking perf follow-up noted (unconditional blueprint pre-pass ~doubles single-shot generation CPU).

PR #1454, opened 2026-06-26
refactor: compact token IDs ([]int → []TokenID/int32)

### LINKED ISSUE #1444: B — Compact token IDs: `[]int` → `[]TokenID` (`int32`) (#1438)
## B — Compact token IDs: `[]int` → `[]TokenID` (`int32`)

**Tracker:** #1438 · **Change B (sole PR)**

Halve the size of every token slice in the simulator by switching from `[]int` (8 bytes/element on amd64) to `[]int32` (4 bytes/element). Vocabulary sizes never approach 2³¹, so 32 bits is plenty.

This is independent of Change A and Change C. It can land in any order relative to them — there is no ordering coupling. It is filed as a single PR because Go does not allow `[]int` and `[]int32` to mix across the same call graph; the conversion has to happen in one atomic change. The diff is large (file count) but mechanical (compiler-driven).

### Why this is one PR

Splitting B by subsystem would require `type TokenID = int32` as a true type alias (interchangeable with `int32`) so partial conversions compile. That defeats the type-safety angle (the compiler stops flagging mixed arithmetic) and converting `[]int` storage still has to happen atomically per slice. The simpler and safer approach is one well-scoped PR with the change made compiler-driven: define the type, change the struct, fix every error the compiler flags, run tests.

### Scope (what this PR does)

1. **Define `TokenID = int32`** as a defined type (NOT a type alias) in a small new file or near `Request`:
   ```go
   // TokenID is the simulator's compact representation of a tokenizer ID.
   // 32 bits is sufficient for all real LLM vocabularies; using a defined
   // type instead of int makes mixing call sites a compile error rather
   // than a silent bug.
   type TokenID int32
   ```
   Place near [`sim/request.go#L30`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/request.go#L30).

2. **Switch the two `Request` token slices** at [`sim/request.go#L33-L34`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/request.go#L33-L34) to `[]TokenID`.

3. **Update `GenerateRandomTokenIDs`** at [`sim/request.go#L120`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/request.go#L120) to return `[]TokenID`.

4. **Fix the compiler-flagged sites.** The largest concentrations are:
   - `sim/workload/generator.go` — [L318, L323, L324, L603, L604, L1030, L1031](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/generator.go#L318)
   - `sim/workload/reasoning.go` — [L49, L50](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/reasoning.go#L49)
   - `sim/workload/replay.go` — [L66, L74, L83, L146, L227, L233, L287, L293](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/sim/workload/replay.go#L66)
   - `sim/workload/multimodal.go`, `sim/workload/session.go`, `sim/workload/client.go`
   - `sim/batch_formation.go`, `sim/kv_store.go` (KV-cache hashing reads token IDs)
   - Trace export/import (`workload.TraceRecord` if it carries token slices)

   `grep -rn 'InputTokens\|OutputTokens\|GenerateRandomTokenIDs\|\[\]int\b' sim/ cmd/` is a good first pass to enumerate sites.

5. **Trace format compatibility.** Check whether the on-disk TraceV2 format encodes token IDs (most likely as integers in YAML or numeric CSV cells). If so, ensure decoding into `[]TokenID` works without precision loss. Add a round-trip test if not already present.

6. **Construction sites for `sim.Request`.** Per project rule R4 (canonical constructors), grep all construction sites before this PR (`grep -rn 'sim.Request{' sim/ cmd/`) and verify each compiles after the type swap. Test fixtures that hand-build requests with `[]int{1, 2, 3}` will need `[]TokenID{1, 2, 3}`.

### Out of scope

- KV-cache hash function changes (the input changes type but the hash is byte-equivalent).
- Performance optimizations beyond the size reduction. No vectorization, no SIMD.

### Acceptance criteria

- `go test ./...` passes.
- Determinism: byte-identical stdout and TraceV2 output for a fixed-seed `blis run` and `blis observe` (INV-6 / INV-13).
- A focused test that constructs a Request with a token slice near `int32`'s positive max and exercises the KV hash path; it must produce the same hash as `main` would for a value within both `int` and `int32` range.
- PR description includes peak-RSS measurements showing the steady-state in-flight memory drop (per #1438's expected ~2× factor on token slices).

### Files expected to change

Many, but each change is mechanical (one type swap). Realistic estimate: 15–25 files. Reviewers should expect the diff to be wide but shallow — one `[]int` → `[]TokenID` per location plus the occasional `int(x)` conversion at integer-arithmetic boundaries.

### Invariants at risk

- **None directly.** The change is representation-only. INV-6 (Determinism) and INV-13 (Run/replay parity) must be verified end-to-end, but no behavioral logic moves.

### Notes for the implementer

- Use a defined type, not an alias. The compile-time check is the whole point.
- If you need to interoperate with code that uses `int` for index arithmetic (e.g., `req.ProgressIndex int`), explicit `int(tok)` / `TokenID(i)` casts at the boundary are correct. Don't change `ProgressIndex`'s type — it's an index, not a token ID.
- Token IDs are unsigned in spirit but the language convention is signed; `int32` is correct (matches Hugging Face, vLLM token-ID conventions).

### Labels

`hardening`, `performance`

namasl:
## Summary

Halves the size of every token slice in the simulator by switching `Request.InputTokens` / `Request.OutputTokens` from `[]int` (8 B/elem on amd64) to `[]TokenID` where `TokenID` is a defined `int32` type. Real LLM vocabularies fit well below 2³¹ (`MaxTokenID = 128000`), so 32 bits is sufficient. Using a *defined* type (not a Go alias) means mixing call sites is a compile error rather than a silent bug — the compiler becomes the audit.

Representation-only change. No behavioral logic moves.

## Behavioral Contracts

- **BC-1:** `TokenID` is a defined type, not a transparent alias of `int32` (`int → TokenID` requires an explicit cast).
- **BC-2:** `Request.InputTokens` and `Request.OutputTokens` are typed `[]TokenID`.
- **BC-3:** `GenerateRandomTokenIDs` returns `[]TokenID`.
- **BC-4:** `hash.HashBlock` digest is byte-identical to the pre-PR `[]int` digest for any value that fits in `int32`. The implementation widens to `int64` via `strconv.AppendInt(buf, int64(t), 10)` before stringification, so `int` vs `int32` with the same numeric value produce identical byte streams.
- **BC-5:** Determinism preserved (INV-6) — verified by diffing two `blis run` invocations with the same seed.
- **BC-6:** Run/replay parity preserved (INV-13) — TraceV2 schema unchanged (it persists token *counts*, not IDs).

## Architecture

`TokenID` lives in a tiny leaf package `sim/internal/tokenid` (no dependencies). Both `sim/` and `sim/internal/hash/` import it, avoiding an import cycle through `sim`. `sim/` re-exports as `type TokenID = tokenid.TokenID` — the Go alias at the re-export boundary preserves the defined-type property, because `tokenid.TokenID` itself is a defined type (not an alias of `int32`). The compile-time check on mixing remains intact.

## Out of scope

- On-disk TraceV2 format is unaffected. It persists token *counts* (`InputTokens int` / `OutputTokens int` in `sim/workload/tracev2.go`), and token IDs are regenerated lazily during replay via `GenerateRandomTokenIDs`. No schema change.
- KV-cache hash function changes only in *input slice type*; the wire format and hex digest are byte-identical for any in-`int32`-range value (BC-4).
- Performance optimizations beyond the size reduction. No vectorization, no SIMD.

## Closes

Fixes #1444. (Change B of tracker #1438.)

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./... -count=1` — all packages pass
- [x] New `sim/request_tokenid_test.go`: BC-1, BC-2, BC-3 + a regression guard against accidental narrowing below `int32` range
- [x] New `TestHashBlock_ByteEquivalentToInt64String` in `sim/internal/hash/hash_test.go`: BC-4 — reconstructs the wire format independently from the algorithm spec (not a golden value)
- [x] Determinism (INV-6) — verified: two `blis run` invocations with identical seed + flags produce byte-identical stdout
- [x] All existing run/replay parity tests pass (INV-13)

### DIFF
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -1087,9 +1087,9 @@ func adaptForSessionManager(original *sim.Request, record *RequestRecord) *sim.R
 	adapted.ProgressIndex = int64(len(original.InputTokens) + outputCount)
 
 	if outputCount > 0 {
-		adapted.OutputTokens = make([]int, outputCount)
+		adapted.OutputTokens = make([]sim.TokenID, outputCount)
 		for i := range adapted.OutputTokens {
-			adapted.OutputTokens[i] = i + 1
+			adapted.OutputTokens[i] = sim.TokenID(i + 1)
 		}
 	}
 
@@ -1099,14 +1099,14 @@ func adaptForSessionManager(original *sim.Request, record *RequestRecord) *sim.R
 // tokensToPrompt converts token IDs into a diverse prompt string using
 // prefixVocabulary. Each token ID selects a vocabulary word via modular
 // indexing, ensuring different token arrays produce different prompts.
-func tokensToPrompt(tokens []int, wordCount int) string {
+func tokensToPrompt(tokens []sim.TokenID, wordCount int) string {
 	vocabLen := len(prefixVocabulary)
 	var b strings.Builder
 	b.Grow(wordCount * 8) // average word ~7 chars + space
 	for i := 0; i < wordCount; i++ {
 		var idx int
 		if i < len(tokens) {
-			idx = tokens[i]
+			idx = int(tokens[i])
 		} else {
 			idx = i
 		}
--- a/cmd/observe_cmd_test.go
+++ b/cmd/observe_cmd_test.go
@@ -136,8 +136,8 @@ func TestObserveOrchestrator_OpenLoop_ConservationAndConcurrency(t *testing.T) {
 		requests[i] = &sim.Request{
 			ID:           fmt.Sprintf("request_%d", i),
 			ArrivalTime:  int64(i) * 10000, // 10ms apart in microseconds
-			InputTokens:  make([]int, 100),
-			OutputTokens: make([]int, 50),
+			InputTokens:  make([]sim.TokenID, 100),
+			OutputTokens: make([]sim.TokenID, 50),
 			MaxOutputLen: 50,
 			State:        sim.StateQueued,
 		}
@@ -307,7 +307,7 @@ func TestObserveOrchestrator_WarmupExclusion(t *testing.T) {
 	for i := range requests {
 		requests[i] = &sim.Request{
 			ID: fmt.Sprintf("request_%d", i), ArrivalTime: int64(i) * 1000,
-			InputTokens: make([]int, 10), OutputTokens: make([]int, 5),
+			InputTokens: make([]sim.TokenID, 10), OutputTokens: make([]sim.TokenID, 5),
 			MaxOutputLen: 5, State: sim.StateQueued,
 		}
 	}
@@ -335,7 +335,7 @@ func TestObserveOrchestrator_WarmupExceedsTotal(t *testing.T) {
 	for i := range requests {
 		requests[i] = &sim.Request{
 			ID: fmt.Sprintf("request_%d", i), ArrivalTime: 0,
-			InputTokens: make([]int, 10), OutputTokens: make([]int, 5),
+			InputTokens: make([]sim.TokenID, 10), OutputTokens: make([]sim.TokenID, 5),
 			MaxOutputLen: 5, State: sim.StateQueued,
 		}
 	}
@@ -374,7 +374,7 @@ func TestObserveOrchestrator_RecordITL_CapturesChunkTimestamps(t *testing.T) {
 	requests := []*sim.Request{
 		{
 			ID: "request_0", ArrivalTime: 0,
-			InputTokens: make([]int, 10), OutputTokens: make([]int, 3),
+			InputTokens: make([]sim.TokenID, 10), OutputTokens: make([]sim.TokenID, 3),
 			MaxOutputLen: 3, State: sim.StateQueued, Streaming: true,
 		},
 	}
@@ -416,7 +416,7 @@ func TestObserveOrchestrator_TimestampOrdering(t *testing.T) {
 
 	requests := []*sim.Request{{
 		ID: "request_0", ArrivalTime: 0,
-		InputTokens: make([]int, 10), OutputTokens: make([]int, 5),
+		InputTokens: make([]sim.TokenID, 10), OutputTokens: make([]sim.TokenID, 5),
 		MaxOutputLen: 5, State: sim.StateQueued,
 	}}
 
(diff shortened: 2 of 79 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
Nice, clean representation-only change — built + vetted locally, the type swap is mechanical across all the call sites, and BC-4's hash byte-equivalence is proven by a genuine independent reconstruction rather than a golden tautology. The defined-type-over-alias choice is the right call: it turns every mixing site into a compile error. A few forward-looking thoughts, none blocking:

**1. `MaxTokenID` ↔ `int32` coupling is unguarded (latent footgun).**
In `GenerateRandomTokenIDs`, `TokenID(rng.Intn(MaxTokenID))` casts an `int` down to `int32`. `MaxTokenID = 128000` today, but if anyone later raises it above `math.MaxInt32` (~2.1B), this cast **silently truncates** — no compile error, and `TestTokenID_HoldsValuesNearInt32Max` only proves `TokenID` *can hold* `MaxInt32`, not that `MaxTokenID` stays within it. Probability is low (no real vocab is that large), but a near-free static assertion would close it and matches this PR's own "compiler-becomes-the-audit" philosophy:
```go
const _ = uint(math.MaxInt32 - MaxTokenID) // fails to compile if MaxTokenID > MaxInt32
```

**2. Merge sequencing vs the delta-encoded-rounds work (#1445).**
That branch's shared session buffer (`sim/workload/session.go`) still uses `[]int` for the buffer, `Prefix`, and the `req.InputTokens` aliasing (with its continuity panic-guard). Once this lands, that code won't compile until those are migrated to `[]TokenID`. That's actually the defined type working as intended — the breakage is *loud, not silent* — but it'd be smoothest to land this first and absorb the type flip during the #1445 rebase. Just flagging for merge order.

**3. Tiny cosmetic nit (no action needed).** The `var buf [20]byte // ... max int64 (19 digits) + pipe` comment in `hash.go` is now oversized for the narrowed `int32` values. Harmless, just slightly stale.

Confirmed safe: no external token-ID ingestion (all serialized fields are token *counts*, never the ID slices), determinism (INV-6) holds since RNG draw counts and hash bytes are unchanged, and run/replay parity (INV-13) is untouched since TraceV2 persists counts and regenerates IDs lazily.

LGTM overall — the only thing I'd genuinely suggest adding is the `MaxTokenID <= MaxInt32` guard from (1).

namasl:
@claude /blis-pr-review

susiejojo (approved):
lgtm

Issue #1455, opened 2026-07-07
Follow-up: migrate blis observe to lazy request generation


namasl:
## Context

PR #1453 (issue #1441, Change A of #1438) landed lazy request generation for \`blis run\` behind \`--lazy-generation\` (alpha, default off). During self-review it was flagged that the PR body incorrectly claimed \`blis observe\` was "already streaming-shaped" — that is not accurate.

\`cmd/observe_cmd.go:386\` calls \`workload.GenerateWorkload(...)\` eagerly and then iterates \`wl.Requests[i]\` to mutate per-request fields (streaming flags, think-time). The #1438 memory spike applies equally to \`observe\` — a large multi-turn cohort workload with a long prefix will materialize the same O(total_requests) slice under \`observe\` that it did under \`run\`.

## What this issue tracks

Extend the lazy generation path to \`blis observe\`. Concretely:

1. Refactor the per-request field mutations in \`cmd/observe_cmd.go\` (\`Streaming\`, \`ThinkTime\`, any late-stage overrides) into a hook that runs at request-consumption time rather than requiring the full slice up front.
2. Add \`--lazy-generation\` to \`observe\` (mirroring \`run\`) and wire it into the request-source construction so that with the flag on, \`observe\`'s HTTP dispatch pump pulls from \`workload.GenerateWorkloadLazy\` instead of a materialized slice.
3. Preserve INV-6 / INV-13 byte-identity for observed traces under the same seed with the flag on vs off, mirroring the run-side property tests in \`sim/workload/stream_test.go\`.
4. Handle the same three unsupported fallback cases (\`ErrLazyUnsupportedTimeVarying\`, \`ErrLazyUnsupportedConcurrency\`, \`ErrLazyUnsupportedMultiSession\`) with a one-line warning and eager fallback.

## Out of scope

- Multi-session reasoning support (SingleSession=false) — deferred separately.
- Concurrency-client streaming — deferred separately.
- Time-varying (per-window parameter) streaming — deferred separately.

## Verification

- \`go test ./cmd/... ./sim/workload/...\` green.
- \`golangci-lint run ./...\` clean.
- Peak-RSS smoke on the same #1438-shape workload used for the run-side comparison (2 clients × 10K-prefix × 200 requests) shows the same ~50%+ reduction under \`observe --lazy-generation\` as we saw under \`run --lazy-generation\`.

## Related

- Issue: #1438 (tracker, all Change A parts)
[...]

PR #1456, opened 2026-07-08
test: property + run/replay parity suite for eager ≡ lazy (#1442)

### LINKED ISSUE #1442: A4 — Property + parity test suite for eager ≡ lazy (and run/replay) (#1438)
## A4 — Property + parity test suite for eager ≡ lazy (and run/replay)

**Tracker:** #1438 · **Change A — PR 4 of 5** · **Depends on:** A3

Lazy generation ships as an alpha flag in A3 (`--lazy-generation`, default off). The central correctness claim is that for any seed and workload spec, eager and lazy modes produce byte-identical request streams — and consequently byte-identical stdout and trace output. This claim is what makes the alpha flag safe and what we need to be confident in before graduating lazy mode to default (A6) and eventually removing eager (A8).

This PR exists to put that claim under continuous test pressure. It does three things:

1. **Adds a Go property test** that randomly samples seeds and workload specs and asserts eager ≡ lazy as request streams. This is the maintainer's explicit ask: *"perhaps we can use go property testing to really guarantee this."*
2. **Hardens INV-13 (run/replay parity)** under both modes. Before A3, parity was structural by construction (both code paths consumed the same eagerly-built list). With lazy mode in play, parity is enforced only by tests.
3. **Codifies a parity matrix** as unit + E2E tests so a regression in either mode trips a clear, well-named assertion.

### Why this is its own PR

A3 is the rewrite. Reviewers of A3 should focus on the lazy generator. The parity-test scaffolding is independent work — it tests the boundary, not the implementation — and lives more cleanly with full attention.

#### 1. Property test: eager ≡ lazy request streams

Using `testing/quick` (or `gopter` if a more structured generator API is desired), write a property test that:

- Samples a seed (uniform `int64`).
- Samples a workload spec from a constrained but representative space: number of cohort members (1–10), max rounds (1–20), arrival process (poisson/gamma/weibull/constant), num_requests (10–500), with/without shared prefix, with/without accumulating context.
- Runs `workload.GenerateWorkload(spec, seed)` in eager mode and lazy mode.
- Asserts the two resulting request streams are equal as ordered sequences: same length; same `Request.ID`, `ArrivalTime`, `InputTokens` (length + content), `OutputTokens` (length + content), `SessionID`, `Round`, and any other `Request` fields that are deterministic functions of `(seed, spec)`.
- Runs ≥ 100 random draws per test invocation (tunable via env var for nightly long runs).

When a draw fails, the test must surface the seed + spec in the failure message so the case can be reproduced and added as a regression unit test.

#### 2. INV-13 run/replay parity, under both modes

For a fixed seed and a representative workload set, assert that:
- `blis run --trace-output X --seed N --model M` (eager mode) produces a trace that replays to byte-identical per-request metrics.
- `blis run --lazy-generation --trace-output X --seed N --model M` produces a trace that replays to byte-identical per-request metrics.
- The two run modes produce byte-identical traces (i.e., the trace file itself is the same regardless of generation mode).

Per-request fields to compare: TTFT, ITL, E2E, completion time, output token count. This is exactly what INV-13 promises, per [`docs/contributing/standards/invariants.md`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/docs/contributing/standards/invariants.md).

#### 3. Coverage matrix

At minimum, run the parity check on:
- A simple chatbot workload (single-turn, Poisson arrivals).
- A multi-turn code-generation workload (the reproducer shape from #1438 — accumulating context, large prefix, cohort-based). This is the case where lazy mode's memory win matters most, so it's the one most worth proving correct.
- One closed-loop session workload (since session blueprints flow through `onRequestDone` and follow-up routing — this is the trickiest path).

These can be small (e.g. `num_requests=50`) so the test runs fast in CI.

#### 4. Determinism harness

A test that runs `blis run --lazy-generation --seed K` twice and asserts byte-identical stdout (INV-6). The eager-mode equivalent should already exist; if not, add it. INV-6 is now load-bearing for A3.

#### 5. Negative test

If a configuration is unsupported by replay (e.g., autoscaler — flagged by INV-13's "logrus.Fatalf at startup" rule), assert that `blis replay` exits with a clear error, not silent degradation. The time-varying fallback added in A3 (lazy → eager when per-window parameters are present) should also have a test asserting the warning fires.

### Out of scope

- New functionality. This PR adds tests only.
- Migrating `observe`-side code to lazy mode (A5).
- Removing eager mode (A8).

### Acceptance criteria

- `go test ./...` passes including the new tests.
- The property test runs ≥ 100 draws per invocation and completes in well under a minute. A nightly/extended mode (env-gated) can run 10× more.
- Each new parity test fails loudly (clear assertion message naming the offending field, plus the seed + spec on property-test failures) if a future PR breaks eager ≡ lazy or INV-13.
- Documentation: add a short note in [`docs/contributing/standards/invariants.md`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/docs/contributing/standards/invariants.md) under INV-13 referencing the new test file as the authoritative enforcement point under lazy mode, and a sibling invariant entry (or a paragraph under INV-6) covering "eager ≡ lazy" as a transient invariant that holds until eager is removed in A8.

### Files expected to change

- New test files:
  - `sim/workload/parity_property_test.go` (or similar) — the eager-vs-lazy property test
  - `cmd/parity_run_replay_test.go` (or extend `cmd/root_test.go`) — the run/replay matrix
- Possibly `docs/contributing/standards/invariants.md` (one paragraph + one invariant addendum)
- Test fixtures (small YAML workloads) — kept inline in tests if reasonable

That's 1-3 files, almost all test code.

### Invariants at risk

None — this PR adds invariant *tests*. It cannot regress any invariant; if it fails to assert what it claims, the worst case is "we still don't have parity coverage," not "we silently broke something."

### Notes for the implementer

- Use `t.Parallel()` where appropriate; these tests are I/O-light.
- For closed-loop / session coverage, the smallest setup that exercises follow-up requests is fine (e.g., 5 sessions × 3 rounds).
- The property test should use a small but non-trivial workload-spec generator — large enough to exercise multi-turn + cohorts, small enough that 100 draws complete quickly. If a draw produces a spec that hits the lazy-mode time-varying fallback (added in A3), the test should either filter it out or assert that both modes are eager and still byte-identical.
- If a metric difference is unavoidable for a *legitimate* reason (e.g., wall-clock fields), the test should mask only those specific fields with a clearly-commented "deliberately ignored" — never blanket tolerance.

### Labels

`test`, `hardening`

namasl:
## Summary

Adds a **property test** plus a **run/replay parity matrix** that put the "eager ≡ lazy" byte-identity claim (and INV-13 run/replay parity under both generation modes) under continuous, automated test pressure. This is the safety net that makes the alpha `--lazy-generation` flag (A3, #1441) trustworthy before it graduates to default (A6) and eager is removed (A8).

**Tests + docs only — no production code changes.** It cannot regress any invariant; the worst case if it under-asserts is "we still lack coverage," never "we silently broke something." This is Change A, PR 4 of 5 (tracker #1438); A3 (#1441) is already merged.

The maintainer's explicit ask — *"perhaps we can use go property testing to really guarantee this"* — is realized by `TestProperty_EagerEqualsLazy_RequestStreams`.

## What's added

- **`sim/workload/parity_property_test.go`** — a randomized property test. A constrained spec generator samples only **lazy-supported** shapes (no per-window params, no concurrency, reasoning always `SingleSession`), so the eager-vs-lazy comparison never degenerates to eager-vs-eager. Each draw builds two independent specs, runs `GenerateWorkload` (eager) and `GenerateWorkloadLazy` (lazy), and asserts the request streams + session blueprints are byte-identical (reusing `assertRequestStreamsEqual`/`drainLazy` from `stream_test.go`). ≥100 draws by default; `BLIS_PROPERTY_DRAWS` env override for nightly 10× runs. Failures print the fixed base seed + draw index + spec label for reproduction.
- **`cmd/parity_run_replay_test.go`** — the CLI-level parity matrix:
  - `TestParity_RunReplay_TraceByteIdentity_Matrix` — eager and lazy `blis run --trace-output` produce byte-identical traces across **chatbot** (single-turn Poisson), **codegen-cohort** (multi-turn accumulate cohort — the #1438 reproducer shape), and **reasoning** (single-session multi-turn) shapes; two lazy runs at the same seed also match (INV-6). Because the trace captures every fresh arrival (#1440), this proves parity of the runtime closed-loop follow-up rounds too.
  - `TestParity_RunReplay_INV13_BothModes` — eager- and lazy-sourced traces replay (via `blis replay --session-mode fixed`) to identical per-request TTFT/E2E (INV-13 under both modes).
  - `TestParity_RunStdout_DeterministicAndEagerLazyIdentical` — the literal INV-6 statement: same seed → byte-identical **stdout**, eager ≡ lazy at the stdout level.
  - `TestParity_LazyTimeVaryingFallback_MatchesEager` — a per-window (unsupported) spec run with `--lazy-generation` falls back to eager **transparently** (same trace bytes, not merely non-fatal).
- **`docs/contributing/standards/invariants.md`** — a transient "eager ≡ lazy" sub-invariant note under INV-6 and a "under lazy generation" paragraph under INV-13, naming the new tests as the authoritative enforcement points. Invariant *statements* are unchanged (evidence-only additions), so no source-of-truth working copies need syncing.

## Behavioral contracts

- **BC-1** — GIVEN a random seed + lazy-supported spec, WHEN generated eagerly and lazily, THEN the request streams and session blueprints are byte-identical.
- **BC-2** — WHEN a property draw diverges, THEN the failure names the base seed + draw index + spec shape for reproduction.
- **BC-3** — GIVEN a representative spec + fixed seed, WHEN `blis run --trace-output` runs eager and lazy, THEN the two TraceV2 files are byte-identical across the coverage matrix.
- **BC-4** — GIVEN a trace exported from a run (eager or lazy), WHEN replayed `--session-mode fixed`, THEN per-request SimResults are produced and are identical between eager-sourced and lazy-sourced replays (INV-13).
- **BC-5** — GIVEN the same seed, WHEN `blis run --lazy-generation` runs twice, THEN stdout and trace bytes are identical (INV-6).
- **BC-6** — GIVEN `--lazy-generation` + a time-varying spec, WHEN `blis run` executes, THEN it falls back to eager transparently (identical trace, no fatal).
- **BC-7** — GIVEN replay + an unsupported feature (autoscaler), WHEN `blis replay` starts, THEN it exits fatally at startup (existing `TestReplayCmd_AutoscalerFlagFatal`; not duplicated).

## Testing / verification

- `go build ./...` — passes.
- `go test ./... -count=1` — passes (all packages).
- `golangci-lint run ./...` — 0 issues.
- `BLIS_PROPERTY_DRAWS=1000 go test ./sim/workload/ -run TestProperty_EagerEqualsLazy_RequestStreams` — extended mode passes in ~2.7s.
- Each new test was mutation-verified to fail loudly on an induced divergence, then restored.

Closes #1442.


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## BLIS Self-Review — PR #1456 (`test: property + run/replay parity suite for eager ≡ lazy`, closes #1442)

**Verdict: ✅ READY TO MERGE**

Independent review against PR head `d77fd1af`. This is a **tests + docs-only** change (`+979 / -1` across 4 files; zero production code). CI fully green (build, lint, all test shards). Both findings from the prior self-review rounds are confirmed resolved on the current head.

---

### What this PR does
Adds the A4 parity suite for Change A (#1438), locking in the load-bearing correctness claim of `--lazy-generation` (#1441): **the lazy streaming generator yields a byte-identical request stream to the eager generator** for every lazy-supported shape. Two layers:
- `sim/workload/parity_property_test.go` — randomized property test (≥100 draws) at the generator boundary.
- `cmd/parity_run_replay_test.go` — CLI-level parity matrix (trace byte-identity, stdout determinism, INV-13, fallback transparency).
- `docs/contributing/standards/invariants.md` — evidence-only annotations under INV-6 and INV-13.

---

### Correctness & test quality (the core review axis for a test PR)
- **Vacuity guards are comprehensive (R20).** The property test fails loudly if no draw produced a non-empty stream, no reasoning draw occurred, or reasoning draws produced no blueprints (`parity_property_test.go:299-310`). The CLI matrix guards `lineCount(dataEager) < 2` and `len(eagerResults) == 0`. This is exactly right: a parity test that can pass on empty input proves nothing.
- **The generator is structurally incapable of emitting unsupported shapes.** `genLazySupportedSpec` never sets Lifecycle/per-window params, never sets `Concurrency`, and always sets `SingleSession: true` (`parity_property_test.go:117`). This keeps the comparison meaningful — a draw that tripped the lazy fallback would compare eager-to-eager and silently prove nothing. Good defensive design.
- **The blueprint-RNG proxy check is sound and catches a real bug class.** `parity_property_test.go:291` draws one `Int63()` from each side's blueprint RNG and asserts equality. This directly guards against the shifted-`blueprintRNG` seed bug from the tight-cap pre-pass (the #1453 bug class), not just structural fields.
- **Independent spec instances per mode.** `buildSpec` reconstructs a fresh spec from the same seed for eager and lazy (`parity_property_test.go:204-206`), removing any chance of cross-contamination via `validateAndExpandSpec` mutation.
- **Reproducibility (INV-6).** Fixed `propertyBaseSeed = 0x1442A4`; a CI failure names base seed + draw index + spec label, enough to reconstruct the offending case as a regression unit test.

### Cross-path parity (run / replay / observe)
- **run + replay: covered.** `TestParity_RunReplay_TraceByteIdentity_Matrix` proves eager≡lazy trace header+data across three shapes (chatbot single-turn, codegen-cohort w/ prefix+accumulate reasoning — the #1438 reproducer, and 2-client single-session reasoning). `TestParity_RunReplay_INV13_BothModes` replays both traces and compares TTFT/E2E per RequestID.
- **INV-13 scoping is justified.** The INV-13 test uses the chatbot shape only, but that is acceptable because trace byte-identity is *already* proven across the full matrix — if the traces are identical, replay cannot diverge. Not a gap.
- **observe: correctly N/A.** observe eagerly generates and dispatches over HTTP; it has no `--lazy-generation` flag and is scoped out of #1441, tracked in #1455. No parity obligation for this PR.

### Fallback transparency (BC-6)
`TestParity_LazyTimeVaryingFallback_MatchesEager` feeds a per-window (`lifecycle.windows`) spec under `--lazy-generation`, which trips `ErrLazyUnsupportedTimeVarying` → eager fallback, and asserts the resulting trace is byte-identical to the eager run. This pins the "silent-but-transparent fallback" contract, not just the happy path.

### Preemption / timeout consistency
N/A — tests-only PR, touches no per-request metrics, state transitions, or deadline/horizon logic.

### Standards adherence
- **R7 (invariant-test companions):** the whole PR *is* the companion law suite for #1441's golden/unit tests — asserts "eager output == lazy output," not fixed values. ✅
- **R20 (degenerate-input guards):** `propertyDraws` rejects `n <= 0` / malformed `BLIS_PROPERTY_DRAWS` and falls back to the default rather than running zero draws. ✅
- **BDD #4/#5 (test laws / refactor-survival):** assertions are observable-behavior (request fields, trace bytes, stdout), not internal structure. ✅

### CLI test-state hygiene
`replaySpecTrace` pairs `captureCmdLevelVars()`/`restore()` with an explicit manual save/restore of the replay-specific package vars including `cacheSignalDelay` and `flowControlEnabled` (`parity_run_replay_test.go:412-422`). `root_test.go` adds `defaultsFile` to the capture/restore set. **Both prior-round findings (the `defaultsFilePath` leak and the missing `cacheSignalDelay` restore) are confirmed fixed on this head.**

### Documentation
`invariants.md` changes are evidence-only: an INV-6 transient sub-invariant ("eager ≡ lazy, holds until A8") and an INV-13 "Under lazy generation" paragraph, each naming the enforcing tests. No invariant *statements* were weakened. Accurate and appropriately scoped.

---

### Non-blocking nano-nits (optional, do not block merge)
1. `parity_run_replay_test.go:360` — `os.Stdout = oldStdout` is not `defer`red, so a panic inside `runCmd.Run` would leave stdout redirected for later tests. Moot in practice because the failure mode is `logrus.Fatalf` → `os.Exit`, but a `defer` would be marginally more robust.
2. The `sawBlueprints` guard proves blueprints were exercised but doesn't *force* a tight-cap truncation draw specifically. That path is already covered deterministically by `stream_test.go`'s dedicated tight-`maxRequests` blueprint-parity test, so this is not a coverage gap — just noting the division of labor.

**Bottom line:** High-quality, non-vacuous parity suite that pins exactly the claim it should. Ready to merge.

susiejojo (approved):
lgtm

PR #1457, opened 2026-07-08
hardening(observe): plumb --lazy-generation through blis observe (#1443)

### LINKED ISSUE #1443: A5 — Plumb `--lazy-generation` through `blis observe` with a parity test (#1438)
## A5 — Plumb `--lazy-generation` through `blis observe` with a parity test

**Tracker:** #1438 · **Change A — PR 5 of 5 (follow-on)** · **Depends on:** A3 (with caveat — see below)

This is the follow-on PR called out in #1438:

> **`observe` benefits.** It is already streaming-shaped (dispatch → wait → record → next). A shared streaming "request source" abstraction between `run` and `observe` becomes natural and would simplify both.

The work is opportunistic — `observe` doesn't suffer from the same memory cliff as `run` because real-server dispatch already paces it — but the architectural payoff is worth tracking explicitly so it doesn't drift indefinitely. It also gives us a parity test on the observe side, which is what the lead maintainer asked for: *"We would want to do something similar for observe as well to the extent possible in tests."*

`blis observe` materializes requests via `wl.Requests` at [`cmd/observe_cmd.go#L391`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L391) and feeds them to `runObserveOrchestrator` at [`cmd/observe_cmd.go#L485`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L485) as a `[]*sim.Request`. After A3, both eager and lazy `RequestSource` implementations exist; observe should consume from whichever is selected by `--lazy-generation`, with eager as default.

This is a small, focused change. Splitting it from A3 keeps A3 reviewable and prevents an "everything migrates at once" PR.

### Dependency on A3 — `Peek()` caveat

A1 (#1439) introduced `RequestSource` with a single `Next()` method and deliberately deferred `Peek()`. The observe orchestrator needs **peek-style access** to merge the pre-generated stream against in-flight session follow-ups by arrival time (see §"Observe-side index-access sites" below). Two options:

- **Preferred:** A3 adds `Peek() (*sim.Request, bool)` to `RequestSource` as part of the streaming generator's natural state model. A5 then consumes it on the observe side. State explicitly in A3's scope whether the interface gains `Peek()`.
- **Fallback:** A3 does **not** add `Peek()`. A5 introduces it (one-line interface change to `sim/cluster/request_source.go` plus a trivial implementation on `SliceRequestSource` and the lazy source). In this case A5's "files expected to change" grows from 2 to 3.

The implementer of A5 should re-check A3's final shape before starting and pick the right option. Either way the orchestrator's merge logic is the only consumer.

### Scope (what this PR does)

1. **Add `--lazy-generation` to `blis observe`.** Same boolean flag, same default (false), same semantics as `run` and `replay`.

2. **Update `runObserveOrchestrator`** at [`cmd/observe_cmd.go#L485`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L485) to take a `RequestSource` (the interface from A1) instead of `[]*sim.Request`.

3. **Update the `observe` call site** to construct an eager or lazy source from `workload.GenerateWorkload`'s spec based on the flag — the same selection logic as `run` and `replay`.

4. **Preserve all observe-specific bits.** The streaming-on prep at [`cmd/observe_cmd.go#L407-L414`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L407-L414) (forcing `Streaming = true` when ITL recording is on) needs to apply per-request as the source emits, not on a pre-materialized slice. See also the two non-trivial index-access sites below.

5. **Add an observe-side parity test (to the extent possible).** A real server is not available in CI, so a true end-to-end byte-identity test isn't feasible. The achievable form is one or both of:
   - **Dispatch-order parity:** with a stubbed/recorded HTTP transport, assert that the sequence of outgoing request payloads (model, prompt tokens, max_tokens, etc.) is byte-identical between eager and lazy modes for the same seed.
   - **Trace-output parity:** observe records its own TraceV2 to disk; with a stubbed transport that returns deterministic responses, assert that the produced TraceV2 header + data is byte-identical between modes.

   The harness for the property tests in A4 can be reused — sample seeds and small workload specs, run observe in both modes with the same stub, compare outputs.

### Observe-side index-access sites the implementer must address

These are the non-trivial places where the orchestrator reaches into `requests` by index today. They don't fall out automatically from "take a `RequestSource` instead of a slice"; each needs an explicit migration decision.

1. **Active-session pre-count** at [`cmd/observe_cmd.go#L904-L909`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L904-L909) iterates the entire `requests` slice once at startup to count distinct `SessionID`s and seed `activeSessionCount`. Options:
   - **Fold counting into the dispatch loop.** Increment `activeSessionCount` the first time a previously unseen `SessionID` is emitted from the source. Simplest; preserves the lazy property; only change is moving the bookkeeping a few dozen lines later.
   - **Attach a session-count metadata field to `RequestSource`.** Cleaner separation but adds API surface; not justified for one call site.
   - **Pre-pass via a two-pass adapter.** Defeats the lazy property; do not do this.

   Recommendation: option 1 (fold into the loop).

2. **Merge-by-`ArrivalTime`** at [`cmd/observe_cmd.go#L1000-L1010`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd.go#L1000-L1010) picks the earlier of `pendingFollowUps[0]` and `requests[preGenIdx]` without consuming either. This is the canonical peek-then-take pattern. Options:
   - **Use `Peek()`** added in A3 (preferred — see "Dependency on A3 — `Peek()` caveat" above).
   - **Buffer one pre-generated request locally.** The orchestrator keeps a `var nextPreGen *sim.Request` field that is refilled lazily via `Next()` after consumption. This requires no interface change but adds local state to the dispatch loop and complicates the exhaustion check.

   Recommendation: lobby for `Peek()` in A3. If A3 ships without it, the local-buffer fallback is acceptable but slightly uglier.

### Out of scope

- The `--max-concurrency` and warmup logic. Those operate at the dispatch layer; they don't care whether the request stream is materialized or streamed.
- `replay` and `calibrate`. Replay reads from disk (the flag is already wired in A3); calibrate is post-processing.
- The eager-removal arc (A6/A7/A8).

### Acceptance criteria

- `go test ./...` passes, including `cmd/observe_cmd_test.go`.
- The two existing observe orchestrator tests at [`cmd/observe_cmd_test.go#L218`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd_test.go#L218) and [L285](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd_test.go#L285) continue to pass with adapter shims as needed (or are migrated to construct sources directly).
- **Default behavior unchanged.** Without `--lazy-generation`, observe trace output is byte-identical to today for at least one workload at a fixed seed (INV-6). **Pinned reference target:** run on the same chatbot preset and seed already used by the existing observe orchestrator test (`workload.WorkloadSpec` with `Seed: 42`, `AggregateRate: 10.0`, the single-client multi-turn shape at [`cmd/observe_cmd_test.go#L249-L270`](https://github.com/inference-sim/inference-sim/blob/d059008f7abb008d69b2f01ff08b31e4372230ea/cmd/observe_cmd_test.go#L249-L270)), under a stubbed deterministic HTTP transport, and `diff` the resulting TraceV2 header+data against a checked-in golden produced before the PR.
- **Eager ≡ lazy on observe.** At least one observe-side parity test passes (dispatch-order parity is the minimum bar; trace-output parity is preferred). Use the same seed/spec as the byte-identical-to-today test for direct cross-reference.

### Files expected to change

- `cmd/observe_cmd.go` (flag, call site, orchestrator signature, active-session counting moved into the loop, merge-by-arrival via `Peek()` or local buffer)
- `cmd/observe_cmd_test.go` (signature update + new parity test + golden trace fixture)
- (Conditional) `sim/cluster/request_source.go` — only if A3 did not already add `Peek()`. One-line interface addition + trivial `SliceRequestSource` impl + lazy-source impl.

That's 2 files in the happy case, 3 if A5 has to add `Peek()` itself.

### Invariants at risk

- **INV-6 (Determinism):** the same seed must produce the same dispatch order under either generation mode. The lazy source already provides this on the `run` side; this PR is plumbing.

### Notes for the implementer

- Re-check A3's final interface shape (does `RequestSource` have `Peek()`?) before starting. This determines whether A5 is 2 files or 3.
- Re-grep the cited line numbers against `main` at implementation time — observe_cmd.go has had drift since this issue was filed (the streaming-on prep moved from L409-L411 to L407-L414 between filing and the latest fetch; the index-access sites at L904-909 and L1000-1010 are stable as of the latest fetch).
- This PR is low risk and high-leverage: after it lands, run-side and observe-side share a generation pipeline and Change C (delta-encoded rounds) immediately benefits both.
- The recorder pipeline (`Recorder.RecordRequest`) is unaffected. Don't restructure it.
- The observe-side parity test does not need to be exhaustive — A4 already covers the generator's correctness. What this test catches is observe-specific plumbing bugs (e.g., a path that reads `req.InputTokens` before the source has produced the request, or the active-session counter missing a session because counting was folded incorrectly into the loop).

### Labels

`hardening`, `refactor`

namasl:
## Summary

Plumbs the `--lazy-generation` alpha switch (introduced for `blis run` in A3, #1453) through **`blis observe`**, completing #1438's "shared streaming request source between `run` and `observe`" promise. `runObserveOrchestrator` now consumes a `cluster.RequestSource` (the A1 interface) instead of a pre-materialized `[]*sim.Request` slice, pulling requests one at a time via `Next()`.

This is **Change A — PR 5 of 5 (A5)** of the lazy-generation arc tracked in #1438.

### What changed

- **`--lazy-generation` flag on `observe`** (default `false`), same semantics and fallback warnings as `run`. Selects an eager `SliceRequestSource` or a lazy `GenerateWorkloadLazy` source; unsupported specs (time-varying, concurrency, multi-session) fall back to eager with a one-line warning — the switch is byte-identical to `run`'s.
- **`runObserveOrchestrator` consumes `cluster.RequestSource`.** The two index-access sites the issue called out are migrated:
  1. **Merge-by-arrival** uses a **one-slot lookahead buffer** (`nextPreGen` + `takePreGen`) rather than `requests[preGenIdx]`. Since A3 shipped `RequestSource` with only `Next()` (no `Peek()`), this PR takes the issue's documented **local-buffer fallback** — no interface change, all merge complexity confined to the single consumer.
  2. **Active-session counting** is folded into the dispatch loop (counted on first sight of each `SessionID`) instead of a startup pre-pass, preserving the lazy property.
- **ITL streaming-on moved into per-request dispatch.** The former slice pre-pass only set `Streaming=true` on round-0 requests; the per-request override now also covers session **follow-up** rounds (created by `SessionManager` with `Streaming=false`), so `--record-itl` + closed-loop sessions now capture ITL on follow-ups too. (Strict improvement; pinned by a new test.)
- **Tie-break extracted** into a pure `preferFollowUp` helper so the `<=` operator is unit-testable.

### Dependency note (`Peek()` caveat)

Per #1443, A3's final `RequestSource` shipped with only `Next()`. This PR therefore uses the **fallback** (local one-slot lookahead buffer), not `Peek()` — keeping the shared interface minimal and staying at 2 source files. Documented in the plan's deviation log.

## Behavioral Contracts

- **BC-1** — GIVEN a supported spec + `--lazy-generation`, WHEN observe generates the workload, THEN requests stream one-at-a-time and match eager mode on the deterministic field surface for the same seed.
- **BC-2** — GIVEN default (eager) mode, WHEN dispatching a hand-built sequence at concurrency 1, THEN each request dispatches exactly once in arrival order with sequential ids (the lookahead reproduces the pre-PR slice-index order; ties go to the follow-up).
- **BC-3** — GIVEN the same spec/seed, WHEN dispatching once eager and once lazy, THEN recorded requests match on the deterministic `TraceRecord` surface (grouped by `(SessionID, RoundIndex)`).
- **BC-4** — GIVEN `--lazy-generation` + an unsupported spec, THEN observe warns and falls back to eager.
- **BC-5** — Session follow-ups still merge by arrival time under the lookahead buffer.
- **BC-6** — `--record-itl` enables streaming per emitted request (incl. follow-ups).
- **BC-9** — a terminal lazy-sampler failure is surfaced via `Fatalf`, not silently swallowed.

## Testing

- **New tests:** `TestPreferFollowUp_TieGoesToFollowUp` (tie-break unit), `TestObserveOrchestrator_MergeOrder_OpenLoopExactSequence` (absolute lookahead order guard), `TestObserveOrchestrator_EagerLazyParity_SameDispatchSequence` (eager ≡ lazy, deterministic surface, exercises a round-1 follow-up), `TestObserveOrchestrator_RecordITL_CoversSessionFollowUps` (follow-up ITL), `TestObserveLazyFallback_UnsupportedSpecSentinels` (all 3 sentinels), `TestObserveCmd_LazyGenerationFlag_Exists`.
- The tie-break and lookahead guards were **mutation-verified** (each fails when the operator/buffer is broken).
- All 10 existing orchestrator tests migrated to the `RequestSource` signature and still pass.
- `go build ./...`, `go test ./...` (incl. `-race` on `cmd/`), and `golangci-lint run ./...` all clean.

## Invariants

- **INV-6 (Determinism):** eager and lazy produce the same observe dispatch order for a fixed seed — now guarded by BC-2/BC-3 on the observe side.

## Notes for reviewers

- The merge rewrite preserves the exact `<=` tie-break (follow-up wins on equal arrival) — verified byte-for-byte against the pre-PR index merge.
- Wall-clock-derived fields (`ArrivalTimeUs`/`RequestID` for follow-up rounds, random `XRequestID`) are deliberately excluded from parity comparison; the determinism surface is request identity + arrival ordering, not raw trace bytes (observe records real timestamps).
- No `sim/` changes; no new struct fields; no `Peek()` added to `RequestSource`.

Closes #1443.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,6 +129,17 @@ go build -o blis main.go
 # multi-client / cohort workloads. Behavior with the flag off is unchanged.
 ./blis run --model qwen/qwen3-14b --lazy-generation
 
+# Observe with lazy request generation (alpha, #1443). Same flag, default, and
+# semantics as `blis run` — streams requests from the generator into the observe
+# dispatch loop instead of pre-generating the full slice, with the same eager
+# fallback (time-varying / concurrency / multi-session). Observe already paces
+# itself against the real server, so the memory win is smaller than run's; the
+# flag mainly makes run and observe share one generation pipeline (#1438). Default
+# (flag off) dispatch behavior is unchanged.
+./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
+  --workload chatbot --rate 10 --num-requests 100 --lazy-generation \
+  --trace-header trace.yaml --trace-data trace.csv
+
 # Convert workload formats
 ./blis convert preset --name chatbot --rate 10 --num-requests 100
 ./blis convert servegen --path data/
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -166,7 +166,7 @@ func init() {
 	observeCmd.Flags().IntVar(&observeTimeout, "timeout", defaultHTTPTimeoutSeconds, "HTTP request timeout in seconds (per request)")
 
 	// ITL recording (opt-in; requires streaming)
-	observeCmd.Flags().BoolVar(&observeRecordITL, "record-itl", false, "Record per-chunk timestamps for ITL calibration (streaming only; forces streaming on non-streaming workloads)")
+	observeCmd.Flags().BoolVar(&observeRecordITL, "record-itl", false, "Record per-chunk timestamps for ITL calibration (forces streaming per request; mutually exclusive with --no-streaming)")
 	observeCmd.Flags().StringVar(&observeITLOutput, "itl-output", "", "Output path for ITL CSV file (default: <trace-data>.itl.csv if --record-itl is set)")
 
 	// Saturation analysis (optional, run/replay parity)
@@ -206,6 +206,21 @@ func validateObserveWorkloadFlags(preset, workloadSpec string, rateChanged bool,
 	}
 	if !rateChanged {
 		return fmt.Sprintf("--workload %q requires --rate (preset synthesis needs a request rate)", preset)
+	}
+	return ""
+}
+
+// validateITLStreamingFlags rejects the incoherent --record-itl + --no-streaming
+// combination. ITL recording captures per-chunk timestamps, which only exist for
+// streaming responses; with --no-streaming the per-request streaming override in
+// runObserveOrchestrator is defeated by requestToPending's `Streaming && !noStreaming`,
+// so no ITL is ever recorded. Fail fast with a clear message instead of emitting a
+// per-request "non-streaming" warning hundreds of times (PR #1457 review).
+// Returns a non-empty error string if the combination is invalid, empty otherwise.
+// Extracted for unit testability (R14).
+func validateITLStreamingFlags(recordITL, noStreaming bool) string {
+	if recordITL && noStreaming {
+		return "--record-itl and --no-streaming are mutually exclusive; ITL recording requires streaming responses"
 	}
 	return ""
 }
@@ -258,6 +273,11 @@ func runObserve(cmd *cobra.Command, _ []string) {
 	// Warn if --itl-output is set without --record-itl (no ITL data will be written)
 	if observeITLOutput != "" && !observeRecordITL {
 		logrus.Warnf("--itl-output is set but --record-itl is not enabled; no ITL data will be written")
+	}
+	// --record-itl requires streaming; reject --no-streaming upfront rather than
+	// silently no-opping the per-request streaming override (PR #1457 review).
+	if msg := validateITLStreamingFlags(observeRecordITL, observeNoStreaming); msg != "" {
+		logrus.Fatalf("%s", msg)
 	}
 	// BC-7: at least one workload input mode must be provided
 	if observeWorkload == "" && observeWorkloadSpec == "" && !cmd.Flags().Changed("rate") && observeConcurrency <= 0 {
--- a/cmd/observe_cmd.go
+++ b/cmd/observe_cmd.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/inference-sim/inference-sim/sim"
+	"github.com/inference-sim/inference-sim/sim/cluster"
 	"github.com/inference-sim/inference-sim/sim/saturation"
 	"github.com/inference-sim/inference-sim/sim/workload"
 	"github.com/sirupsen/logrus"
@@ -482,7 +483,7 @@ func runObserve(cmd *cobra.Command, _ []string) {
 
 	// Run orchestrator
 	startTime := time.Now()
-	runObserveOrchestrator(ctx, client, recorder, sessionMgr, wl.Requests, observeNoStreaming, observeMaxConcur, observeWarmup, prefixes, prefixLengths, observeUnconstrainedOutput, observeRecordITL, tokensPerWord)
+	runObserveOrchestrator(ctx, client, recorder, sessionMgr, cluster.NewSliceRequestSource(wl.Requests), observeNoStreaming, observeMaxConcur, observeWarmup, prefixes, prefixLengths, observeUnconstrainedOutput, observeRecordITL, tokensPerWord)
 	logrus.Infof("Observation wall-clock time: %.3fs", time.Since(startTime).Seconds())
 
 	// Resolve goodput SLO targets (#1413, BC-1, BC-7). Observe has no trace
@@ -865,14 +866,30 @@ func runPrewarm(ctx context.Context, client *RealClient, duration time.Duration)
 	}
 }
 
+// preferFollowUp reports whether the pending session follow-up should dispatch
+// before the buffered pre-generated request. Ties go to the follow-up (<=),
+// preserving the pre-lazy merge order (former index-based merge in
+// runObserveOrchestrator). Extracted as a pure function so the tie-break is
+// unit-testable — a live orchestrator run cannot construct a deterministic tie
+// because follow-up arrival times are wall-clock-derived (session.go). (#1443)
+func preferFollowUp(followUp, preGen *sim.Request) bool {
+	return followUp.ArrivalTime <= preGen.ArrivalTime
+}
+
 // runObserveOrchestrator implements the dispatch loop with session support.
 // This is the core orchestration function, extracted for testability.
+//
+// Requests are pulled from source (a cluster.RequestSource — eager slice adapter
+// or lazy streaming generator, selected by --lazy-generation) one at a time via
+// Next(), and merged against in-flight session follow-ups by arrival time using a
+// one-slot lookahead buffer (nextPreGen). This preserves the O(in-flight) memory
+// property of lazy generation on the request stream (#1443, #1438 Change A5).
 func runObserveOrchestrator(
 	ctx context.Context,
 	client *RealClient,
 	recorder *Recorder,
 	sessionMgr *workload.SessionManager,
-	requests []*sim.Request,
+	source cluster.RequestSource,
 	noStreaming bool,
 	maxConcurrency int,
 	warmupCount int,
@@ -882,9 +899,6 @@ func runObserveOrchestrator(
 	recordITL bool,
 	tokensPerWord float64,
 ) {
-	if len(requests) == 0 {
-		return
-	}
 
 	semaphore := make(chan struct{}, maxConcurrency)
 	var wg sync.WaitGroup
@@ -897,16 +911,17 @@ func runObserveOrchestrator(
 	// Completion channel for session serialization (BC-8, D7)
 	completionCh := make(chan completionEvent, maxConcurrency)
 
-	// Active session tracking for drain (count unique session IDs)
+	// Active session tracking for drain. Counting is folded into the dispatch
+	// loop (incremented on the first sighting of each SessionID among the
+	// pre-generated requests) rather than pre-scanned, so the lazy source is
+	// never fully materialized up front (#1443). seenSessions dedups so a
+	// session is counted exactly once; follow-ups reuse an existing SessionID
+	// and arrive via followUpCh (never through the pre-gen selection site), so
+	// they cannot double-count.
 	activeSessionCount := int64(0)
+	var seenSessions map[string]bool
 	if sessionMgr != nil {
-		sessionIDs := make(map[string]bool)
-		for _, req := range requests {
-			if req.SessionID != "" && !sessionIDs[req.SessionID] {
-				sessionIDs[req.SessionID] = true
-				activeSessionCount++
-			}
-		}
+		seenSessions = make(map[string]bool)
 	}
 
 	// Session serializer goroutine (BC-8: single-threaded OnComplete)
@@ -967,10 +982,37 @@ func runObserveOrchestrator(
 	}
 
 	// Merge pre-generated requests and follow-ups, dispatch in arrival order.
-	// Follow-ups are buffered in a local slice and merged by arrival time
-	// with pre-generated requests (deterministic, no select/default race).
-	preGenIdx := 0
+	// Pre-generated requests are pulled from the source one at a time through a
+	// one-slot lookahead buffer (nextPreGen); follow-ups are buffered in a local
+	// slice and merged by arrival time (deterministic, no select/default race).
 	var pendingFollowUps []*sim.Request
+
+	// nextPreGen holds the next pre-generated request pulled from source (the
+	// one-slot lookahead that replaces the former requests[preGenIdx] random
+	// access). nil once the source is exhausted.
+	var nextPreGen *sim.Request
+	if r, ok := source.Next(); ok { // prime the lookahead
+		nextPreGen = r
+	}
+
+	// takePreGen returns the buffered pre-generated request and refills the
+	// lookahead from source. It also folds active-session counting into the
+	// loop: the first time a session's (round-0) request is selected, the
+	// session is counted. Both pre-gen consumption branches route through this
+	// single helper so the counting cannot diverge between them.
+	takePreGen := func() *sim.Request {
+		req := nextPreGen
+		if sessionMgr != nil && req.SessionID != "" && !seenSessions[req.SessionID] {
+			seenSessions[req.SessionID] = true
+			atomic.AddInt64(&activeSessionCount, 1)
+		}
+		if r, ok := source.Next(); ok {
+			nextPreGen = r
+		} else {
+			nextPreGen = nil // exhausted; req above already captured, not dropped
+		}
+		return req
+	}
 
 	drainFollowUps := func() {
 		for {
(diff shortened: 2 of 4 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## BLIS Self-Review — PR #1457 (`hardening(observe): plumb --lazy-generation through blis observe`, closes #1443)

**Verdict: ✅ READY TO MERGE**

Independent review against PR head `427c9487` (`+919 / -57`, 4 files: `cmd/observe_cmd.go`, `cmd/observe_cmd_test.go`, `docs/guide/observe-replay-calibrate.md`, `CLAUDE.md`). This is **Change A — PR 5 of 5 (A5)** and it closes the observe-parity gap that was deferred as the follow-up during the A3/A4 reviews. CI fully green; all three prior `@claude` self-review rounds' findings confirmed addressed on this head. I re-derived the correctness argument independently and ran the lazy path locally rather than trusting the prior rounds.

---

### Independent correctness verification (all pass)
I specifically re-checked the paths the prior rounds touched only lightly:
- **Empty-source termination is leak-free.** Removing the old `if len(requests)==0 { return }` early-return is safe: with an empty source and `sessionMgr != nil`, `activeSessionCount` stays 0, so the `activeSessionCount > 0` drain guard (`observe_cmd.go:1166`) is false → `break` → `drain:` → `close(completionCh)` unblocks the serializer's `for range`, `<-serializerDone` returns. No hang, no goroutine leak.
- **`takePreGen` precondition always holds.** Both call sites (`:1158`, `:1161`) are under `hasPreGen := nextPreGen != nil`; no `goto drain` path calls it. The `req.SessionID` deref is safe.
- **`wl.Requests` nil-safety.** Only 4 references; each is guarded by `lazySource == nil` or is `len(nil)==0`-safe. The prefix-string loop iterates `spec.Clients`, not `wl.Requests`.
- **Incremental session counting is correct.** `activeSessionCount` is now folded into `takePreGen` (first-sight of each round-0 `SessionID`) instead of a startup pre-pass. This threads a real needle: the count is only *read* at the drain guard when the pre-gen source is exhausted — by which point every session's round-0 has necessarily been counted. Folding it one line earlier, or reading the guard while pre-gen still has items, would drain prematurely. It's correct as written.
- **`classifyLazyGenError` `default → lazyFatal` is the safe design** (unknown error aborts; it does *not* silently succeed). Note for the record: round 3's prose described this as a "default = success" pattern — that description is inaccurate, but the actual code (`observe_cmd.go:95-97`) is correct and safe.
- **BC-9**: `lazySource.Err()` is surfaced post-dispatch, before trace export; `logrus.Fatalf → os.Exit(1)` yields no trace, matching eager's abort-at-generation.

### Test quality — non-vacuous and mutation-resistant
- `TestObserveOrchestrator_MergeOrder_OpenLoopExactSequence` — absolute hand-written dispatch-order guard *including a tie* (5000,5000); catches drop/dup/reorder that a cross-mode parity test structurally cannot.
- `TestObserveOrchestrator_EagerLazyParity_SameDispatchSequence` — has a real vacuity guard (`hasRound(...,1)` in **both** modes) ensuring the follow-up merge path is actually exercised; correctly excludes wall-clock fields with documented justification.
- `TestObserveOrchestrator_RecordITL_CoversSessionFollowUps` — asserts the round-0 precondition (non-streaming) so the test genuinely proves the per-request override reached the follow-up (BC-6, the strict improvement).
- `TestClassifyLazyGenError` — exhaustive: nil→stream, 3 sentinels→fallback (with reason substring), **wrapped** sentinel via `fmt.Errorf("%w")`→fallback, generic error→fatal.
- Subprocess Fatalf tests (`RecordITLNoStreaming_FatalNoTrace`, `LazySamplerError_FatalNoTrace`) — assert exit-1 + reason substring + **neither trace file written**. The lazy-sampler trigger (a `ReasonRatioDist` that passes `Validate()` but fails the streaming sampler) is a clever way to reach the `Err()` path.

### Local verification (I ran the lazy path on `blis run`, per request)
| Case | Result |
|---|---|
| chatbot (×2 seeds), summarization, multidoc, contentgen, distribution | **stdout byte-identical** eager vs lazy (INV-6) |
| `--trace-output` (chatbot, 100 req) | **TraceV2 header + data byte-identical** (stronger than stdout — full request stream) |
| single-session reasoning (blueprint pre-pass) | streamed lazily (no fallback); **trace byte-identical** |
| `--concurrency` (closed-loop) | eager fallback + correct warning; stdout identical |
| multi-session reasoning (`SingleSession=false`) | eager fallback + correct warning; stdout identical |
| `go test ./cmd/ -race` (observe suite) + `./sim/workload` lazy tests + full `./cmd` | all **PASS** |

The trace byte-identity run is a stronger check than the PR's own parity test: `blis run` is pure DES (no wall-clock in output), so I could diff the *entire* exported trace, closing the gap the observe-side test necessarily leaves open (observe records real timestamps).

### Cross-path parity
This PR **closes** the observe gap. `--lazy-generation`: `run` ✅ (A3) · `replay` ✅ (accepted no-op, CLI symmetry) · `observe` ✅ (this PR). Fallback sentinels are the same `sim/workload` values matched via `errors.Is` in both `run` and `observe`. No new gaps.

### Preemption / timeout
- **Preemption: N/A** — observe dispatches real HTTP; no DES event loop, no `StateQueued→Running→Completed`, no re-queue.
- **Timeout: not touched** — `--lazy-generation` changes *generation*, not the HTTP client timeout or rate-pacing.

### Docs
`observe-replay-calibrate.md` (flag row + updated `--record-itl` mutual-exclusion note) and `CLAUDE.md` (observe lazy example with the "smaller memory win, shared pipeline" rationale) are accurate and appropriately scoped.

---

### Non-blocking observations (nano-nits; do not block merge)
1. **Empty-trace warning asymmetry (new).** In lazy mode the `No requests generated — writing empty trace` warning is suppressed (`observe_cmd.go:499` guards on `lazySource == nil`), so a lazy run that produces zero requests writes an empty trace silently, whereas eager warns. Purely diagnostic; not a correctness issue. Could log the same warning after dispatch when `recorder` saw zero records.
2. **Acceptance-criteria form.** #1443 suggested a checked-in golden trace under a stubbed transport for the "byte-identical-to-today" criterion; the PR meets the issue's explicit *minimum bar* (dispatch-order / deterministic-surface parity) instead. Within scope, and my local full-trace byte-identity runs cover the stronger form in practice.

**Bottom line:** Correct, well-tested, parity-closing. The lazy plumbing mirrors `blis run` exactly, every behavioral contract has a meaningful non-vacuous test, and I verified byte-identity locally across presets, reasoning, and trace output. Ready to merge.

susiejojo:
## Reproduction + verification: memory reduction is working

Reproduced the peak-memory pain point locally against the current tree (`blis run`, `qwen/qwen3-14b`, single instance) and measured peak RSS via `/usr/bin/time -l` (macOS). Both memory-representation changes are merged and always-on, and Change A delivers a large, deterministic win on supported workloads.

### Change status confirmed in-tree

| Change | Effect | State |
|---|---|---|
| **B** — `[]int` → `[]TokenID` (`int32`) | 2× per-token tax removed everywhere | ✅ merged, always-on (`sim/internal/tokenid/tokenid.go:11`) |
| **C** — delta-encoded multi-turn rounds | O(R²) per-session accumulation removed | ✅ merged, always-on (#1452) |
| **A** — lazy (streaming) generation | eager materialization + `perClientCap=2×maxRequests` over-generation removed | ✅ available via `--lazy-generation` (#1453) |

### Reproducer 1 — the issue's exact shape (multi-turn `accumulate`, 30K prefix)

40-member cohort, `max_rounds: 15`, 30K-token shared prefix, 1500-in / 425-out per round, `num_requests: 500`, `single_session: true`.

| Mode | Peak RSS |
|---|---|
| eager (default) | 100.4 MB |
| `--lazy-generation` | 90.4 MB |

`--lazy-generation` is **supported** for this shape (no eager fallback), and stdout is **byte-identical** between modes (INV-6 holds). The peak is already modest here because **Change C is always-on** — the O(R²) accumulation that drove the original tens-of-GB spike is gone independent of the flag.

### Reproducer 2 — isolating the eager-generation factors Change A removes

To exercise factors (1) eager materialization and (2) per-client over-generation directly: 40 single-shot clients, unlimited horizon, `num_requests: 1000` (so each client generates up to `2×1000` candidates → ~80,000 materialized, then truncated to 1000). Modest 2K-token shared prefix.

| Mode | Peak RSS |
|---|---|
| eager (default) | **1,750.9 MB (1.75 GB)** |
| `--lazy-generation` | **66.3 MB** |

**≈ 26× peak-RSS reduction**, stdout **byte-identical**, no fallback, both runs complete 1000/1000 requests. This lands within the issue's predicted "10–100× (workload-dependent)" for Change A.

### Commands

```bash
go build -o blis main.go

# Reproducer 2 (over-generation dominant)
/usr/bin/time -l ./blis run --model qwen/qwen3-14b --workload-spec overgen.yaml
/usr/bin/time -l ./blis run --model qwen/qwen3-14b --workload-spec overgen.yaml --lazy-generation
```

<details><summary>overgen.yaml</summary>

```yaml
version: "2"
seed: 42
category: language
aggregate_rate: 400.0
num_requests: 1000
cohorts:
  - id: "single-shot-cohort"
    population: 40
    rate_fraction: 1.0
    slo_class: "standard"
    arrival: { process: poisson }
    input_distribution:  { type: constant, params: { value: 1000.0 } }
    output_distribution: { type: constant, params: { value: 100.0 } }
    prefix_group: "shared-catalog"
    prefix_sharing: "shared"
    prefix_length: 2000
    streaming: false
```
</details>

**Summary:** B and C are merged and reduce memory on every run; A produces a ~26× peak-RSS reduction with byte-identical output on supported workloads. The memory-representation and streaming mechanisms all behave as designed.

susiejojo (approved):
lgtm

PR #1461, opened 2026-07-10
hardening(workload): lazy generation for multi-session reasoning (#1458)

### LINKED ISSUE #1458: Extend lazy request generation to multi-session reasoning (single_session=false) (#1438)
## A-coverage — Extend lazy request generation to multi-session reasoning (`single_session: false`)

**Tracker:** #1438 · **Change A — lazy-coverage 1 of 3** · **Blocks:** A8 (#1451, eager removal)

### Background (for a newcomer)

`blis` builds the workload — the full stream of requests to simulate — before the simulation runs. Historically it did this **eagerly**: it materialized every request into one big `[]*sim.Request` slice up front. For large workloads (many clients, long multi-turn conversations, big shared prefixes) that slice can grow to tens of GB and crash the machine. Issue #1438 ("Change A") introduced an alternative **lazy** (streaming) generator behind the `--lazy-generation` flag: requests are produced one at a time as the simulation consumes them, so peak memory is bounded by the *in-flight* working set instead of the *total* request count.

Lazy generation shipped as an opt-in alpha covering a **subset** of workloads. For anything outside that subset, `blis` silently falls back to the eager generator (with a one-line stderr warning) so the run still succeeds — it just doesn't get the memory benefit. There are three such unsupported classes today, each guarded by a sentinel error. This issue covers **one** of them: **multi-session reasoning**.

### What "multi-session" means

A reasoning (multi-turn) client models a conversation that accumulates context across rounds. It has two modes, set by `single_session`:

- **`single_session: true`** — the client is *one* persistent conversation cycling through its rounds sequentially. Round N+1 cannot arrive until round N completes. This is the shape used by the inference-perf catalog and #1438's reproducer, and it **is** supported by lazy today.
- **`single_session: false`** (the default when omitted) — the client's arrival process spawns *many independent conversations over time*, which overlap and interleave. This is **not** supported by lazy; it falls back to eager.

Example of a spec that hits the fallback (a multi-turn cohort that omits `single_session`, so it defaults to `false`):

```yaml
cohorts:
  - id: "code-generation"
    population: 40
    reasoning:
      multi_turn:
        max_rounds: 15
        think_time_us: 15000000
        context_growth: accumulate
        # single_session omitted -> false -> multi-session -> eager fallback
```

Running this with `--lazy-generation` today prints:

```
[workload] --lazy-generation ignored: workload has multi-session reasoning (SingleSession=false); using eager generator (issue #1441)
```

### Where this is in the code (permalinks pinned to `main` @ `1f1af36`)

- The sentinel that triggers the fallback: [`sim/workload/stream.go` L29–L38](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L29-L38)
- The gate that rejects multi-session and returns the sentinel: [`sim/workload/stream.go` L528–L531](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L528-L531)
- The eager behavior that must be reproduced byte-for-byte — the multi-session loop that spawns successive sessions across the horizon: [`sim/workload/generator.go` L163–L224](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L163-L224)
- The lazy per-client state machine that already contains a multi-session `else` branch (see "Why this is more achievable than the comment suggests"): [`sim/workload/stream.go` L204–L312](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L204-L312)
- The CLI fallback switches that consume the sentinel: [`cmd/root.go` L1510–L1515](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/cmd/root.go#L1510-L1515) and [`cmd/observe_cmd.go` L249–L254](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/cmd/observe_cmd.go#L249-L254)

### Does lazy actually save memory for multi-session? (yes — the current code comment is too pessimistic)

The sentinel's doc comment ([`stream.go` L29–L38](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L29-L38)) claims reproducing the eager interleaving "would require pushing every session's anchor onto the heap up front — defeating the memory savings." That overstates the difficulty:

- **Eager** generates *every* session across the *entire horizon* for each client, then globally sorts and truncates — resident cost `O(total sessions over the whole run)`.
- A **correct lazy** implementation only needs the sessions that have *started but not finished* at the current sim-time (the concurrent working set). A client draws its next session's start time one step ahead (exactly as the eager loop advances `currentTime += iat` at [`generator.go` L167–L177](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L167-L177)). Resident cost ≈ `arrival_rate × session_duration` (Little's law) — **independent of horizon**.

So the savings ratio is roughly `horizon / session_duration`, which is still a large win precisely when the memory cliff bites. Multi-session lazy is **not** `O(total)` and is worth doing. The `stream.go` comment should be corrected as part of this work (or a small standalone edit) to say "deferred pending a per-client internal merge + parity tests," not "defeats the memory savings."

### What is genuinely hard (and it's not memory)

1. **Per-client internal ordering.** Eager relies on the *global* sort to interleave one client's overlapping sessions correctly (e.g., session 1 round 5 at t=100 vs session 2 round 1 at t=50). A streaming client must yield *its own* requests in non-decreasing arrival order — a small per-client sub-merge across its live sessions, rather than the single linear cursor the current single-session model uses.
2. **Byte-identical determinism (INV-6).** The load-bearing requirement is that lazy produces a request stream **byte-identical** to eager for the same seed. That means matching eager's exact RNG draw order (per-session `SampleIAT`, then `GenerateReasoningRequests`) and its `perClientCap = 2 * maxRequests` over-generation + global-sort-truncate behavior ([`generator.go` L56–L58](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L56-L58)). This is the real work.
3. **Closed-loop blueprint pre-pass.** For closed-loop multi-turn, the lazy path builds `SessionManager` blueprints and reproduces eager's "generate all rounds → truncate → filter round-0" sequencing. That pre-pass and its cap-counting must be audited to hold when a client emits multiple sessions rather than one.

### Scope

1. Teach the lazy per-client state machine to stream multiple sessions per client in arrival order without materializing all of a client's sessions at once (per-client internal merge over live sessions).
2. Remove the `ErrLazyUnsupportedMultiSession` gate at [`stream.go` L528–L531](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L528-L531) once support lands; the sentinel and the two CLI fallback branches become dead for this class and should be removed.
3. Correct the `stream.go` L29–L38 comment.

### Acceptance criteria

- A property/parity test in the style of the A4 harness ([`sim/workload/parity_property_test.go`](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/parity_property_test.go)) proves eager ≡ lazy **byte-identical** request streams across sampled seeds for multi-session reasoning specs (INV-6). This is the gate that made the single-session alpha safe; multi-session must clear the same bar.
- A peak-RSS measurement (`/usr/bin/time -v` or equivalent) on a multi-session, long-horizon workload shows a material reduction under `--lazy-generation` vs eager (the `horizon / session_duration` win above), reported before/after in the PR.
- `go test ./...` and `golangci-lint run ./...` clean.
- The `stream.go` L29–L38 comment no longer claims multi-session defeats the memory savings.

### Alternative (explicit decision point for maintainers)

If the maintainers decide multi-session lazy support is not worth the implementation + parity-proof cost, the alternative is to **consciously accept that multi-session workloads hard-error after eager is removed in A8 (#1451)** — i.e., `blis` would refuse to run them rather than silently falling back. That is a product decision, not a default. This issue assumes the implementation path; if the decision goes the other way, it should be recorded here and A8 updated to hard-error (with a clear message) instead of removing the fallback. Either way, **A8 must not delete the eager path while multi-session silently depends on it.**

### Out of scope

- Concurrency clients (separate lazy-coverage issue).
- Time-varying / per-window parameters (separate lazy-coverage issue).

### Labels

`hardening`

namasl:
## Summary

Extend lazy request generation (`--lazy-generation`, #1441) to **multi-session reasoning** clients (`single_session: false` — the default multi-turn shape). Previously these silently fell back to the eager generator with a warning; that fallback blocked eager removal (tracker #1438, **AC1** / A8 hard-blocker).

The blocker was **per-client ordering, not memory**. The lazy min-heap holds one entry per client and requires each client's sub-stream to be arrival-monotonic. Multi-session sessions **overlap** (session N+1's round 0 can precede session N's later rounds), so draining one session before the next produced a non-monotonic per-client stream that broke the global merge.

**Fix:** each multi-session client runs a small internal merge over its **live** (built-but-not-drained) sessions. Sessions are built in eager's index order (identical RNG draws) and emitted in arrival order behind a **safety gate**: a live head is emitted only once no not-yet-built session could precede it (future session starts are strictly `> currentTime`). Resident sessions are bounded by the concurrent working set (≈ `arrival_rate × session_duration`, Little's law) — independent of horizon. This corrects the too-pessimistic `stream.go` comment (which claimed multi-session lazy "defeats the memory savings").

## Issue

Closes #1458

## How byte-identity (INV-6) is preserved

1. Sessions are **built in eager's index order** → identical RNG draw order (`SampleIAT` then `GenerateReasoningRequests`).
2. **Replicates eager's `perClientCap = 2 × maxRequests`** per-client build budget. This is *required* for multi-session (unlike single-shot, where build order == arrival order and the global popped-cap suffices): build order ≠ arrival order here, so an un-capped build could emit a round eager truncated away.
3. Emit key `(arrival, sessionIdx, roundIndex)` reproduces eager's global `sort.SliceStable` tie-break (append order = client, then session-build-index, then round-index).

The blueprint pre-pass and closed-loop round-0 suppression are **unchanged** — they flow through the fixed `produceNext` in cloned states.

## Cross-Path Parity

| `blis run` | yes | yes | Removes the multi-session fallback switch branch; multi-session now streams |
| `blis observe` | yes | yes | Removes `classifyLazyGenError`'s multi-session case; mirrors run |
| `blis replay` | no | N/A | Reads requests from a trace file; does not generate |
| `blis calibrate` | no | N/A | Post-processing |

## Invariants Affected

- **INV-6 (determinism):** Preserved. Verified: eager and lazy produce **byte-identical stdout** for a multi-session spec; the property test asserts eager ≡ lazy across ≥100 draws (2000 draws pass locally) now sampling `single_session=false`, guarded by a `sawMultiSession` coverage check.
- **INV-13 (run/replay parity):** Multi-session is now lazy-supported and flows through the same per-arrival trace hook; the existing `parity_run_replay_test.go` suite covers the path.

## Behavioral Contracts

- **BC-1** — GIVEN a lazy-supported spec with `single_session: false` reasoning clients, WHEN the lazy stream is drained and compared to eager, THEN the two request streams and session blueprints are byte-identical.
- **BC-2** — GIVEN one multi-session client with overlapping sessions, WHEN it produces requests, THEN each request's arrival time is `>=` the previous (arrival-monotonic sub-stream).
- **BC-3** — GIVEN a spec whose only unsupported trait was `single_session: false`, WHEN `GenerateWorkloadLazy` runs, THEN it returns a usable source with `nil` error and no fallback warning.
- **BC-4** — GIVEN a multi-session spec where `maxRequests`/`horizon` binds, WHEN eager and lazy both generate, THEN the surviving requests and blueprints match.

## Peak-RSS measurement

Short-session, long-horizon multi-session workload (8000-token shared prefix, `max_rounds=3`, `think_time=20ms`, 100k requests, 1-hour horizon), peak RSS via `/proc/<pid>/VmHWM`:

| Mode | Peak RSS |
|------|----------|
| eager | 8721 MB |
| `--lazy-generation` | 3690 MB |

**2.4× reduction (−58%)**, matching the `horizon / session_duration` thesis. (The residual lazy footprint is dominated by cluster-side in-flight state — 100k requests with large prefixes flowing through the DES — which this PR does not change; the generator-side spike is what is eliminated.)

## Test Plan

- [x] Unit tests pass (`go test ./...`)
- [x] Lint passes (`golangci-lint run ./...` — 0 issues)
- [x] `TestLazyRequestSource_Reasoning_MultiSession_MatchesEager` (open- and closed-loop)
- [x] `TestLazyRequestSource_MultiSession_PerClientMonotonic`
- [x] `TestLazyRequestSource_MultiSession_MultiClientTightCap_MatchesEager` (cohort + tight cap)
- [x] `TestProperty_EagerEqualsLazy_RequestStreams` extended to `single_session=false` (2000 draws byte-identical)
- [x] Eager stdout == lazy stdout for a multi-session spec (INV-6)

---

Generated with [Claude Code](https://claude.com/claude-code)


namasl:
@claude /blis-pr-review

namasl:
Pushed `fae85bf0` addressing self-review findings (round 1): added multi-session + lifecycle parity coverage (IMPORTANT), guarded the empty-session heap push, extended the stream comparator with ReasonRatio, added a `reasoning-multi-session` row to the CLI run/replay parity matrix (INV-13), and corrected the CLAUDE.md memory-complexity wording. Build/test/lint clean.

@claude /blis-pr-review

susiejojo:
### BLIS PR Self-Review — #1461 (independent verification pass)

Thanks for the thorough write-up and the two prior self-review rounds — they made this pass much easier to follow. This review targets the two areas of interest: **cross-path parity (run / replay / observe / calibrate)** and **whether multi-session reasoning is addressed adequately** (the byte-identity claim vs the eager generator). Verified on `fae85bf0`; build clean, `sim/workload` + `cmd` test suites green.

**Verdict: READY TO MERGE** — no BLOCKING or IMPORTANT findings.

---

#### Cross-path parity ✅

| Path | Status | Notes |
|------|--------|-------|
| `blis run` | ✅ Streams | Multi-session fallback branch removed (`cmd/root.go`) |
| `blis observe` | ✅ Mirrors run | `classifyLazyGenError` multi-session case removed (`cmd/observe_cmd.go`); help text + comments updated |
| `blis replay` | ✅ N/A (correct) | Trace-driven; parity enforced by the new `reasoning-multi-session` matrix row (INV-13) |
| `blis calibrate` | ✅ N/A (correct) | Post-processing only |

Help-text and doc wording is now consistent across `cmd/root.go`, `cmd/observe_cmd.go`, and `docs/guide/observe-replay-calibrate.md`, and `invariants.md` INV-13 names the new matrix entry. No drift found.

#### Multi-session correctness (byte-identity vs eager oracle `generator.go:163-220`) ✅

Traced point-by-point:

1. **Build order / RNG** — `buildNextSession` guard order (horizon → cap → `SampleIAT` → `+=iat` → horizon → lifecycle → build → count) exactly mirrors eager's `for currentTime < horizon` loop. Identical draw order → INV-6 preserved.
2. **Per-client cap** — `msClientCap = 2×maxRequests` with the same int64 overflow guard; `msReqCount += len(reasoningReqs)` counts **before** horizon/lifecycle filtering (matches `generator.go:205`), and is threaded into **both** Phase 3 and the blueprint pre-pass clone.
3. **Tie-break** — `liveSessionHeap.Less = (arrival, sessionIdx)` reproduces eager's `sort.SliceStable` append order; the subtle same-timestamp-across-sessions case resolves correctly (earlier-built session wins, consistent with eager).
4. **Emit-safety gate** — `head.ArrivalTime > currentTime → hold` is provably safe: next session start = `currentTime + iat` with `iat ≥ 1`, so the `>` boundary (emit when `≤`) can never emit a round a future session could precede.
5. **`heap.Fix(0)`** after `cursor++` correctly re-heapifies the mutated min.
6. **Lifecycle-skip re-entry** — returns false without `msBuildDone`; no infinite loop (clock advances each call, `lastWindowEndUs` terminates).
7. **Per-client monotonicity (BC-2)** — the sub-stream fed to the global heap is arrival-monotonic, so the global merge stays correct.
8. **Zero-round session** — guarded before push yet still counted toward the cap, matching eager. No `head()` panic path.

#### Preemption / timeout

Purely a workload-generator-layer change (`sim/workload/`) — no per-request metrics, state transitions, or aggregate counters touched, so preemption semantics are unaffected. `buildSession` sets `req.Deadline = computeDeadline(req.ArrivalTime, client.Timeout, true)` identically to eager for multi-session. No new timeout/horizon semantics.

---

#### Optional follow-ups (non-blocking, none required for merge)

- **MINOR (pre-existing, not introduced here):** `enumerateSurvivingSessionsPerClient` accepts a `prefixes` parameter that is unused. Could be dropped in a future cleanup — noting only for awareness, not as a blocker for this PR.
- **Nice-to-have test:** a case where the **per-client** build cap (`2×maxRequests`) itself binds for a single multi-session client (distinct from the global truncation cap), to directly guard the `msClientCap`/pre-pass-threading parity. Existing coverage exercises this indirectly via the multi-client tight-cap test.

Nice work — the implementation is clean, well-commented, and the parity reasoning holds.

_Independent verification pass; did not re-run the dedicated test-coverage agent._

susiejojo (approved):
lgtm

Issue #1460, opened 2026-07-09
Extend lazy request generation to time-varying / per-window workloads (#1438)


namasl:
## A-coverage — Extend lazy request generation to time-varying / per-window workloads

**Tracker:** #1438 · **Change A — lazy-coverage 3 of 3** · **Blocks:** A8 (#1451, eager removal)

### Background (for a newcomer)

`blis` builds the workload (the full stream of requests to simulate) before the run. The **eager** generator materializes every request into one `[]*sim.Request` slice up front, which can reach tens of GB and crash the machine on large workloads. Issue #1438 ("Change A") added a **lazy** (streaming) generator behind `--lazy-generation` that produces requests one at a time, bounding peak memory by the in-flight working set.

Lazy shipped as an opt-in alpha covering a subset of workloads; anything outside the subset silently falls back to eager (with a stderr warning) so the run still succeeds without the memory benefit. There are three unsupported classes, each guarded by a sentinel error. This issue covers **time-varying / per-window** workloads.

### What "time-varying / per-window" means

A client can have a **lifecycle** made of time **windows**, and each window can **override** the client's parameters for that slice of the run — a different arrival rate, a different input/output token distribution, or a different arrival process. This is how diurnal patterns, traffic spikes, and ServeGen-style time-series workloads are modeled: the same client behaves differently at t=0–1h than at t=1h–2h. A workload is "time-varying" when any window carries per-window overrides (`trace_rate`, `arrival`, `input_distribution`, or `output_distribution`).

A spec that hits the fallback (two windows with different per-window rates):

```yaml
clients:
  - id: "diurnal"
    input_distribution:  { type: constant, params: { value: 512 } }
    output_distribution: { type: constant, params: { value: 128 } }
    lifecycle:
      windows:
        - { start_us: 0,        end_us: 3600000000, trace_rate: 2.0 }
        - { start_us: 3600000000, end_us: 7200000000, trace_rate: 20.0 }
```

Running with `--lazy-generation` today prints:

```
[workload] --lazy-generation ignored: workload has per-window parameters; using eager generator (issue #1441)
```

### Where this is in the code (permalinks pinned to `main` @ `1f1af36`)

- The sentinel: [`sim/workload/stream.go` L20](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L20)
[...]

Issue #1451, opened 2026-06-26
A8 — Remove eager generation (#1438)


namasl:
## A8 — Remove eager generation

**Tracker:** #1438 · **Change A — Graduation step 3 of 3** · **Depends on:** #1450 (A7) + a removal-grace window **AND the lazy-coverage gap being closed — #1458, #1459, #1460 (see hard blocker below)**

After A7's deprecation period, the eager code path is removed. This is where the full benefit of #1438 finally lands in code: there's only one generation strategy, the abstraction tax is gone, and the `perClientCap = 2 * maxRequests` over-generation factor disappears entirely.

> ### ⚠️ HARD BLOCKER — lazy coverage gap
>
> The lazy generator shipped as an alpha covering a **subset** of workloads. Three workload classes are **not** supported by lazy today and currently fall back to eager (each guarded by a sentinel error): **multi-session reasoning** (#1458), **concurrency clients** (#1459), and **time-varying / per-window** workloads (#1460).
>
> **This PR cannot delete the eager path while any of those three classes silently depends on it** — doing so would make those workloads hard-error with no working generation path. Before A8 can land, each of #1458 / #1459 / #1460 must reach one of two resolved states:
>
> 1. **Lazy support implemented** (the workload now streams; the sentinel + fallback are removed), OR
> 2. **A recorded maintainer decision to hard-error** that class after eager removal — in which case A8 replaces the silent fallback with an explicit, clear error message rather than deleting it blindly.
>
> The original draft of this issue (item 4 below) only flagged the *time-varying* fallback. That was an oversight: **all three** unsupported classes are eager-removal blockers, not just one. They are now tracked as #1458, #1459, #1460 and feed into this issue in the tracker DAG (#1438).

### What this PR does

1. **Delete the eager generation code path** in `sim/workload/generator.go` — including the eager `[]*sim.Request` build-up, the `perClientCap = 2 * maxRequests` cap, the post-merge sort, and the truncation step.
2. **Delete the eager `RequestSource` adapter** introduced in A1. The lazy streaming source is the only implementation.
3. **Remove the `--lazy-generation` CLI flag** from `blis run`, `blis replay`, and `blis observe`. Lazy is no longer optional.
[...]

PR #1462, opened 2026-07-14
hardening(workload): lazy generation for concurrency clients (#1459)

### LINKED ISSUE #1459: Extend lazy request generation to concurrency clients (#1438)
## A-coverage — Extend lazy request generation to concurrency clients (`concurrency > 0`)

**Tracker:** #1438 · **Change A — lazy-coverage 2 of 3** · **Blocks:** A8 (#1451, eager removal)

### Background (for a newcomer)

`blis` builds the workload (the full stream of requests to simulate) before the run. The **eager** generator materializes every request into one `[]*sim.Request` slice up front, which can reach tens of GB and crash the machine on large workloads. Issue #1438 ("Change A") added a **lazy** (streaming) generator behind `--lazy-generation` that produces requests one at a time, bounding peak memory by the in-flight working set instead of the total request count.

Lazy shipped as an opt-in alpha covering a subset of workloads; anything outside the subset silently falls back to eager (with a stderr warning) so the run still succeeds without the memory benefit. There are three unsupported classes, each guarded by a sentinel error. This issue covers **concurrency clients**.

### What "concurrency clients" means

A **concurrency client** models a fixed pool of *closed-loop virtual users* rather than an open-loop arrival rate. Instead of "requests arrive at rate λ," it means "N users each send a request, wait for the response (+ think time), then send the next — indefinitely." It's set via `concurrency: N` on a client (with `rate_fraction: 0`), and it's how `blis observe --concurrency` and closed-loop load tests are modeled. The number of in-flight requests is capped by N, and follow-up requests are generated reactively at completion time by the `SessionManager`, not scheduled up front.

A spec that hits the fallback:

```yaml
clients:
  - id: "load-test-users"
    concurrency: 50        # 50 closed-loop virtual users
    input_distribution:  { type: constant, params: { value: 512 } }
    output_distribution: { type: constant, params: { value: 128 } }
```

Running with `--lazy-generation` today prints:

```
[workload] --lazy-generation ignored: workload has concurrency clients; using eager generator (issue #1441)
```

### Where this is in the code (permalinks pinned to `main` @ `1f1af36`)

- The sentinel: [`sim/workload/stream.go` L27](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L27)
- The gate that rejects concurrency clients and returns the sentinel: [`sim/workload/stream.go` L525–L527](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L525-L527)
- The eager concurrency path — seed-request generation + `SessionBlueprint`/`FollowUpBudget` construction that lazy must reproduce: [`sim/workload/generator.go` L489–L510](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L489-L510) and the `FollowUpBudget` contract at [`generator.go` L327–L331](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L327-L331)
- The CLI fallback switches: [`cmd/root.go` L1512](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/cmd/root.go#L1512) and [`cmd/observe_cmd.go` L251](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/cmd/observe_cmd.go#L251)

### Memory consideration

Concurrency clients are the *most* naturally streaming-friendly of the three unsupported classes: by construction only the seed requests (one per virtual user, so `O(N)`) exist up front, and follow-ups are generated reactively at completion by the `SessionManager`. There is no horizon-length pre-materialization to eliminate the way there is for open-loop clients — the eager cost here is dominated by the seed set plus whatever the `SessionManager` holds. So the memory win from lazy is smaller for pure-concurrency workloads than for open-loop ones. **However**, this class still forces the *entire run* onto the eager path whenever a concurrency client is present alongside other clients (mixed specs), so removing the fallback is primarily about (a) not disqualifying an otherwise-streamable mixed workload and (b) letting eager be deleted in A8. The PR should measure and report the actual peak-RSS delta so the benefit is stated honestly rather than assumed.

### What is genuinely hard

1. **Byte-identical determinism (INV-6).** Lazy must reproduce eager's seed-generation RNG draw order and the `FollowUpBudget` computation exactly. See [`generator.go` L489–L510](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/generator.go#L489-L510).
2. **SessionManager interaction.** Concurrency follow-ups are produced reactively (not from a pre-built blueprint list the way closed-loop multi-turn round-0 is). The lazy source must hand off to the same `SessionManager` follow-up machinery the eager path uses, preserving the budget cap and the closed-loop pacing, without materializing the pool's full request history.

### Scope

1. Teach the lazy path to emit concurrency seed requests and integrate with the `SessionManager` follow-up generation, bounded to `O(N virtual users)` resident state.
2. Remove the `ErrLazyUnsupportedConcurrency` gate at [`stream.go` L525–L527](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/stream.go#L525-L527) once support lands; the sentinel and the two CLI fallback branches become dead for this class.

### Acceptance criteria

- A property/parity test in the style of the A4 harness ([`sim/workload/parity_property_test.go`](https://github.com/inference-sim/inference-sim/blob/1f1af3683fad2b664f6a12999d126e4e2234550e/sim/workload/parity_property_test.go)) proves eager ≡ lazy **byte-identical** request streams across sampled seeds for concurrency-client specs, including the follow-up rounds the `SessionManager` generates (INV-6).
- A peak-RSS measurement on a mixed workload (concurrency client + open-loop clients) shows the run is now streamable under `--lazy-generation`; report before/after honestly (the win may be modest for pure-concurrency specs — see "Memory consideration").
- `go test ./...` and `golangci-lint run ./...` clean.

### Alternative (explicit decision point for maintainers)

If concurrency lazy support is judged not worth the cost, the alternative is to **consciously accept that concurrency workloads hard-error after eager removal in A8 (#1451)** rather than silently falling back. That is a recorded product decision, not a default. This issue assumes the implementation path; if the decision goes the other way, record it here and update A8 to hard-error with a clear message. **A8 must not delete the eager path while concurrency clients silently depend on it.**

### Out of scope

- Multi-session reasoning (separate lazy-coverage issue).
- Time-varying / per-window parameters (separate lazy-coverage issue).

### Labels

`hardening`

namasl:
## Summary

Extends BLIS's **lazy** (streaming) request generator to support **concurrency clients** (`concurrency > 0`, closed-loop virtual-user pools), removing the `ErrLazyUnsupportedConcurrency` fallback. This is **lazy-coverage 2 of 3** for Change A (#1438) and a **hard blocker for A8** (#1451, eager removal).

Before this PR, any spec with a concurrency client silently fell back to the eager generator under `--lazy-generation` (with a stderr warning). Now it streams through the lazy source, and `--lazy-generation` produces **byte-identical** stdout, traces, and follow-up rounds vs. eager (INV-6, INV-13). The only remaining unsupported class is time-varying (per-window) workloads (#1460).

Fixes #1459.

## How it works

The concurrency seed-request + `SessionBlueprint` construction is **extracted into one shared helper** (`generateConcurrencySeedsAndBlueprints`) called by both the eager `GenerateWorkload` and the lazy `GenerateWorkloadLazy` (R4 — single construction site), so the RNG-draw order (`concurrencyRNG = spec.Seed + 10007`, per-user `userSeed → token samples → bpSeed`) and the seed cap are identical by construction.

Lazy pushes **each concurrency seed as its own heap entry**, keyed `(arrival, len(allClients), generationIndex)`, onto the same global arrival heap the open-loop states feed. The global min-heap then reproduces eager's `sort.SliceStable`-by-arrival over `[open-loop…, seed_0, seed_1, …]` byte-for-byte:
- `len(allClients)` (computed post-cohort-expansion) is strictly greater than every real client index, so open-loop pops before seeds at equal arrival;
- the generation index breaks ties among seeds in eager's stable order.

A **single naive FIFO seed emitter would be wrong** here: two concurrency pools with different `think_time`/`concurrency` produce interleaved seed arrivals (e.g. A={0,100,200}, B={0,50}), and a FIFO head hides B's earlier-arriving seed behind A's. Individual heap entries fix this.

**Seed cap parity (the subtle part).** Eager's seed loop stops at `len(round0Only) + len(seeds) >= maxRequests`. The lazy path obtains `keptOpen` (the open-loop count that survives the cap) as a byproduct of the blueprint pre-pass dry-run and feeds it as `alreadyKept`. By the no-displacement invariant — *if any seed is emitted, all open-loop was kept, so `genOpen + nSeeds <= maxRequests`* — the global popped-cap is non-binding whenever seeds exist, so pushing pre-capped seeds is exact.

Seed heap entries carry a single shared **always-exhausted sentinel** `clientStreamState`, which makes `lazyRequestSource.Next` panic-free at both `e.state` dereference sites with **zero changes to `Next()`** (`produceNext()` returns false → no re-push; `isClosedLoop == false` → seed emits directly).

The `horizon <= 0` fast path is now concurrency-aware: eager emits concurrency seeds even at `horizon <= 0` (horizon-independent, and without running `validateAndExpandSpec`), so lazy mirrors that exact sequence rather than silently returning an empty source.

## Behavioral contracts

- **BC-1** — GIVEN a valid concurrency spec, WHEN both generators run, THEN the ordered request streams are byte-identical field-by-field.
- **BC-2** — session blueprints match (length, per-index `SessionID`/`ClientID`/`UnlimitedRounds`, identically-seeded RNGs).
- **BC-3** — `FollowUpBudget` matches eager (shared `concurrencyFollowUpBudget` formula).
- **BC-4** — seed cap parity under a binding `maxRequests` (both emit exactly `maxRequests`; `FollowUpBudget == 0` when seeds consume the remainder).
- **BC-5** — determinism: two lazy runs at the same seed are byte-identical.
- **BC-6** — run/replay + eager/lazy CLI trace byte-identity (INV-13).
- **BC-7** — the `ErrLazyUnsupportedConcurrency` sentinel + its two CLI fallback branches are removed.

## Testing

- `TestGenerateWorkloadLazy_ConcurrencyClient_MatchesEager` — 6-case parity table: mixed, **multi-concurrency** (≥2 pools with distinct stagger — the ordering regression test), all-concurrency, binding-cap (BC-4), **horizon-zero**, and **combined-cap** (open-loop partially kept + seeds partially truncated).
- `TestProperty_EagerEqualsLazy_RequestStreams` — extended to sample concurrency specs (~¼ of draws), with `sawConcurrency` + `sawMultiConcurrency` anti-vacuity guards and a scoped `FollowUpBudget` parity assertion. Clean at 100 and 1000 draws.
- `TestParity_RunReplay_TraceByteIdentity_Matrix` — new two-pool `concurrency` row proving eager==lazy trace + lazy determinism at the CLI seam.
- Fallback unit tests replaced with positive parity assertions (`GenerateWorkloadLazy` now returns a usable source, not a sentinel).

Verification: `go build ./...`, `go test ./...`, and `golangci-lint run ./...` all clean.

## Peak-RSS (issue acceptance criterion #2 — reported honestly)

Measured via `/proc/<pid>/status` VmHWM, `blis run` eager vs `--lazy-generation`:

| Spec | Eager peak | Lazy peak |
|---|---|---|
| Mixed (open-loop + concurrency, 400k reqs, large tokens) | 3091 MB | 3071 MB |
| Mixed (generator-dominated, 1.5M reqs, tiny tokens) | 2463 MB | 2391 MB |
| Pure-concurrency (200 users, 100k reqs) | 85 MB | 90 MB |

As the issue predicted, the win is **modest** for concurrency: concurrency seeds are `O(N virtual users)` and exist up front in **both** modes (there is no horizon-length pre-materialization to stream away), and cluster-side in-flight state dominates peak RSS at scale. **The primary value is architectural**: a mixed spec is no longer *disqualified* from streaming by the presence of a concurrency client, and eager can now be deleted in A8. Reported per the issue's "Memory consideration" — stated rather than assumed.

## Review

Plan and code each passed the project convergence review (Steps 2.5 / 4.5): the plan review caught two CRITICAL design flaws pre-implementation (multi-client seed mis-ordering; `horizon<=0` silent divergence), both resolved in the design; the code review caught a guard-reorder divergence (empirically confirmed and fixed), an undocumented load-bearing invariant (now documented), and a stale docstring. The self-audit (Step 4.75) additionally caught two stale `--help` strings.


namasl:
@claude /blis-pr-review

namasl:
## 🔎 BLIS PR Self-Review — verdict: **READY TO MERGE**

Reviewed against issue #1459 and tracker #1438, with focus on invariants, cross-path parity, preemption safety, and timeout/deadline consistency. `go build ./...`, `go test ./...`, and `golangci-lint run ./...` are clean.

### Requirements coverage (#1459 acceptance criteria)
- ✅ **Lazy path emits concurrency seeds + integrates with `SessionManager` follow-up generation**, bounded to `O(N virtual users)` resident state. Follow-ups remain the `SessionManager`'s job (mode-agnostic, untouched) — the PR produces byte-identical `SessionBlueprint`s + `FollowUpBudget`, which is all the `SessionManager` consumes.
- ✅ **`ErrLazyUnsupportedConcurrency` gate + sentinel + both CLI fallback branches removed.**
- ✅ **Property/parity test in the A4 style** proves eager ≡ lazy byte-identical across sampled seeds, including the `SessionManager`-generated follow-up rounds (via blueprint RNG-seed parity). 100 + 1000 draws clean.
- ✅ **Peak-RSS measured and reported honestly** (see PR body) — modest for pure-concurrency (seeds are `O(N)` up front in both modes), value is architectural (unblocks A8).

### Invariants
- **INV-6 (determinism / byte-identity):** shared `generateConcurrencySeedsAndBlueprints` helper guarantees identical RNG-draw order; individual-seed-heap-entries reproduce eager's `sort.SliceStable`; verified by 6-case table + 1000-draw property test + CLI trace matrix. A load-bearing invariant (`keptOpen == len(round0Only)` only under the concurrency⊥multi-turn mutual exclusion) is now explicitly documented in code.
- **INV-13 (run/replay parity):** new `concurrency` row in `TestParity_RunReplay_TraceByteIdentity_Matrix` proves byte-identical traces.

### Cross-path parity (run / replay / observe) — ✅ complete
- **run** + **observe** both consume `GenerateWorkloadLazy`; both CLI fallback branches (`cmd/root.go`, `cmd/observe_cmd.go`) updated symmetrically; both `--help` strings corrected.
- **replay** loads traces (does not generate workloads) and accepts `--lazy-generation` as a documented no-op (`cmd/replay.go` unchanged) — correct, no parity gap.
- No CLI flags, spec fields, defaults, or metrics diverge across paths.

### Preemption safety — ✅ N/A (verified)
The diff touches **only workload generation** (`sim/workload/`) + CLI wiring + tests + docs. Zero changes to per-request metrics, state transitions, or aggregate counters (grep for TTFT/ITL/E2E/ProgressIndex/State*/CompletedRequests/recordRequestCompletion/RunningBatch in the diff returns nothing). No preemption exposure.

### Timeout / deadline consistency — ✅ preserved
Concurrency seeds get `Deadline` via the single shared `computeDeadline(arrivalTime, client.Timeout, true)` call inside the extracted helper — identical for eager and lazy. `DefaultTimeoutUs` (300s) and HTTP timeout defaults are untouched. Trace pipeline deadline semantics unchanged (covered by the trace byte-identity matrix).

### Design & standards
- **R4 (single construction site):** the concurrency seed/blueprint block is now extracted into one helper called by both generators — a net R4 improvement, not a violation.
- **R1/R6 (error handling):** the helper returns `error` (sampler failures) and both callers propagate; no `logrus.Fatalf` in `sim/`; no silent `continue`.
- **Sentinel-state safety:** the shared `exhaustedSentinelState` is unexported, immutable-in-practice (exhausted path performs no writes), never added to `states` — verified panic-free at both `e.state` deref sites in `Next()` with zero `Next()` changes.
- **Docs:** CLAUDE.md, observe-replay-calibrate.md, both `--help` strings, and all relevant godoc updated; time-varying is now the only remaining fallback class.

### Prior review comments
This is a fresh PR; the plan (Step 2.5, 4 rounds) and code (Step 4.5, 2 rounds) convergence reviews and the Step 4.75 self-audit are all resolved (see PR body). No open reviewer threads.

### Verdict
**Ready to merge.** No CRITICAL or IMPORTANT findings survive. Optional future items (out of scope): the pre-existing large-function trend in `generator.go`/`stream.go`, and the pre-existing open-loop `FollowUpBudget` 0-vs-(-1) cosmetic difference (inert — never read without sessions).

---
*Self-review generated via the `blis-pr-review` workflow.*

susiejojo:
## Self-review verdict: READY TO MERGE ✅

Thanks for this — it's a clean, well-reasoned extension of `--lazy-generation` to concurrency clients (#1459), and it reads as a natural sibling to the multi-session PR (#1458). I ran the full BLIS review pass (correctness/invariants, cross-path parity, preemption, timeout/deadline, tests, docs) plus an independent verification of the byte-identity claim against the eager oracle. Build + `sim/workload` + `cmd` tests all pass locally. No blocking or important issues — only three non-blocking documentation/comment nits below.

### Correctness (byte-identity vs eager) — verified sound
- **Shared construction site (R4):** `generateConcurrencySeedsAndBlueprints` is the single source of seeds + blueprints for both eager (`generator.go:501`) and lazy (`stream.go:896`). `concurrencyRNG` (`seed+10007`) is created fresh inside the helper, so the per-user draw order (`userSeed → token samples → bpSeed`) is invariant to prior open-loop draws → INV-6 holds by construction.
- **Load-bearing invariant `keptOpen == len(round0Only)`:** correct, and correctly documented. It holds only because concurrency ⊥ multi-turn is a hard error in `spec.Validate` (`spec.go:359-361`). Without that guard, the lazy dry-run's `popped` (counts intermediate closed-loop rounds, `stream.go:1109` before the round-0 filter at `:1114`) would diverge from eager's round-0-only `round0Only` (`generator.go:476-486`). The exclusivity guard makes them equal. Good defense-in-depth.
- **No-displacement invariant:** the seed cap-check (`alreadyKept + len(seeds) >= maxRequests → break`, `generator.go:587`) guarantees `keptOpen + nSeeds <= maxRequests`, so the global popped-cap in `lazyRequestSource.Next` is non-binding whenever a seed exists → no truncation, matching eager's non-truncating `sort.SliceStable`.
- **Emit tie-break:** seeds pushed with `clientIdx = len(allClients)` (> all real indices) and `perClientSeq = g` reproduce eager's stable order over `[open-loop…, seed_0, seed_1, …]` at equal arrival exactly.
- **`exhaustedSentinelState`:** clean, panic-free at both `e.state` deref sites with zero `Next()` changes; holds no per-seed state so sharing one instance is safe.

### Cross-path parity (run / replay / observe / calibrate)
- **run:** fallback branch removed in `cmd/root.go`; now streams. ✓
- **observe:** `ErrLazyUnsupportedConcurrency` case removed from `classifyLazyGenError`; help text mirrors run. ✓
- **replay:** trace-driven (N/A for generation), and the new `concurrency` row in `paritySpecShapes()` exercises INV-13 byte-identity + both-mode replay. ✓
- **calibrate:** post-processing (N/A). ✓

### Preemption safety
Purely a workload-generator-layer change (`sim/workload/`). No touch to per-request metrics, request-state transitions, or aggregate counters — preemption semantics are unaffected.

### Timeout / deadline consistency
`generateConcurrencySeedsAndBlueprints` sets `Deadline: computeDeadline(arrivalTime, client.Timeout, true)` (`generator.go:620`), identical to eager. No new timeout/horizon semantics.

### Dead-code & test quality
- `ErrLazyUnsupportedConcurrency` fully removed — remaining references are documentation-only comments noting its removal. ✓
- Old `TestGenerateWorkloadLazy_FallsBackOnConcurrencyClient` replaced by positive parity tests, not just deleted. ✓
- Property test has real vacuity guards (`sawConcurrency`, `sawMultiConcurrency`) plus a `FollowUpBudget` parity assertion — cannot pass vacuously. ✓

---

### Non-blocking follow-ups (MINOR — documentation/comment drift only)

1. **`cmd/root_test.go:1650-1652,1656`** — the assertion-site inline comments still say `"eager fallback"` / `"written by eager fallback"`, but the test now exercises the lazy *streaming* path (the whole point of the `_Streams` rename). The test's top docstring (`:1575-1587`) was updated correctly; just the inline comments lagged.

2. **`docs/contributing/standards/invariants.md:182`** — the INV-13 "under lazy generation" coverage-matrix enumeration lists *"single-turn chatbot, multi-turn accumulate cohort, single-session reasoning, and multi-session reasoning"* but not `concurrency`. #1458 added its shape here; #1459 should too. (This file isn't in the PR's changed set.)

3. **`cmd/parity_run_replay_test.go:10-12`** — the top-of-file BC-3 docstring still lists only three shapes and is now missing both `reasoning-multi-session` (inherited from #1458) and `concurrency`. Note the `paritySpecShapes()` docstring at `:61-76` *is* fully up to date, so this is just the file-header summary.

None of these affect behavior, and all three could be swept into a single trivial docs commit. I'd suggest fixing #1 and #2 before merge (they're in this PR's own files / directly describe this PR's invariant) and #3 opportunistically.

Nice work — the load-bearing-invariant comment and the shared-helper extraction are exactly the right calls here.

namasl:
Addressed the three documentation/comment-drift findings from @susiejojo's review (MINOR-1/2/3) in 8b11a0b6:

- `cmd/root_test.go` — the renamed `TestRunCmd_LazyGeneration_Concurrency_Streams` inline comments + `t.Fatalf` message now describe the lazy streaming path (were stale "eager fallback").
- `cmd/parity_run_replay_test.go` — file-header BC-3 docstring now names multi-session reasoning (#1458) and concurrency (#1459), pointing to `paritySpecShapes` as authoritative.
- `docs/contributing/standards/invariants.md` — INV-13 coverage-matrix enumeration now includes `concurrency clients` and #1459.

No behavior change; build + `cmd` tests + lint clean.

@claude /blis-pr-review

susiejojo (approved):
lgtm

PR #1473, opened 2026-07-16
feat(lora): PR2 — per-instance resident-adapter set + capacity + LRU eviction (#1465)

atantawi:
## Summary

PR2 of the LoRA control-plane subsystem (epic #1464). Adds the **per-instance resident-adapter set** (User Story 2): a capacity-bounded LRU over adapter ids, wired into the core `Simulator`, plus per-adapter load/eviction metrics.

**State-only:** residency has no latency or gating effect yet — a cold request still runs immediately. The cold-load pre-admission gate that consumes this state is a follow-up PR.

> ✅ **PR1 (#1472) merged to `main` (2026-07-23); this branch has been rebased onto merged `main`.** The stale PR1 commits it previously carried are gone — the diff now shows the single PR2 commit alone. Part of the LoRA control-plane stack (epic #1464); PR3–PR7 follow and will rebase onto this once it lands.

Closes #1465.

## What it does

- **`sim/lora/resident_set.go`** — capacity-bounded LRU reusing the `cpuTier` hash + doubly-linked-list pattern (`store`/`touch`/`unlink`/`appendToTail`), with reference-counted pinning (`Pin`/`Unpin`/`EvictLRU` proven now so the follow-up gate inherits a tested primitive). Deterministic, no RNG (INV-6).
- **`sim/resident_adapter_set.go`** — `ResidentAdapterSet` seam interface + `NewResidentAdapterSetFunc`, registered by `sim/lora`'s `init()` (mirrors `NewAdapterRegistryFunc` / `NewLatencyModelFunc`; no reverse import). Interface is intentionally minimal (`IsResident`/`Touch`/`Store`/`Len`); `Pin`/`Unpin`/`EvictLRU` join it when the gate consumes them (R13).
- **`sim/simulator.go`** — `Simulator.residentAdapters`, wired in `NewSimulator` when adapters + capacity are configured (capacity validated `> 0`, returns an **error** not a panic — R6); `recordAdapterResidency` hook off the existing `NewlyScheduled` loop (cold → `Store`, warm → `Touch`).
- **`sim/metrics.go` / `metrics_utils.go`** — `AdapterLoadCounts`/`AdapterEvictionCounts`, surfaced through `buildAdapterMetrics` (sorted-id; returns nil for adapter-blind runs).
- **`sim/cluster/cluster.go`** — `aggregateMetrics` sums the per-adapter counts across instances (adapter ids legitimately recur across instances, unlike globally-unique request ids → sum, not merge-and-warn).

## Invariants

[...]

PR #1472, opened 2026-07-15
feat(lora): PR1 — adapter identity plumbing + config/registry + per-adapter metrics (#1464)

atantawi:
Implements **PR1 (#1471)** of the LoRA control-plane subsystem epic **#1464** — the no-op-safe foundation: adapter identity plumbing, config + registry, and per-adapter metrics (spec **US1**).

## No-op default (INV-6, SC-001)
With no adapters configured the subsystem is fully inert and stdout is **byte-identical** to a pre-feature build, verified by a golden subprocess test against `specs/007-lora-control-plane/testdata/baseline_noop.json`.

## What's included
**Config & registry**
- `LoRAConfig` — 7th `SimConfig` sub-config (R16) — + `AdapterSpec` + `StepOverheadTier` + `Validate()` (R3/R11 numeric guards, R9 pointer types)
- `sim/lora` package: adapter `Registry` (`id → rank[, base model]`, key-sorted R2) wired into `sim` via `NewAdapterRegistryFunc` (`init()` seam, no reverse import — Principle I)
- inert `lora:` defaults in `defaults.yaml`

**CLI (run + replay parity, INV-13)**
- `--lora-config` file + `--lora-adapter-capacity` / `--lora-load-base-latency-us` / `--lora-load-bandwidth-bytes-us` / `--lora-footprint-bytes-per-rank`
- `resolveLoRAConfig` composes `defaults.yaml < file < flags` (R18 `Flags().Changed`); adapters declared with `adapter_capacity: 0` ⇒ `logrus.Fatalf`

**Identity & metrics**
- `Request.Adapter` threaded 1:1 with `Model` at all 10 workload construction sites (R4)
- `ValidateAdapterReferences` (unknown id + base-model mismatch) on the run path
- `RequestMetrics.Adapter` + `MetricsOutput.Adapters` (per-adapter TTFT p50/p99 + throughput); `omitempty` so an adapter-blind run adds no stdout fields

## Testing
TDD throughout (tests fail first). `go build ./...`, `go test ./...`, and `golangci-lint run ./...` all clean. Determinism and a 2-adapter end-to-end (`adapters` block emitted, sorted keys) verified.

## Deliberate deviations from the tasks.md literal text (reviewers please note)
1. `AdapterSpec` gained an optional `BaseModel` field — `data-model.md` requires it for the base-model-match validation (T012), though T004 named only `{ID, Rank}`.
[...]

Issue #1464, opened 2026-07-15
feat(lora): LoRA control-plane subsystem


atantawi:
## LoRA control-plane subsystem

Give BLIS enough LoRA awareness to research **adapter placement, migration, and adapter-aware routing** in a deterministic, GPU-free harness — so a policy developed against BLIS transfers to llm-d.

**Classification** (Extension Framework): **Subsystem Module** (design doc required, no-op default mandatory) + one **Policy Template** (the routing scorer).

**Approach**: graft an adapter identity onto the existing `Model`-tag plumbing; add a per-instance resident-adapter set (capacity + LRU eviction); port the Digital Twin's three cost terms (cold-load latency by rank, per-step compute overhead by unique-adapters × max rank, per-adapter HBM footprint) as *deltas* onto BLIS's calibrated engine; expose a `lora-affinity` routing scorer; surface per-adapter metrics.

### Spec artifacts
On feature branch `007-lora-control-plane`, `specs/007-lora-control-plane/`: `spec.md` (+ Clarifications), `plan.md` (Constitution Check), `research.md`, `data-model.md`, `contracts/`, `quickstart.md`, `tasks.md`.

### Mandatory constraint
**No-op default (INV-6)**: with no adapters configured, output MUST be byte-identical to today. Every PR below is individually no-op-safe (Small-tier).

> **Design-gate status (2026-07-15): APPROVED** (human gate, tantawi). Convergence ran to the 10-round cap + a confirmation round; design *substance* validated against the codebase. Not a formal 0/0 — residuals are documentation/consistency (not behavioral), all applied; the final edits are unverified, to be confirmed at the per-PR micro-plan + code-review gates. **Cleared to proceed to PR1 (#1471).**
>
> **Authoritative source:** the committed `tasks.md` + design doc (`docs/plans/2026-07-15-lora-control-plane-design.md`, branch `007-lora-control-plane`) are authoritative. The task checklists embedded in the PR sub-issues are **indicative and predate the convergence review** — re-derive tasks from `tasks.md` at micro-plan time.

> **C1 prerequisite**: the 7th `SimConfig` sub-config (`LoRAConfig`) requires ratifying a Principle VI amendment ("one sub-config per module", constitution v1.1.0) in the design-doc convergence review before PR1 merges.

### Sub-issues (PR order)
- [x] #1471 — PR1 Adapter identity plumbing + config/registry + per-adapter metrics (MVP)
[...]

PR #1474, opened 2026-07-17
feat(lora): PR3 — adapter cold-load latency via pre-admission gate (#1466)

### LINKED ISSUE #1466: feat(lora): PR3 — Adapter cold-load latency via pre-admission gate (#1464)
**Parent epic**: #1464

### Scope
Adapter **cold-load latency** via a **pre-admission gate** (spec **US3**, load-latency part) + shared cost-model core.

### Tasks (TDD)
- [x] T025 Cost-model contract tests (`sim/lora/cost_model_test.go`): `LoadLatency≥0`, footprint monotonic in rank, `StepOverheadFactor=1.0` when no adapters; pure/deterministic
- [x] T026 Implement `sim/lora/cost_model.go` (pure query methods) + `init()`
- [x] T027 Cold-load contract test: cold adds LoadLatency to TTFT (once), warm zero (SC-004); loads serialized per instance
- [x] T028 Implement pre-admission gate (`sim/batch_formation.go` + `sim/simulator.go`): hold cold request out of batch; schedule serialized load completion at `now + base + ceil(bytes/bw)`
- [x] T029 Companion invariants: INV-5 (enqueue≤schedule), INV-8 (work-conserving), INV-3

### Invariants / rules
INV-3/5/8; load latency ≥ 0 charged only on cold; determinism (no RNG). Resolves design-doc §9 gating (research R1).

### Depends on
PR2 (#1465 resident set — cold/warm state). ✅ Merged as `42e29187`; PR3 rebased onto `main` in #1474.

### Definition of done
Verification gate clean; cold-vs-warm TTFT gap demonstrable.

atantawi:
## Summary

PR3 of the LoRA control-plane epic (#1464). Implements US3's cold-load part: a **serialized per-instance cold-load pre-admission gate** plus the **shared adapter cost-model core**. Closes #1466.

> **✅ Rebased onto merged `main` (2026-07-23).** PR1 (#1472) and PR2 (#1473) have merged, so this PR is now a **single clean commit** (`78fe5d01`) on top of `main` — the diff is PR3 only. PR2's review-round hardening was carried across the rebase (see the two `NewSimulator`/`recordAdapterResidency` conflict resolutions: the cost-model wiring now sits inside PR2's factory-nil guard + warning branches).

## What's in it

**Cost model** — `sim/lora/cost_model.go` (T025/T026)
- Pure, deterministic queries: `LoadLatency(id) = base + ceil(footprint/bw)`, `FootprintBytes(id)`, `StepOverheadFactor(batch) = 1 + (K6(r_max)/K7(r_max))·A_B`. `StepOverheadFactor`/`FootprintBytes` are built now but consumed by PR4/PR5.
- Fail-fast construction rejects nil / non-positive / non-finite (NaN, ±Inf) coefficients and any declared rank whose load latency would overflow a safe int64 bound (R3/R20, protects INV-3).
- Exposed to `sim/` via the new `sim.AdapterCost` seam (`LoadLatency` only — R13-minimal), registered by `sim/lora` `init()`.

**Cold-load gate** — `sim/simulator.go` + `sim/batch_formation.go` (T027/T028)
- `maybeStartAdapterLoad`: cold wait-queue head with no in-flight load → commit the LRU eviction victim (reserve a slot) and schedule an `AdapterLoadCompletionEvent` at `now + LoadLatency`; serialized per instance.
- `FormBatch` holds any not-yet-resident adapter out of the batch via `BatchContext.AdapterResident` (DT-faithful **blocking model**: the gated head stalls requests behind it for the load duration, design §7).
- New `AdapterLoadCompletionEvent`, priority-ordered ahead of a co-timed `StepEvent` (`PriorityAdapterLoad` inserted between `Queued` and `Step`; all existing relative orderings preserved ⇒ INV-6 byte-identity holds for adapter-blind runs).
- Completion makes the adapter resident, charges the load once (INV-L3), clears the in-flight marker, and ensures a step forms (INV-8) on **every** path.
- **Pinning (INV-L5)**: pin on running-batch entry (once per request via `Request.adapterPinned`, persists through preemption), released at every terminal path — completion, timeout, and **gateway in-flight eviction** (new exported `ReleaseAdapterPin`, called from cluster `EvictRequest`). `scheduleNextStep` suppresses empty-step spinning while a load is in flight.
- `ResidentAdapterSet` interface gains `AtCapacity`/`EvictLRU`/`Pin`/`Unpin` (consumed by the gate).

**Config** — `LoRAConfig.Validate` now requires the cost coefficients when adapters are declared with a positive capacity, so the CLI catches the gap at its validation gate rather than failing later in `NewSimulator`.

## Tests (T029)
Cold-vs-warm TTFT gap, load serialization, same-adapter coalescing, INV-5/INV-8/INV-6/determinism, no-deadlock under capacity pressure, cost-model laws (monotonicity, no-op normalization, malformed-config rejection incl. NaN/Inf/overflow), and a **cluster eviction pin-release regression** (provably fails without the fix). The PR2-inherited `TestResidentAdapterSet_PreemptionDoesNotDoubleCountLoad` was updated to supply cost coefficients (PR3 now requires them) and still asserts LoadCount==1 + `IsResident` across a preemption. Full `go test ./...` + `golangci-lint run ./...` clean; no-op byte-identity golden holds.

## Invariants
INV-3 (clock monotonicity — load completion `> now`, overflow-guarded), INV-5 (gate delays schedule not arrival), INV-6 (byte-identity when inert), INV-8 (work-conserving across the load), INV-L3 (load charged once per cold transition), INV-L5 (pinned adapters never evicted; pin released on all terminal paths).

### DIFF
--- a/sim/adapter_cost.go
+++ b/sim/adapter_cost.go
@@ -1,0 +1,24 @@
+package sim
+
+// AdapterCost is the read-only query bridge over the LoRA adapter cost model. It
+// is owned by sim/ so the cold-load pre-admission gate can charge load latency
+// without importing sim/lora (Principle I: no reverse import); the concrete
+// model (rank-derived cost terms) lives in sim/lora and is wired in via
+// NewAdapterCostFunc, mirroring NewAdapterRegistryFunc / NewResidentAdapterSetFunc.
+//
+// The interface is intentionally scoped to what the gate consumes today (R13):
+// only LoadLatency. The concrete cost model additionally derives the per-step
+// compute-overhead factor and the HBM footprint; those enter this interface when
+// the latency and memory PRs consume them.
+type AdapterCost interface {
+	// LoadLatency returns the one-time cold-load latency of an adapter id in µs
+	// (>= 0). An empty (base-model) or unregistered id returns 0 — it never gates.
+	LoadLatency(id string) float64
+}
+
+// NewAdapterCostFunc builds an AdapterCost from a LoRAConfig (rank registry +
+// cost coefficients). Registered by sim/lora's init() (import sim/lora to wire
+// it); nil until then, in which case the Simulator leaves cold-load gating inert
+// (INV-6). Returns an error for a malformed config (missing/non-positive divisor
+// coefficients), which the caller maps to fatality.
+var NewAdapterCostFunc func(cfg LoRAConfig) (AdapterCost, error)
--- a/sim/batch_formation.go
+++ b/sim/batch_formation.go
@@ -31,6 +31,14 @@ type BatchContext struct {
 	Now                   int64
 	StepCount             int
 	ComputedTokens        map[string]int64
+
+	// AdapterResident is the cold-load pre-admission gate predicate (#1466): it
+	// reports whether a request's LoRA adapter is currently resident on the
+	// instance. A new prefill request whose adapter is NOT resident is held out of
+	// the batch (blocking model: it also stalls the requests behind it) until the
+	// kernel-scheduled adapter load completes. nil ⇒ no LoRA gating; admission is
+	// byte-identical to a pre-feature build (INV-6).
+	AdapterResident func(id string) bool
 }
 
 // ScheduledRequest carries metadata about a newly scheduled request.
@@ -157,6 +165,17 @@ func (v *VLLMBatchFormation) FormBatch(ctx BatchContext) BatchResult {
 	// Phase 2: Dequeue new requests from wait queue
 	for len(result.RunningBatch.Requests) < int(ctx.MaxRunningReqs) && ctx.WaitQ.Len() > 0 && tokenBudget > 0 && !result.PreemptionHappened {
 		next := ctx.WaitQ.Peek()
+
+		// Cold-load pre-admission gate (LoRA, #1466): a new prefill request whose
+		// adapter is not yet resident on this instance is held out of the batch
+		// until its adapter load completes (FR-007, §7). Because Phase 2 inspects
+		// only the queue head and breaks on the first non-admittable request, this
+		// realizes the DT-faithful blocking model — a gated head stalls the warm
+		// requests behind it for the load duration. Decode sub-requests (PD) are
+		// already past the gate (their prefill ran with the adapter resident).
+		if ctx.AdapterResident != nil && !next.IsDecodeSubRequest && next.Adapter != "" && !ctx.AdapterResident(next.Adapter) {
+			break
+		}
 
 		// Handle decode-only requests (PD disaggregation: KV pre-allocated by transfer).
 		// IsDecodeSubRequest is set exclusively by KVTransferStartedEvent when it
--- a/sim/cluster/adapter_metrics_aggregation_test.go
+++ b/sim/cluster/adapter_metrics_aggregation_test.go
@@ -17,9 +17,15 @@ func TestClusterSimulator_AdapterMetrics_AggregatedAcrossInstances_E2E(t *testin
 	config := newTestDeploymentConfig(2)
 	config.RoutingPolicy = "round-robin"
 	capVal := 8 // capacity >> 1 adapter, so no evictions
+	// Cost coefficients are required once the cold-load gate consumes them (#1466);
+	// production fills them from defaults.yaml when adapters are configured.
+	base, bw, fp := 1000.0, 2.0e6, 2.0e6
 	config.LoRAConfig = sim.LoRAConfig{
-		AdapterCapacity: &capVal,
-		Adapters:        []sim.AdapterSpec{{ID: "adapter_shared", Rank: 8}},
+		AdapterCapacity:       &capVal,
+		LoadBaseLatencyUs:     &base,
+		LoadBandwidthBytesUs:  &bw,
+		FootprintBytesPerRank: &fp,
+		Adapters:              []sim.AdapterSpec{{ID: "adapter_shared", Rank: 8}},
 	}
 
 	requests := newTestRequests(6)
--- a/sim/cluster/instance.go
+++ b/sim/cluster/instance.go
@@ -477,6 +477,10 @@ func (i *InstanceSimulator) DrainWaitQueue() []*sim.Request {
 func (i *InstanceSimulator) EvictRequest(req *sim.Request) bool {
 	if i.sim.WaitQ.Remove(req) {
 		i.sim.KVCache.ReleaseKVBlocks(req)
+		// Release the LoRA adapter pin (#1466): a queued request is not pinned, so
+		// this is a no-op here, but calling it keeps eviction symmetric with the
+		// running-batch branch below and robust if pinning semantics change.
+		i.sim.ReleaseAdapterPin(req)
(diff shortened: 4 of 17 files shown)


atantawi:
## Self-review report (convergence-review, 4 rounds)

Ran a multi-perspective convergence review (haiku 10-agent rounds across substance / code-quality / tests / DES / vLLM / distributed / performance / security / getting-started, plus two synchronous higher-tier consolidated passes). **9 fixes applied; converged.** Highlights:

- **R1 (4 IMPORTANT):** unchecked `Store()` result in `completeAdapterLoad`; asymmetric `residentAdapters`/`adapterCost` wiring; missing same-adapter coalescing test; NaN/Inf coefficient guard.
- **R2 (2 IMPORTANT):** `completeAdapterLoad` now preserves INV-8 on the (unreachable) `Store`-fail path; overflow finite-guard on cost coefficients.
- **R3 (2):** finite-but-huge `LoadLatency` could still saturate `int64` → `now+loadTicks` negative (INV-3) — added `maxLoadLatencyUs` bound; `LoRAConfig.Validate` now requires cost coefficients when adapters+capacity>0 (CLI catches the gap instead of `NewSimulator`).
- **R4 (1 CRITICAL):** gateway **in-flight eviction leaked the adapter pin** — `EvictRequest` removed a running (pinned) request without releasing its pin, and `releaseAdapterPin` was unexported, so under `--in-flight-eviction` + cluster + LoRA the resident slot stayed falsely pinned forever → `EvictLRU` stuck → **permanent silent cold-load stall**. Fixed via exported `ReleaseAdapterPin` called from `EvictRequest`, with regression test `TestInstanceSimulator_EvictRequest_ReleasesAdapterPin` (verified to fail without the fix). Confirmed `EvictRequest` is the only live-instance running-request removal path (autoscaler drain discards the instance → harmless; preemption keeps the pin by design).

Rejected-with-rationale (not defects): the `NewAdapterCostFunc` DI-var (matches the established `NewResidentAdapterSetFunc` registration pattern), head-of-line blocking (documented DT-faithful modeling choice, design §7/D1), `StepOverheadFactor` per-call allocation (not on any hot path until PR4), user-facing CLI docs/examples (deferred to the epic's polish PR, T046/PR8), and config↔constructor dual validation (intentional R6 library-boundary defense).

`go build ./...`, `go test ./...` (all packages), and `golangci-lint run ./...` are clean; the INV-6 no-op byte-identity golden holds.

atantawi:
### ✅ Rebased onto merged `main` — ready for review

PR1 (#1472) and PR2 (#1473) have merged. Rebased this branch with `git rebase --onto upstream/main <old-PR2-tip>`, dropping the 5 stale PR1+PR2 commits it carried. PR3 is now a **single commit** `78fe5d01` on top of `main` — the diff is PR3 only (17 files, +1076/−56).

**Two conflicts resolved (both from PR2's review-round hardening overlapping PR3):**
1. `sim/simulator.go` `NewSimulator` wiring — PR3 wires the cost model alongside the resident set; PR2's later rounds added a factory nil-return guard and two silent-misconfig warning branches. Kept **both**: the cost-model build + `s.adapterCost = ac` now sit inside PR2's nil-guarded block, and the "sim/lora not linked" warning branch now checks either factory being unregistered.
2. `sim/simulator.go` `recordAdapterResidency` / `releaseAdapterPin` — PR2's old inline `Store` cold-load block is superseded by PR3's gate (the load now happens in `completeAdapterLoad`), so `recordAdapterResidency` is Touch+Pin only and `releaseAdapterPin` does the Unpin. Took PR3's structure.

**One rebase-integration fix (folded into the commit):** the PR2-added `TestResidentAdapterSet_PreemptionDoesNotDoubleCountLoad` (new on `main`, unseen by PR3's original commit) now supplies cost coefficients, which PR3 requires when adapters+capacity are configured. Its invariant is unchanged — a preemption still fires and LoadCount stays 1 with `"a"` resident across the reschedule.

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` all clean locally. CI running.

atantawi:
@claude /blis-pr-review

atantawi:
### Review response — K6 validation gap closed (`daf82dfb`)

Addressed the one Important finding; the two minor observations resolved as noted below.

**1. K6 validation inconsistency (Important) — fixed.** Confirmed: `LoRAConfig.Validate` treats `step_overhead_tiers[*].k6` as optional (it only validates K6 when non-nil, but requires K7 as the divisor), while `NewCostModel` rejected a nil K6 — so a config with `k7` set and `k6` omitted passed the CLI gate then failed at `NewSimulator` with a confusing `adapter cost model:` constructor error. Took the suggested direction (default nil K6 → 0, not tighten `Validate`), because `Validate`'s K7-required / K6-optional asymmetry already encodes the intended semantic: K7 is the structural divisor, K6 is a numerator coefficient that legitimately defaults to 0 (→ `StepOverheadFactor` stays 1.0, no overhead). `NewCostModel` now honors that; K6 is still checked finite/≥0 when present. Added `TestNewCostModel_NilK6DefaultsToZero` for the previously-untested nil-K6 tier (builds cleanly, factor 1.0 even with a distinct adapter in the batch).

**2. Phantom load on gated-request timeout (minor) — no action.** Agreed with the review's own conclusion: this is the documented §7 "commit eviction victim at load-start" approximation; INV-L3 (charged once) and INV-8 (no spinning) both hold. No correctness impact.

**3. `scheduleNextStep` comment (minor) — clarified.** The comment already stated "nothing else can be admitted (blocking model)", which is the HOL implication, but I made it explicit that this includes warm-adapter and base-model requests queued behind the gated head (intentional, design §7/D1).

`go build ./...`, `go test ./...` (all packages), and `golangci-lint run ./...` are clean.

atantawi:
@claude /blis-pr-review

atantawi:
### Review response (round 2) — two minor observations (`6181d361`)

Verdict was READY TO MERGE; both observations non-blocking. Doc-only touches, no behavior change.

**1. `Store()` internal double-eviction path (theoretical) — clarifying comment added.** Confirmed unreachable in PR3: the eviction + `AdapterEvictionCounts` increment happen at load-start (`maybeStartAdapterLoad` → `EvictLRU`, `simulator.go:882`), and under the blocking model the set is at capacity−1 when `completeAdapterLoad` calls `Store`, so `Store`'s internal `EvictLRU` never fires. Added a comment at the call site making explicit that the discarded `evicted` return is always `""` here and that any future non-gate path calling `Store` at capacity must account for the eviction — exactly the future-maintainer note flagged.

**2. `tierForRank` tie-break — no change.** As the review notes, this is correct: the strict `d < bestDist` check keeps the first (lowest, since `tierRanks` is sorted) tier on a distance tie, matching the documented contract. The comment already states it accurately.

**Also:** fixed the stale doc comment on `TestColdLoadGate_Determinism` (it opened with a nonexistent `TestColdLoadGate_INV3_LoadCompletionInFuture` name).

`go build ./...`, `go vet ./sim/...`, and the gate tests are clean.

vishakha-ramani:
@claude /blis-pr-review

vishakha-ramani (approved):

Issue #1480, opened 2026-07-17
PR-C: session-pool replay (--concurrent-sessions / --total-sessions) for agentic traces


namasl:
**Part of #1477 — PR-C.** Structurally independent of PR-A (#1478) and PR-B (#1479) — touches the pool driver + CLI wiring, not the trace format — so it can be reviewed in parallel. For the complete agentic-replay story you want PR-A's accumulate fidelity alongside this, but neither blocks the other's build or tests.

## What external behavior are we replicating?

**System:** inference-perf (Kubernetes-SIGs). **Behavior:** the **fixed pool of N concurrent sessions** load model, with duplicate-to-fill scaling. inference-perf keeps at most `concurrent_sessions` sessions active; when one completes, the next is admitted. When the corpus has fewer sessions than the target, it duplicates them (with a cache-busting marker so clones don't share KV).

This PR ports that load model to BLIS as `blis replay --concurrent-sessions N --total-sessions L`: it adds **session-level** closed-loop on top of BLIS's existing **round-level** closed-loop.

**Vocabulary** (used throughout): a **round** is one LLM call (one prompt→response). A **session** is one full replay of a captured trace — its rounds run serially with the recorded think-time gap between them. The **corpus** is the M distinct traces supplied. `--concurrent-sessions N` = how many sessions run at once; `--total-sessions L` = how many session-replays complete in total (corpus duplicated to fill when L > M).

## Reference implementation (MANDATORY)

- `concurrent_sessions` semantics ("at most N active; when one completes, next is activated"): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/config/loadgen/config.py#L88-L95
- Pool admission gate (`should_start_next_session`) + dispatch: https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/loadgen/load_generator.py#L496-L514
- Duplicate-to-fill (clones with cache-busting): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/replay_graph_session_datagen.py#L1021-L1024
[...]

Issue #1479, opened 2026-07-17
PR-B: session→TraceV2 encoder core + blis convert otel (OTel agentic trace JSON)


namasl:
**Part of #1477 — PR-B.** ⚠️ **Depends on PR-A (#1478).** This PR writes the `session_context_growth` header field that PR-A adds to the `TraceHeader` struct; it cannot merge until PR-A lands, or the traces it produces will fail BLIS's strict-YAML header load.

## What external behavior are we replicating?

**System:** inference-perf (Kubernetes-SIGs). **Behavior:** converting an OpenTelemetry agent trace into a replayable per-call schedule. inference-perf's `otel_trace_to_replay_graph.py` reads OTel spans, keeps LLM-call spans, orders them, and derives per-event token counts and inter-call wait times. This PR ports the *ingestion + per-call token/timing extraction* into a BLIS `convert` subcommand that emits a TraceV2 file (BLIS's native replay format), rather than inference-perf's graph JSON.

## Reference implementation (MANDATORY)

- LLM-span filtering — a span is an LLM call if `name` starts with `chat ` or attributes contain `gen_ai.input.messages`; error spans (`status.code == 2`) are dropped unless `include_errors`: https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/otel_trace_to_replay_graph.py#L288-L299
- ISO-8601 timestamp parsing (must handle tz-aware and tz-naive): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/otel_trace_to_replay_graph.py#L84-L89
- Raw-call extraction (token counts from `gen_ai.usage.*_tokens`, start/end times, model): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/otel_trace_to_replay_graph.py#L365-L438
- `wait_ms` = gap between last predecessor end and this call's start, clamped ≥ 0 (BLIS carries this as inter-round arrival deltas): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/otel_trace_to_replay_graph.py#L1089-L1095
[...]

Issue #1477, opened 2026-07-17
Tracking: replay captured agentic traces (Exgentic/OTel + Weka) in BLIS


namasl:
## Summary

Add support for replaying captured **agentic LLM traces** through the BLIS discrete-event simulator, from **two source families**:

1. **Exgentic / OpenTelemetry** — OTel span JSON (e.g. the Exgentic `agent-llm-traces` dataset).
2. **SemiAnalysis Weka CC traces** — real Claude Code proxy traffic in the bespoke `WekaTrace` JSONL schema ([`cc-traces-weka-062126`](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-062126), v7, 393 traces; [`cc-traces-weka-with-subagents-051926`](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-with-subagents-051926), v5, 219 traces).

The goal:

> *If N users ran these captured agent sessions against **this** serving config, what would the inference engine see?*

Each captured session is a **linear chain of LLM calls**: call N+1's prompt is call N's prompt, plus call N's generated output, plus new content (a user/tool message whose size comes from the trace). BLIS re-simulates each call's inference on the serving config under test, preserves the recorded think/tool gap between calls, and runs a **fixed pool of N such sessions in parallel** — independent sessions, each internally serial.

**Architecture — one spine, pluggable front-ends.** The format coupling lives entirely in the converter. Every downstream stage (accumulate fidelity, session pool, observe pool) consumes BLIS's native **TraceV2** format and is source-agnostic. So "support another trace family" = *add one more `blis convert` front-end* over a shared session→TraceV2 encoder — the same multi-front-end pattern `blis convert` already uses for `preset`/`servegen`/`inference-perf`. Both families normalize to the same per-round `{input_tokens (delta), output_tokens, think_time, arrival}` before they ever reach the engine.

## Motivation

BLIS today can replay traces captured by `blis observe` (`blis replay --session-mode closed-loop`), but there is no path to ingest **externally captured agentic traces** (multi-turn tool-using agent sessions). These are the most realistic available model of production agent workloads: large, growing prompts with heavy KV-cache reuse and long think/tool gaps between calls.

- **Exgentic/OTel** gives broad benchmark coverage (SWE-bench, AppWorld, TAU-Bench-style harnesses).
[...]

PR #1484, opened 2026-07-21
feat(workload): session_context_growth header drives accumulate replay (PR-A of #1477)

### LINKED ISSUE #1478: PR-A: closed-loop replay fidelity (session_context_growth + recorded think time) for agentic trace replay
**Part of #1477 — PR-A.** This is the first PR; it unblocks PR-B (#1479) and PR-F (#1604).

PR-A carries **two closed-loop replay fidelity fixes**: (1) the strictly-growing shared prefix (`session_context_growth`, detailed under "Proposed change" below); and (2) a first-class optional **`think_time_us` trace column** so a converter can supply **recorded** per-round think time instead of the lossy start-to-start derivation. Fix (2) is what lets Weka (#1604) replay its explicitly-recorded think time correctly — see "Second fidelity fix" below.

**System:** inference-perf (Kubernetes-SIGs) agentic trace replay. **Behavior:** faithful reproduction of the strictly-growing shared prefix within an agent session. In a captured agent session, each LLM call re-sends the entire prior conversation plus new content, so call N+1's prompt begins with call N's prompt verbatim and appends to it. inference-perf preserves this at replay time by substituting each predecessor's real output back into the successor's prompt and reconstructing the shared prefix (`shared`/`output`/`unique` input segments).

This PR gives BLIS the equivalent for its **token-count-driven** model: a session where round N+1's synthetic input is byte-identical to round N's input for the first `len(input_N)` tokens, then strictly appended — so the block-hash prefix cache produces the same reuse a real growing conversation would.

inference-perf reconstructs the shared prefix and injects predecessor output via input segments:

- Input-segment types (`shared` = KV-reusable prefix, `output` = predecessor output injected, `unique` = new): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/replay_graph_types.py#L46-L64
- Segment substitution at replay (shared prefix reconstructed from predecessor's recorded messages; output segment replaced with live generated text): https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a27cd2231a8770b3acd21dc93b6bd4028cb/inference_perf/datagen/replay_graph_session_datagen.py#L436-L634

BLIS already has the mechanism to model this — the accumulate session buffer, which stores one growable token slice per session and hands each round a prefix-sharing sub-slice:

- Accumulate buffer layout + per-round slice (`inputTokens = sess.buf.Slice(0, inputEnd)` — same backing array, always from index 0): https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/session.go#L197-L263
- `SessionBlueprint.ContextGrowth` field (`"accumulate"` or `""`): https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/session.go#L29-L47
- Closed-loop blueprint loader (currently builds **non-accumulate** blueprints, so replayed sessions get fresh independent inputs and zero cross-round KV reuse): https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/replay.go#L131-L282
- Strict-YAML header parsing (why a new header field must be a declared struct field): https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/tracev2.go#L253

## Problem this PR solves

`blis replay --session-mode closed-loop` builds **non-accumulate** session blueprints: each round gets a *fresh* random input of the recorded size, so there is **zero cross-round KV reuse**. For agentic traces — whose entire point is a large, growing, re-sent prefix — this makes prefix-cache hit rate and KV-memory pressure badly wrong. This PR adds an opt-in that makes closed-loop replay model the growing prefix exactly.

## Proposed change

1. Add an optional field to `TraceHeader` (`sim/workload/tracev2.go`):
   ```go
   // SessionContextGrowth controls per-round input assembly in closed-loop replay.
   // "accumulate" → each session reuses one growing token buffer so round N+1's
   // input is byte-identical to round N's plus new tokens (strict prefix reuse for
   // KV-cache modeling). Empty (default) = per-round-independent inputs (existing
   // behavior). Set by `blis convert otel`.
   SessionContextGrowth string `yaml:"session_context_growth,omitempty"`
   ```
   `omitempty` + zero-value default means existing traces (which lack the field) load unchanged under `KnownFields(true)`. No trace-version bump needed — trace `Version` stays `3`; unlike `GoodputSLOTargets` (#1409, which bumped by convention), no bump is required here because `Version` is not gated on load and the field is additive-optional.

2. In `LoadTraceV2SessionBlueprints` (`sim/workload/replay.go`), set `ContextGrowth: trace.Header.SessionContextGrowth` on each built `SessionBlueprint`. Everything downstream (the accumulate buffer in `SessionManager.OnComplete`) already exists and is exercised by the reasoning-workload path.

## Second fidelity fix: first-class recorded think time (`think_time_us`)

**Problem.** Closed-loop replay schedules round N+1 at `completion_sim[N] + think`. Today `LoadTraceV2SessionBlueprints` derives `think` from **start-to-start** arrival gaps (`ArrivalTimeUs[i] − ArrivalTimeUs[i-1]`), which the loader's own doc comment ([`replay.go:124-128`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/replay.go#L124-L128)) flags as equal to `service_time[i-1] + client_think` — **not pure think**. The OTel converter cannot do better because the response-complete time (`end_time`) was a fake `start+1ms` placeholder. Applied on top of the *simulated* completion, this **double-counts inference** in the inter-round spacing. Weka (#1604) records the pure gap explicitly, so we add a channel to carry it.

**Change.** Add an optional `ThinkTimeUs int64` to `TraceRecord` (`sim/workload/tracev2.go` — the struct has no think field today), serialized as a **conditional trailing CSV column** using the format's existing optional-column machinery: the writer conditionally appends `vllm_priority`/`slo_target_us`/`x_request_id` after the fixed `traceV2Columns` ([`tracev2.go:167-182`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/tracev2.go#L167-L182)) and the reader detects optional columns from the header row (`FieldsPerRecord = -1`, [`tracev2.go:266-327`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/tracev2.go#L266-L327)). Add `think_time_us` the same way, so a trace without it is byte-identical to a pre-column trace. In `LoadTraceV2SessionBlueprints`, when a session's records carry `think_time_us`, build its think `SequenceSampler` from that column ([`replay.go:208-219`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/replay.go#L208-L219) is the current gap-derivation to branch around) instead of from arrival gaps. Precedence:

```
--think-time-dist (CLI sampler)  >  recorded think_time_us  >  arrival-gap derivation (today)
```

Absent column ⇒ today's arrival-gap path, byte-identical (INV-6). The OTel converter (#1479) leaves it empty; the Weka converter (#1604) populates it. This stays within PR-A's two files (`tracev2.go`, `replay.go`).

### Precondition: accumulate interprets `input_tokens` as a per-round **delta**

This is the load-bearing contract for the whole feature and must be stated explicitly. In accumulate mode the closed-loop loader builds each round's `InputSampler` from the record's `input_tokens` via `SequenceSampler` ([`replay.go:273`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/replay.go#L273); `SequenceSampler.Sample` returns recorded values by index, ignoring the RNG — [`distribution.go:207`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/distribution.go#L207)), and the accumulate buffer **appends** that many freshly-generated tokens each round ([`session.go:262`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/session.go#L262), `buf.Append(newInputTokens)`). So `input_tokens` is treated as *the new tokens this round adds*, **not** the cumulative prompt size.

This is correct only for **delta-encoded** traces — exactly what PR-B (`blis convert otel`, #1479) emits: `input_tokens_N = totalInput_N − totalInput_{N-1} − output_{N-1}` (tracked in #1477). Applying `session_context_growth: accumulate` to a **totals-encoded** trace (e.g. one produced by `blis observe`, whose per-round `input_tokens` are cumulative prompt sizes) would make the buffer grow as the running *sum of totals* — badly wrong input sizes. PR-A does not gate against this at load time (the guard is a documentation contract, not a runtime check): `session_context_growth: accumulate` is meaningful only for delta-encoded corpora written by the converter.

## Key behaviors to preserve

- [ ] Within a session, round N+1's input tokens `[0:len(input_N)]` are **element-wise identical** to round N's full input tokens (strict growing prefix). (reference: replay_graph_session_datagen.py#L436-L634; BLIS mechanism: session.go#L197-L263)
- [ ] Across two different sessions, prefixes are entirely disjoint (each session has its own per-session RNG).
- [ ] Header field absent / empty ⇒ behavior **byte-for-byte identical** to today's closed-loop replay (existing `blis observe` traces unaffected).
- [ ] `think_time_us` column absent ⇒ per-round think **byte-identical** to today's arrival-gap derivation; column present ⇒ round N+1 arrives at `completion_sim[N] + think_time_us`; CLI `--think-time-dist` overrides both.
- [ ] **Run/replay field source (cross-path parity):** `session_context_growth` is a **replay-only** trace-header field, consumed solely by `LoadTraceV2SessionBlueprints` (`replay.go`). `blis run` already supports accumulate via the separate *spec-level* `context_growth` field ([`spec.go:218`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/spec.go#L218), plumbed at [`generator.go:459`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/generator.go#L459)) — it does **not** read the trace header. Do not wire the header field into `blis run`, nor the spec field into replay. `blis observe` writes trace headers (`observe_cmd.go`) but never sets this field, so its traces decode to `""` (`omitempty`) and stay on the existing path.
- [ ] **`think_time_us` column — replay-consumed only (cross-path parity):** the new column is read solely by `LoadTraceV2SessionBlueprints` (`replay.go`). `blis run` uses its own spec-level think time (`MultiTurnSpec.ThinkTimeUs`) and neither emits nor reads this column; `blis observe` writes traces but does **not** populate it in this PR (the OTel converter #1479 leaves it empty; the Weka converter #1604 populates it). So `run` and `observe` are unaffected, and a trace without the column is byte-identical to today.

- BLIS synthesizes token IDs; it never sends real message text. So it models the *identical-and-growing* prefix cleanly (not the "near-identical" reality where interleaved tool-result text differs). This is deliberate and more faithful to what a KV cache sees.
- KV-cache hits are **block-granular** (BLIS's `PrefixCacheIndex` chains block hashes, like vLLM), so round N+1's modeled hit covers `floor(len(input_N)/block_size)` blocks, not a token-exact prefix. Faithful to real engines.

## Target parity version

- [x] Workload generation (`sim/workload/`)
- [ ] Core simulator (`sim/`)
- [ ] CLI (`cmd/`)

## Extension friction check

- **Files changed:** 2 (`sim/workload/tracev2.go`, `sim/workload/replay.go`) + tests. No new interface — extends the existing `TraceHeader` (+ one optional `TraceRecord` column) and reuses the existing accumulate buffer and think `SequenceSampler`.
- **Invariants:** INV-6 (determinism) preserved — per-session RNG seeded from master seed in fixed order; `think_time_us` is a conditional trailing CSV column (omitted when unset), so existing traces are byte-identical. INV-13 (run/replay parity) preserved — default values keep the existing path unchanged.

- [ ] `session_context_growth: accumulate` in the header ⇒ blueprints built with `ContextGrowth: "accumulate"`.
- [ ] Absent field ⇒ `ContextGrowth: ""` (regression: existing closed-loop behavior byte-identical).
- [ ] **Invariant test (strict prefix identity + absolute length):** drive a 2-round session through `SessionManager.OnComplete`; assert **both**:
  - (a) round 1's first `len(input_0)` tokens equal round 0's input tokens exactly (strict growing prefix); and
  - (b) round 1's **total** input length equals the reconstructed cumulative total `len(input_0) + actual_output_0 + delta_1`.

  The absolute-length assertion (b) guards against append-accounting drift (e.g. a token region seeded twice, or an off-by-`len(prefix)` in the round-0 seed) that a prefix-identity-only check would silently pass. (The round-0 seed at [`session.go:214-215`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/session.go#L214-L215) deliberately strips `len(bp.Prefix)` tokens from round-0's input before appending, because [`replay.go:228-232`](https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/sim/workload/replay.go#L228-L232) prepends the prefix into round-0 `InputTokens`; assertion (b) is what pins that accounting.)
- [ ] **Think-time column round-trip:** a `TraceRecord` with `think_time_us` set exports and re-loads with the value intact; a record without it exports **no** trailing column and re-loads byte-identical to a pre-column trace.
- [ ] **Think-time precedence:** with a `think_time_us` column present, the built think `SequenceSampler` uses those values (not arrival gaps); passing `--think-time-dist` overrides the column; neither present ⇒ arrival-gap derivation unchanged.
- [ ] Full `go test ./sim/workload/...` + clean `go build`.

Cross-repo port; tracked by #1477. Blocks PR-B (`blis convert otel`, #1479), which writes the `session_context_growth` header field, and PR-F (`blis convert weka`, #1604), which populates the `think_time_us` column added here.

namasl:
> **Draft — part of the OTel agentic trace replay stack (#1477). PR-A of 3.**

Closes #1478.

## What this adds

A `session_context_growth` field on `TraceHeader`. When set to `accumulate`, closed-loop replay routes each session's per-round input through the existing accumulate token buffer, so round N+1's input is **byte-identical** to round N's plus the newly-added tokens (a strictly-growing shared prefix → realistic KV-cache reuse). Default (empty) preserves today's behavior exactly.

- `SessionContextGrowth string` with `yaml:"session_context_growth,omitempty"` — absent field decodes to `""`, so existing traces are unaffected (strict-YAML safe, no version bump).
- `LoadTraceV2SessionBlueprints` propagates the header value to each blueprint's `ContextGrowth`.

## Invariants

- **INV-13 (run/replay parity):** field is replay-only and additive; `blis run` uses the separate spec-level `context_growth`. Default path byte-for-byte unchanged.
- **INV-6 (determinism):** no new randomness.

## Stack

This is the base of a 3-PR stack (all based on `main`):

1. **PR-A (this)** — accumulate replay fidelity → #1478
2. **PR-B** — `blis convert otel` → #1479
3. **PR-C** — `--concurrent-sessions` / `--total-sessions` pool replay → #1480

## Tests

`go test ./...` green; the load-bearing check is the strict-prefix-identity test (round 1's first N tokens == round 0's input).

### DIFF
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -185,6 +185,9 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 	var requests []*sim.Request
 	var blueprints []SessionBlueprint
 
+	// Growth mode from header (design §5): "accumulate" → strict growing prefix.
+	contextGrowth := trace.Header.SessionContextGrowth
+
 	for _, sessionID := range sessionOrder {
 		sr := sessionMap[sessionID]
 		rounds := sr.records
@@ -268,6 +271,7 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 			SessionID:        sessionID,
 			ClientID:         r0.ClientID,
 			MaxRounds:        len(rounds),
+			ContextGrowth:    contextGrowth,
 			ThinkTimeSampler: sessionThinkTimeSampler,
 			Horizon:          horizon,
 			InputSampler:     &SequenceSampler{values: inputSeq[1:]},  // rounds 1..N
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -118,6 +118,14 @@ func LoadTraceV2Requests(trace *TraceV2, seed int64) ([]*sim.Request, error) {
 // carries a recorded think_time_us (#1478). Round 0 has no predecessor, so its think
 // time is meaningless and ignored. When true, closed-loop replay prefers the recorded
 // per-round think time over arrival-gap derivation.
+//
+// LOSSY SENTINEL (F1, #1484 review): a value of 0 means "not recorded", so a session
+// whose every round has a genuinely-zero recorded think time (real for Weka's
+// overlap-clamped rounds) reads as false here and falls back to the arrival-gap path
+// — which bundles service time into the gap (see the gap-derivation note below), the
+// very effect think_time_us exists to avoid. The impact is bounded (a ~0 recorded
+// think replaced by the recorded arrival gap); a non-lossy encoding is deferred to the
+// Weka converter (#1604). Pinned by TestLoadTraceV2SessionBlueprints_AllZeroRecordedThink_FallsBackToGap.
 func sessionHasRecordedThinkTime(rounds []TraceRecord) bool {
 	for i := 1; i < len(rounds); i++ {
 		if rounds[i].ThinkTimeUs != 0 {
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -200,7 +200,13 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 	var blueprints []SessionBlueprint
 
 	// Growth mode from header (design §5): "accumulate" → strict growing prefix.
+	// Validate up front: an unrecognized value (e.g. a typo like "Accumulate") would
+	// otherwise fall through to the non-accumulate branch silently, disabling the
+	// feature with no feedback — an operator footgun. Fail loudly instead (R1).
 	contextGrowth := trace.Header.SessionContextGrowth
+	if contextGrowth != "" && contextGrowth != "accumulate" {
+		return nil, nil, fmt.Errorf("session_context_growth: unknown value %q (valid: \"accumulate\" or empty)", contextGrowth)
+	}
 
 	for _, sessionID := range sessionOrder {
 		sr := sessionMap[sessionID]
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -114,6 +114,19 @@ func LoadTraceV2Requests(trace *TraceV2, seed int64) ([]*sim.Request, error) {
 	return requests, nil
 }
 
+// sessionHasRecordedThinkTime reports whether any non-round-0 record in a session
+// carries a recorded think_time_us (#1478). Round 0 has no predecessor, so its think
+// time is meaningless and ignored. When true, closed-loop replay prefers the recorded
+// per-round think time over arrival-gap derivation.
+func sessionHasRecordedThinkTime(rounds []TraceRecord) bool {
+	for i := 1; i < len(rounds); i++ {
+		if rounds[i].ThinkTimeUs != 0 {
+			return true
+		}
+	}
+	return false
+}
+
 // LoadTraceV2SessionBlueprints groups trace records by session and builds
 // SessionBlueprints with SequenceSamplers for deterministic token replay.
 // Returns round-0 requests (plus all non-session requests) for initial injection,
@@ -204,11 +217,29 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 			outputSeq[i] = rec.OutputTokens
 		}
 
-		// Build think time: use provided sampler or derive from inter-round arrival gaps.
+		// Build think time. Precedence (#1478):
+		//   1. caller-provided sampler (CLI --think-time-dist / --think-time-ms)
+		//   2. recorded per-round think_time_us column (set by agentic-trace converters):
+		//      pure client think time, decoupled from inference — preferred over the
+		//      arrival-gap derivation, which bundles service time into the gap
+		//   3. inter-round arrival gaps (existing default; NOT pure client think time)
 		var sessionThinkTimeSampler LengthSampler
-		if thinkTimeSampler != nil {
+		switch {
+		case thinkTimeSampler != nil:
 			sessionThinkTimeSampler = thinkTimeSampler // stateless: safe to share across sessions
-		} else if len(rounds) > 1 {
+		case sessionHasRecordedThinkTime(rounds):
+			// think_time_us on round i is the recorded gap BEFORE round i (from round
+			// i-1's end). Round 0 carries no think; rounds 1..N supply the sequence.
+			thinkTimes := make([]int, len(rounds)-1)
+			for i := 1; i < len(rounds); i++ {
+				t := rounds[i].ThinkTimeUs
+				if t < 0 {
+					t = 0 // defensive; LoadTraceV2 already rejects negatives (INV-3)
+				}
+				thinkTimes[i-1] = int(t)
+			}
+			sessionThinkTimeSampler = &SequenceSampler{values: thinkTimes}
+		case len(rounds) > 1:
 			thinkTimes := make([]int, len(rounds)-1)
 			for i := 1; i < len(rounds); i++ {
 				gap := rounds[i].ArrivalTimeUs - rounds[i-1].ArrivalTimeUs
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -186,6 +186,9 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 	var requests []*sim.Request
 	var blueprints []SessionBlueprint
 
+	// Growth mode from header (design §5): "accumulate" → strict growing prefix.
+	contextGrowth := trace.Header.SessionContextGrowth
+
 	for _, sessionID := range sessionOrder {
 		sr := sessionMap[sessionID]
 		rounds := sr.records
@@ -270,6 +273,7 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 			SessionID:        sessionID,
 			ClientID:         r0.ClientID,
 			MaxRounds:        len(rounds),
+			ContextGrowth:    contextGrowth,
 			ThinkTimeSampler: sessionThinkTimeSampler,
 			Horizon:          horizon,
 			InputSampler:     &SequenceSampler{values: inputSeq[1:]},  // rounds 1..N
--- a/sim/workload/replay_test.go
+++ b/sim/workload/replay_test.go
@@ -3,6 +3,8 @@ package workload
 import (
 	"path/filepath"
 	"testing"
+
+	"github.com/inference-sim/inference-sim/sim"
 )
 
 func TestLoadTraceV2Requests_CorrectTokenCounts(t *testing.T) {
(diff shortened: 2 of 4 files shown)


namasl:

namasl:

susiejojo:
## BLIS PR Self-Review (correctness · observe-replay-calibrate parity · Weka suitability · Archon)

**PR #1484** — `feat(workload): session_context_growth header drives accumulate replay (PR-A of #1477)` @ `fee061f5`

Verified locally: `go build ./...` clean, `go test ./sim/workload/...` green on the PR HEAD. This review focuses on the four dimensions requested (correctness, run/replay/observe parity, Weka-corpus suitability, Archon consistency) and adds findings **beyond** the two prior self-reviews rather than repeating them.

---

### Verdict: **APPROVE (ready to merge) — with 1 important + 3 minor follow-ups, all deferrable**

The two prior self-reviews correctly resolved the unknown-value validation gap. The core feature is correct, minimal, and byte-identical on the default path. The follow-ups below are **not merge-blockers for PR-A's stated scope**, but two of them (F1, F2) are load-bearing for the downstream Weka work (#1604) and should be tracked before that PR lands.

---

### Correctness & invariants ✅ (with one caveat)

- **INV-6 / INV-13**: Both features are additive, replay-only, and default to byte-identity. `session_context_growth` absent → validation passes on `""` → `ContextGrowth=""` → accumulate branch (`session.go:199`) never entered. `think_time_us` absent → `thinkTimeUsIdx=-1` → `sessionHasRecordedThinkTime` false → arrival-gap path unchanged. `includeThinkTime` gate omits the column entirely when no record carries it. Confirmed the accumulate machinery itself is **pre-existing (#1445)** — this PR only wires the header field into it, so blast radius is small. ✅
- **INV-3**: Negative `think_time_us` rejected at parse (`tracev2.go`), with a belt-and-suspenders clamp in the loader. ✅
- **CSV min-column accounting**: `minCols` correctly increments for `thinkTimeUsIdx >= 0` (`tracev2.go:457-459`), and the parser resolves the column by absolute header position, so the vllm_priority/slo_target mid-row shifts don't corrupt the trailing index math. ✅

**F1 (IMPORTANT) — the `0` think-time value is an overloaded sentinel; silent fallback for legitimately-zero think times.**
`sessionHasRecordedThinkTime` (`replay.go:121-130`) treats `ThinkTimeUs == 0` as "not recorded." But a session with genuinely rapid back-to-back turns (real think time ≈ 0) records `0` for every round — indistinguishable from an absent column — so the `switch` at `replay.go:236` drops to the **arrival-gap** branch, which bundles service time (per the loader's own doc comment) on top of the *simulated* completion. That is exactly the lossy double-counting this feature exists to eliminate. This is directly relevant to Weka (see F2): Weka clamps `think_time` to `≥ 0` on response overlap, so all-zero sessions are a real, expected input class. Recommend: (a) document "`0` = not recorded" as a lossy sentinel on both the `TraceRecord.ThinkTimeUs` field and the loader, and (b) pin the actual behavior with a test. Not blocking for PR-A, but should be resolved before the Weka converter (#1604) relies on this channel.

---

### Weka-corpus suitability (`cc-traces-weka-with-subagents-051926`, v5, 219 traces)

I checked the dataset schema against what PR-A provides. **PR-A is a correct, necessary enabler but is not by itself sufficient to replay Weka** — and that is by design (the actual ingestion is PR-F/#1604). Key facts:

- ✅ **`think_time` channel**: Weka records a per-request `think_time` (pure post-response gap, `t[N] − t[N-1] − api_time[N-1]`, clamped ≥0). PR-A's new `think_time_us` column + 3-tier precedence is precisely the vessel for it — this is the "second fidelity fix" the tracking issue calls out as *the reason* for this PR. Correct fit. ✅
- ⚠️ **F2 (SUITABILITY CONTRACT, doc-only, no runtime guard)**: accumulate mode interprets `input_tokens` as a **per-round delta** (the buffer *appends* that many tokens each round — `session.go:262`). Weka's `in` field is a **cumulative** o200k block-quantized count, **not** a delta. Feeding raw Weka `in` values into `session_context_growth: accumulate` would grow the buffer as a running *sum of totals* — badly wrong input sizes. The PR (correctly) documents this as a contract and pushes delta computation (`in_N − in_{N-1} − out_{N-1}`) to the converter (#1604). But there is **no load-time guard** distinguishing delta- from totals-encoded traces, so a `blis observe`-produced trace (cumulative) + a hand-set `accumulate` header would silently corrupt. Acceptable for PR-A scope; flag loudly for #1604.
- ✅ **Token-unit caveat**: Weka `in`/`hash_ids` use o200k (~1.67× Anthropic); `out` is Anthropic-reported. This is a converter concern (#1604), out of PR-A scope. Correctly deferred.
- ✅ **Out-of-scope items handled cleanly**: Weka `hash_ids` (BLIS derives block hashes internally; sessions prefix-disjoint by design) and `type:"subagent"` fan-out (linear-chain only) are both explicitly deferred (PR-E). No PR-A surface touches them. ✅

**Net**: PR-A is corpus-agnostic and correct for its scope; it unblocks Weka rather than ingesting it. F1 (zero-think sentinel) is the one PR-A-level correctness item that Weka will actually exercise.

---

### Cross-path parity (run / replay / observe) ✅

- `session_context_growth`: replay-only trace-header field, consumed solely by `LoadTraceV2SessionBlueprints`. `blis run` uses the separate **spec-level** `context_growth` (`spec.go:220` → `generator.go:460`); `blis observe` never sets the header field (decodes to `""`). Asymmetry is intentional and documented. ✅
- `think_time_us`: read only by the replay closed-loop path. `RequestsToTraceRecords` (the run/observe export) does **not** populate it, so re-exported traces omit the column and stay byte-identical. Verified no wiring leak into `blis run`/`observe`. ✅
- **Calibrate path**: unaffected — neither field feeds the observe→replay→calibrate metric comparison. The eventual value (Weka's real `api_time`/`ttft` as a calibration target) is a PR-D concern, not here. ✅

---

### Test quality — solid happy-path coverage; behavioral corners missing

Strong: divergent-value precedence test (recorded 300/400 vs gaps 5000/7000), loud unknown-value error with message assertion, header YAML round-trip (present+absent), three-way trailing-column coexistence, negative-value rejection via corrupt-file round-trip (real parse path). All law-tests, not structural. ✅

Gaps (all MINOR, deferrable):
- **F3**: `TestAccumulateReplay_StrictPrefixIdentity` pins the prefix `[0:100]` and total length (150) but leaves the interior segment `[100:150]` (round-0 actual output + delta) unverified — precisely the `ProgressIndex`-derived accounting at `session.go:242-264`. Also untested: the **length-capped** accumulate path (`actualOutputLen < len(OutputTokens)` truncation) and **3+ round transitivity** (round 2's prefix == round 1's full input as the buffer keeps growing).
- **F4**: No test for **mixed-session** traces (session A carries think times, session B all-zero) — the export gate is global (`includeThinkTime`) but dispatch is per-session (`sessionHasRecordedThinkTime(rounds)`), so heterogeneous selection is unexercised. This compounds F1.
- **F5 (cosmetic, previously flagged, still open)**: the `parseTraceRecord` docstring (`tracev2.go:472-476`) still lists only `x_request_id` and `adapter` — `think_time_us` / `thinkTimeUsIdx` not mentioned. The last self-review noted this; it wasn't picked up.

---

### Archon consistency ✅

Both Archon passes returned `NO_CHANGE` (internal-only, fast-track eligible, 2 guarded promises within the existing boundary — the `SessionContextGrowth` default-empty invariant and the `ThinkTimeUs` additive-schema promise). Consistent with this analysis; both guarded promises now carry explicit round-trip tests.

---

### Preemption safety / timeout consistency ✅ — N/A

No new per-request metrics, state transitions, aggregate counters, or deadline/timeout logic. `ContextGrowth` and `ThinkTimeUs` affect only request *assembly* upstream of the DES. Preemption re-run cannot double-count anything introduced here.

---

**Summary of follow-ups** (none block PR-A merge):
1. **F1 (important)** — document the `0`-means-not-recorded sentinel and pin all-zero-think behavior; resolve before #1604 leans on the column.
2. **F2 (track for #1604)** — the delta-vs-totals accumulate contract has no runtime guard; the Weka converter must compute deltas and should fail loudly on totals-encoded input.
3. **F3/F4 (minor)** — interior-accumulation, length-capped, 3+ round, and mixed-session tests.
4. **F5 (cosmetic)** — update the `parseTraceRecord` docstring.

namasl:

susiejojo (approved):
lgtm

PR #1485, opened 2026-07-21
feat(cmd): blis convert otel — OTel agentic trace JSON to TraceV2 (PR-B of #1477)

### LINKED ISSUE #1479: PR-B: session→TraceV2 encoder core + blis convert otel (OTel agentic trace JSON)

namasl:
> **Draft — part of the OTel agentic trace replay stack (#1477). PR-B of 3.**

Closes #1479.

## What this adds

`blis convert otel` — converts OpenTelemetry agentic trace JSON (one agent session per trace; file, directory, or `.jsonl`) into a TraceV2 pair. One CSV row per LLM chat call; per-round input token counts stored as **deltas** (`max(0, in_{N+1} − in_N − out_N)`) so the accumulate buffer (PR-A) reconstructs each round's exact recorded total.

- `sim/workload/otel_convert.go` — pure converter (`ConvertOTelTrace`), LLM-span filtering, tz-aware/naive timestamps, error-span drop, think-time cap, min-rounds skip.
- `cmd/convert_otel.go` — the cobra subcommand, global request-ID assignment across sessions, deterministic corpus ordering.
- `TraceRecord.Model` is left **empty** by design: it is routing-significant (`buildRouterState` filters instances by it), so writing a recorded cross-model name would drop every request at routing. All calls simulate under `--model`.

## ⚠️ Reviewer note — stacked on PR-A

Based on `main`, but the diff **includes PR-A's commit** (`session_context_growth` header) because that field must exist first for strict-YAML load of the emitted corpus. **New in this PR:** the 3 `convert otel` commits (`3c93b511`, `92624677`, `b9891267`, `54d44194`). Review PR-A (#1478) separately; merge it first.

Emitted corpora **must** be replayed with `--session-mode closed-loop` (delta encoding requires the accumulate buffer) — PR-C's `--concurrent-sessions` auto-promotes to it.

## Stack

1. PR-A — accumulate replay fidelity → #1478
2. **PR-B (this)** — `blis convert otel` → #1479
3. PR-C — pool replay → #1480

## Tests

`go test ./...` green; delta-reconstruction law verified against a real Exgentic trace; independent-growth + skip paths covered.

### DIFF
--- a/cmd/convert_otel.go
+++ b/cmd/convert_otel.go
@@ -1,0 +1,177 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/inference-sim/inference-sim/sim/workload"
+)
+
+var (
+	otelInputPath     string
+	otelTraceOutput   string
+	otelContextGrowth string
+	otelMaxThinkTime  time.Duration
+	otelIncludeErrors bool
+	otelMinRounds     int
+)
+
+var convertOtelCmd = &cobra.Command{
+	Use:   "otel",
+	Short: "Convert OTel agentic trace JSON to a TraceV2 file for closed-loop replay",
+	Long: `Convert OpenTelemetry agentic trace JSON (one agent session per trace) into a
+TraceV2 pair (<prefix>.yaml + <prefix>.csv) suitable for 'blis replay --concurrent-sessions'.
+
+Input may be a single JSON file ({"spans":[...]}), a directory of such files, or a
+.jsonl file (one trace per line). Each trace becomes one session; each LLM chat span
+becomes one round. Per-round input token counts are stored as deltas so accumulate
+replay reconstructs the exact growing prompt with a strictly-identical shared prefix.
+
+Note: unlike other 'blis convert' subcommands (which emit a WorkloadSpec to stdout),
+'convert otel' writes TraceV2 files, because it targets 'blis replay' (fixed per-call
+token counts) rather than 'blis run' (distribution sampling).`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if otelContextGrowth != "accumulate" && otelContextGrowth != "independent" {
+			logrus.Fatalf("--context-growth must be \"accumulate\" or \"independent\", got %q", otelContextGrowth)
+		}
+		if otelMinRounds < 1 {
+			logrus.Fatalf("--min-rounds must be >= 1, got %d", otelMinRounds)
+		}
+		if otelMaxThinkTime < 0 {
+			logrus.Fatalf("--max-think-time must be >= 0, got %s", otelMaxThinkTime)
+		}
+		opts := workload.OTelConvertOptions{
+			ContextGrowth:  otelContextGrowth,
+			MaxThinkTimeUs: otelMaxThinkTime.Microseconds(),
+			IncludeErrors:  otelIncludeErrors,
+			MinRounds:      otelMinRounds,
+		}
+		if err := runConvertOtel(otelInputPath, otelTraceOutput, opts); err != nil {
+			logrus.Fatalf("OTel conversion failed: %v", err)
+		}
+	},
+}
+
+// collectTraceInputs returns the raw bytes of each trace found at path. A
+// directory yields one entry per *.json file; a *.jsonl file yields one entry
+// per non-empty line; any other file is treated as a single trace.
+func collectTraceInputs(path string) ([][]byte, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if info.IsDir() {
+		entries, err := os.ReadDir(path)
+		if err != nil {
+			return nil, err
+		}
+		var names []string
+		for _, e := range entries {
+			if !e.IsDir() && strings.HasSuffix(e.Name(), ".json") {
+				names = append(names, e.Name())
+			}
+		}
+		sort.Strings(names) // deterministic corpus order (INV-6)
+		var out [][]byte
+		for _, n := range names {
+			b, err := os.ReadFile(filepath.Join(path, n))
+			if err != nil {
+				return nil, err
+			}
+			out = append(out, b)
+		}
+		return out, nil
+	}
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasSuffix(path, ".jsonl") {
+		var out [][]byte
+		for _, line := range strings.Split(string(b), "\n") {
+			line = strings.TrimSpace(line)
+			if line != "" {
+				out = append(out, []byte(line))
+			}
+		}
+		return out, nil
+	}
+	return [][]byte{b}, nil
+}
+
+// runConvertOtel reads all traces at inputPath, converts each to session
+// records, assigns global request IDs, and writes a TraceV2 pair to
+// <outPrefix>.yaml / <outPrefix>.csv.
+func runConvertOtel(inputPath, outPrefix string, opts workload.OTelConvertOptions) error {
+	if inputPath == "" {
+		logrus.Fatalf("--input is required")
+	}
+	if outPrefix == "" {
+		logrus.Fatalf("--trace-output is required")
+	}
+	inputs, err := collectTraceInputs(inputPath)
+	if err != nil {
+		return err
+	}
+
+	var allRecords []workload.TraceRecord
+	nextID := 0
+	skipped := 0
+	for _, raw := range inputs {
+		recs, err := workload.ConvertOTelTrace(raw, opts)
+		if err != nil {
+			logrus.Warnf("skipping unparseable trace: %v", err)
+			skipped++
+			continue
+		}
+		if recs == nil {
+			skipped++
+			continue // below MinRounds
+		}
+		for i := range recs {
+			recs[i].RequestID = nextID
+			nextID++
+		}
+		allRecords = append(allRecords, recs...)
+	}
+
+	if len(allRecords) == 0 {
+		logrus.Fatalf("no usable sessions found in %q (skipped %d)", inputPath, skipped)
+	}
+
+	growth := opts.ContextGrowth
+	if growth == "independent" {
+		growth = "" // empty header value = per-round-independent inputs
+	}
+	header := &workload.TraceHeader{
+		Version:              3,
+		TimeUnit:             "microseconds",
+		Mode:                 "generated",
+		SessionContextGrowth: growth,
+	}
+	if err := workload.ExportTraceV2(header, allRecords, outPrefix+".yaml", outPrefix+".csv"); err != nil {
+		return err
+	}
+	logrus.Infof("Wrote %d records from %d sessions to %s.{yaml,csv} (skipped %d)",
+		len(allRecords), len(inputs)-skipped, outPrefix, skipped)
+	return nil
+}
+
+func init() {
+	convertOtelCmd.Flags().StringVar(&otelInputPath, "input", "", "Path to OTel trace JSON file, directory of *.json, or *.jsonl (required)")
+	convertOtelCmd.Flags().StringVar(&otelTraceOutput, "trace-output", "", "Output TraceV2 prefix; writes <prefix>.yaml + <prefix>.csv (required)")
+	convertOtelCmd.Flags().StringVar(&otelContextGrowth, "context-growth", "accumulate", "Prefix model: \"accumulate\" (strict growing shared prefix) or \"independent\"")
+	convertOtelCmd.Flags().DurationVar(&otelMaxThinkTime, "max-think-time", 15*time.Second, "Cap on per-round inter-call gap (think time); 0 = no cap")
+	convertOtelCmd.Flags().BoolVar(&otelIncludeErrors, "include-errors", false, "Include spans with error status (status.code == 2)")
+	convertOtelCmd.Flags().IntVar(&otelMinRounds, "min-rounds", 1, "Skip sessions with fewer than N usable LLM calls")
+	_ = convertOtelCmd.MarkFlagRequired("input")
+	_ = convertOtelCmd.MarkFlagRequired("trace-output")
+
+	convertCmd.AddCommand(convertOtelCmd)
+}
--- a/cmd/convert_otel.go
+++ b/cmd/convert_otel.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
@@ -58,10 +59,19 @@ token counts) rather than 'blis run' (distribution sampling).`,
 	},
 }
 
-// collectTraceInputs returns the raw bytes of each trace found at path. A
-// directory yields one entry per *.json file; a *.jsonl file yields one entry
-// per non-empty line; any other file is treated as a single trace.
-func collectTraceInputs(path string) ([][]byte, error) {
+// traceInput pairs a raw trace payload with a human-readable source label, so
+// skip/parse-error diagnostics can name the offending file (and line, for
+// .jsonl inputs) instead of failing silently.
+type traceInput struct {
+	name string
+	raw  []byte
+}
+
+// collectTraceInputs returns each trace found at path, paired with a source
+// label. A directory yields one entry per *.json file (name = filename); a
+// *.jsonl file yields one entry per non-empty line (name = "<path>:line<N>");
+// any other file is treated as a single trace (name = path).
+func collectTraceInputs(path string) ([]traceInput, error) {
 	info, err := os.Stat(path)
 	if err != nil {
 		return nil, err
@@ -78,13 +88,13 @@ func collectTraceInputs(path string) ([][]byte, error) {
 			}
 		}
 		sort.Strings(names) // deterministic corpus order (INV-6)
-		var out [][]byte
+		var out []traceInput
 		for _, n := range names {
 			b, err := os.ReadFile(filepath.Join(path, n))
 			if err != nil {
 				return nil, err
 			}
-			out = append(out, b)
+			out = append(out, traceInput{name: n, raw: b})
 		}
 		return out, nil
 	}
@@ -93,16 +103,16 @@ func collectTraceInputs(path string) ([][]byte, error) {
 		return nil, err
 	}
 	if strings.HasSuffix(path, ".jsonl") {
-		var out [][]byte
-		for _, line := range strings.Split(string(b), "\n") {
+		var out []traceInput
+		for i, line := range strings.Split(string(b), "\n") {
 			line = strings.TrimSpace(line)
 			if line != "" {
-				out = append(out, []byte(line))
+				out = append(out, traceInput{name: fmt.Sprintf("%s:line%d", path, i+1), raw: []byte(line)})
 			}
 		}
 		return out, nil
 	}
-	return [][]byte{b}, nil
+	return []traceInput{{name: path, raw: b}}, nil
 }
 
 // runConvertOtel reads all traces at inputPath, converts each to session
@@ -123,14 +133,15 @@ func runConvertOtel(inputPath, outPrefix string, opts workload.OTelConvertOption
 	var allRecords []workload.TraceRecord
 	nextID := 0
 	skipped := 0
-	for _, raw := range inputs {
-		recs, err := workload.ConvertOTelTrace(raw, opts)
+	for _, in := range inputs {
+		recs, err := workload.ConvertOTelTrace(in.raw, opts)
 		if err != nil {
-			logrus.Warnf("skipping unparseable trace: %v", err)
+			logrus.Warnf("skipping unparseable trace %s: %v", in.name, err)
 			skipped++
 			continue
 		}
 		if recs == nil {
+			logrus.Debugf("skipping trace %s: below --min-rounds", in.name)
 			skipped++
 			continue // below MinRounds
 		}
--- a/cmd/convert_otel.go
+++ b/cmd/convert_otel.go
@@ -119,11 +119,15 @@ func collectTraceInputs(path string) ([]traceInput, error) {
 // records, assigns global request IDs, and writes a TraceV2 pair to
 // <outPrefix>.yaml / <outPrefix>.csv.
 func runConvertOtel(inputPath, outPrefix string, opts workload.OTelConvertOptions) error {
+	// Return errors (do not Fatalf): the cobra Run wrapper owns the CLI exit
+	// boundary and Fatalf's on a returned error (R6 — library-style helpers
+	// surface errors, the CLI layer terminates). --input/--trace-output are also
+	// MarkFlagRequired'd, so these guards are defense-in-depth.
 	if inputPath == "" {
-		logrus.Fatalf("--input is required")
+		return fmt.Errorf("--input is required")
 	}
 	if outPrefix == "" {
-		logrus.Fatalf("--trace-output is required")
+		return fmt.Errorf("--trace-output is required")
 	}
 	inputs, err := collectTraceInputs(inputPath)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


namasl:

namasl:

susiejojo:
## BLIS PR Self-Review (correctness · code quality · parent-issue relevance · run/replay/observe parity · Weka convertibility · Archon)

**PR #1485** — `feat(cmd): blis convert otel — OTel agentic trace JSON to TraceV2 (PR-B of #1477)` @ `230777cf`

Verified locally: `go build ./...` clean; `go test ./sim/workload/ ./cmd/` green. I read all 6 changed files, built the binary, and exercised the converter + replay end-to-end.

---

### Verdict: **APPROVE (ready to merge)** — one important cross-command footgun (F1) worth a follow-up guard; rest are test-hardening.

The PR faithfully ports inference-perf's OTel ingestion, the delta-encoding + shared-encoder factoring matches issue #1479's mandated design exactly, and routing-safety (empty `Model`) is correctly enforced and tested. Nothing blocks merge for PR-B's scope.

---

### ⚠️ Direct answer to the Weka question (`cc-traces-weka-with-subagents-051926`)

**No — this PR cannot convert those Weka traces, and is not intended to.** I verified empirically: pointing `blis convert otel` at a Weka-format JSONL line produces `skipping unparseable trace …: no session_id or trace_id found` → `no usable sessions found` (exit 1). This is **correct, graceful behavior**, not a bug:

- Weka's schema is a session-per-line `{id, models[], block_size, hash_id_scope, requests[]}` with per-request `{t, in, out, hash_ids, api_time, think_time, ttft}` and `type:"subagent"` groups. The OTel reader only understands `{"spans":[{...gen_ai.usage.*...}]}`, so a Weka line deserializes to an empty span list and is skipped.
- **Weka support is PR-F (#1604), a separate stacked PR that does not exist yet.** This PR's real contribution toward Weka is the **shared encoder seam** (`NormalizedRound` + `EncodeSessionToTraceRecords` in `trace_encode.go`): PR-F will add a Weka JSONL *reader* over this same encoder, with **no engine or encoder changes**. That factoring is present, clean, and source-agnostic — the encoder takes absolute per-round input and computes deltas itself, so no reader duplicates the delta law. ✅
- Two Weka-specific concerns are already correctly anticipated in code comments but **out of scope here**: (a) `think_time` → the `ThinkTimeUs` column (encoder writes it when a reader sets it; OTel leaves it 0); (b) the lossy-zero-sentinel for `think_time` and the o200k/×64 cumulative-token caveat, both deferred to #1604. When reviewing PR-F, the load-bearing checks will be: Weka `in` is **cumulative o200k**, so the reader must pass it as `InputTokensAbs` (the encoder deltas it) and must NOT itself pre-delta; and `type:"subagent"` groups must be skipped (linear-chain only, per #1477).

**Bottom line:** PR-B is a necessary enabler for Weka but not sufficient; the correct way to test Weka convertibility is against PR-F once it lands.

---

### Correctness & invariants ✅

- **Delta law**: round 0 = full input; round N+1 = `max(0, in_{N+1} − in_N − out_N)`. Verified end-to-end (in=1000→1200, out=50 ⇒ delta 150). The monotone reconstruction law is asserted behaviorally by re-summing deltas+outputs against recorded totals (refactor-survivable). Non-monotone clamp-to-0 tested. ✅
- **Routing safety (INV-13-adjacent)**: `TraceRecord.Model` left empty at both encoder and converter level, with explicit tests asserting it — a recorded cross-model name would zero out routable instances and silently drop every request. Correctly the single most-defended invariant. ✅
- **INV-6 determinism**: directory corpus sorted (`sort.Strings`) before global request-ID assignment; `.jsonl` processed in line order. ✅
- **Timestamps**: tz-aware (RFC3339) and tz-naive (→UTC) both handled; think-time (arrival gap) capped at `--max-think-time` (default 15s), matching the reference. ✅
- **Error handling boundaries (R6)**: `runConvertOtel` returns errors; the cobra `Run` owns the `Fatalf` exit boundary. One bad trace in a corpus warns + skips rather than aborting the batch — good resilience. Real exit code on total failure is 1 (verified). ✅

---

### F1 (IMPORTANT) — delta-encoded corpus is silently misread in the DEFAULT replay mode

`blis convert otel` writes `session_context_growth: accumulate` into the header, and the emitted `input_tokens` are **deltas** (small: 150, not 1200). These are only correct under `--session-mode closed-loop`, whose `LoadTraceV2SessionBlueprints` reads the accumulate header. But `blis replay` **defaults to `--session-mode fixed`**, which routes through `LoadTraceV2Requests` — a path that **never reads `SessionContextGrowth`** and treats each `input_tokens` as an absolute per-round size.

I confirmed empirically: replaying the accumulate-tagged corpus in default fixed mode runs to completion with **no warning**, silently misinterpreting delta 150 as a 150-token prompt. Issue #1479 explicitly flags this ("Consider whether a fixed-mode load of a `session_context_growth: accumulate` header should warn") — the guard was not added. Recommend: in `cmd/replay.go`, when `trace.Header.SessionContextGrowth == "accumulate"` and mode is `fixed`, emit a `logrus.Warnf` (or Fatalf) that the corpus is delta-encoded and requires closed-loop. Low-effort, prevents a silent-wrong-result footgun that every OTel/Weka corpus will be exposed to. Not a PR-B-internal correctness bug, so not merge-blocking, but should land before or alongside PR-C.

---

### Cross-path parity (run / replay / observe) ✅

`convert otel` is a trace **producer** targeting `blis replay --session-mode closed-loop`, not a run/observe feature — so classic per-flag parity doesn't apply. Correctly scoped:
- `blis run` uses distribution-sampled workloads + spec-level `context_growth`; it neither reads nor emits this corpus. ✅
- `blis observe` produces its own traces (same-model, cumulative inputs); it does not consume OTel corpora. The doc note that observe traces are totals-encoded (not delta) is why `accumulate` must never be hand-applied to an observe trace — consistent with PR-A's documented contract. ✅
- The one genuine cross-command coupling is F1 above (producer in `cmd/convert_otel.go`, consumer default in `cmd/replay.go`).

---

### Code quality & design ✅

- The two-layer split (source-agnostic `EncodeSessionToTraceRecords` + thin `otelSpan` reader) is exactly issue #1479's mandated factoring and is the cleanest possible seam for PR-F. Public surface (`NormalizedRound`, `EncodeSessionToTraceRecords`, `ConvertOTelTrace`, `OTelSessionID`, `OTelConvertOptions`) is minimal and purposeful.
- `collectTraceInputs` cleanly unifies file/dir/jsonl with source-labeled diagnostics (`<path>:line<N>`).
- Doc comments are unusually thorough and explain *why* (routing safety, think-time absence) rather than *what*.
- Deliberately does not use `DisallowUnknownFields` (real traces carry unmodeled attributes) — correct call, documented.

---

### Archon review

Archon returned **`ARCHITECTURAL_CHANGE`** (1 edge strengthened `cmd → sim/workload`, 5 new public symbols, 1 schema, 2 invariant/test promises). This is **expected and sensible**: the change deliberately introduces the shared-encoder public surface that PR-F reuses. No package boundary is violated (dependency direction `cmd → sim/workload` is correct per the layering rule), no interface-contract membership changed, and no cycle. The "architectural" flag here reflects intentional, well-motivated new surface — not a red flag. Consistent with my analysis. ✅

---

### Test quality — strong happy paths; several guard/error paths unexercised (all MINOR, hardening)

Solid: behavioral delta-law (re-sum, not field-peek), both `isLLMSpan` branches, tz-aware+naive, non-monotone clamp, end-to-end round-trip for single-file/jsonl/directory, global-ID assignment, Model-empty at both levels.

Gaps to harden (none blocking):
- **T1 (highest)**: the start-time **sort is never forced to reorder** — every test trace is already in start order, so a regression dropping the sort would pass the suite. Add an out-of-order-spans trace asserting `RoundIndex`/delta reflect time order. (Delta correctness depends on this sort.)
- **T2**: nil-token-count drop (`InputTokens==nil`) untested — the error span uses `0`, not an absent key.
- **T3**: `sessionIDFromSpans` "no id" error + its warn/skip propagation untested (only malformed-JSON skip is covered).
- **T4**: "no usable sessions found" hard-failure path untested.
- **T5**: unparseable-timestamp abort (whole-trace fail vs per-span drop asymmetry) untested.
- **T6**: the documented clamp **over-count** magnitude is asserted only as `delta==0`; pin the reconstructed absolute so a future clamp-semantics change is caught.
- **T7**: `OTelSessionID` (exported) has zero coverage.
- **T8**: `--max-think-time 0` (uncapped) branch not asserted; **T10**: directory-determinism test uses already-sorted filenames, so `sort.Strings` isn't proven to reorder.

---

### Preemption safety / timeout consistency ✅ — N/A

Pure offline trace converter; no DES execution, per-request metrics, state transitions, or timeout/deadline logic. Deltas affect request *assembly* at replay (upstream of the DES), not execution.

---

**Follow-ups (none block PR-B):**
1. **F1 (important)** — add a fixed-mode-vs-accumulate-header warning in `cmd/replay.go` (issue #1479 already flagged this); land before/with PR-C.
2. **T1 (test)** — force the start-time sort to reorder in a test.
3. **T2–T10 (test hardening)** — cover the untested guard/error paths above.
4. **Weka (#1604)** — this PR's encoder seam is ready; Weka convertibility is verifiable only once PR-F adds the reader.

namasl:
Re the F1 follow-up (default \`--session-mode fixed\` silently misreading an accumulate-delta corpus as absolute per-round inputs):

This is already handled by **PR-C (#1486)**, which adds a hard guard in \`cmd/replay.go\`:

\`\`\`go
if traceData.Header.SessionContextGrowth == "accumulate" && replaySessionMode != "closed-loop" {
    logrus.Fatalf("trace header has session_context_growth=accumulate (per-round input_tokens are deltas that only reconstruct correctly in closed-loop replay), but --session-mode is %q. Re-run with --session-mode closed-loop, or use --concurrent-sessions N for pooled replay.", replaySessionMode)
}
\`\`\`

That is a hard \`Fatalf\` (stronger than the suggested warn), and it lands **with PR-C** — matching the "before or alongside PR-C" recommendation. PR-B is the trace/encoder layer and intentionally does not touch the replay CLI, which is why the guard is not in this diff. I have added this guard to #1480 (PR-C's issue) as an explicit requirement so it survives the rebase.

The T1–T10 converter test-hardening items are noted as non-blocking; they will be handled separately.

namasl:

Context: the only change since the prior review (`230777cf` → `102570f3`) is **test-only** — it folds in the T1–T10 converter test-hardening items from the earlier follow-up (out-of-order-span sort coverage, absent-token-key drop, no-session-id error + corpus warn/skip propagation, unparseable-timestamp abort, non-monotone clamp reconstruction magnitude, `OTelSessionID` coverage, uncapped `--max-think-time 0`, and filename-sort ID determinism). No production code changed, so archon's `ARCHITECTURAL_CHANGE` is unchanged from before (PR-B's existing encoder surface).

susiejojo (approved):
lgtm

Issue #1483, opened 2026-07-21
session-pool replay: Unstarted() undercounts dropped sessions under an explicit --horizon hard cap


namasl:
## Summary

In session-pool replay (`blis replay --concurrent-sessions N`, added in PR-C of #1477 / #1480), the `SessionPoolDriver.Unstarted()` accounting can **undercount** dropped sessions when an explicit `--horizon` hard cap truncates the run mid-drain. The self-draining path (no `--horizon`) is exact; only the hard-cap edge case is affected.

## Detail

`SessionPoolDriver.OnComplete` admits the next queued session by pushing its round-0 request into the cluster and incrementing `totalStarted`. But the cluster's horizon guard silently discards any `ClusterArrivalEvent` whose timestamp exceeds `config.Horizon`. So when a user sets `--horizon` small enough to truncate the run:

1. A session completes near the cap.
2. `OnComplete` pushes the next refill and counts it as *started*.
3. The cluster discards that refill (past horizon) — it never actually arrives or runs.
4. `Unstarted() == totalCount - totalStarted` therefore over-counts `totalStarted`, so it **undercounts** the sessions that were effectively never admitted, and the "pool drained early" warning may not fire.

The invariant `totalStarted + Unstarted() == TotalSessions()` still holds arithmetically, but `totalStarted` includes phantom admissions, so the reported numbers understate the truncation.

## Why deferred

The common/default path — self-draining (`--horizon` unset, cluster horizon = `math.MaxInt64`) — admits every session exactly once and is exact. The undercount only occurs under an explicit hard cap that truncates, which is an opt-in escape hatch, not the primary mode. PR-C ships with a code comment documenting the limitation and this issue tracking the fix.

## Proposed fix (sketch)

Detect discarded refills so `totalStarted` reflects only admissions the cluster actually accepted. Options:
- Have the driver learn which pushed refills the cluster dropped (e.g. reconcile against completed/observed arrivals post-run), or
- Compute `Unstarted()` from `nextQueued` vs. actually-arrived sessions rather than the push-time counter, or
- Gate the increment on an arrival confirmation callback rather than counting at push time.

## Acceptance criteria

- [ ] Under `--concurrent-sessions N --total-sessions L --horizon <small>` that truncates mid-drain, the reported un-admitted count equals the true number of sessions the cluster never ran.
- [ ] Self-draining path (no `--horizon`) remains exact and unchanged.
- [ ] Regression test covering the hard-cap truncation accounting.

## Context

[...]

PR #1486, opened 2026-07-21
feat(replay): --concurrent-sessions / --total-sessions fixed-pool replay (PR-C of #1477)

### LINKED ISSUE #1480: PR-C: session-pool replay (--concurrent-sessions / --total-sessions) for agentic traces

namasl:
> **Draft — part of the OTel agentic trace replay stack (#1477). PR-C of 3.**

Closes #1480.

## What this adds

`blis replay --concurrent-sessions N --total-sessions L` — a fixed pool of N concurrent closed-loop sessions drawn from the trace corpus, duplicating the corpus (with cache-busting) to reach L total. Ports inference-perf's `concurrent_sessions` load model.

- `sim/workload/session_pool.go` — `SessionPoolDriver` wraps `SessionManager`: `BuildSessionPool` round-robin duplicate-to-fill (each clone gets a unique SessionID, unique round-0 request ID, a cache-busting prefix token, and its own cloned stateful samplers); `OnComplete` admits the next queued session when one terminates, keeping the pool at N.
- `cmd/replay.go` — the two flags (gated so default replay is byte-identical, INV-13), auto-promote to closed-loop, and the self-draining horizon (`math.MaxInt64` when `--horizon` unset, so the pool drains all L sessions rather than truncating at the 600 s auto-fallback). `--horizon` remains available as an optional hard cap.

## ⚠️ Reviewer note — top of the stack

Based on `main`, but the diff **includes PR-A + PR-B's commits**. **New in this PR:** the pool commits (`cae87644`, `df2a189e`, `d654b1b9`, `8308269f`, `7bb6e4ae`, `e893b722`, `b6b5509b`) plus a one-line `#1483` reference (`1460de5b`). Review/merge #1478 and #1479 first.

## Known limitation

Under an explicit `--horizon` hard cap that truncates mid-drain, `Unstarted()` can undercount dropped sessions (a refill pushed but discarded by the cluster horizon guard is still counted as started). The self-draining default is exact. Tracked in #1483; documented inline.

## Stack

1. PR-A — accumulate replay fidelity → #1478
2. PR-B — `blis convert otel` → #1479
3. **PR-C (this)** — pool replay → #1480

## Validation

`go test ./...` green. Real-data: Exgentic parquet → `convert otel` → pool replay, 216/216 sessions completed with duplicate-to-fill at ~97% KV cache hit. Self-draining multi-wave and INV-13 default-path parity both covered by tests.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -163,6 +163,9 @@ go build -o blis main.go
 
 # Replay a fixed pool of N concurrent closed-loop sessions from the corpus
 # (--total-sessions duplicates the corpus with cache-busting to fill the target)
+# Real agentic prompts can be huge (tens of thousands to >100K tokens) and grow
+# each round — raise --max-model-len (default ~41K) and scale --total-kv-blocks
+# accordingly, or every request drops as unservable (completed_requests: 0).
 ./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
   --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,6 +157,15 @@ go build -o blis main.go
 ./blis convert inference-perf --spec spec.yaml
 ./blis compose --from spec1.yaml --from spec2.yaml
 
+# Convert OTel agentic trace JSON to a TraceV2 corpus (accumulate = strict growing prefix)
+./blis convert otel --input otel_json/ --trace-output corpus \
+  --context-growth accumulate --max-think-time 15s
+
+# Replay a fixed pool of N concurrent closed-loop sessions from the corpus
+# (--total-sessions duplicates the corpus with cache-busting to fill the target)
+./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
+  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200
+
 # Run with gateway queue flow control (utilization-based saturation gating)
 ./blis run --model qwen/qwen3-14b --flow-control --saturation-detector utilization \
   --queue-depth-threshold 5 --kv-cache-util-threshold 0.8
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -207,6 +207,15 @@ go build -o blis main.go
 ./blis convert inference-perf --spec spec.yaml
 ./blis compose --from spec1.yaml --from spec2.yaml
 
+# Convert OTel agentic trace JSON to a TraceV2 corpus (accumulate = strict growing prefix)
+./blis convert otel --input otel_json/ --trace-output corpus \
+  --context-growth accumulate --max-think-time 15s
+
+# Replay a fixed pool of N concurrent closed-loop sessions from the corpus
+# (--total-sessions duplicates the corpus with cache-busting to fill the target)
+./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
+  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200
+
 # Run with gateway queue flow control (utilization-based saturation gating)
 ./blis run --model qwen/qwen3-14b --flow-control --saturation-detector utilization \
   --queue-depth-threshold 5 --kv-cache-util-threshold 0.8
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -213,6 +213,9 @@ go build -o blis main.go
 
 # Replay a fixed pool of N concurrent closed-loop sessions from the corpus
 # (--total-sessions duplicates the corpus with cache-busting to fill the target)
+# Real agentic prompts can be huge (tens of thousands to >100K tokens) and grow
+# each round — raise --max-model-len (default ~41K) and scale --total-kv-blocks
+# accordingly, or every request drops as unservable (completed_requests: 0).
 ./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
   --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200
 
--- a/cmd/convert_otel.go
+++ b/cmd/convert_otel.go
@@ -1,0 +1,177 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/inference-sim/inference-sim/sim/workload"
+)
+
+var (
+	otelInputPath     string
+	otelTraceOutput   string
+	otelContextGrowth string
+	otelMaxThinkTime  time.Duration
+	otelIncludeErrors bool
+	otelMinRounds     int
+)
+
+var convertOtelCmd = &cobra.Command{
+	Use:   "otel",
+	Short: "Convert OTel agentic trace JSON to a TraceV2 file for closed-loop replay",
+	Long: `Convert OpenTelemetry agentic trace JSON (one agent session per trace) into a
+TraceV2 pair (<prefix>.yaml + <prefix>.csv) suitable for 'blis replay --concurrent-sessions'.
+
+Input may be a single JSON file ({"spans":[...]}), a directory of such files, or a
+.jsonl file (one trace per line). Each trace becomes one session; each LLM chat span
+becomes one round. Per-round input token counts are stored as deltas so accumulate
+replay reconstructs the exact growing prompt with a strictly-identical shared prefix.
+
+Note: unlike other 'blis convert' subcommands (which emit a WorkloadSpec to stdout),
+'convert otel' writes TraceV2 files, because it targets 'blis replay' (fixed per-call
+token counts) rather than 'blis run' (distribution sampling).`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if otelContextGrowth != "accumulate" && otelContextGrowth != "independent" {
+			logrus.Fatalf("--context-growth must be \"accumulate\" or \"independent\", got %q", otelContextGrowth)
+		}
+		if otelMinRounds < 1 {
+			logrus.Fatalf("--min-rounds must be >= 1, got %d", otelMinRounds)
+		}
+		if otelMaxThinkTime < 0 {
+			logrus.Fatalf("--max-think-time must be >= 0, got %s", otelMaxThinkTime)
+		}
+		opts := workload.OTelConvertOptions{
+			ContextGrowth:  otelContextGrowth,
+			MaxThinkTimeUs: otelMaxThinkTime.Microseconds(),
+			IncludeErrors:  otelIncludeErrors,
+			MinRounds:      otelMinRounds,
+		}
+		if err := runConvertOtel(otelInputPath, otelTraceOutput, opts); err != nil {
+			logrus.Fatalf("OTel conversion failed: %v", err)
+		}
+	},
+}
+
+// collectTraceInputs returns the raw bytes of each trace found at path. A
+// directory yields one entry per *.json file; a *.jsonl file yields one entry
+// per non-empty line; any other file is treated as a single trace.
+func collectTraceInputs(path string) ([][]byte, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if info.IsDir() {
+		entries, err := os.ReadDir(path)
+		if err != nil {
+			return nil, err
+		}
+		var names []string
+		for _, e := range entries {
+			if !e.IsDir() && strings.HasSuffix(e.Name(), ".json") {
+				names = append(names, e.Name())
+			}
+		}
+		sort.Strings(names) // deterministic corpus order (INV-6)
+		var out [][]byte
+		for _, n := range names {
+			b, err := os.ReadFile(filepath.Join(path, n))
+			if err != nil {
+				return nil, err
+			}
+			out = append(out, b)
+		}
+		return out, nil
+	}
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasSuffix(path, ".jsonl") {
+		var out [][]byte
+		for _, line := range strings.Split(string(b), "\n") {
+			line = strings.TrimSpace(line)
+			if line != "" {
+				out = append(out, []byte(line))
+			}
+		}
+		return out, nil
+	}
+	return [][]byte{b}, nil
+}
+
+// runConvertOtel reads all traces at inputPath, converts each to session
+// records, assigns global request IDs, and writes a TraceV2 pair to
+// <outPrefix>.yaml / <outPrefix>.csv.
+func runConvertOtel(inputPath, outPrefix string, opts workload.OTelConvertOptions) error {
+	if inputPath == "" {
+		logrus.Fatalf("--input is required")
+	}
+	if outPrefix == "" {
+		logrus.Fatalf("--trace-output is required")
+	}
+	inputs, err := collectTraceInputs(inputPath)
+	if err != nil {
+		return err
+	}
+
+	var allRecords []workload.TraceRecord
+	nextID := 0
+	skipped := 0
+	for _, raw := range inputs {
+		recs, err := workload.ConvertOTelTrace(raw, opts)
+		if err != nil {
+			logrus.Warnf("skipping unparseable trace: %v", err)
+			skipped++
+			continue
+		}
+		if recs == nil {
+			skipped++
+			continue // below MinRounds
+		}
+		for i := range recs {
+			recs[i].RequestID = nextID
+			nextID++
+		}
+		allRecords = append(allRecords, recs...)
+	}
+
+	if len(allRecords) == 0 {
+		logrus.Fatalf("no usable sessions found in %q (skipped %d)", inputPath, skipped)
+	}
+
+	growth := opts.ContextGrowth
+	if growth == "independent" {
+		growth = "" // empty header value = per-round-independent inputs
+	}
+	header := &workload.TraceHeader{
+		Version:              3,
+		TimeUnit:             "microseconds",
+		Mode:                 "generated",
+		SessionContextGrowth: growth,
+	}
+	if err := workload.ExportTraceV2(header, allRecords, outPrefix+".yaml", outPrefix+".csv"); err != nil {
+		return err
+	}
+	logrus.Infof("Wrote %d records from %d sessions to %s.{yaml,csv} (skipped %d)",
+		len(allRecords), len(inputs)-skipped, outPrefix, skipped)
+	return nil
+}
+
+func init() {
+	convertOtelCmd.Flags().StringVar(&otelInputPath, "input", "", "Path to OTel trace JSON file, directory of *.json, or *.jsonl (required)")
+	convertOtelCmd.Flags().StringVar(&otelTraceOutput, "trace-output", "", "Output TraceV2 prefix; writes <prefix>.yaml + <prefix>.csv (required)")
+	convertOtelCmd.Flags().StringVar(&otelContextGrowth, "context-growth", "accumulate", "Prefix model: \"accumulate\" (strict growing shared prefix) or \"independent\"")
+	convertOtelCmd.Flags().DurationVar(&otelMaxThinkTime, "max-think-time", 15*time.Second, "Cap on per-round inter-call gap (think time); 0 = no cap")
+	convertOtelCmd.Flags().BoolVar(&otelIncludeErrors, "include-errors", false, "Include spans with error status (status.code == 2)")
+	convertOtelCmd.Flags().IntVar(&otelMinRounds, "min-rounds", 1, "Skip sessions with fewer than N usable LLM calls")
+	_ = convertOtelCmd.MarkFlagRequired("input")
+	_ = convertOtelCmd.MarkFlagRequired("trace-output")
+
+	convertCmd.AddCommand(convertOtelCmd)
+}
--- a/cmd/convert_otel.go
+++ b/cmd/convert_otel.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
@@ -58,10 +59,19 @@ token counts) rather than 'blis run' (distribution sampling).`,
 	},
 }
 
-// collectTraceInputs returns the raw bytes of each trace found at path. A
-// directory yields one entry per *.json file; a *.jsonl file yields one entry
-// per non-empty line; any other file is treated as a single trace.
-func collectTraceInputs(path string) ([][]byte, error) {
+// traceInput pairs a raw trace payload with a human-readable source label, so
+// skip/parse-error diagnostics can name the offending file (and line, for
+// .jsonl inputs) instead of failing silently.
+type traceInput struct {
+	name string
+	raw  []byte
+}
+
+// collectTraceInputs returns each trace found at path, paired with a source
+// label. A directory yields one entry per *.json file (name = filename); a
+// *.jsonl file yields one entry per non-empty line (name = "<path>:line<N>");
+// any other file is treated as a single trace (name = path).
+func collectTraceInputs(path string) ([]traceInput, error) {
 	info, err := os.Stat(path)
 	if err != nil {
 		return nil, err
@@ -78,13 +88,13 @@ func collectTraceInputs(path string) ([][]byte, error) {
 			}
 		}
 		sort.Strings(names) // deterministic corpus order (INV-6)
-		var out [][]byte
+		var out []traceInput
 		for _, n := range names {
 			b, err := os.ReadFile(filepath.Join(path, n))
 			if err != nil {
 				return nil, err
 			}
-			out = append(out, b)
+			out = append(out, traceInput{name: n, raw: b})
 		}
 		return out, nil
 	}
@@ -93,16 +103,16 @@ func collectTraceInputs(path string) ([][]byte, error) {
 		return nil, err
 	}
 	if strings.HasSuffix(path, ".jsonl") {
-		var out [][]byte
-		for _, line := range strings.Split(string(b), "\n") {
+		var out []traceInput
+		for i, line := range strings.Split(string(b), "\n") {
 			line = strings.TrimSpace(line)
 			if line != "" {
-				out = append(out, []byte(line))
+				out = append(out, traceInput{name: fmt.Sprintf("%s:line%d", path, i+1), raw: []byte(line)})
 			}
 		}
 		return out, nil
 	}
-	return [][]byte{b}, nil
+	return []traceInput{{name: path, raw: b}}, nil
 }
 
 // runConvertOtel reads all traces at inputPath, converts each to session
@@ -123,14 +133,15 @@ func runConvertOtel(inputPath, outPrefix string, opts workload.OTelConvertOption
 	var allRecords []workload.TraceRecord
 	nextID := 0
 	skipped := 0
-	for _, raw := range inputs {
-		recs, err := workload.ConvertOTelTrace(raw, opts)
+	for _, in := range inputs {
+		recs, err := workload.ConvertOTelTrace(in.raw, opts)
 		if err != nil {
-			logrus.Warnf("skipping unparseable trace: %v", err)
+			logrus.Warnf("skipping unparseable trace %s: %v", in.name, err)
 			skipped++
 			continue
 		}
 		if recs == nil {
+			logrus.Debugf("skipping trace %s: below --min-rounds", in.name)
 			skipped++
 			continue // below MinRounds
 		}
--- a/cmd/convert_otel_test.go
+++ b/cmd/convert_otel_test.go
@@ -51,7 +51,8 @@ func TestConvertOtel_DirectoryDeterministicIDs(t *testing.T) {
 	if err := os.MkdirAll(inDir, 0755); err != nil {
 		t.Fatal(err)
 	}
-	// Two single-call sessions in separate files; one below min-rounds session dropped.
+	// Two single-call sessions in separate files, both kept (MinRounds: 1);
+	// verifies deterministic global request-ID assignment in sorted-filename order.
 	write := func(name, tid string, in, out int) {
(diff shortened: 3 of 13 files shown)


namasl:

namasl:

namasl:

namasl:

Context for this review: PR-C has just been **rebased onto merged main** (`ffc7aaee`, which now includes PR-A #1484 and PR-B #1485). The 12 PR-C-unique commits (SessionPoolDriver, `--concurrent-sessions`/`--total-sessions`, the fixed-mode/accumulate replay guard, `--shuffle-corpus`, and the pool-refill deadline fix) replayed cleanly. One post-rebase fix: an earlier PR-C commit dropped `OTelSessionID` as unused, but merged PR-B keeps it and covers it with `TestOTelSessionID` (the T7 hardening item) — I restored `OTelSessionID` to match merged main exactly, so `sim/workload/otel_convert.go` now carries zero PR-C diff. Please focus on the pool/shuffle/guard integration against the current `cmd/replay.go` surface, INV-6/INV-11/INV-13, and the accumulate-header fixed-mode guard. Locally: `go build ./...`, `go test ./...`, and `golangci-lint` are all green.

namasl:

Context: this push (`a86de78a` → `1bf047c1`) addresses the two remaining SHOULD-FIX items from your round-4 review:
- **F4** (refactor-survival): removed the write-only `SessionPoolDriver.activeCount`/`totalTerm` fields (never read for control flow — the pool bound is maintained structurally by admitting ≤1 replacement per termination; `totalStarted` stays, read by `Unstarted()`), and rewrote `TestSessionPool_RefillAndConservation` to assert via observable outputs only (injected slice, per-completion `len(next) ≤ 1`, `Unstarted()`/`TotalSessions()`).
- **F5** (clear error): added a pool-mode pre-check in `cmd/replay.go` that `Fatalf`s with an actionable message when the corpus contains non-session records, pre-empting `BuildSessionPool`'s internal "count mismatch"; covered by the new subprocess test `TestReplayCmd_PoolRejectsNonSessionRecords`.

The `--shuffle-corpus` help-text wording you flagged as minor is left as-is. Locally `go build ./...`, `go test ./...`, and `golangci-lint run ./...` are all green on `1bf047c1`.

susiejojo:
### BLIS PR Self-Review — PR #1486 (independent pass: correctness · observe/replay/calibrate parity · Weka suitability · Archon)

**Reviewing:** `feat(replay): --concurrent-sessions / --total-sessions fixed-pool replay (PR-C of #1477)`
**Base:** `ffc7aaee` (merged main, includes PR-A #1484 + PR-B #1485) · **HEAD:** `1bf047c1`

Local gates on the checked-out branch: `go build ./...` ✅ · `go test ./sim/workload/ ./cmd/` ✅ · `go vet` ✅ · `gofmt -l` clean ✅.

This PR has already converged through 5 self-review rounds (F1–F5) plus repeated Archon passes. I re-verified each resolved finding against the current tree and did a fresh correctness/parity/Weka sweep. **No new blocking issues.** Verdict below, with three non-blocking observations.

---

#### Prior findings — all still resolved on `1bf047c1`

| # | Finding | Status |
|---|---------|--------|
| F1 | `OnComplete` rebased `ArrivalTime` but not `Deadline` → mass-cancel wave 2+ on deadline-bearing (run/observe) corpora | ✅ Fixed — `session_pool.go:183-186` shifts `Deadline` by the same offset; regression test `TestSessionPool_RefillRebasesDeadline` |
| F2 | Pool + `--trace-output` / `--results-path` silent truncation | ✅ Fixed — `--trace-output` now `Fatalf`s in pool mode; `--results-path` warns loudly (`cmd/replay.go:138-146`) |
| F3 | `Unstarted()` docstring didn't acknowledge pre-instance drops | ✅ Fixed — docstring + drain warning now name both causes |
| F4 | Write-only `activeCount`/`totalTerm` fields; test read internal struct | ✅ Fixed — fields removed; `TestSessionPool_RefillAndConservation` asserts via observable outputs only (refactor-survival) |
| F5 | Mixed session/non-session trace → opaque internal "count mismatch" | ✅ Fixed — actionable pre-check (`cmd/replay.go:135`) + subprocess test `TestReplayCmd_PoolRejectsNonSessionRecords` |

#### Correctness & invariants (independently verified)

- **INV-1 conservation:** refills from `OnComplete` are injected via `cluster.go:596 pushArrival`, the same path SessionManager follow-ups use, so `injectedByClass`/`pendingArrivals` increment correctly. Refills are counted, never leaked.
- **INV-11 completeness on the failure path:** the unservable-drop guards (`simulator.go:537/553/567/584`) all invoke `OnRequestDone` before returning, so a session whose oversized prompt is dropped still terminates and refills its pool slot. This is the exact failure mode the new docs warn about for large agentic prompts — good that it degrades cleanly rather than abandoning the slot.
- **INV-6 determinism:** clone RNGs seeded in fixed corpus order; shuffle drawn from a *distinct* XOR-salted stream (`seed^0x53485546`) so it doesn't perturb per-session token/clone RNGs; `TestReplayCmd_SessionPool_Deterministic` byte-compares two full runs. ✅
- **INV-13 run/replay parity:** pool mode is replay-only and fully gated (`replayConcurrentSessions > 0`); flags-unset replay is untouched. `math.MaxInt64` self-drain horizon is safe against the only arithmetic site (`cluster.go:1026 min(clock, Horizon)`); all other horizon uses are comparisons. ✅
- **Cache-busting fidelity:** the clone's cache-buster is prepended to its *actual* round-0 `InputTokens`, and the accumulate buffer seeds from that (`session.go:214`), so the growing shared prefix diverges from round 0 onward — the KV-reuse divergence is real, not just cosmetic on round 0.

Correctly scoped: pool replay is a **replay-only** load model (there is no synthetic-workload or real-server analogue on `run`). The observe-side twin is **explicitly tracked as PR-D (#1487, draft PR #1499)**, which the issue states reuses `SessionPoolDriver` verbatim. `--trace-output` is hard-failed and `--results-path` warned in pool mode, so no silent cross-artifact drift. Parity obligation satisfied.

#### Weka corpus suitability (the specific ask)

PR-C is the **source-agnostic replay engine**; it consumes TraceV2 and is indifferent to whether the corpus came from `blis convert otel` or `blis convert weka`. Assessment against `cc-traces-weka-with-subagents-051926`:

1. **Recorded `think_time`** (Weka's key differentiator) — carried by PR-A's `think_time_us` column with the documented 3-tier precedence; the pool driver leaves intra-session follow-ups to `SessionManager` unchanged, so recorded think times flow through. ✅
2. **Large ISL / growing prefix** — the `--max-model-len` / `--total-kv-blocks` sizing caveat is documented in a warning admonition, and the drop path is conservation-safe (above). ✅
3. **Subagent fan-out** — the linked corpus is the **with-subagents** variant, but the linear converter **skips `type:"subagent"`** and DAG session admission is deferred to **PR-E**. So today this corpus replays *only its main-agent streams*; subagent traffic is silently absent from the load. This is by design and documented, but it means the with-subagents corpus is **not yet fully represented** — worth stating plainly to anyone calibrating against it.
4. **Blocker on the datapath:** PR-C alone cannot ingest Weka — it needs **PR-F (#1604 `blis convert weka`)**. PR-C is suitable *and ready* as the engine; end-to-end Weka replay is gated on PR-F.

#### Non-blocking observations

1. **[docs] OTel-only framing.** The new guide section is titled "Replaying OTel agentic traces" and the CLAUDE.md snippet is OTel-specific, yet PR-C's whole value is being source-agnostic (issue #1477). A one-line note that the same `--concurrent-sessions` flow applies to Weka corpora (once #1604 lands) would prevent the reader from assuming pool replay is OTel-bound. Cheap, and it advertises the shared spine.
2. **[minor correctness] Initial-wave clones are not deadline-rebased.** When `--concurrent-sessions > len(corpus)`, the initial wave includes clones injected at the *source's* `ArrivalTime`/`Deadline` (`BuildSessionPool` injects `d.queued[i]` directly), unlike refills which rebase to `tick`. Harmless for OTel/Weka corpora (round-0 arrival 0, no deadline), but a run/observe-derived corpus with non-zero round-0 arrivals + tight deadlines could see initial-wave clones dropped on the enqueue guard. Consider rebasing the initial wave for symmetry with the refill path, or a one-line comment noting the assumption.
3. **[test gap] No pool-level conservation test over the drop-unservable path.** The docs call out `completed_requests: 0` / `dropped_unservable` as the likely Weka failure mode, and I verified the refill hook fires on that path — but there's no e2e test asserting `injected == completed + dropped + unstarted` when an oversized round drops. Given it's the headline real-world risk for this corpus, a small fixture (one round sized past `--max-model-len`, assert conservation + a refill) would lock the behavior.

#### Archon

Verdict `ARCHITECTURAL_CHANGE` (`1 surface · 2 invariant`) reflects the `cmd → sim/workload` surface strengthening already baked into merged main via PR-A/PR-B; PR-C adds no new package edge. No cycle, correct dependency direction (`cmd → sim/cluster → sim`). Consistent with the layering in principles.md.

---

### Verdict: **READY TO MERGE.**

Correctness, invariants (INV-1/6/11/13), and cross-path parity all hold; the feature is well-tested and the failure paths degrade cleanly. The three observations above are non-blocking polish — I'd suggest folding in #2 (initial-wave rebase note) and #3 (conservation-on-drop test) since they directly harden the Weka path this stack exists to serve, but neither should gate the merge. Full end-to-end Weka replay remains gated on PR-F (#1604) by design.

susiejojo:
### Note on `--total-sessions` sampling semantics (for corpus subsetting)

A clarification worth capturing for anyone subsetting a large corpus (e.g. the Weka `cc-traces-weka-*` sets): **`--total-sessions` on its own does deterministic, ordered selection — it is not a random sample.** To get a representative random subset you must pair it with `--shuffle-corpus`.

**Two regimes in `BuildSessionPool` (`sim/workload/session_pool.go:123-134`), both deterministic:**

- `--total-sessions ≤ corpus size` → **front-truncation**: replays the first `total` sessions *in corpus/file order* (`srcIdx = i`); the tail is never replayed.
- `--total-sessions > corpus size` → all originals in order, then **round-robin clones** (`i % len(blueprints)`) with cache-busting to fill the remainder.

Neither path draws randomly — so `--total-sessions 50` on a 200-session corpus silently replays only the **first 50 by file order**. For a time-sorted or harness-grouped corpus that is a biased sample, not a representative one.

**How to shuffle (seeded random subset):**

```bash
# Representative random 50-of-200 subset, reproducible from --seed:
  --model qwen/qwen3-14b \
  --concurrent-sessions 8 --total-sessions 50 \
  --shuffle-corpus --seed 42
```

`--shuffle-corpus` applies a seeded Fisher-Yates permutation (`ShuffleSessions`) to the corpus **before** `BuildSessionPool`, so:

- **`--shuffle-corpus` + `--total-sessions N < corpus`** → permute, then front-truncate → a **seeded-random subset of N distinct sessions**.
- **`--shuffle-corpus` + `--total-sessions ≥ corpus`** → every session still runs; only the pool's step/admission order is randomized.

The shuffle RNG is drawn from a **distinct stream** (`seed ^ 0x53485546`), separate from the per-session token and clone RNGs (`cmd/replay.go`). That is deliberate: shuffle order varies independently of *what tokens* each session emits, so two runs differing only in `--shuffle-corpus` stay comparable, and the "random" subset is fully reproducible from `--seed` (INV-6).

**Suggestion:** the current `--shuffle-corpus` help text ("With `--total-sessions < corpus` this yields a random subset; every session still runs otherwise.") captures this, but the docs guide's OTel example doesn't mention it. A one-line note in the guide — *"to subset a large corpus representatively, add `--shuffle-corpus`; `--total-sessions` alone takes the first N in file order"* — would steer calibration users away from accidentally measuring the head of the file.

namasl:
Thanks — good catch on the calibration footgun. Added a `!!! tip "Subsetting a large corpus: add --shuffle-corpus"` admonition to the OTel/pool section of `docs/guide/observe-replay-calibrate.md` (commit `681b8d2d`): it spells out that `--total-sessions N` alone is deterministic front-of-file selection (a biased sample for time-sorted/harness-grouped corpora) and that `--shuffle-corpus` turns it into a seeded-random, reproducible `N`-of-corpus subset. Left the `--shuffle-corpus` flag help text as-is per your note that it already captures the semantics.

namasl:

Context: the only change since the round-5 READY TO MERGE (`1bf047c1` → `681b8d2d`) is **documentation-only** — a `!!! tip` admonition in `docs/guide/observe-replay-calibrate.md` clarifying that `--total-sessions N` alone selects the first N sessions in file order and that `--shuffle-corpus` is needed for a representative seeded-random subset (addressing the corpus-subsetting note above). No code changed; archon's `ARCHITECTURAL_CHANGE` is the unchanged cumulative pool surface. `go build ./...`, `go test ./...`, and `golangci-lint run ./...` remain green.

susiejojo (approved):
lgtm. Same behavior in BLIS observe/run needs to be addressed in PR-D.

Issue #1491, opened 2026-07-22
feat(lora): B-3 — Eviction seam + lru default (Backend Swap A) (#1488)


atantawi:
Part of #1488. Stories: US2, US4. Backend Swap Phase A.

**Goal**: Extract an `EvictionPolicy` seam from the hardcoded LRU; register `lru` byte-identical default; introduce the eviction context wiring `AdapterRegistry` (rank) into the instance simulator (D2 — rank is unreachable at the eviction site today).

**Scope** (tasks.md Phase 5):
- Failing tests: `lru` default byte-identical (Phase-A gate); INV-L2/L5 pin-safety; INV-8 no-deadlock (all-pinned ⇒ waiting request runs once a pin clears).
- Define `EvictionPolicy` seam + registry in `sim/lora/eviction/`; route the single eviction call site (`maybeStartAdapterLoad`) through it.
- Eviction context exposing unpinned candidates + rank/reload-cost via the registry (not by widening the frozen `AdapterCost` — R13).

**Gate**: §5.4 Phase-A — all existing tests pass, factory returns `lru` by default. Own micro-plan.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1488, opened 2026-07-22
feat(lora): LoRA placement-policy seams (Part B) — epic


atantawi:
## Epic — LoRA placement-policy seams (Part B)

Makes LoRA adapter **placement policy** a first-class, selectable concern along three
independent, **byte-identical-by-default** seams, plus a `(trigger → policy)` taxonomy,
named strategy bundles, and run-level provenance. This is **Part B** — the BLIS-side Go
seams that the landed `lora-control` harness policy names (route-to-holder, pre-placement,
rank-aware evictor) resolve to (lora-control issue #12).

Builds directly on the merged LoRA control-plane subsystem (epic #1464): adapter identity &
registry, per-instance resident set + LRU/pin/capacity, cold-load gate, adapter cost model,
static HBM accounting, `lora-affinity` scorer, per-adapter metrics.

### Seams (each with a byte-identical no-op default — INV-6/INV-L1)
- **Routing** — which instance serves a request. New: `route-to-holder` (strict).
- **Eviction** — victim under capacity pressure. New: `rank/cost-aware`.
- **Creation** — start-of-run residency + admit-on-miss. New: `pre-placement` (t=0 seeding).

Plus: reactive `(trigger → policy)` taxonomy (implemented) + an inert scaffolded `Periodic`
trigger; named strategy bundles; run-level effective-policy-triple provenance.

### Classification
Subsystem Module (Creation) + Backend Swap (Eviction) + Policy Templates (Routing) —
design doc required, no-op default mandatory.

### Artifacts (branch `008-lora-policy-seams` on fork `atantawi/inference-sim`, based on the assembled LoRA stack)
- Design doc (**APPROVED**; convergence review PASSED — 6 rounds, 0 CRITICAL / 0 IMPORTANT, 8 sonnet perspectives): https://github.com/atantawi/inference-sim/blob/008-lora-policy-seams/docs/plans/2026-07-22-lora-policy-seams-design.md
- Spec / plan / tasks / contracts: https://github.com/atantawi/inference-sim/tree/008-lora-policy-seams/specs/008-lora-policy-seams

### New invariants
- **INV-PS1** strict-routing honesty · **INV-PS2** pre-placement conservation · **INV-PS3** periodic-trigger inertness. Preserves INV-1/6/8/9/13 + INV-L1..L7.

### PR sub-issues (Small-tier, each independently mergeable + no-op-safe)
[...]

PR #1496, opened 2026-07-23
feat(lora): B-1 — routing scorer switch → init()-registered registry (#1489)

atantawi:
## B-1: Routing scorer `switch` → `init()`-registered registry

First PR of the **LoRA placement-policy seams (Part B)** stack — epic #1488. Converts `sim/routing_scorers.go`'s hardcoded scorer `switch` into a name→constructor registry so adding a routing scorer/policy is a localized additive change. **Byte-identical by default** (INV-6/INV-L1): all 11 existing scorers are registered as defaults; `DefaultScorerConfigs()` and `newScorerWithObserver` behavior is unchanged.

Closes #1489.

### What changed
- **Registry** (`scorerRegistry`, unexported per R8): `map[string]scorerConstructor` populated by a single `init()` (the one canonical registration site, R4). `registerScorer` panics on empty/duplicate name.
- **Validity is now derived from the registry keys** — `IsValidScorer`/`ValidScorerNames` read the registry; the hand-maintained `validScorerNames` map is deleted (closes the validity/constructibility drift gap).
- `stateless(fn)` helper wraps the 8 stateless scorers as `(fn, nil)` constructors, exactly matching the pre-registry switch arms. Stateful/param-backed scorers (prefix-affinity, precise-prefix-cache, no-hit-lru) pass constructors preserving their exact `(scorer, observer)` pairing.
- `sortedScorerNames()` sorts keys explicitly before any output-feeding use (R2, INV-6).
- Docs: `extension-recipes.md` "Adding New Scorers" recipe updated to the registry (single `registerScorer` call; no separate name list).

### Tests (TDD — failing first, commit `e0503974`)
- `TestScorerRegistry_ReproducesSwitchSelection` — byte-identity guard across all 11 scorers vs an **independent per-name reference** (exact `assert.Equal`, not `InDelta`); observer non-nil only for {prefix-affinity, no-hit-lru}.
- `_UnknownPanics`, `_ValidityDerivedFromRegistry` (bidirectional set-equality), `_DefaultsRegistered`, `_RegisterAddsName` (isolated via `t.Cleanup`), `TestRegisterScorer_{Duplicate,Empty}NamePanics`, `TestParseScorerConfigs_InvalidNameErrorIsDeterministic`.

### Gate
- Code-review convergence (10 perspectives): **CONVERGED, 0 CRITICAL / 0 IMPORTANT** (Round 1 fixed 1 IMPORTANT — a stale extension-recipes recipe; Round 2 clean).
[...]

PR #1500, opened 2026-07-23
feat(lora): B-4 — add rank-aware LoRA eviction policy (#1492)

### LINKED ISSUE #1492: feat(lora): B-4 — rank/cost-aware eviction (Backend Swap B) (#1488)
Part of #1488. Stories: US2. Depends on B-3. Backend Swap Phase B.

**Goal**: Add `rank/cost-aware` eviction consuming the eviction context (victim criterion provisional, D2/§14).

**Scope** (tasks.md Phase 6):
- Failing rank-sensitivity test: scenario where LRU-order and rank/cost-order **provably disagree** (skewed popularity + differing ranks) ⇒ victim = lowest-reload-cost unpinned, monotonic, deterministic id tie-break; never a pinned adapter.
- Implement + register the policy; `--eviction-policy rank-aware` CLI (run + replay).

**Gate**: no-op golden holds. Own micro-plan; victim rule confirmed vs `Li2025` (Toppings) — may be "inspired by" not "reproduces" (design §5/§14).

atantawi:
## Summary

Adds the **`rank-aware`** LoRA eviction policy behind the B-3 `sim.EvictionPolicy` seam, plus a `--eviction-policy` CLI selector. Backend Swap **Phase B** (design §5.4) of the eviction chain (US2 eviction ablation). Consumes the frozen `EvictionContext.RankOf` accessor from B-3 without widening the context (D-4-1).

Closes #1492 (sub-issue of epic #1488).

## What changed

- **Policy** (`sim/lora/eviction/rank_aware.go`): victim = argmin over unpinned candidates by the strict total order **(rank ASC, id ASC)** — evict the lowest-rank adapter (cheapest to reload), tie-broken by lexicographically smallest id. An unregistered candidate (`RankOf` ok=false) maps to sentinel `math.MaxInt` so it sorts last and is never preferred over a registered adapter (**D-4-2**). Pure victim-selection function — no RNG, no map iteration, no wall-clock.
- **CLI** (`cmd/root.go`): `--eviction-policy` registered on the shared `registerSimConfigFlags` (run **and** replay, R23) and resolved through the shared `resolveLoRAConfig` (INV-13 sync-point). Applied only when `Changed()` (R18); empty ⇒ `lru` at two canonical fallback sites. Unknown names fail fast at the CLI boundary with the listed valid set (Principle V); `sim/` stays free of the `sim/lora/eviction` import via the nil-safe `ValidEvictionPolicyNamesFunc` hook set in `register.go` `init()`.

## Invariants & provenance

- **INV-6 / INV-13**: rank is a pure function of static adapter metadata ⇒ every instance ranks identically; decisions reproducible across instances and run/replay. Multi-instance no-op byte-identity test proves **unset == lru == rank-aware** stdout when LoRA is inert (**D-4-3**).
- **INV-8**: empty candidate set ⇒ `("", false)`, no victim, retry once a pin clears (no deadlock).
- **§14 provenance**: `rank-aware` is **BLIS-native**, *inspired by* the Li2025 (Toppings) rank-aware placement signal — **not** a reproduction. Production S-LoRA/vLLM use recency-style eviction.

## Testing

Rank-sensitivity (victim differs from LRU under skewed ranks), tie-break, unregistered-sorts-last, empty candidates, all-pinned no-deadlock, and cross-instance rank parity covered by contract + parametric (24-permutation) tests. `go build ./...`, `go test ./...`, and B-4 files under `golangci-lint` all green.

## Stacking note

Stacked on **B-3** (#1491). Shown against `main`, so the diff appears as CONFLICTING / large cumulative — inherited from the lora-integration stack; resolved by the planned whole-stack rebase, not per-PR.


docs/plans/2026-07-22-lora-policy-seams-design.md
```diff
@@ -0,0 +1,782 @@
+# LoRA Placement-Policy Seams — Design
```
jgchn: Do we want to commit this file?

sim/lora/eviction/rank_aware.go
```diff
@@ -0,0 +1,44 @@
...
+	// so it is never chosen ahead of any registered adapter (D-4-2). Candidate order
+	// does not affect the result — the total order is strict on (rank, id).
+	minID := ""
+	minRank := 0
+	first := true
+	for _, id := range ctx.Candidates {
+		rank, ok := ctx.RankOf(id)
+		if !ok {
+			rank = math.MaxInt
+		}
```
jgchn, line 44: It seems possible that two or more unregistered adapters can get the same rank. Will this be a problem?

docs/plans/2026-07-22-lora-policy-seams-design.md
atantawi: Good catch — no, we don't. `docs/plans/` has been gitignored since #1322 (which also purged 22+ previously-committed plan and design docs, with #1352 following up), so this was force-added past that rule and shouldn't be in the diff. Removed.

It's added at the bottom of the stack by `b8552178`, so the removal went into **B-1 (#1496)** as `git rm --cached` — the file stays on disk as the local working artifact it was always meant to be — and I rebased B-2…B-7 on top, which clears it from all seven diffs. Verified: `git diff --stat main..<branch> -- docs/plans/` is empty for every branch in the stack.

Two things worth flagging separately:
- `CLAUDE.md` still advertises `docs/plans/` as the home for active plans, which directly contradicts `.gitignore:132`. That stale line is plausibly what invited the force-add in the first place — filing a separate fix.
- `docs/plans/2026-07-15-lora-control-plane-design.md` (the Part A design doc) is already on main via the same slip, and is currently the sole surviving tracked file in that ignored directory. Folding cleanup into #1504.

`specs/008-lora-policy-seams/` is unaffected — `specs/` isn't ignored, and it already carries the durable spec/plan/research record.

sim/lora/eviction/rank_aware.go
atantawi: Yes, they can — and it's handled, though the comment didn't say so explicitly, which is a fair hit. I've extended it in `a2f8ec14`.

Several unregistered candidates all map to `math.MaxInt`, and then the comparator's second clause separates them by lexicographically smallest id — the same tie-break used for any other rank tie (C-2). The order is strict and total on `(rank, id)`, so the victim is deterministic regardless of candidate order (INV-6). Covered by `eviction_test.go:244-252` (the all-unregistered context) and `TestRankAware_CandidateOrderIndependence`, which runs every permutation against an independent reference argmin.

They also can't collide with a *registered* adapter: `lora.NewRegistry` rejects `Rank <= 0` and no declared rank approaches `MaxInt`, so the sentinel stays strictly above every real rank and unregistered ids always sort last (D-4-2) — `TestRankAware_UnregisteredSortsLast` pins that even against a registered adapter at `MaxInt-1`.

The one real limitation, now recorded in the comment: if *no* candidate is registered, rank-aware has no reload-cost signal at all and degenerates to smallest-id, ignoring recency. That's inherent to the policy without ranks rather than a defect in the tie-break — recency is `lru`'s job, and `lru` remains the default.

Issue #1511, opened 2026-07-31
Bug: decode-side-dropped PD parents (transfer-start & transfer-complete) enter TTFT/E2E distributions


namasl:
## Summary

On the PD-disaggregated path, a parent request that was **dropped on the decode side without ever emitting a token** is nonetheless treated as `completed == true` by `projectPDMetrics`, so it receives a prefill-only TTFT entry and an E2E entry — polluting the latency distributions with a request that produced no output.

This affects **all decode-side drop paths**, not just drop-at-transfer-start (see *Scope* below). The originally-filed version of this issue described only the transfer-start case; it has been broadened after code review.

## Root cause

`projectPDMetrics` (`sim/cluster/cluster.go:1789`) gates metric projection on:

```go
// sim/cluster/cluster.go:1799
completed := parent.CompletionTime > 0 && parent.DecodeInstanceID != ""
```

Both conditions are satisfied by a decode-side-dropped parent:

- `DecodeInstanceID` is set **upfront at routing** (decode-first routing, `sim/cluster/cluster_event.go:488`; also `cluster.go:2256`) and is **never cleared** on any drop path.
- Every decode-side drop stamps `parent.CompletionTime` (drop-at-transfer-start: `sim/cluster/pd_events.go:205`; late-drops: `pd_events.go:368` and `:379`).

So the parent passes the `completed` gate. Its `DecodeSubReq` is `nil` (the drop paths either never attach it or set it back to `nil`), so:

- **TTFT** — it fails the primary guard `hasPrefillTTFT && hasDecodeDelay && parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0` (`cluster.go:1955`) and falls through to the **prefill-only fallback** (`cluster.go:1973–1978`), recording `RequestTTFTs[pid]` for a request whose decode never produced a token.
- **E2E** — with the decode sub-request's own per-instance metrics absent, the E2E block takes the **fallback** `parent.CompletionTime - parent.ArrivalTime` (`cluster.go:1885–1903`), recording `RequestE2Es[pid]`.

Both values are `> 0`, so they survive `CollectPDMetrics`'s BC-11 zero-exclusion (`sim/cluster/pd_metrics.go:120`) and enter the reported `ParentTTFT` / E2E distributions.

[...]

PR #1512, opened 2026-07-31
fix(cluster): correct PD-disaggregated TTFT — drop phantom OTPT, add decode-queue wait (#1510)

### LINKED ISSUE #1510: Bug: PD-disaggregated TTFT double-counts OutputTokenProcessingTime and omits the decode-queue wait
## Summary

On the disaggregated prefill/decode (PD) path, the user-visible **time-to-first-token (TTFT)** that BLIS reports is computed with a formula that has two independent defects:

1. **It double-counts the per-output-token processing cost** (`OutputTokenProcessingTime`, the streaming/detokenization overhead). That cost is added once too often.
2. **It omits the decode-queue wait** — the time the request spends waiting to be scheduled on the *decode* instance after its KV cache has been transferred. Under load this is the larger error, and it makes the reported TTFT *smaller* than the latency a real user would see.

Both defects live inside the single line that composes the parent request's TTFT in `projectPDMetrics`.

This was found during code review, not by a failing test — and the reason no test caught it is itself a problem (see **The existing test does not guard this**, below).

---

## Background for readers new to this code

In **PD disaggregation**, one user request is split into two sub-requests that run on different instances:

- a **prefill** sub-request (processes the prompt, produces the KV cache), and
- a **decode** sub-request (receives the KV cache over a transfer, then generates output tokens).

The first token the user actually sees is produced by the **decode** instance, *after*: prefill runs → the KV cache transfers → the decode sub-request waits in the decode instance's queue → the decode instance runs its first step.

So the correct user-visible TTFT is:

```
TTFT = (arrival → first token emitted by the decode instance)
     = prefill_queue_wait
     + prefill_step
     + kv_transfer
     + decode_queue_wait      ← the term that is currently missing
     + first_decode_step
     + one OutputTokenProcessingTime   ← should appear exactly ONCE
```

`OutputTokenProcessingTime` (call it **OTPT**) is a fixed per-output-token overhead defined here:
https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/latency/trained_physics_model.go#L524

---

## Where the bug is

The parent request's TTFT is assembled here, in `projectPDMetrics`:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/cluster.go#L1808-L1830

The key line is:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/cluster.go#L1823

```go
newTTFT := prefillTTFT + transferDuration + firstDecodeStep
```

To see why this is wrong, we have to look at what each of the two "black-box" terms — `prefillTTFT` and `firstDecodeStep` — actually contains.

### Defect 1 — `OutputTokenProcessingTime` is counted twice

**`prefillTTFT`** is the `FirstTokenTime` recorded for the *prefill* sub-request. It is computed here:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/simulator.go#L998

```go
req.FirstTokenTime = now + currStepAdvance + sim.latencyModel.OutputTokenProcessingTime() - req.ArrivalTime
```

So `prefillTTFT` already includes **one** OTPT. (Note: the prefill sub-request produces zero output tokens, so charging it a token-streaming cost is itself questionable — but the point here is simply that the term is present.)

**`firstDecodeStep`** is `ITL[0]` of the decode sub-request. The per-step inter-token latency is recorded here:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/simulator.go#L987

```go
req.ITL = append(req.ITL, currStepAdvance+sim.latencyModel.OutputTokenProcessingTime())
```

So `firstDecodeStep` also includes **one** OTPT — and this is the legitimate one (the decode instance really does stream the first token).

Adding the two terms therefore lands **two** OTPTs in TTFT when there should be **one**.

### Defect 2 — the decode-queue wait is missing entirely

Look again at the composed formula. It contains `prefillTTFT` (arrival → prefill done), `transferDuration` (KV transfer), and `firstDecodeStep` (one decode step). **There is no term for the time the decode sub-request spends waiting in the decode instance's queue** between the transfer completing and the decode instance actually scheduling it.

The decode sub-request is enqueued onto the decode instance here (it goes into a wait queue, exactly like any other request):

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/simulator.go#L621-L648

When the decode instance is busy, that wait can be milliseconds. Because the formula never adds it, **the reported TTFT is smaller than reality** exactly when the system is under load — the regime capacity planning cares about most.

---

## Worked example

Suppose (all values in µs): `prefill_queue_wait = 100`, `prefill_step = 400`, `OTPT = 50`, `kv_transfer = 200`, `decode_queue_wait = 3000` (busy decode pod), `first_decode_step = 300`.

- `prefillTTFT = 100 + 400 + 50 = 550`
- `transferDuration = 200`
- `firstDecodeStep = 300 + 50 = 350`
- **Current formula:** `550 + 200 + 350 = 1100`
- **Correct value:** `100 + 400 + 200 + 3000 + 300 + 50 = 4050`

The reported TTFT (1100) is off by nearly 4× here — understating by the 3000 µs queue wait, and separately carrying a phantom extra 50 µs OTPT. At *low* load (`decode_queue_wait ≈ 0`) the formula instead slightly *over*-states TTFT by that phantom OTPT. So the error changes sign with load, but the important, larger error is the under-statement under load.

---

## Proposed fix

Both defects can be removed in one change, because the engine already records exactly the quantity we need.

For **every** scheduled request, the engine records its scheduling delay (arrival → first scheduled onto an instance) here:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/simulator.go#L824

```go
sim.Metrics.RequestSchedulingDelays[s.Request.ID] = now - s.Request.ArrivalTime
```

Crucially, the decode sub-request's `ArrivalTime` is set to the **parent's original arrival time**:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/pd_events.go#L139

```go
ArrivalTime:        orig.ArrivalTime,
```

Therefore `RequestSchedulingDelays[decodeSubReqID]` already spans the entire chain from the user's arrival up to the moment the decode instance schedules the first decode step:

```
decodeSchedulingDelay = prefill_queue_wait + prefill_step + kv_transfer + decode_queue_wait
```

(The prefill and decode sub-requests run on instances that share the cluster clock, so `now - ArrivalTime` is a valid arrival-to-schedule span across the two hops.)

So the fix is to replace the composed formula with:

```go
newTTFT := decodeSchedulingDelay + firstDecodeStep
```

This:

- **adds** the previously-missing `decode_queue_wait`, and
- **drops** the phantom extra OTPT (we no longer add `prefillTTFT`, which carried the second copy; `firstDecodeStep` still contributes the one correct OTPT).

The result equals the correct TTFT defined at the top of this issue. It is also structurally identical to how TTFT is derived on the non-PD path (`scheduling delay + first-token step`), which is a good consistency signal.

A secondary benefit: `RequestSchedulingDelays` is recorded unconditionally by the engine for *any* scheduled request, so this works for every disaggregating configuration, not only specific router setups.

### Don't forget the dependent bookkeeping

The same block keeps a running `TTFTSum` in sync with the per-request adjustment here:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/cluster.go#L1826

```go
m.TTFTSum += int64(newTTFT - prefillTTFT)
```

Whatever new formula is adopted, this line (and the "missing decode data" fallback branch just below it) must be updated to stay consistent with the new `newTTFT`, so the aggregate mean TTFT matches the corrected per-request values.

---

## The existing test does not guard this

There is a test that appears to cover this exact formula:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/disaggregation_test.go#L895

Its core assertion is:

https://github.com/inference-sim/inference-sim/blob/f54a50aaa524b473ec854fbe1b5e804d1752e39c/sim/cluster/disaggregation_test.go#L939

```go
expectedTTFT := origPrefillTTFT + float64(transferDuration) + float64(firstDecodeStep)
```

This is the **production formula retyped**, reading the same fields in the same order. The "expected" value on the right-hand side is derived by re-running the implementation's own arithmetic, so the assertion is a tautology: `code_output == recompute(code_output)`. It stays green for *any* formula the production code contains — including both wrong ones above — and it decomposes neither `origPrefillTTFT` (so it can't see the doubled OTPT hidden inside) nor accounts for `decode_queue_wait` (a term absent from both sides). It also runs a light, near-empty-queue scenario, so it wouldn't exercise queue wait even if it were written independently.

Per this repo's testing guidance ("test laws, not just values" — CLAUDE.md, BDD/TDD section), the fix should replace this with an **independent oracle**: reconstruct the expected TTFT from event timestamps (e.g. `decode-first-token timestamp − parent.ArrivalTime`), computed via a different path than `projectPDMetrics`, and run it under a **loaded** decode instance so a nonzero `decode_queue_wait` is actually present.

---

## Impact

- **Severity:** correctness / metric fidelity. TTFT is a primary SLO dimension, so this directly affects goodput/SLO-attainment analysis, calibration against real servers, and any capacity-planning conclusion drawn on the PD path.
- **Direction:** under-reports TTFT under load (missing queue wait dominates); slightly over-reports at very low load (phantom OTPT).
- **Scope:** disaggregated (PD) path only; the non-PD path is unaffected.
- **Related invariant:** INV-5 (causality, `arrival ≤ schedule ≤ completion`) is not violated, but the reported TTFT is inconsistent with the true arrival→first-token span the DES actually simulated.

## Which invariant is violated?

- [x] None / unknown (metric-fidelity bug on the PD path; no hard invariant tripped)

## How this was discovered

Code review of the PD TTFT projection in `projectPDMetrics`.

## Environment

- Applies to upstream `main` at commit `f54a50aaa524b473ec854fbe1b5e804d1752e39c` (all permalinks above are pinned to it).

namasl:
## Summary

On the PD-disaggregated (prefill/decode split) path, the user-visible **time-to-first-token (TTFT)** BLIS reports was assembled in `projectPDMetrics` with a formula that had two independent defects:

1. **Double-counted `OutputTokenProcessingTime` (OTPT)** — `prefillTTFT` already carries one OTPT (`sim/simulator.go:998`) and `firstDecodeStep = ITL[0]` carries a second (`sim/simulator.go:987`), so summing both landed **two** when exactly **one** is correct (the decode pod streams the first real token once).
2. **Omitted the decode-queue wait** — the time the decode sub-request spends queued on the decode instance between the KV transfer completing and its first decode step. Under load this term dominates, so reported TTFT was *smaller* than a real user sees — exactly the regime capacity planning cares about.

## The fix

Replace the composed formula with the decode sub-request's own scheduling delay plus the first decode step:

```go
newTTFT = decodeSchedulingDelay + firstDecodeStep
```

`decodeSchedulingDelay = RequestSchedulingDelays[decodeSubReqID]`. Because the decode sub-request's `ArrivalTime` is set to the parent's original arrival time (`sim/cluster/pd_events.go:139`), that delay already spans `prefill_queue + prefill_step + kv_transfer + decode_queue_wait` — adding the previously-missing wait — and `firstDecodeStep = ITL[0]` contributes the single legitimate OTPT. This is structurally identical to the non-PD TTFT (`scheduling_delay + first-token step`) and works for **every** disaggregating configuration (`RequestSchedulingDelays` is recorded unconditionally for any scheduled request).

Worked example from the issue (µs): `prefill_queue=100, prefill_step=400, OTPT=50, kv_transfer=200, decode_queue_wait=3000, first_decode_step=300`. Old formula reported **1100**; correct value is **4050**. The fix now yields the correct value.

## Behavioral contracts

- **BC-1** — GIVEN a completed PD parent with a recorded decode scheduling delay and non-empty decode ITL, WHEN `projectPDMetrics` runs, THEN reported TTFT = `decodeSchedulingDelay + firstDecodeStep` (the true arrival → first-token-emitted span, containing exactly one OTPT).
- **BC-2** — GIVEN a decode instance under load (positive decode-queue wait), WHEN TTFT is reported, THEN the wait is included (reported TTFT strictly exceeds the old formula's value). Verified under a **pinned loaded** decode instance, not by re-typing the formula.
- **BC-3** — GIVEN any PD run, WHEN projection completes, THEN `TTFTSum` stays consistent with the sum/mean of projected `RequestTTFTs` (pre-projection baseline per parent is `prefillTTFT`; the decode sub-request contributes 0, so the delta identity is unchanged), and `TTFT > 0`.
- **Guard** now gates on `hasPrefillTTFT && hasDecodeDelay && DecodeSubReq != nil && len(ITL) > 0` (the two transfer-timestamp guards are replaced by `hasDecodeDelay`; `hasPrefillTTFT` is retained as the `TTFTSum` baseline and to preserve the three outcome branches). Added a defensive `newTTFT < 0` fallback mirroring the adjacent E2E block.

## Why the tests changed (test epistemology)

The prior test (`disaggregation_test.go:939`) set `expectedTTFT` to the **production formula retyped** — a tautology (`code == recompute(code)`) that stayed green for *any* formula, including both buggy ones, and never exercised queue wait. Per this repo's "test laws, not values" standard it is replaced with:

1. **`TestDisaggregation_TTFT_IncludesTransferAndDecode`** — a residual decomposed against **independent** `ParentRequest` phase timestamps (`DecodeEnqueueTime − ArrivalTime`, recorded by a different mechanism than the scheduling-delay map the fix reads) plus a **differential vs the old formula** (`reported < old` at low load, proving the phantom OTPT is gone).
2. **`TestDisaggregation_TTFT_IncludesDecodeQueueWait`** — a **pinned loaded** scenario (1 decode instance, `maxRunningReqs=1`, 6 overlapping short requests) that deterministically forces `decode_queue_wait > OTPT`, asserts the residual proves the wait is included and `reported > old`, and `t.Fatal`s (never `t.Skip`) if no parent queued.
3. **`TestDisaggregation_TTFT_ProjectionBranches`** — unit-tests the primary / prefill-fallback / Branch-C / negative-guard branches and the `TTFTSum` delta.

## Testing verification

- `go build ./...` — passes
- `go test ./...` — all packages pass (11 ok, 0 fail)
- Determinism (INV-6) verified: 3 identical runs produce byte-identical `TTFTSum` and `sum(RequestTTFTs)`.

## Parity & invariants

Applies identically on `blis run` and `blis replay` (both reach the shared `projectPDMetrics` via `ClusterSimulator.Run()` — no separate replay TTFT path), so **INV-13** run/replay parity is preserved by construction. **INV-6** (determinism) holds (verified above). No `logrus.Fatalf` added to library code. Golden tests do not exercise the PD path, so their values are unaffected.

> **Note on INV-5 (`TTFT ≤ E2E`):** this PR does not claim INV-5 holds on the PD path. Review surfaced a **pre-existing** bug on `main` (independent of this change) in which the parent E2E under-counts the decode step, so `TTFT ≤ E2E` can be violated for short-output PD requests — filed as #1513. This PR corrects the TTFT numerator (reducing the discrepancy) but cannot eliminate the violation, since the breach is on the E2E side and out of scope here.

## Discovered Issues

- **#1511** — Pre-existing (not introduced here): drop-at-transfer-start PD parents evaluate `completed == true` and enter the TTFT/E2E distributions. The old formula routed them to the same prefill-only fallback, so this PR does not change the behavior; it is documented in a code comment and filed separately.
- **#1513** — Pre-existing (not introduced here): parent E2E under-counts the decode step advance on the PD path, so `TTFT ≤ E2E` (INV-5) can be violated for short-output requests. The breach is on the E2E side; this PR corrects the TTFT numerator (reducing the gap) but cannot close it, and the E2E fix is out of scope here.

Fixes #1510

### DIFF
--- a/docs/guide/results.md
+++ b/docs/guide/results.md
@@ -232,7 +232,7 @@ When PD disaggregation is active (`--prefill-instances > 0`), BLIS prints a `===
 | **Prefill Throughput** | Sub-request completion rate on prefill instances (sub-req/s) |
 | **Decode Throughput** | Sub-request completion rate on decode instances (sub-req/s) |
 | **Load Imbalance Ratio** | `max(prefill_load, decode_load) / min(...)` — `1.0` = perfectly balanced; `inf` = one pool has no completions |
-| **Parent TTFT** | Client-visible TTFT (prefill TTFT + KV transfer duration + first decode step); distribution in microseconds |
+| **Parent TTFT** | Client-visible TTFT (decode-sub-request scheduling delay + first decode step) — the arrival → first-token-emitted-by-decode span, including the decode-queue wait and exactly one output-token processing overhead (#1510); distribution in microseconds |
 | **KV Transfer Duration** | Time to transfer KV blocks from prefill to decode instance; distribution in microseconds |
 | **Peak Concurrent Transfers** | Maximum simultaneous in-flight KV transfers (only with `--pd-transfer-contention`) |
 | **Mean Transfer Queue Depth** | Average queue depth at the transfer bandwidth bottleneck (only with `--pd-transfer-contention`) |
--- a/sim/cluster/cluster.go
+++ b/sim/cluster/cluster.go
@@ -1844,6 +1844,13 @@ func (c *ClusterSimulator) projectPDMetrics() {
 		delete(m.RequestTTFTs, pfx)
 		delete(m.RequestTTFTs, dec)
 		if completed {
+			// A decode sub-request that emitted a first token and THEN timed out mid-generation
+			// still has a recorded scheduling delay and a non-empty ITL, so it takes this
+			// primary branch and reports a real TTFT. That is intentional and correct: the user
+			// did receive that first token, so its arrival→first-token span is a genuine
+			// measurement (INV-5 holds — the first token preceded the timeout ≤ completion). A
+			// decode sub-request that timed out while still queued has an empty ITL and falls to
+			// the prefill fallback below. Both match the pre-#1510 behavior.
 			if hasPrefillTTFT && hasDecodeDelay && parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0 {
 				firstDecodeStep := float64(parent.DecodeSubReq.ITL[0])
 				newTTFT := float64(decodeDelay) + firstDecodeStep
--- a/sim/cluster/cluster.go
+++ b/sim/cluster/cluster.go
@@ -1805,29 +1805,69 @@ func (c *ClusterSimulator) projectPDMetrics() {
 			}
 		}
 
-		// TTFT: user-visible time-to-first-token for PD disaggregation.
-		// In llm-d, the first token reaches the user from the decode pod, not
-		// prefill: prefill completes → KV transfers → decode pod recomputes last
-		// prompt token and samples first output token. User-visible TTFT =
-		// prefillTTFT + transferDuration + firstDecodeStep. See issue #930.
+		// TTFT: user-visible time-to-first-token for PD disaggregation (issue #1510,
+		// correcting the earlier #930 composition). In llm-d the first token reaches the
+		// user from the decode pod, not prefill: prefill completes → KV transfers →
+		// decode pod queues, recomputes the last prompt token, and samples the first
+		// output token. The correct user-visible TTFT is the arrival → first-token-emitted
+		// span:
+		//
+		//   TTFT = decodeSchedulingDelay + firstDecodeStep
+		//
+		// where decodeSchedulingDelay = RequestSchedulingDelays[decodeSubReqID] = the
+		// decode sub-request's (schedule − arrival). Because the decode sub-request's
+		// ArrivalTime is the parent's original arrival time (pd_events.go), that delay
+		// already spans prefill_queue + prefill_step + kv_transfer + decode_queue_wait —
+		// including the decode-queue wait that the previous
+		// (prefillTTFT + transferDuration + firstDecodeStep) formula omitted. It also
+		// carries exactly ONE OutputTokenProcessingTime: prefillTTFT (dropped here) held a
+		// second, phantom copy; firstDecodeStep = ITL[0] contributes the single legitimate
+		// one (the decode pod streams the first real token exactly once). This is
+		// structurally identical to the non-PD TTFT (scheduling_delay + first-token step).
 		//
 		// Read prefill TTFT before deleting sub-request keys (R1: no silent data loss).
-		// Gate on completed: dropped-request TTFTs must not enter the distribution.
+		// prefillTTFT is retained as the TTFTSum baseline: pre-projection TTFTSum holds
+		// exactly prefillTTFT for this parent (the decode sub-request never sets
+		// FirstTokenTime, so it contributes 0). Read the decode scheduling delay here too,
+		// before the scheduling-delay block below deletes it.
+		//
+		// Gate on `completed` (CompletionTime > 0 && DecodeInstanceID != ""). A
+		// late-drop (decode pod unroutable at transfer complete) leaves DecodeInstanceID
+		// set but nils DecodeSubReq, so it fails the primary guard and takes the
+		// prefill-only fallback. NOTE: a drop-at-transfer-start parent is still
+		// `completed == true` (DecodeInstanceID is set upfront at routing and dropAtStart
+		// stamps CompletionTime), so it too receives a prefill-only TTFT here — a
+		// pre-existing behavior unchanged by this fix (the old guard also routed these to
+		// the same fallback). Tracked separately in issue #1511.
 		prefillTTFT, hasPrefillTTFT := m.RequestTTFTs[pfx]
+		decodeDelay, hasDecodeDelay := m.RequestSchedulingDelays[dec]
 		delete(m.RequestTTFTs, pfx)
 		delete(m.RequestTTFTs, dec)
 		if completed {
-			if hasPrefillTTFT && parent.TransferStartTime > 0 && parent.TransferCompleteTime >= parent.TransferStartTime && parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0 {
-				transferDuration := float64(parent.TransferCompleteTime - parent.TransferStartTime)
+			if hasPrefillTTFT && hasDecodeDelay && parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0 {
 				firstDecodeStep := float64(parent.DecodeSubReq.ITL[0])
-				newTTFT := prefillTTFT + transferDuration + firstDecodeStep
-				m.RequestTTFTs[pid] = newTTFT
-				// BC-3: Keep TTFTSum consistent with the TTFT adjustment.
-				m.TTFTSum += int64(newTTFT - prefillTTFT)
+				newTTFT := float64(decodeDelay) + firstDecodeStep
+				if newTTFT < 0 {
+					// Defensive parity with the E2E block above: never emit a negative
+					// TTFT (a headline SLO metric). Unreachable in normal operation —
+					// decodeDelay ≥ 0 by shared-clock event ordering and firstDecodeStep > 0
+					// — but guards against a hypothetical clock regression rather than
+					// silently reporting a negative value.
+					logrus.Errorf("[cluster] projectPDMetrics: negative TTFT for %s (decodeDelay=%d firstDecodeStep=%.0f); using prefill TTFT",
+						pid, decodeDelay, firstDecodeStep)
+					m.RequestTTFTs[pid] = prefillTTFT // delta 0 vs baseline: TTFTSum untouched
+				} else {
+					m.RequestTTFTs[pid] = newTTFT
+					// BC-3: Keep TTFTSum consistent with the TTFT adjustment. The baseline
+					// prefillTTFT is replaced by newTTFT for this parent.
+					m.TTFTSum += int64(newTTFT - prefillTTFT)
+				}
 			} else if hasPrefillTTFT {
-				// Defensive fallback: use prefill-only TTFT if decode data unavailable.
+				// Defensive fallback: use prefill-only TTFT if decode data unavailable
+				// (no decode scheduling delay, nil DecodeSubReq, or empty ITL). TTFTSum
+				// unchanged (projected value equals the baseline).
 				m.RequestTTFTs[pid] = prefillTTFT
-				logrus.Warnf("[cluster] projectPDMetrics: parent %s missing decode ITL or TransferCompleteTime; using prefill TTFT", pid)
+				logrus.Warnf("[cluster] projectPDMetrics: parent %s missing decode scheduling delay, DecodeSubReq, or ITL; using prefill TTFT", pid)
 			} else {
 				logrus.Warnf("[cluster] projectPDMetrics: completed parent %s has no prefill TTFT (key %s)", pid, pfx)
 			}
--- a/sim/cluster/disaggregation_test.go
+++ b/sim/cluster/disaggregation_test.go
@@ -895,8 +895,8 @@ func TestDisaggregation_MetricProjection_E2ECorrectness(t *testing.T) {
 // (prefillTTFT + transferDuration + firstDecodeStep), which double-counted OTPT and
 // omitted the decode-queue wait.
 //
-// Test independence (see docs/plans/pr1510-pd-ttft-plan.md Part 3): the assertions never
-// recompute the production formula. Instead they use two orthogonal guards —
+// Test independence: the assertions never recompute the production formula (the pre-#1510
+// test did, making it a tautology). Instead they use two orthogonal guards —
 //  (1) a residual reconstructed from ParentRequest phase timestamps
 //      (DecodeEnqueueTime − ArrivalTime), recorded by a different mechanism than the
 //      RequestSchedulingDelays map the fix reads; and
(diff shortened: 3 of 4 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

Addressed the two non-blocking notes from the previous review plus the guard-logic observation in commit f3f39a76:
- Removed the dangling `docs/plans/pr1510-pd-ttft-plan.md` reference from a test comment (plan docs are gitignored and must not be referenced in committed code).
- Replaced the derived `wantSum` in `TestDisaggregation_TTFT_ProjectionBranches` with an explicit per-case field modeling production exactly (delta applied only on the primary `newTTFT` path; all fallbacks incl. the negative guard leave `TTFTSum` untouched); also assert `TTFTSum` for Branch C.
- Documented the timed-out-decode-with-partial-ITL case in `projectPDMetrics` (it correctly takes the primary branch — the first token was genuinely emitted; INV-5 holds) and pinned it with a new test case so it can't silently regress.

No production formula change.

sriumcp:
@claude Shouldn't we be reporting both in the PD disagg setting? Specifically, report the time it took for the prefill instance to finish with a request (this includes queueing at the prefill instance), and separately report the time it takes for the first decode to occur at the decode instance (this includes the queuing delay at the decode instance).

Of course, when a request is not disaggregated at all, things should stay as they already are (i.e., time for the first token in the non-disaggregated instance, which includes queueing).

WDYT?

jgchn (approved):

susiejojo:
## Review finding: INV-5 (`TTFT ≤ E2E`) is still violated for short-output PD requests

I reviewed this PR and want to flag one substantive inaccuracy before merge. **The core TTFT fix is correct and a real improvement** — the double-counted OTPT and the missing decode-queue wait are both genuinely addressed, the `decodeSchedulingDelay + firstDecodeStep` reduction is sound (`decodeSubReq.ArrivalTime = orig.ArrivalTime` at `pd_events.go:139` makes the scheduling delay span the full arrival→schedule window), and the `TTFTSum` baseline/delta bookkeeping is consistent (decode sub-requests never set `FirstTokenTime` because they start at `ProgressIndex == InputLen`, `batch_formation.go:185`, so the `simulator.go:996` TTFT-set condition never fires for them).

**However, the PR description and BC-4 claim "`TTFT ≤ E2E` (INV-5) … hold[s]". That is false for short-output PD requests.** The existing tests use ~10 output tokens, which is above the violation band, so they don't catch it.

### Reproduction

Drop this into `sim/cluster/inv5_probe_test.go` (on this branch) and run it:

```go
package cluster

import (
	"fmt"
	"testing"

	sim "github.com/inference-sim/inference-sim/sim"
)

func runProbe(t *testing.T, outTokens int) {
	config := newTestDisaggDeploymentConfig(4, 2, 2) // alphas [100,1,100]: OTPT=100, PostDecode=1
	requests := make([]*sim.Request, 5)
	for i := 0; i < 5; i++ {
		requests[i] = &sim.Request{
			ID:           fmt.Sprintf("request_%d", i),
			InputTokens:  make([]sim.TokenID, 20),
			OutputTokens: make([]sim.TokenID, outTokens),
			State:        sim.StateQueued,
			ArrivalTime:  int64(i * 100),
		}
	}
	cs := NewClusterSimulator(config, NewSliceRequestSource(requests), nil)
	mustRun(t, cs)
	m := cs.AggregatedMetrics()

	violations := 0
	for _, parent := range cs.ParentRequests() {
		if parent.CompletionTime == 0 || parent.DecodeInstanceID == "" {
			continue
		}
		pid := parent.ID
		ttft, ok := m.RequestTTFTs[pid]
		if !ok {
			continue
		}
		e2e := m.RequestE2Es[pid]
		t.Logf("[out=%d] %s: TTFT=%.1f E2E=%.1f (TTFT-E2E)=%.1f ITL=%v",
			outTokens, pid, ttft, e2e, ttft-e2e, parent.DecodeSubReq.ITL)
		if ttft > e2e {
			violations++
		}
	}
	if violations > 0 {
		t.Errorf("[out=%d] INV-5 violated in %d parents", outTokens, violations)
	}
}

func TestProbe_INV5_1(t *testing.T)  { runProbe(t, 1) }
func TestProbe_INV5_2(t *testing.T)  { runProbe(t, 2) }
func TestProbe_INV5_3(t *testing.T)  { runProbe(t, 3) }
func TestProbe_INV5_10(t *testing.T) { runProbe(t, 10) }
```

```
go test ./sim/cluster/ -run 'TestProbe_INV5_' -v
```

Result on this branch:

| output tokens | TTFT | E2E | INV-5 |
|---|---|---|---|
| 1 | 451.0 | 152.0 | ❌ violated (TTFT > E2E by 299µs) |
| 2 | 451.0 | 152.0 | ❌ violated |
| 3 | 451.0 | 352.0 | ❌ violated |
| 10 | 451.0 | 1752.0 | ✅ holds |

### Root cause — it's the parent E2E, not this PR's new TTFT

The violation is a **pre-existing E2E undercount** that this PR neither introduces nor fixes (I confirmed it also fails on the merge-base, where the old formula gave an even larger `TTFT=751 > E2E=152`). Instrumenting the 1-token case:

- **Non-PD baseline**, same 1-token request: `TTFT=400, E2E=401` ✅ (correct).
- **PD path**, same request: the decode sub-request's *own* per-instance E2E is **301** (correct), but `projectPDMetrics` discards it and reports `parent.CompletionTime − ArrivalTime = 152`.
- `parent.CompletionTime` is stamped at the completion-*detection* tick + `PostDecodeFixedOverhead` (`cluster.go:1303`) and does **not** include the decode step's own advance — so the parent E2E (152) ends up **smaller than a single decode step** (`ITL[0]=300`).

Timing trace for `request_0` (1 output token):
```
arrival=0 prefillEnq=0 transferStart=100 transferComplete=151 decodeEnq=151 completion=152
TTFT=451.0  E2E=152.0  ITL=[300]
```

So `TTFT` is actually the *correct* arrival→first-token value here; `E2E=152` is the wrong one. This is why the PR reduces the violation (751→451) but can't eliminate it — the fix is on the TTFT numerator, but the invariant breach lives in the E2E metric.

### Suggested actions

1. **Correct the INV-5 claim** in the PR body and BC-4 — either scope it (holds only when the decode span exceeds `OTPT − PostDecodeFixedOverhead`) or acknowledge the residual violation. As written it advertises a guarantee the code doesn't provide.
2. **File a separate issue for the parent-E2E undercount** (sibling to #1511). The aggregated PD E2E distribution is under-reported vs non-PD independently of TTFT, and it's the actual blocker for full INV-5 parity.
3. **Add a short-output (1–2 token) case** to the projection tests to pin this boundary, even if only to document the known residual.

Net: the TTFT change is good on its own merits; the ask is to stop claiming INV-5 holds and to file the E2E undercount separately.

*(Reproduced in an isolated worktree at PR head `f3f39a76`; no changes to my local branch.)*

susiejojo:
Filed the E2E under-count finding as a standalone bug: #1513 (pre-existing on `main`, orthogonal to this PR's TTFT change). Reproduction and the INV-5 table are in the issue.

Issue #1516, opened 2026-08-04
Saturation part 1: --detectors <name> single-detector trace + --saturation-config; remove 13 legacy flags


susiejojo:
## Goal (one sentence)

Replace the sprawling saturation CLI (13 flags) with **one selector flag `--detectors <name>`** plus **one config file `--saturation-config`**, and make `--saturation-report` write a **per-event verdict trace** for the selected detector. Put **nothing** saturation-related on stdout.

**Hard dependency — do not start until #1515 is merged.** #1515 makes backlog-drift a streaming detector (`Observe`/`Detect`). This issue's uniform replay loop assumes all three detectors stream; on today's `main` backlog-drift's `Observe`/`Detect` are still no-ops (`backlog_drift.go`), so running the loop before #1515 lands would emit a trace of constant `STABLE` verdicts for it. #1519 adds the bank (`--detectors all`), #1517 adds the stdout final label.

---

## SCOPE — read before writing code

**IN scope:**
1. Add a strict-YAML saturation config (one file replacing the 11 tuning flags).
2. Add a trace-sink abstraction (a no-op default + an in-memory collector).
3. Add the machinery to replay one detector over the completed requests and write its per-event trace to a file.
4. Rework the CLI on run/replay/observe: add `--detectors <name>` and `--saturation-config`; redefine `--saturation-report` to write the trace; remove the 13 legacy flags; share one helper across the three commands.
5. Remove `Classify` from the `Detector` interface (all detectors stream after #1515; the batch path's last callers are deleted here) — final interface is `Name / Observe / Detect / Reset`.
6. Delete the code that becomes dead.
7. Tests for the new pieces; update `CLAUDE.md`.

**OUT of scope (other issues — building them here is a mistake):**
- ❌ The detector **bank** / multiple detectors / `--detectors all` or comma-lists → **#1519**. Here `--detectors` takes exactly ONE name; `all`/lists must ERROR.
- ❌ Any **stdout `saturation`** output / final label → **#1517**. Here stdout has nothing saturation-related.
- ❌ Changing detector *algorithms* — `composite.go`, `threshold.go`, `backlog_drift.go` stream as-is (backlog-drift became streaming in #1515). You edit the `Detector` interface only to remove `Classify`.
- ❌ Editing the detector algorithms or the `Detector` interface — used as-is.
- ❌ `calibrate` — it has no saturation wiring on `main`; leave it entirely alone.

[...]

PR #1520, opened 2026-08-05
feat(saturation): make backlog-drift a streaming detector (#1515)

### LINKED ISSUE #1515: Saturation part 0: make backlog-drift a streaming detector (Observe/Detect)
## Goal (one sentence)

Make the `backlog-drift` detector **streaming** — give it real per-event `Observe`/`Detect` (a running in-flight count + online slope) instead of the current no-ops — so all three saturation detectors produce a per-event verdict uniformly.

This is the **lead-off** of the saturation rework (depends only on `main`). It unblocks #1516 (config + CLI + single-detector trace), which can then treat every detector as streaming and drop the batch `Classify` path.

---

## SCOPE — read before writing code

**IN scope:**
1. Implement `Observe`/`Detect` in `sim/saturation/backlog_drift.go` so the detector maintains an incremental in-flight count and emits an evolving verdict per event.
2. Make `Reset()` clear the new streaming state.
3. Add `NewBacklogDriftDetectorWithConfig(config workload.BacklogDriftConfig) Detector` so tests (and later #1516) can set a small `window_size` (see **Testing note** below — the two existing constructors both hardwire `DefaultBacklogDriftConfig()`'s 60s window, which is impractical for a unit test).
4. Unit tests driving events directly and asserting the verdict evolves (rising backlog → toward OVERLOADED; draining → toward STABLE).

**OUT of scope (other issues — do not build here):**
- ❌ The detector **bank** / `--detectors` flag / CLI wiring → **#1516/#1519**. This issue changes only the detector; nothing is user-visible yet.
- ❌ Trace sinks / report files / stdout output → **#1516**.
- ❌ Removing `Classify` from the `Detector` interface → **#1516** (once its last callers are gone).
- ❌ Touching `composite.go` / `threshold.go` / `detector.go`.

**Done when:** `backlog-drift`'s `Observe`/`Detect` produce an evolving per-event `Result`, `Reset` clears state, new unit tests pass, and `go test ./... && golangci-lint run ./...` are green. No behavior elsewhere changes (`Classify` still exists and still works this issue).

---

## Background (facts)

- `sim/saturation/backlog_drift.go` on `main`: `Observe` (line ~51) and `Detect` (line ~56) are **no-ops** returning flat STABLE; the real verdict lives in `Classify`, which runs `workload.AnalyzeBacklogDriftWithClassifier` (a whole-trace, non-causal regression: relativize time → window → trim warmup/tail → OLS slope → drain-ratio classifier). `Reset` (line ~164) is a no-op.
- `composite` and `threshold` already implement real streaming `Observe`/`Detect` (accumulate events, recompute a verdict) — use them as the pattern.
- The whole-trace `Classify` is non-causal (needs the tail), so it **cannot** be reused per-event. The streaming path must be a separate, causal computation.

---

## What to build

Add incremental state to `BacklogDriftDetector` and drive it from `Observe`:
- Track running arrival and completion counts → **in-flight = arrivals − completions** (arrivals/completions come from the `Event.Type`).
- Maintain a bounded trailing window of `(timestamp, in-flight)` samples — one sample per `window_size` bucket, carrying the last value forward across empty buckets, so memory is O(windows) not O(events). Reuse the configured `window_size` (default from `workload.DefaultBacklogDriftConfig()`).
- `Detect()` computes an online slope of in-flight over the trailing window and returns a `Result`:
  - signals: at least `in_flight`, `arrivals`, `completions`, `running_slope`, and a `noise_floor` (`1/√arrivals`, mirroring composite).
  - **`running_slope` units (pin this):** slope of in-flight **per window bucket** (Δ in-flight per `window_size`), computed by OLS over the trailing `(bucket_index, in_flight)` samples. Bucket-indexed (not per-microsecond) so the value is scale-stable and does not depend on the absolute timestamp magnitude. Fewer than 2 samples ⇒ slope `0`.
  - **level bands (pin this — mirror composite's two-threshold structure):** compare `running_slope` against `noise_floor`. `running_slope <= noise_floor` → **STABLE**; `noise_floor < running_slope <= K·noise_floor` → **BACKLOGGED**; `running_slope > K·noise_floor` → **OVERLOADED**. Use `K = 3` as the "clearly rising" multiplier (document it as a tunable constant, not a calibrated value). Document the whole band scheme as an **online heuristic** — explicitly NOT the drain-ratio classifier `Classify` uses. The two may legitimately disagree.
  - **`Result.Score` (pin this):** normalized slope magnitude in `[0, 1]` — `min(1.0, max(0.0, running_slope) / (K·noise_floor))` — so score crosses ~1.0 exactly as the level reaches OVERLOADED. Draining (negative slope) ⇒ score `0`. This mirrors the "normalized slope magnitude, capped at 1.0" convention already used by `Classify` (`backlog_drift.go:151-163`).
  - **confidence:** reuse composite's `min(1.0, arrivals/20.0)` so the three streaming detectors agree on the confidence ramp.
- `Reset()` clears all the new state.
- Leave `Classify` as-is (still used until #1516 removes its callers).

Keep it deterministic: state only accumulates counts/sums over the events it's fed in order — no clock, no map iteration in the verdict.

---

## Testing note (window size)

The `BacklogDriftDetector` struct holds a `config workload.BacklogDriftConfig`, but on `main` the only public constructors (`NewBacklogDriftDetector`, `NewBacklogDriftDetectorWithClassifier`) hardwire `DefaultBacklogDriftConfig()`, whose `WindowSize` is **60s**. With a 60s bucket, any reasonably-sized synthetic test trace lands in one or two buckets → the online slope is undefined/flat and the "verdict evolves" contract passes vacuously (it never leaves the zero-window degenerate case).

Resolution for this PR: **add `NewBacklogDriftDetectorWithConfig`** (IN scope, item 3) and have the evolving-verdict tests inject a small `WindowSize` (e.g. `100*time.Millisecond`) via `NewBacklogDriftConfig`, so a handful of directly-fed events span ≥ `MinWindows` buckets and the slope is deterministically exercised. The default-config constructors keep the 60s window unchanged for production callers.

---

## Contract

- Feeding a rising-backlog event sequence (arrivals outpacing completions) makes `Detect()`'s in-flight and slope rise, and the level moves off STABLE.
- A draining sequence (completions catching up) trends the level back toward STABLE.
- `Reset()` returns the detector to its initial state (next `Detect()` on no events → STABLE, zero confidence).
- No events observed → STABLE, zero confidence (no panic, R20).
- `Classify` output is unchanged from `main` (this issue doesn't touch the batch path).
- **run/replay/observe parity:** all three consume the detector as `sim.BatchClassifier` (`Classify`-only; no `Detect()` call sites exist in production), so this PR changes no command's output — INV-13 holds trivially. Streaming gets wired (and parity becomes testable) only in #1516.

---

## Files

- **MODIFY:** `sim/saturation/backlog_drift.go` (implement `Observe`/`Detect`/`Reset`; add `NewBacklogDriftDetectorWithConfig`), `sim/saturation/backlog_drift_test.go` (replace the "Observe/Detect are no-op" test with evolving-verdict tests)
- **DO NOT TOUCH:** `composite.go`, `threshold.go`, `detector.go`, `sim/workload/saturation.go`

---

## Note

`Observe` is currently dead code in production (only tests call it). After #1516 wires the replay, this streaming path becomes the source of backlog-drift's trace. Lead-time claims from post-hoc traces remain invalid regardless (replay sees each request's final E2E).

susiejojo:
Closes #1515.

## Summary

Makes the `backlog-drift` detector **streaming** — it now has real per-event `Observe`/`Detect` (a running in-flight count + online slope) instead of no-ops, so all three saturation detectors produce a per-event verdict uniformly. This is the lead-off of the saturation rework and unblocks #1516.

## What changed (`sim/saturation/backlog_drift.go`)

- **`Observe`**: maintains running `arrivals`/`completions` counts → `in_flight = arrivals − completions`, folded into a bounded trailing window of **one sample per `WindowSize` bucket**, carrying the last value forward across empty buckets (O(windows) memory, not O(events)). Deterministic — counts/snapshots only, no clock, no map iteration.
- **`Detect`**: computes an online **OLS slope of in-flight per window bucket** over the trailing samples (bucket-index x-axis, so it's scale-stable and independent of absolute timestamp magnitude; <2 samples ⇒ slope 0). Bands mirror composite's two-threshold structure against a `noise_floor = 1/√arrivals` with `K = 3`:
  - `slope ≤ noise_floor` → **STABLE**
  - `noise_floor < slope ≤ K·noise_floor` → **BACKLOGGED**
  - `slope > K·noise_floor` → **OVERLOADED**
  - `Score = min(1, max(0, slope)/(K·noise_floor))` (draining ⇒ 0); `Confidence = min(1, arrivals/20)`.
  - Signals: `in_flight`, `arrivals`, `completions`, `running_slope`, `noise_floor`.
  - Documented as an **online heuristic** — explicitly NOT the drain-ratio classifier `Classify` uses; the two may legitimately disagree.
- **`Reset`**: clears all new streaming state.
- **`NewBacklogDriftDetectorWithConfig`**: new constructor so tests (and #1516) can inject a small `WindowSize`; the default-config constructors keep the 60s production window. All constructors route through one canonical private constructor (R4).

## What did NOT change

- `Classify` (the non-causal batch regression path) is untouched — its output is byte-identical to `main`.
- `composite.go`, `threshold.go`, `detector.go`, `sim/workload/saturation.go` untouched (per scope).
- No detector bank / `--detectors` flag / CLI / trace wiring (that's #1516).

## Observe/run/replay parity (INV-13)

All three commands consume the detector as `sim.BatchClassifier` (`Classify`-only). A repo-wide grep confirms **no production code calls `Detect()`/`Observe()`** on this detector, so this PR changes no command's output — INV-13 holds trivially. Streaming gets wired (and parity becomes testable) in #1516.

## Tests (`sim/saturation/backlog_drift_test.go`)

Replaced the "Observe/Detect are no-op" test with evolving-verdict tests driving events directly through a small-window config:
- rising backlog → positive slope, level moves off STABLE;
- steep sustained climb → OVERLOADED, score ~1.0;
- build-then-drain → negative slope, back to STABLE, score 0;
- no events → STABLE, zero confidence (R20, no panic);
- `Reset` → initial state.

## Verification

- `go build ./...` ✅
- `go test ./...` ✅
- `golangci-lint run ./...` (v2.9.0, matching CI) → `0 issues` ✅

### DIFF
--- a/sim/saturation/backlog_drift.go
+++ b/sim/saturation/backlog_drift.go
@@ -19,8 +19,11 @@ const backlogDriftSlopeK = 3.0
 
 // BacklogDriftDetector wraps the workload.AnalyzeBacklogDriftWithClassifier logic
 // as a post-hoc saturation detector (Issue #7).
-// This detector is stateless - it performs regression analysis over
-// completed request intervals during Classify().
+//
+// Two independent paths: the Classify (batch) path is stateless — it performs
+// whole-trace regression over completed request intervals on each call. The
+// Observe/Detect streaming path (#1515) carries incremental per-event state (the
+// fields below).
 //
 // The classifier is configurable (#1392): defaults to drain-ratio (matching the
 // CLI default for --saturation-classifier) so that the post-hoc detector path
@@ -37,12 +40,17 @@ type BacklogDriftDetector struct {
 	arrivals    int64 // running count of Arrival events
 	completions int64 // running count of Completion events
 
-	// Bounded trailing window of in-flight samples, one per WindowSize bucket.
-	// The last in-flight value is carried forward across empty buckets so memory
-	// is O(windows), not O(events). buckets holds the in-flight value at the end
-	// of each observed bucket; empty intervening buckets are forward-filled so
-	// the samples stay evenly spaced, letting Detect use bucket position as the
-	// OLS x-axis (scale-stable, independent of absolute timestamp magnitude).
+	// In-flight samples, one per WindowSize bucket spanned. buckets[i] holds the
+	// in-flight value at the end of bucket i; empty intervening buckets are
+	// forward-filled with the last value so the samples stay evenly spaced,
+	// letting Detect use bucket position as the OLS x-axis (scale-stable,
+	// independent of absolute timestamp magnitude).
+	//
+	// Memory is O(buckets spanned) = O(elapsed_span / WindowSize), i.e. it grows
+	// with the observation horizon, not the event count — a saving over O(events)
+	// when buckets are densely populated, but NOT a fixed bound (a sparse stream
+	// over a long horizon forward-fills many empty buckets). #1516 may add a
+	// trailing cap when it wires this to output; today the whole span is kept.
 	buckets       []int64
 	curBucketIdx  int64 // absolute index of the bucket currently being filled
 	curBucketInit bool  // whether curBucketIdx has been established yet
@@ -130,10 +138,9 @@ func (b *BacklogDriftDetector) Observe(event Event) {
 	}
 
 	// Advanced to a later bucket. Carry the last value forward across any empty
-	// intervening buckets so samples stay evenly spaced (one per bucket), keeping
-	// memory at O(windows). Events must arrive in non-decreasing timestamp order;
-	// an out-of-order earlier event folds into the current bucket rather than
-	// rewriting history.
+	// intervening buckets so samples stay evenly spaced (one per bucket spanned).
+	// Events must arrive in non-decreasing timestamp order; an out-of-order
+	// earlier event folds into the current bucket rather than rewriting history.
 	if bucketIdx > b.curBucketIdx {
 		lastVal := b.buckets[len(b.buckets)-1]
 		for gap := b.curBucketIdx + 1; gap < bucketIdx; gap++ {
--- a/sim/saturation/backlog_drift.go
+++ b/sim/saturation/backlog_drift.go
@@ -201,6 +201,14 @@ func (b *BacklogDriftDetector) Detect() Result {
 	// so it crosses ~1.0 exactly as the level reaches OVERLOADED. Draining
 	// (negative slope) ⇒ 0. Mirrors Classify's "normalized slope magnitude,
 	// capped at 1.0" convention.
+	//
+	// Note the shared boundary: at exactly slope == K·noiseFloor the band switch
+	// (<=) still reports BACKLOGGED while the score reaches 1.0, so Score==1.0 can
+	// co-occur with Level==BACKLOGGED. This is a measure-zero float coincidence,
+	// and both the band inequality and the score formula are pinned verbatim by
+	// #1515 — kept as-is so callers get the contracted values rather than a
+	// locally-nudged epsilon; Score is a magnitude, Level is the authoritative
+	// band.
 	score := 0.0
 	denom := backlogDriftSlopeK * noiseFloor
 	if denom > 0 {
--- a/sim/saturation/backlog_drift.go
+++ b/sim/saturation/backlog_drift.go
@@ -2,9 +2,20 @@
 package saturation
 
 import (
+	"math"
+	"time"
+
 	"github.com/inference-sim/inference-sim/sim"
 	"github.com/inference-sim/inference-sim/sim/workload"
 )
+
+// backlogDriftSlopeK is the "clearly rising" multiplier for the streaming band
+// classifier (#1515): running_slope in (noiseFloor, K*noiseFloor] → BACKLOGGED,
+// running_slope > K*noiseFloor → OVERLOADED. It is a tunable heuristic constant,
+// NOT an empirically calibrated value. The streaming bands are an online
+// heuristic and are explicitly NOT the drain-ratio classifier used by Classify;
+// the two computations may legitimately disagree.
+const backlogDriftSlopeK = 3.0
 
 // BacklogDriftDetector wraps the workload.AnalyzeBacklogDriftWithClassifier logic
 // as a post-hoc saturation detector (Issue #7).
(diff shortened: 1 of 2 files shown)


susiejojo:
## Self-Review Summary (blis-pr-review)

Ran the pr-review-toolkit agents (code-reviewer, pr-test-analyzer, comment-analyzer) against this PR vs. issue #1515.

**Verdict: Ready to merge.** No Critical or Important code-correctness issues. The implementation matches the pinned spec exactly, follows the `composite.go` reference pattern, respects the do-not-touch boundary (`composite.go`/`threshold.go`/`detector.go`/`sim/workload/saturation.go` untouched), and is clean against R1–R23 and the invariants.

### Verified against pinned spec (EXACT match)
- `running_slope` = OLS over bucket-index x-axis (0,1,2,…), per-bucket units, `<2 samples ⇒ 0`.
- Bands with K=3: `slope≤noise`→STABLE, `noise<slope≤K·noise`→BACKLOGGED, `slope>K·noise`→OVERLOADED.
- `Score = min(1, max(0,slope)/(K·noise))`; `confidence = min(1, arrivals/20)`; `noise_floor = 1/√arrivals`.
- Signals set = exactly {in_flight, arrivals, completions, running_slope, noise_floor}.
- R4 (canonical ctor), R11 (guarded division), R20 (no-events → STABLE/0/0), INV-6 (deterministic; no map iteration/clock in verdict), INV-13 (no production `Detect()` callers; `Classify` unchanged).

### Findings addressed in commit 9efafe2b
- **[docs]** Removed the inaccurate "bounded trailing window" / "O(windows), not O(events)" claims — memory is actually O(buckets spanned) = O(horizon/WindowSize), a saving over O(events) only when buckets are dense. Also fixed the stale type-doc "stateless" claim (only the Classify path is stateless).
- **[tests]** Added coverage for three documented behaviors #1516 will build on:
  - forward-fill across **empty** buckets (plateau stays flat → STABLE);
  - **out-of-order** earlier-timestamp event (folds into current bucket, no drop/panic);
  - the **BACKLOGGED** middle band (hand-computed boundaries → `Level==Backlogged`, score ∈ (0,1)).
  - Strengthened RisingBacklog to assert the STABLE→off-STABLE *transition*.

### Verification
`go build ./...` ✅ · `go test ./...` ✅ · `golangci-lint run ./...` (v2.9.0) → `0 issues` ✅

susiejojo:
@claude /blis-pr-review

jgchn:
@claude /blis-pr-review

sriumcp:
@claude Does this mean all the detectors in the codebase now satisfy the following ... 

1) All of them are streaming, so they can declare saturation verdicts at different instants in time throughout a run
2) All of them achieve run/replay/observe parity

susiejojo:
@claude Is this PR good to merge?

sim/saturation/backlog_drift.go
```diff
@@ -19,48 +33,215 @@ import (
...
+	//   slope > K·noiseFloor           → OVERLOADED
+	var level Level
+	switch {
+	case runningSlope <= noiseFloor:
+		level = Stable
+	case runningSlope <= backlogDriftSlopeK*noiseFloor:
+		level = Backlogged
+	default:
+		level = Overloaded
+	}
```
jgchn, line 198: Minor nit (non-blocking): at the exact boundary `runningSlope == K*noiseFloor`, the `<=` in the switch gives `Level=BACKLOGGED` but the score formula yields `K*noise / K*noise = 1.0`. So you can get `Score=1.0` with `Level=BACKLOGGED`. Unlikely to hit in practice with float64, but if you want them to agree, either use `< in the BACKLOGGED` case or use a strict < denominator in the score (e.g. `score = slope / (K*noise + epsilon)`). Fine to defer, just flagging the semantic gap.

susiejojo: Good catch — real semantic gap. I've kept both formulas as-is (68e18dec) since the `<=` band boundary and the `slope/(K·noise)` score are both pinned verbatim in #1515, and I'd rather not have BLIS diverge from the contracted values that #1516 builds on. Added a comment documenting the coincidence: it's measure-zero in float64, Score is a magnitude while Level is the authoritative band. Happy to switch to a strict `<` in the score denominator if you'd prefer them to strictly agree.

susiejojo: Addressed.

jgchn (approved):
LGTM!

Issue #893, opened 2026-03-31
fix(latency): update ModelHardwareConfig.HWConfig for roofline backends when NodePools gpu_type differs from --gpu


atantawi:
## Context

PR #892 deferred `InstanceSimulator` construction until after `PlaceInstance` and sets `simCfg.GPU = matchedGPUType` (pool-authoritative GPU label). This fixes hardware calibration for the **blackbox** backend (which does not use `HWConfig`).

## Problem

For **roofline** and **trained-roofline** backends, `NewLatencyModel` uses `ModelHardwareConfig.HWConfig` (the `TFlopsPeak`, `BwPeakTBs`, etc. struct), which is loaded at CLI startup via `GetHWConfig(hwConfigPath, --gpu)`. PR #892 does not update `HWConfig` — it only updates the `GPU string` label.

**Impact**: When a NodePool has `gpu_type: A100` and the user passes `--gpu H100`, instances report `GPU() == "A100"` (correct) but roofline latency math still uses H100 hardware coefficients (incorrect). The blackbox backend is unaffected.

## Root cause

`HWConfig` is loaded once in `cmd/root.go` using the CLI `--gpu` flag and stored in `DeploymentConfig.SimConfig.ModelHardwareConfig.HWConfig`. `NewClusterSimulator` and `NodeReadyEvent.Execute` do not have access to `hwConfigPath` to reload per pool GPU type.

## Proposed fix

Pre-compute a map `gpuType → HardwareCalib` for all pool GPU types in `cmd/root.go` (after pool configs are parsed) and pass this map into `DeploymentConfig`. In `NewClusterSimulator` and `NodeReadyEvent.Execute`, look up the pool's calibration from the map and set `simCfg.ModelHardwareConfig.HWConfig` before calling `NewInstanceSimulator`.

## Expanded scope (added — closes gap deferred from PR #892)

PR #892 review noted: *"No integration test exercises a full `Run()` with NodePools where `gpu_type ≠ --gpu` and verifies E2E metric correctness. All 4 new tests are unit-level. A full E2E integration test would be appropriate in #893 when roofline HWConfig is fixed."*

This issue therefore also covers:
[...]

PR #1523, opened 2026-08-06
fix(cluster): per-instance KV capacity for mixed-GPU node pools

### LINKED ISSUE #1522: bug(cluster): mixed-GPU node pools inherit one global KV-block capacity
**Describe the bug**

For analytical backends, a cluster with node pools containing different GPU memory capacities gives every instance the same global KV-block capacity.

`cmd/root.go` auto-calculates `totalKVBlocks` once from the global `--hardware` selection before node-pool placement. Every instance then inherits `SimConfig.TotalKVBlocks`. Node-pool placement can select a different `gpu_type`, and `HWConfigByGPU` can correctly replace that instance's execution coefficients, but neither path recalculates its KV capacity. `node_pools[].gpu_memory_gib` is retained as inventory metadata and does not affect instance KV capacity.

This makes a mixed GPU pool internally inconsistent: execution time can reflect the placed GPU while KV capacity reflects the global GPU. Role-wide `PrefillOverrides`/`DecodeOverrides` cannot represent different capacities when, for example, H100 and L40S instances are both decode candidates.

**CLI command to reproduce**

Use an analytical backend and a policy bundle containing two node pools:

```yaml
node_pools:
  - name: h100
    gpu_type: H100
    gpus_per_node: 8
    gpu_memory_gib: 80
    initial_nodes: 1
    min_nodes: 1
    max_nodes: 1
  - name: l40s
    gpu_type: L40S
    gpus_per_node: 4
    gpu_memory_gib: 48
    initial_nodes: 1
    min_nodes: 1
    max_nodes: 1
```

```bash
./blis run \
  --model meta-llama/Llama-3.3-70B-Instruct \
  --hardware H100 \
  --tp 4 \
  --policy-config hetero.yaml \
  --num-instances 3 \
  --prefill-instances 1 \
  --decode-instances 2 \
  [workload flags]
```

Alternatively, this can be reproduced directly in a cluster test by constructing `DeploymentConfig` with the two `NodePools` and matching `HWConfigByGPU` entries, then comparing `TotalKVBlocks()` on the placed H100 and L40S instances.

With the standard L40S/H100 hardware definitions, Llama-3.3-70B TP4, default block size, and default memory utilization, the independently calculated capacities are:

```text
H100 80 GiB: 121801 blocks
L40S 48 GiB:  27429 blocks
```

Both placed instances instead receive 121801 blocks when the global hardware is H100.

**Expected behavior**

When `--total-kv-blocks` is not explicitly supplied, each placed instance should receive a KV-block capacity calculated from its actual GPU type, model, TP, block size, memory utilization, and weight precision. The behavior must also apply to deferred placement and autoscaler-created instances.

An explicit global `--total-kv-blocks` can retain its current override semantics, but the precedence should be documented.

**Actual behavior**

`totalKVBlocks` is calculated once from the global hardware in `cmd/root.go` and copied to every instance. Changing only `node_pools[].gpu_memory_gib` does not change any instance's KV cache.

- [x] None of the listed conservation invariants. This violates pool-config consistency (INV-P2-1): an instance's GPU label/execution calibration and KV capacity describe different devices.

**Environment**

- OS: Fedora Linux, kernel 7.0.12, x86_64
- Go version: 1.25.6
- Upstream commit: `f54a50aaa524b473ec854fbe1b5e804d1752e39c`

**Additional context**

Discovered while auditing heterogeneous H100+L40S simulations. BLIS's bundled `hardware_config.json` correctly defines L40S with 48 GiB; this is not an incorrect L40S hardware definition.

Related but distinct issues:

- #633 added per-role GPU/TP/KV overrides. Those work when all instances in a role share hardware, but cannot express mixed hardware within one role.
- #893 applies per-GPU execution coefficients after node-pool placement. There is no corresponding per-GPU KV-capacity application.
- #1315 covers the inverse CLI-path problem: role-specific KV capacity is correct, but latency coefficients are not. This issue concerns node-pool placement where per-instance capacity is wrong.

Suggested regression coverage should include startup placement, deferred placement, and direct-actuator/autoscaling paths with H100 and L40S instances serving the same role.

namasl:
## Summary

Mixed-GPU node pools previously gave **every** instance the same global KV-block capacity, computed once from the global `--hardware` GPU. Node-pool placement (#893) correctly overrides each instance's GPU type and execution coefficients (`HWConfig`), but never recalculated KV capacity — so an H100 pool and an L40S pool both received the H100 capacity. This restores **INV-P2-1** (pool-config consistency): an instance's GPU calibration and its KV capacity now describe the same device.

- Add `cluster.KVAutoCalcConfig` (opt-in carrier on `DeploymentConfig`) + `applyPerInstanceKVCapacity` helper that recomputes `TotalKVBlocks` from the **placed** pool's `gpu_memory_gib`, TP, DP, block size, `--gpu-memory-utilization`, and weight precision.
- Wire it at **all three** placement sites, immediately after the existing `HWConfigByGPU` override — so the placed GPU is authoritative for KV capacity just as it is for execution calibration:
  - startup placement (`NewClusterSimulator`)
  - deferred placement (`NodeReadyEvent.Execute`)
  - autoscaler scale-up (`DirectActuator.scaleUp`)
- CLI (`blis run`) enables it only when `--total-kv-blocks` is **not** explicitly set and an analytical backend with extractable KV params is in use.

### Reproduction (before → after)

With the issue's H100 (80 GiB) + L40S (48 GiB) pools, `qwen3-14b`, TP1:

```
[cluster] per-instance KV auto-calc for GPU "H100": total-kv-blocks=15909 (GPU=80 GiB, TP=1, DP=1)
[cluster] per-instance KV auto-calc for GPU "L40S": total-kv-blocks=4113  (GPU=48 GiB, TP=1, DP=1)
```

Before this fix both instances received the H100 value. (Exact counts scale with model/util; the point is they now differ per GPU.)

## Issue

Closes #1522

## Precedence (documented)

- **Explicit `--total-kv-blocks`** → per-instance recalc disabled; every instance keeps the uniform global capacity (backward-compatible).
- **Node pools + PD per-pool overrides both present** → placement-derived per-GPU capacity wins (mirrors how `HWConfigByGPU` unconditionally overrides the resolved `HWConfig`; placement is GPU-authoritative, SC-004).
- **Per-GPU capacity < `--max-model-len`** → that instance's `MaxModelLen` is auto-capped to the KV-feasible maximum (mirrors the existing CLI per-pool auto-cap), so it constructs instead of tripping the KV-too-small check.
- **Missing pool memory / calc error** → fall back to the inherited global capacity with a warning (R1: observable, never a panic).

## Cross-Path Parity

| `blis run` | yes | yes | Node pools + autoscaler live only here; all three placement paths covered. |
| `blis replay` | no | N/A | Replay rejects `node_pools` with `logrus.Fatalf`; `KVAutoCalc` left zero-value (off). INV-13 parity N/A. |
| `blis observe` | no | N/A | Dispatches to a real server; no simulated KV cache / node pools. |

## Invariants Affected

- **INV-P2-1 (pool-config consistency):** restored for node-pool placement. Evidence/verification note added in `invariants.md`.
- **INV-4 (KV conservation):** preserved per instance.
- **INV-6 (determinism):** preserved — recalc is a pure function of config; zero-value `KVAutoCalc` = no-op = byte-identical to prior builds; no map iteration for output.
- **INV-13 (run/replay parity):** N/A (node pools are run-only).

## Test Plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (all packages pass)
- [x] `golangci-lint run ./...` (0 issues)
- [x] Behavioral tests in `sim/cluster/kv_autocalc_test.go`:
  - BC-1 startup per-GPU capacity (distinct H100 vs L40S; each == an independent `CalculateKVBlocks` — a two-way law, not a golden constant)
  - BC-2 deferred placement (`NodeReadyEvent`)
  - BC-3 autoscaler scale-up
  - BC-4 precedence (explicit capacity → uniform global)
  - BC-5 zero-value inert
  - BC-6 `MaxModelLen` auto-cap
  - BC-7 graceful fallback (missing memory / calc error, no panic)
- [x] End-to-end CLI run verified (log lines above); precedence verified (`--total-kv-blocks` suppresses recalc).


namasl:
@claude /blis-pr-review

namasl:
Addressed the non-blocking test-coverage suggestion from the review: added `TestNoNodePools_EnabledIsInert` verifying that `KVAutoCalc.Enabled=true` with no node pools is inert (instances keep the global capacity). No production code changed.

@claude /blis-pr-review

namasl:
Addressed the remaining non-blocking review follow-ups:
- Added tests: `AdapterReservedBytes` forwarding, `MaxModelLen=0` stays unconstrained after recalc, and deferred placement with two distinct GPU pools.
- Documented in `configuration.md` that `gpu_type` is the hardware-identity key (use distinct types for variants that differ in `gpu_memory_gib`; first-match keying matches the existing `cost_per_hour`/`hw_config_by_gpu` behavior).

No production code changed — tests + docs only.

@claude /blis-pr-review

sriumcp:
@claude /blis-pr-review does this PR correctly handle all the comments here? https://github.com/inference-sim/inference-sim/issues/1522#issuecomment-5199809980

sriumcp (approved):
lgtm

Issue #1517, opened 2026-08-04
Final saturation classification: uniform last-window plurality reducer over detector traces


susiejojo:
> **Sequence:** Depends on #1516 (which emits per-detector verdict **traces** and puts **nothing** saturation-related on stdout) and composes with #1519 (bank / `--detectors all`). This issue adds the **final saturation label back on stdout**, derived **uniformly from the trace** by one reducer — the same rule for a single detector (#1516) and for the whole bank (#1519), rather than each detector's own `Classify`.

**What problem does this solve?**

As of #1516, saturation runs emit per-detector verdict **traces** to `--saturation-report` and put **nothing** on stdout — there is no single end-of-run label anywhere (whether `--detectors <name>` or, from #1519, `--detectors all`). But operators/experiments still need a one-line answer: "was this run saturated, per detector?"

Previously each detector computed its own final verdict in its `Classify` method, using a *different* algorithm per detector (threshold: mean-E2E; composite: rate-deficit+trend; backlog-drift: whole-trace drain-ratio regression). That makes cross-detector comparison unfair — a verdict difference could reflect the detector's *final-labeling rule* rather than its per-instant signal. It also duplicates "how do we decide the headline label" logic across detectors.

---

## Ground truth — types this issue builds on (already on `main`, do not invent new ones)

The reducer consumes the **existing** trace record type from #1516 (`sim/saturation/sink.go`). There is **no `VerdictRecord`** — use `TraceRecord`:

```go
// sim/saturation/sink.go — EXISTING, unchanged by this issue
type TraceRecord struct {
    Timestamp int64  `json:"timestamp"` // MICROSECONDS (µs). Populated as int64(ArrivedAt * 1e6).
    Detector  string `json:"detector"`  // detector name, e.g. "composite" — this is the grouping key
    Result    Result `json:"result"`
}
type Result struct {                    // EXISTING
    Level      Level              `json:"level"`      // STABLE | BACKLOGGED | OVERLOADED
    Score      float64            `json:"score"`
    Confidence float64            `json:"confidence"`
    Signals    map[string]float64 `json:"signals"`
}
```

The records are obtained from the **existing** collector already wired in `cmd` for #1516/#1519:

```go
// sim/saturation/sink.go — EXISTING
func (c *InMemoryCollector) Records() []TraceRecord  // stream order, tagged by Detector
```

[...]

Issue #1413, opened 2026-06-09
feat(metrics): goodput wiring — CLI flags, MetricsOutput population, run/replay/observe/calibrate (#1409 PR2)


namasl:
> Sub-issue of #1409. Implements **PR2: Wiring** — `cmd/` changes that turn the foundation built in #1412 into user-visible goodput numbers. **Depends on #1412 / PR #1410 merging first.**

## Why this is real

After #1412 lands, every piece needed to compute goodput exists in `sim/`:
- `cluster.SLOAttainmentMultiDim(results, injectedByClass, targets)` is callable
- `RawMetrics.InjectedByClass` carries the per-class denominator
- `WorkloadSpec.GoodputSLOTargets` and `TraceHeader.GoodputSLOTargets` carry the thresholds
- `MetricsOutput.GoodputRPS`, `SLOAttainment`, `PerClass` are reserved JSON slots

…but **nothing reads or populates any of it**. Run a build off the PR1 branch with a workload spec that contains `goodput_slo_targets` and stdout will not contain `goodput_rps`. That's the gap PR2 closes.

Verify the gap on a post-PR1 checkout:
```bash
grep -rn 'SLOAttainmentMultiDim\|GoodputSLOTargets' cmd/ --include='*.go'
```
Returns nothing on the PR1 branch.

## Proposed approach (PR2 scope)

PR2 is `cmd/`-only. No additional changes to `sim/`. The four BLIS commands each gain three flags, target-resolution logic, and goodput emission; one helper change in `cmd/root.go` fixes the per-SLO print suppression for single-class workloads.

### What lands

| File | Change |
|------|--------|
| `cmd/root.go` (`blis run`) | Add `--slo-ttft`, `--slo-itl`, `--slo-e2e` flags. After simulation, build `targets map[string]workload.SLODimTargets` with precedence **CLI > workload spec** (no trace header in `run`). Call `cluster.BuildLatencyResults(cs.AggregatedMetrics())` and `cluster.SLOAttainmentMultiDim(results, cs.InjectedByClass(), targets)`. Populate `MetricsOutput.GoodputRPS = good/runtimeSec`, `MetricsOutput.SLOAttainment = overall`, `MetricsOutput.PerClass = perClass`. Write `targets` into the `TraceHeader` when `--trace-output` is used. Fix `printPerSLOMetrics` suppression guard (currently `len(sloMetrics) <= 1` skips the section; with goodput targets configured it must print even for one class) and add the ITL line. |
[...]

PR #1524, opened 2026-08-06
feat(saturation): --detectors all | subset detector bank fans one replay to N detectors (#1519)

### LINKED ISSUE #1519: Detector bank: --detectors all | subset-list fans one replay out to selected detectors
> **Sequence — part 2 (after #1515 → #1516).** Depends on #1516 (unified `Detector` interface, `SaturationConfig`, `--detectors <name>`, `--saturation-config`, `--saturation-report`, and the single-detector streaming-trace machinery: `TraceSink`/`CollectingSink`, event replay, `WriteCombinedReport`/`ValidateReportPath`). This issue adds the **detector bank** — fan the same replay out to the roster — behind **`--detectors all`** or an arbitrary **comma-list subset** (absorbs #1518). Prereq for #1517.

#1516 can record one detector's per-event trace (`--detectors <name> --saturation-report`), but you still can't run detectors **together on identical traffic** — comparing them means separate runs where each sees different traffic. This issue adds the bank: one deterministic replay, fanned out to every detector, so all are scored on byte-identical events in a single run.

**Proposed solution — the bank multiplexes #1516's replay.**

Nearly all the machinery already exists from #1516; this issue adds the multiplexing layer:
- `Bank` — a struct that *holds and drives* the roster and satisfies only `sim.BatchClassifier` (reimplements no `Detector` method). `NewBank(cfg, sink)` builds the fixed roster from #1516's `SaturationConfig`. Its replay is #1516's single-detector driver generalized to **fan each event out to N detectors** (`fanout`), reusing the same `TraceSink`, `WriteCombinedReport`, and `ValidateReportPath`.
- **`--detectors all`** *or* **`--detectors <name,name,...>`** — new accepted values of #1516's `--detectors` flag. `all` runs the full roster; a comma-list runs exactly the named subset through the bank (same replay, only the selected detectors driven). A single `<name>` (from #1516) still runs one detector via streaming replay. Unknown name → hard error listing valid names (R1); a bare single name that could run either way is treated as #1516's single-detector path for byte-identical continuity.
- **`--saturation-report`** now accepts the bank's multi-detector `{trace}` (records tagged by detector name) in addition to #1516's single-detector trace.

```bash
./blis run --model qwen/qwen3-14b --detectors all --saturation-report saturation.json
./blis run --model qwen/qwen3-14b --detectors all --saturation-config saturation.yaml
```

```
   saturation.Detector  ← THE one interface (from #1516)
       ┌──────────────┬───────┴────────┐
   CompositeDetector  ThresholdDetector  BacklogDrift
   (streaming)        (streaming)        (streaming)
       └──────────────┴────────┬────────┘
                    held & driven by
                          │
                   saturation.Bank  (struct; implements sim.BatchClassifier only)
                          │  Classify(...) → multi-detector verdict trace (NO stdout map)
```

**Bank methods:** `Classify(requests, totalArrivals)` (sole public method — it satisfies `sim.BatchClassifier`; replays once, fans each event out to every detector's `Observe`/`Detect`, writes trace records), `Close()`, private `replayEvents`/`fanout`. (Note: the *bank* keeps a `Classify` method for the `BatchClassifier` seam; the *detectors* no longer have one after #1516 — the bank drives them purely by `Observe`/`Detect`.)

**Requirements:**
- **`--detectors all` or a comma-list triggers the bank** (off by default); drives exactly the selected detectors over one identical, sorted event sequence — reusing #1516's replay ordering `(t_us, type, id)`. `all` ≡ the full roster. Only selected detectors run; the trace contains records only for them. Selecting a subset yields byte-identical records for those detectors as `all` would (selection filters *which* run, never *how* they see traffic — INV-6/INV-13). Each selected detector's `--saturation-config` block is validated before the run (R1); a config block for an unselected detector is ignored.
- **`--saturation-report`** writes the multi-detector `{trace}` via #1516's `WriteCombinedReport` (sorted detector + signal keys, INV-6); requires a detector selection; unwritable path fails up front.
- **`calibrate` is untouched** (no saturation wiring on `main`).
- **stdout carries NOTHING saturation-related under `--detectors all`** — the bank's output is the trace (file-only); `MetricsOutput.Saturation` stays unset ⇒ `omitempty` drops it. (The single-detector `--detectors <name>` stdout `Result` from #1516 is unaffected. The per-detector final-label map on stdout returns in #1517.)
- **Vacuous-pass guard** — `TestBank_ClassifyActuallyReplaysEvents`: a streaming detector observes exactly `2*N` events.
- **Shared helper (R23)** — extend #1516's helper to resolve `all`/subset → `Bank`, `<name>` → single detector, on run/replay AND observe (observe still uses its own `TraceRecordsToRequestMetrics` input and passes nil into `printObserveMetrics`, per #1516).
- **Determinism (INV-6) / parity (INV-13)** — identical input+config ⇒ byte-identical report; run vs. replay-of-exported-trace ⇒ byte-identical.

- [x] Core simulator (`sim/`) — new `sim/saturation/bank.go` (`Bank` + `NewBank` roster + `fanout`); reuses #1516's `sink.go`/config/report writer
- [ ] Cluster / Workload / KV / Trace
- [x] CLI (`cmd/`) — accept `--detectors all`; route through #1516's shared helper (R23)
- [ ] New package needed — no

**Behavioral contracts (sketch)**
- `--detectors all --saturation-report r.json` ⇒ `{trace}` with records from every detector; `--detectors composite,threshold` ⇒ trace has only those two (no backlog-drift work done). Both non-empty, deterministic.
- `--detectors all` ≡ `--detectors composite,threshold,backlog-drift` (byte-identical output). A subset detector's records are byte-identical to its records under `all`.
- Unknown name in the list ⇒ hard error naming it + valid names (R1).
- `--saturation-report` without a selection ⇒ hard error.
- Bank does not alter sim results (BC-8): every non-saturation stdout field byte-identical with/without `--detectors all`.
- Run/replay parity (INV-13): exported trace replays byte-identically.
- Trace never on stdout.

- **Depends on #1516** (reuses its interface/config/replay/sink/report). **Part 2.**
- **Absorbs #1518** (`--detectors` subset lists) — the comma-list is a trivial extension of the flag parsing this issue already does. Prereq for #1517 (stdout final-label map). All detectors already stream (#1515), so the bank's fan-out has no special cases.

---

**Design update (post-review, #1524 discussion): the `sim.BatchClassifier` coupling is transitional.**

This issue's design has `Bank` satisfy `sim.BatchClassifier`, but that seam is now understood to be *temporary scaffolding*, not part of the desired end state:

- The seam exists to break a `sim ↔ saturation` import cycle so **`sim`** can call a classifier. But the bank is driven from **`cmd`** (`saturationTracer.trace → Bank.Classify`), and `cmd` already imports both packages — so no interface is needed at the actual call site. In #1524 `Bank.Classify` returns `nil` and ignores `totalArrivals`; it is a side-effecting driver wearing a classifier signature.
- **#1517** now adopts the goodput stdout pattern (`cmd` mutates `MetricsOutput.Saturation` between `BuildOutput` and `EmitOutput`) and therefore **retires `sim.BatchClassifier`**: delete `sim/classifier.go`, drop the `nil` param from `BuildOutput`/`SaveResults`, and rename `Bank.Classify` → `Bank.Run() error`.

**Scope for #1524 (this issue) is unchanged** — the bank ships satisfying `sim.BatchClassifier` as specified above, for byte-identical continuity with #1516. The seam removal is deliberately deferred to #1517, where the stdout wiring makes it a natural, focused change rather than widening this PR into `sim/metrics.go` and ~30 call sites. Read the `BatchClassifier` references in this issue as "the transitional seam #1517 removes," not "the intended architecture."

susiejojo:
## Summary

Closes #1519 (part 2 of the saturation-detector sequence, after #1515 → #1516).

Adds the **detector bank** on top of #1516's single-detector streaming trace. `--detectors all` runs the full roster; a comma-list (e.g. `composite,threshold`) runs exactly the named subset. Both route through a new `sim/saturation` `Bank` that holds and drives the roster over **one deterministic replay**, fanning each event out to every selected detector — so all detectors are scored on a byte-identical event sequence in a single run, instead of separate runs where each sees different traffic. A single `<name>` still uses #1516's single-detector streaming path for byte-identical continuity.

```bash
./blis run --model qwen/qwen3-14b --detectors all --saturation-report saturation.json
./blis run --model qwen/qwen3-14b --detectors composite,threshold --saturation-report saturation.json
```

## What changed

- **`sim/saturation/bank.go`** (new): `Bank` — a struct that *holds and drives* the roster and satisfies **only** `sim.BatchClassifier` (reimplements no `Detector` method). `NewBank(names, cfg, sink)` validates/de-dups names and orders the roster canonically (`composite`, `threshold`, `backlog-drift`); `Classify` replays once and `fanout` fans each event out to every detector's `Observe`/`Detect`. Reuses #1516's `TraceSink`, `WriteCombinedReport`, `ValidateReportPath`. `AllDetectorNames()` expands `all`.
- **`sim/saturation/replay.go`**: extract `buildSortedEvents` so `ReplayOneDetector` (single, #1516) and the `Bank` (multi, #1519) fan out over an identical, identically-sorted `(t_us, type, id)` event sequence.
- **`sim/saturation/config.go`**: split `buildDetector` (no ownership check — the bank ignores a config block for an unselected detector, since another detector may own it) from `BuildDetector` (single-detector path keeps #1516's block↔detector ownership enforcement). A value error in a *selected* detector's own block still surfaces (R6).
- **`cmd/saturation.go`**: `resolveSaturation` now returns a `saturationTracer` that unifies the single-detector and bank paths behind **one** `trace()` call site (R23); run/replay/observe all updated to use it. Handles `all`/comma-list parsing, unknown-name hard error (R1), empty-selection error, and up-front report-path validation.

## Contracts verified end-to-end with the binary

- **Determinism (INV-6)**: two identical `--detectors all` runs → byte-identical trace.
- **`all` ≡ scrambled full comma-list** → byte-identical trace (selection order/spelling never change output).
- **Subset ≡ under-`all`**: a subset detector's records are byte-identical to its records filtered out of the `all` trace (selection filters *which* detectors run, never *how* they see traffic).
- **Run/replay parity (INV-13)**: exported trace replayed with `--detectors all` → byte-identical.
- **BC-8**: stdout byte-identical with/without `--detectors` — saturation stays file-only (`omitempty`); no stdout label until #1517.
- **Errors**: unknown name (single or in a list) is a hard error listing valid names (R1); `--saturation-report`/`--saturation-config` without `--detectors` error.

## Tests

- **Vacuous-pass guard** (issue-mandated): `TestBank_ClassifyActuallyReplaysEvents` — a streaming detector observes exactly `2*N` events.
- Bank unit tests: unknown-name error, empty-selection error, canonical order + de-dup, `all`==explicit-list, subset==under-`all`, determinism, foreign-block-ignored, selected-block-value-error-surfaces.
- CLI tests: `all`/comma-list route to the bank, single name to #1516 path, block-ownership preserved for single vs ignored for bank, trace-file record counts and byte-identity.

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` (v2.9.0) all pass.

## Not in scope

`calibrate` is untouched (no saturation wiring on `main`). The stdout per-detector final-label map returns in #1517.

---

## Note on the `sim.BatchClassifier` seam (transitional — retired in #1517)

Per the review discussion below, the `Bank`→`sim.BatchClassifier` coupling in this PR is **intentional transitional scaffolding**, kept for byte-identical continuity with #1516. It is *not* the intended end state:

- `Bank.Classify` here returns `nil` and ignores `totalArrivals` — it is a side-effecting driver, called directly from `cmd` (`saturationTracer.trace`), never through `sim`. Since `cmd` imports both `sim` and `sim/saturation`, the call site needs no interface; the seam only exists to break a `sim ↔ saturation` cycle that this call path doesn't actually cross.
- **#1517** adopts the goodput stdout pattern (`cmd` mutates `MetricsOutput.Saturation` between `BuildOutput` and `EmitOutput`, exactly as `emitGoodput` does for #1413) and **retires the seam**: delete `sim/classifier.go`, drop the `nil` param from `BuildOutput`/`SaveResults`, rename `Bank.Classify` → `Bank.Run() error`.

**This PR's scope is unchanged** — deferring the removal to #1517 keeps this diff focused on the bank (`sim/saturation/` + `cmd/`) instead of widening it into `sim/metrics.go` and ~30 call sites. Issue #1519 has been annotated to match; issue #1517 owns the removal.


susiejojo:
## 🔍 Self-Review: PR #1524 (detector bank, #1519)

Ran a multi-agent convergence review (general code, tests, silent-failure, type-design, comments) plus manual checks of the three BLIS-specific dimensions (cross-path parity, preemption safety, timeout/deadline). Every finding below was reproduced against the built binary. This is a self-review — no code was changed.

### Verdict: **Approve with one recommended fix** (not a merge-blocker)

The PR correctly and fully implements #1519. All six issue requirements are met and covered by behavioral tests, including the mandated vacuous-pass guard and the `all ≡ explicit-list` / `subset ≡ under-all` byte-identity laws. `go build`, `go test ./...`, and `golangci-lint` (v2.9.0) all pass. No Critical issues. No R6 violation (`sim/saturation` is `Fatalf`/`panic`-free). Determinism (INV-6), run/replay parity (INV-13), and stdout parity (BC-8) all verified end-to-end. The one finding below is a pre-existing-style hardening gap, not a correctness bug.

---

### ⚠️ Important — config-block ownership is not enforced on bank (multi-name/comma) selections (R1 parity gap)

`isBankSelection` (cmd/saturation.go:153-155) routes any selection containing a comma through the bank, which builds detectors via `buildDetector` (no `checkBlockOwnership`) rather than `BuildDetector`. Result: a config block for a **detector the user did not select** is silently ignored on a bank selection, while the identical mistake on the single-detector path is a hard error.

Reproduced against the binary:

| Command (+ a `threshold:` block in the config) | Behavior |
|---|---|
| `--detectors composite` | ❌ hard error: *"threshold block is not valid for --detectors composite"* |
| `--detectors composite,` *(trailing comma)* | ✅ silently accepted, trace written, **no warning** |
| `--detectors composite,threshold` + a `backlog_drift:` block | ✅ `backlog_drift` block silently dropped |

The trailing-comma case is the sharpest: one invisible character flips the validation contract. This is exactly the R1 "silent drop of user tuning" the single-detector path's `checkBlockOwnership` exists to prevent. Two independent review agents flagged it.

**Note:** for `--detectors all` the current behavior is *correct by design* — every block is owned by some selected detector, so no foreign block is possible. The gap is specific to explicit subset selections.

**Recommended fix:** for a bank selection that is **not** `all`, enforce that every present config block belongs to some selected detector (generalize `checkBlockOwnership` to the selected set), erroring otherwise. Alternatively, if a selection collapses to a single name after trim/dedup, route it through the single-detector path so ownership is enforced. This restores single-vs-bank parity while preserving the shared-config convenience for `all`.

---

### Suggestions (non-blocking)

- **Test — zero/degenerate input (R20) is unguarded.** `Bank.Classify` with zero requests, and `WriteCombinedReport`'s `nil → []TraceRecord{}` empty-file branch, have no direct test. The docs promise a valid `{"trace":[]}`; add a bank test with `[]sim.RequestMetrics{}` and a CLI test asserting the file is exactly `{"trace":[]}`.
- **Test — BC-8 stdout parity has no automated regression guard.** The property holds today (`BuildOutput` passed `nil` + `omitempty`), but nothing pins it. This matters specifically because **#1517 will re-wire that exact seam** to repopulate `output.Saturation`. A subprocess test diffing stdout of `run --detectors all --saturation-report …` vs plain `run` at the same seed would lock it.
- **Type — `saturationTracer`'s "exactly one of detector/bank" invariant is doc-only** (cmd/saturation.go:38-42). The `else` branch in `trace()` would nil-deref `ReplayOneDetector(nil, …)` if both were ever nil. A one-line guard (`if (t.bank == nil) == (t.detector == nil) { … }`) converts a latent nil-deref into an explicit contract violation cheaply.
- **Comment rot — `ReplayOneDetector` doc (replay.go:21-33)** still calls itself *"the single uniform loop the whole feature relies on."* No longer true now that `Bank` has its own `fanout` loop; the genuinely shared piece is `buildSortedEvents`. Reword.
- **Comment — `totalArrivals` rationalization (cmd/saturation.go:61-65)** claims passing `len(requests)` "matches what the single-detector path sees," but `ReplayOneDetector` takes no such arg, and `BatchClassifier` documents `totalArrivals` as *injected* (completed + timed-out + dropped), not completed. Harmless now (bank ignores it), but #1517 could inherit the wrong count — state the caveat honestly.
- **Minor doc precision:** "comma-list of ≥2 names → bank" is imprecise (any comma routes to the bank); CLAUDE.md attributes `buildSortedEvents` to #1516 though it's a new #1519 extraction; `bank.go:41-43` "completed run's events" under-describes the observe path.

---

### Verified strengths / non-issues

- **Cross-path parity (run/replay/observe):** ✅ full parity. `registerDetectorFlags` on all three commands; identical `resolveSaturation → nil-guard → satTracer.trace()` sequence; only the input `[]RequestMetrics` differs (R23 shared transformation).
- **Preemption safety:** ✅ N/A — the diff touches zero request-lifecycle/metrics/state code; the bank is pure post-hoc analysis over completed `RequestMetrics`.
- **Timeout/deadline consistency:** ✅ N/A — no timeout/deadline/horizon code touched.
- **Bank satisfies `sim.BatchClassifier` only** (compile-time assert at bank.go:56); does not accidentally satisfy `Detector`. ✅ requirement #3.
- **Determinism / byte-identity:** canonical `rosterOrder` sort + dedup + shared `buildSortedEvents` + json-sorted `Signals` maps (R2). Detector independence makes `fanout` order affect only record interleaving — the basis for subset ≡ under-`all`. ✅
- **`AllDetectorNames()` returns a defensive copy** (R8). ✅
- **Vacuous-pass guard is genuine** — `countingDetector` asserts exactly `2*N` Observe/Detect + 1 Reset.

---

susiejojo:
## ✅ Review findings addressed (commit `2fb6e8b`)

All findings from the [self-review above](https://github.com/inference-sim/inference-sim/pull/1524#issuecomment-5207839600) are resolved. Every invariant re-verified with the binary; `go build`, `go test ./...`, and `golangci-lint` (v2.9.0) pass.

### ⚠️ Important — config-block ownership on bank selections — **FIXED**
Added `checkBlockOwnershipSet` (config.go): a tuning block whose owning detector is not in the selected set is now a hard error, matching the single-detector path (R1). `NewBank` calls it after validate/dedup/canonicalize. `all` selects every owner so it passes trivially; only subsets that omit a detector whose block was supplied error.

Verified end-to-end:
| Command (+ `threshold:` block) | Before | After |
|---|---|---|
| `--detectors composite,` *(trailing comma)* | silently accepted | ❌ hard error ✓ |
| `--detectors composite,threshold` | (n/a) | ✅ accepted (threshold selected) |
| `--detectors all` | ✅ | ✅ (unchanged) |

Inverted the two tests that asserted the old "ignored" behavior; added trailing-comma, selected-block-accepted, and all-accepts-every-block cases.

### Suggestions — all addressed
- **R20 zero/degenerate input** — added `TestBank_ZeroRequestsEmptyTrace` + `TestSaturationTracer_ZeroRequestsWritesEmptyTrace`: zero requests write a valid `{"trace":[]}` (not `{"trace":null}`) on both paths.
- **BC-8 stdout parity guard** — new `saturation_stdout_parity_test.go` re-execs `run` plain vs `--detectors composite` vs `--detectors all` at the same seed and requires **byte-identical stdout**. Locks the exact seam #1517 will re-wire.
- **`saturationTracer` exactly-one invariant** — `trace()` now guards `(bank==nil)==(detector==nil)` and errors loudly instead of nil-deref'ing `ReplayOneDetector(nil, …)`.
- **Comment rot** — `ReplayOneDetector` doc corrected ("single uniform loop" → single-detector drive loop; bank has its own fanout, shares `buildSortedEvents`).
- **`totalArrivals` comment** — corrected: `len(requests)` is the *completed* count, not the injected total the `BatchClassifier` contract documents; explicitly flagged for #1517 to supply the true arrival total.
- **Doc precision** — comma routing (any comma → bank, incl. trailing), `buildSortedEvents` attributed to #1519 (not #1516), `Bank` described over run/replay/observe metrics.

### Not changed (deferred, as noted in the review)
The `interface{}` return of `BatchClassifier` (nil boxed) is inherited seam debt — correctly resolved when #1517 gives the return an actual shape, not in this PR.

susiejojo:
@claude /blis-pr-review

cmd/saturation.go
```diff
@@ -11,65 +11,135 @@ import (
...
+	detector  saturation.Detector // single-detector path (#1516); nil when the bank is used
+	bank      *saturation.Bank    // multi-detector path (#1519); nil when a single detector is used
+}
+
+// trace streams the resolved selection over requests and writes the per-event
+// trace to --saturation-report. It is a no-op when no report path was given (the
+// trace would be discarded anyway), mirroring #1516. The []sim.RequestMetrics
+// input is the only per-command difference (sim-derived for run/replay,
+// server-derived for observe); the pipeline is identical.
+func (t *saturationTracer) trace(requests []sim.RequestMetrics) error {
```
jgchn, line 57: The `trace` method reads `saturationReport` (line 52) and `detectorName` (line 60) from package-level vars rather than storing them as fields at construction. This creates a contract that globals remain set between `resolveSaturation()` and the later `trace()` call. In the current cobra lifecycle it works, but it's fragile. If a future refactor clears globals between construction and use, `trace()` silently no-ops or logs the wrong name.

Consider storing reportPath and selection on the struct at construction time.

sim/saturation/bank_test.go
```diff
@@ -0,0 +1,273 @@
...
+		t.Fatalf("record count differs: %d vs %d", len(a), len(b))
+	}
+	for i := range a {
+		if a[i].Timestamp != b[i].Timestamp || a[i].Detector != b[i].Detector ||
+			a[i].Result.Level != b[i].Result.Level || a[i].Result.Score != b[i].Result.Score ||
+			a[i].Result.Confidence != b[i].Result.Confidence {
+			t.Errorf("record %d differs: %+v vs %+v", i, a[i], b[i])
+		}
+	}
+}
```
jgchn, line 303: The helper compares `Timestamp`, `Detector`, `Level`, `Score`, `Confidence`, but not the `S`ignals map. This means `TestBank_AllEqualsExplicitList` and `TestBank_SubsetMatchesRecordsUnderAll` don't verify signal-level parity at the unit layer. It's covered indirectly by the CLI-level `TestSaturationTracer_AllEqualsExplicitList` which compares serialized JSON bytes, but a `reflect.DeepEqual` on `Signals` here would make the unit test self-sufficient.

cmd/saturation_stdout_parity_test.go
```diff
@@ -0,0 +1,81 @@
...
+		switch v {
+		case "plain":
+			// no saturation flags
+		case "single":
+			args = append(args, "--detectors", "composite", "--saturation-report", os.Getenv("BLIS_BC8_REPORT"))
+		case "bank":
+			args = append(args, "--detectors", "all", "--saturation-report", os.Getenv("BLIS_BC8_REPORT"))
+		}
+		rootCmd.SetArgs(args)
+		_ = rootCmd.Execute()
```
jgchn: If `Execute()` fails without calling `logrus.Fatalf` (e.g., a cobra flag-parsing error), the subprocess exits 0 with partial/empty stdout, and the test compares that against the other variants. The plain `== ""` guard (line 70) catches total failure but not partial output from a non-fatal error. Safer:

```go
if err := rootCmd.Execute(); err != nil {
    os.Exit(1)
}
```

CLAUDE.md
```diff
@@ -422,14 +432,16 @@ BLIS includes post-hoc saturation detection for analyzing completed runs. This i
...
+**CLI flags** (`run`, `observe`, `replay`; #1516, #1519):
+- `--detectors <selection>`: empty = off. A single name (`composite`, `threshold`, `backlog-drift`) runs #1516's single-detector streaming path. `all` runs the full roster; a comma-list (e.g. `composite,threshold`) runs exactly the named subset. `all` and comma-lists route through the **detector bank** (#1519). Unknown name — single or inside a list — is a hard error listing valid names (R1).
+- `--saturation-config <path>`: strict-YAML tuning file with optional `threshold:` and `backlog_drift:` blocks. composite has no block. For a **single** detector the config must carry only that detector's block — a foreign block errors (no silent drop, R1). For the **bank**, a block for an *unselected* detector is ignored (several detectors legitimately share one config); a value error inside a *selected* detector's own block still errors. Absent block = defaults; a partial block overrides only named fields; an unknown key or out-of-range value errors naming the field; an empty file = all defaults.
```
jgchn: This is wrong. Commit 2 added `checkBlockOwnershipSet` (sim/saturation/config.go:163) which makes a block for an unselected detector a hard error. The tests confirm: `TestResolveSaturation_BankUnselectedBlockErrors`, `TestNewBank_UnselectedBlockErrors`. The CLAUDE.md text was written for commit 1's behavior and not corrected in commit 2.

Similarly, the inline comment at cmd/saturation.go:113 still says:

```
// NewBank validates names, de-dups, canonicalizes order, and builds each
// detector; a foreign config block (for an unselected detector) is ignored,
```

This also contradicts the code one line later that calls `NewBank` which calls `checkBlockOwnershipSet` and returns an error.

Fix: Both spots should say `"is a hard error (R1)" instead of "is ignored"`.

cmd/saturation.go
susiejojo: Fixed in c0fc6af7. `saturationTracer` now captures `reportPath` and `selection` as fields at construction in `resolveSaturation`; `trace()` no longer reads the flag globals, so it's self-contained once resolved.

sim/saturation/bank_test.go
susiejojo: Fixed in c0fc6af7. `assertRecordsEqual` now compares `Result.Signals` via `reflect.DeepEqual`, so the unit tests verify signal-level parity directly. Applied the same to the CLI-level `SubsetMatchesRecordsUnderAll` comparison for consistency.

cmd/saturation_stdout_parity_test.go
susiejojo: Fixed in c0fc6af7. The subprocess now `os.Exit(1)` on a non-fatal `Execute()` error, so partial/empty stdout from a flag-parse failure fails the test instead of being compared as valid.

CLAUDE.md
susiejojo: Good catch — fixed in c0fc6af7. Both spots (CLAUDE.md and the cmd/saturation.go inline comment) now say the block is a hard error (R1) via checkBlockOwnershipSet, not ignored. Verified with the binary: `--detectors composite,threshold` + a backlog_drift block errors with "backlog-drift is not among the selected detectors".

susiejojo:
## 🔁 Re-review after fixes — verdict: **Ready to merge**

Re-ran the four-agent convergence review (general code, tests, silent-failure, comments) plus independent manual verification, focused on whether @jgchn's four comments and the earlier convergence findings are correctly and completely resolved, and whether the two fix commits (`2fb6e8b`, `c0fc6af7`) introduced any regression. `go build`, `go test ./...`, and `golangci-lint` (v2.9.0) all pass.

### @jgchn's four comments — all verified resolved

| # | Comment | Status |
|---|---|---|
| 1 | `trace()` reads flag globals | ✅ **Fixed** — `reportPath`/`selection` captured on the struct at construction; `trace()` reads no globals (grep-confirmed). No "used-before-set" hazard; both construction branches capture correctly. |
| 2 | `assertRecordsEqual` skips `Signals` | ✅ **Fixed** — now compares `Result.Signals` via `reflect.DeepEqual`; same applied to the CLI-level comparator. Unit tests are now self-sufficient for signal parity. |
| 3 | BC-8 subprocess exits 0 on non-fatal error | ✅ **Fixed** — `os.Exit(1)` on `Execute()` error, so a flag-parse failure now `Fatalf`s the parent instead of comparing partial stdout. |
| 4 | Stale "block is ignored" docs (called *wrong*) | ✅ **Fixed** — no remaining "ignored" claim about bank blocks; CLAUDE.md + inline comment now say "hard error (R1)" and match `checkBlockOwnershipSet` exactly. Verified with the binary. |

### Earlier convergence findings — confirmed still fixed
- **R1 silent-drop (config-block ownership on bank):** fully resolved. `checkBlockOwnershipSet` is called from the sole `NewBank` path, over the deduped/canonical set; `all` passes, subset-omitting-owner errors, trailing-comma enforces ownership. No bypass.
- **R20 zero-input, BC-8 parity, Signals parity, ownership inversion tests:** all present and behavioral.
- **R6** clean (no `Fatalf`/`os.Exit`/`panic` in `sim/saturation`); **R8** clean (`AllDetectorNames` returns a copy); **INV-6/INV-13** upheld via shared `buildSortedEvents`.

### Two minor observations (non-blocking, no action required for merge)
- **Test (low, 4/10):** the field-capture decoupling has no direct regression guard — every test sets globals then calls `trace()` immediately, so reverting `trace()` to read globals would still pass. A ~10-line test that clears the globals between `resolveSaturation()` and `trace()` would lock the invariant. Optional.
- **Pre-existing (not introduced here):** the byte-identity comparators use `==` on `Score`/`Confidence` (and now `DeepEqual` on `Signals`), which treat `NaN != NaN` — a NaN score would make two identical runs compare unequal. Current detectors don't emit NaN; noted only for completeness.

No Critical or Important findings remain. All four reviewer comments and all prior convergence findings are resolved with no regressions.

cmd/saturation.go
susiejojo: Added a regression guard in `feeb15b8`: `TestSaturationTracer_DecoupledFromGlobals` resolves the tracer, clears the flag globals, then traces — asserting the trace is still written to the path captured at construction. Verified non-vacuous: reverting `trace()` to read the global makes it fail. This locks the decoupling this comment asked for.

susiejojo:
@claude /blis-pr-review

susiejojo:
## Archon architecture check (run manually)

We ran [Archon](https://github.ibm.com/ai-native-systems/repoevolve) (`archon-go`) manually against this PR locally — comparing the merge-base with `main` (`18c2c926`) against the PR head (`feeb15b8`). Archon triages at **package altitude**: it ignores line-level churn and only fires when a dependency edge, public surface, or interface contract changes.

**Verdict:** `NEEDS ARCHITECTURE REVIEW — a boundary moved.` The one substantive item is that `saturation.Bank` now implements `sim.BatchClassifier` but no contract test binds that interface yet (the seam is wired in #1517, where a polymorphic contract test should land). Direction is correct (`cmd → sim/saturation → sim`), blast radius is 1 direct dependent (`cmd`), no cycles.

### Delta — what this PR changed architecturally

The entire delta at package altitude is a single new *implements* edge (green); both packages already existed (grey). +1144/−194 LOC collapse to one edge — the signature of a well-scoped change.

```mermaid
graph LR
  n0["sim"]
  n1["saturation"]
  n1 -. implements .-> n0
  linkStyle 0 stroke:#1a7f37,stroke-width:3px
  n1 -.-> n0
  linkStyle 1 stroke:#c8c8c8,stroke-width:2px
  classDef unchanged fill:#eef3fb,stroke:#4a6fa5;
  class n0,n1 unchanged
```

_Green = added edge, grey = unchanged, dotted = implements._

### Full architecture at PR head (for context)

```mermaid
graph LR
  n0["env:HF_TOKEN"]
  n1["flag:update-golden"]
  n2["inference-sim"]
  n3["cmd"]
  n4["sim"]
  n5["cluster"]
  n6["hash"]
  n7["testutil"]
  n8["tokenid"]
  n9["util"]
  n10["kv"]
  n11["latency"]
  n12["lora"]
  n13["saturation"]
  n14["trace"]
  n15["workload"]
  n2 --> n3
  n3 -. config .-> n0
  n3 --> n4
  n3 --> n5
  n3 --> n11
  n3 -.-> n12
  n3 --> n13
  n3 --> n14
  n3 --> n15
  n4 --> n6
  n4 -.-> n8
  n4 --> n9
  n5 --> n4
  n5 -. implements .-> n4
  n5 --> n10
  n5 --> n11
  n5 --> n14
  n5 -.-> n15
  n6 -.-> n8
  n7 -. config .-> n1
  n10 --> n4
  n10 -. implements .-> n4
  n10 --> n6
  n10 --> n9
  n11 --> n4
  n11 -. implements .-> n4
  n12 -. implements .-> n4
  n12 -.-> n4
  n13 -. implements .-> n4
  n13 -.-> n4
  n13 --> n15
  n15 --> n4
```

_Solid arrow = import, dotted = implements/config. `saturation` (n13) sits at the leaf — depended on only by `cmd`, depending only on `sim` and `workload`._

<sub>🤖 Archon run and comment generated with [Claude Code](https://claude.com/claude-code)</sub>

susiejojo:
## Archon architecture review (run manually)

Verdict: **✅ APPROVE** (near fast-track). Archon flags "a boundary moved" — that review is done below; the move is single, intentional, and low-risk.

**What changed:** one meaningful new edge — `sim/saturation → sim` (`Bank ⊨ sim.BatchClassifier`); plus a stdlib `strings` import. New surface: `NewBank`, `Bank`, `Bank.Classify`, `Bank.Close`, `AllDetectorNames`. No new/removed packages. Direction is correct (`cmd → sim/saturation → sim`); `BatchClassifier` already lived in `sim/` (from #1516) to avoid an import cycle.

**Blast radius:** minimal — `sim/saturation` has 1 direct dependent (`cmd`), 2 transitive. Nothing outside the `cmd → saturation` line is affected.

**Contracts:** zero failing contract tests. `saturation.Bank` is flagged as an uncovered implementer of `sim.BatchClassifier` — real, but it's 1 of 29 uncovered implementers repo-wide (Archon reports most factory-driven/multi-implementer interfaces this way), and `Bank` is currently driven as a concrete type (the seam is passed `nil` until #1517). PR also adds 30+ guards incl. a vacuous-pass guard.

**Intentional:** yes — the description explicitly scopes `Bank` to satisfy *only* `BatchClassifier` and defers the stdout wiring to #1517; the `var _ sim.BatchClassifier = (*Bank)(nil)` assertion documents it.

**Non-blocking follow-up (#1517):** when `Bank` is wired into `SaveResults` (classifier no longer `nil`), add a contract test that drives it *through* `sim.BatchClassifier` rather than the concrete type — closing the one seam Archon flagged.

<sub>🤖 Archon run + review generated with [Claude Code](https://claude.com/claude-code)</sub>

susiejojo:
@claude /blis-pr-review

susiejojo:
## Design observation: the `sim.BatchClassifier` seam is now speculative-only, and `Bank.Classify` doesn't really fit it

First — the bank itself looks great: one deterministic replay fanned to N detectors, canonical ordering, the `all ≡ scrambled-list` and `subset ≡ under-all` byte-identity contracts are exactly right. This is **not** a blocker, it's a seam-shape question worth settling before #1517 builds on it.

**What I'm seeing.** `Bank` satisfies `sim.BatchClassifier`, but in this PR the interface's whole purpose isn't exercised:

```go
func (b *Bank) Classify(requests []sim.RequestMetrics, _ int) interface{} {
    for _, det := range b.detectors { det.Reset() }
    b.replayEvents(buildSortedEvents(requests))
    return nil                       // returns nil, and ignores totalArrivals (_ int)
}
```

- It's called **directly from `cmd`** (`saturationTracer.trace` → `t.bank.Classify(...)`), never through `sim`. `sim.BuildOutput` still receives `nil`.
- It **returns `nil`** and **ignores its `totalArrivals` param** — so it's shaped like a `BatchClassifier` but behaves like a side-effecting driver (`run() error` would describe it honestly).

The reason `BatchClassifier` lives in `sim/` (not `sim/saturation/`) is to break the `sim ↔ saturation` import cycle so **`sim`** can call into a classifier without importing `saturation`. But `cmd` already imports both packages — it has no cycle to break — so the call from `cmd` needs no interface at all. Right now `BatchClassifier` is satisfied by exactly one type, called from a place that doesn't need it, returning `nil`. It's kept alive (incl. `var _ sim.BatchClassifier = (*Bank)(nil)`) purely to reserve the seam for #1517's stdout label.

**There's already a blessed pattern for "add a field to stdout" that doesn't need this seam** — goodput (#1413), one call away in `cmd/root.go`:

```go
clusterOutput := aggregated.BuildOutput("cluster", nil)
emitGoodput(&clusterOutput, aggregated, ...)   // cmd mutates the returned struct
aggregated.EmitOutput(clusterOutput, metricsPath)
```

`EmitOutput`'s own doc comment endorses it: *"Callers that want goodput fields populated should call BuildOutput, mutate the returned struct, then call this method (#1413)."* And `MetricsOutput.Saturation interface{}` (`json:"saturation,omitempty"`) is already a public, cycle-safe field. So the codebase currently has **two different patterns for the same job**: goodput fills its stdout field via cmd-mutation; saturation is being routed through a callback interface that's threaded as `nil` through ~25 call sites.

**Suggested update to #1517** (which already leans this way — its "Open question" prefers a standalone `saturation.ReduceFinal(records, window)` library function, and its CLI task is *"wire reducer output into stdout `saturation`"* from `cmd`):

```go
out := aggregated.BuildOutput("cluster", nil)
out.Saturation = saturation.ReduceFinal(collectedRecords, window)  // pure fn, cmd sets the field
aggregated.EmitOutput(out, metricsPath)
```

This lets #1517 produce the stdout label with **zero** reliance on `BatchClassifier`, unifies saturation with the goodput pattern, and lets `Bank.Classify` become an honest `run() error`. `sim.BatchClassifier` (and the `nil` param on `BuildOutput`/`SaveResults`) could then be **retired entirely** — unless there's a reason the reducer must run *inside* `sim.BuildOutput` that I'm missing.

**Question for the author:** is the `BatchClassifier` seam being kept for a reason beyond "#1517 will use it"? If #1517 adopts the goodput mutation pattern (its own preferred placement), the seam looks removable — and dropping it would delete an interface + a 25-site `nil` thread rather than build more on top of it.

cmd/saturation.go
```diff
@@ -11,65 +11,146 @@ import (
...
-//   - unknown detector name: lists the valid single names.
 //   - --saturation-config / --saturation-report given without --detectors.
-//   - bad config file (unknown YAML key, out-of-range param).
+//   - unknown detector name (lists the valid names, R1).
+//   - an empty selection after trimming (e.g. "," or ", ,").
+//   - bad config file (unknown YAML key, out-of-range param), including a config
+//     block that does not belong to the single selected detector (#1516 ownership).
 //   - unwritable --saturation-report path (checked up front so the run fails fast).
-func resolveSaturation() (saturation.Detector, *saturation.InMemoryCollector, error) {
+func resolveSaturation() (*saturationTracer, error) {
```
jgchn, line 109: If a user writes `--detectors "composite,all"` or `--detectors "all,threshold"`, `isBankSelection` routes it to `parseDetectorSelection` (because of the comma). Since `sel != "all"` (it contains a comma), the function splits on comma and passes the literal "all" as a detector name to NewBank. NewBank then rejects it with:

`unknown saturation detector "all"; valid: composite, threshold, backlog-drift`

This isn't a silent failure. It errors, but the error message is misleading because the user clearly knows about all as a valid keyword. A user might reasonably try `--detectors "composite,all"` expecting it to expand to the full roster.

Suggested fix (not a merge-blocker, but worth a follow-up): After trimming, check if any entry in names equals "all" and return a targeted error like "'all' cannot be combined with individual detector names; use `'--detectors all' alone"`.

jgchn (commented):
Minor nit, otherwise LGTM

cmd/saturation.go
susiejojo: Good catch — fixed in fddce610. `parseDetectorSelection` now detects `all` appearing among comma-separated entries (after trimming) and returns a targeted error:

> `--detectors "composite,all": "all" cannot be combined with individual detector names; use "--detectors all" alone to select the full roster`

instead of letting `NewBank` reject the literal `all` with the misleading `unknown saturation detector` message. Covered by `TestResolveSaturation_AllMixedWithNames` (asserts the error names `all` and is *not* the unknown-detector message), across `composite,all`, `all,threshold`, and a spaced `composite, all , threshold`. Valid paths (`all` alone, plain subset lists) unchanged.

jgchn (approved):

PR #1525, opened 2026-08-07
docs(blog): add latency-model post to the simulation series

susiejojo:
## Summary

Adds the third post in the *simulate before you scale* blog series, following **Why Simulate Before You Scale** and **The Physics of High-Fidelity Distributed Inference Platform Simulation**. It picks up the open question the second post left ("can the predictions be trusted?") and answers it at the latency-model level.

The post covers:

- **Two latency models, one basis** — the analytical roofline model and the default trained-physics model (roofline basis functions + learned scaling coefficients + additive overhead terms), with the exact equations from the paper rendered via MathJax in a collapsible aside.
- **How we train it** — a single coefficient set that generalizes across model architecture, GPU type, and TP degree, fit from **client-side latency only** (`observe` against an unmodified vLLM/llm-d deployment), contrasted with profiling-based simulators (AIConfigurator, Vidur, LLMServingSim) and the purely-analytical llm-optimizer.
- **How we evaluate it** — the `observe`/`replay`/`calibrate` loop on models and GPUs held out of training.
- **Accuracy, honestly** — 6.7% median E2E, ~200× speedup, P90/P99 within 7.5%/9.2%, plus the weak spots (TTFT, long-decode reasoning, NVLink all-reduce contention). Includes the roofline-vs-trained-physics ablation figure.
- **Where it's headed** — moving from point-estimate coefficients toward a Bayesian posterior (particle filters / ABC), teased for a forthcoming companion paper.

## Changes

- New post: `docs/blog/posts/how-blis-learns-latency.md`
- New figure: `docs/blog/posts/images/roofline-vs-trained-physics.png` (the ablation)
- Updated the previous post's "next in series" footer to link here

## Notes

- `mkdocs build --strict` succeeds; no new broken links or nav warnings introduced.
- Equations use `pymdownx.arithmatex` + MathJax (already configured); technical detail (equations, loss/training loop, tail-latency table) lives in collapsible `??? note` asides to keep the main read ~8 min.


docs/blog/posts/how-blis-learns-latency.md
```diff
@@ -0,0 +1,331 @@
...
+
+    | Statistic | Median MAPE |
+    |-----------|-------------|
+    | E2E mean  | 6.7%        |
+    | E2E P90   | 7.5%        |
+    | E2E P99   | 9.2%        |
+    | TTFT mean | 17.6%       |
+    | TTFT P90  | 21.3%       |
+    | TTFT P99  | 29.0%       |
+    | ITL mean  | 7.1%        |
```
jgchn, line 294: The graph shows mean errors for roofline vs. trained-physics. Can we do the same here? Is the median MAPE reported for roofline or trained physics? Also why median instead of mean for this table?

jgchn (commented):
Nice article!

docs/blog/posts/how-blis-learns-latency.md
susiejojo: Good catch, clarified the table in 7e229866. Answers to the three:

- **Trained-physics.** "BLIS" throughout means the default trained-physics model; roofline only shows up in the ablation. Relabeled the table to say so explicitly.
- **Why median (not mean).** The aggregation is *across the 36 experiments*, which span very different models/GPUs/workloads. A couple of hard configs (long-decode reasoning ~25%, TP=2 contention) would pull a mean well above the typical experiment, so the median reports the typical case. Note each row is a two-level aggregate: e.g. "E2E mean = 6.7%" is the *median over experiments* of each experiment's *mean-E2E* MAPE. Added a one-line note in the post too.
- **Roofline vs trained-physics here too.** The ablation figure right above already does that comparison, but only for the *means* (E2E/TTFT/ITL). This table's value-add is the *tail percentiles* (P90/P99), which the figure doesn't cover, and I don't have roofline P90/P99 aggregates (they're not in the paper's Table 3). Happy to add a roofline column if you have those numbers, but I didn't want to read them approximately off the chart.

jgchn: If you don't like the tail latency numbers for roofline, maybe we can just add a line that says since trained physics outperforms roofline, we are evaluating trained physics for tail latency accuracy.

susiejojo: Done in f0190cac. Added a line to the table aside: "Since trained-physics is the clearly stronger model above, we report tail-latency accuracy for it alone." That frames the roofline omission as intentional rather than a gap, so we don't need to invent roofline P90/P99 numbers.

susiejojo: Addressed.

jgchn (dismissed):
The blog lgtm!

jgchn:
Curious if there's reproduction artifacts?

susiejojo:
We can release the training and evaluation datasets, but the training code itself lies in a branch. @sriumcp can confirm if we can release the training code.

susiejojo:
Added links to the training and evaluation data matrix READMEs, as well as [evaluation repo link](https://github.com/inference-sim/sim-to-real-accuracy-validation/tree/main), this should improve transparency and reproducibility.

mtoslalibu:
Some feedback:
1. I would surface impressive result earlier, probably directly after the opening paragraphs. Right now the reader has to get through roofline, trained physics, equations, training procedure, etc. before discovering how impressive the result is.

2. Be more careful with “generalizes to unseen GPUs” So the 6.7% headline is true across the full 36 experiments, but a skeptical reader could interpret the final sentence as saying cross-GPU generalization itself is 6.7%.
```
Across all 36 held-out experiments, median E2E error is 6.7%. More importantly, without any A100 or L40S training data, cross-GPU E2E error remains 15.7% and 13.3%, respectively.
```

3. I would substantially shorten the Bayesian ending. This is the one section where the post suddenly feels like a research proposal instead of a BLIS results post.

susiejojo:
Thanks @mtoslalibu — addressed all three in f0d49d32:

1. **Surfaced the result earlier.** Added a headline teaser right after the opening paragraphs (above the `<!-- more -->` fold, so it also shows in the blog listing excerpt): 6.7% median E2E, ~200× speedup, and real-cluster validation — before any roofline/equations/training detail.

2. **Disambiguated cross-GPU generalization.** The bottom line now states the two numbers distinctly: 6.7% median is across *all 36* held-out experiments, while cross-GPU error is 15.7% (A100) / 13.3% (L40S). No sentence can now be read as claiming cross-GPU generalization is 6.7%.

3. **Shortened the Bayesian ending.** "Where this goes" went from three paragraphs to one, dropping the research-proposal tone while keeping the through-line (point estimates → posterior → ranges/incremental updates/policy error bars → companion paper).

Also linked the [admission-controller](https://ai-native-systems-research.github.io/ai-native-systems-research/blog/2026/05/13/from-simulation-to-production-how-an-ai-native-pipeline-discovered-a-better-admission-controller-for-llm-d/) and [soft-reflective-flow-control](https://ai-native-systems-research.github.io/ai-native-systems-research/blog/2026/07/27/from-simulation-to-production-part-ii-soft-reflective-flow-control-for-llm-d/) posts at the first mention of real-cluster validation. `mkdocs build --strict` passes.

mtoslalibu (approved):

PR #1533, opened 2026-08-07
fix(latency): accept SiTU-GLU ("situ") in KV-capacity SwiGLU allowlist

### LINKED ISSUE #1526: Bug: hidden_act "situ" (Kimi-K3 SiTU-GLU) rejected by KV-capacity SwiGLU allowlist
**Describe the bug**
`blis run` fatals during KV-capacity auto-calculation for any model whose HF `config.json` declares `hidden_act: "situ"` (Kimi-K3's SiTU-GLU activation). The run aborts before the simulation starts:

```
--latency-model: KV capacity auto-calculation failed: CalculateKVBlocks:
unsupported activation "situ"; only SwiGLU-family activations (silu, swiglu, geglu) are supported
```

**CLI command to reproduce**
```bash
./blis run --model moonshotai/Kimi-K3 --hardware H100 --tp 16
```
(with `model_configs/kimi-k3/config.json` carrying `text_config.hidden_act: "situ"`)

**Expected behavior**
The run should proceed. SiTU (Sigmoid Tanh Unit) / SiTU-GLU is Kimi-K3's successor to K2's SwiGLU. It is a **3-matrix gated GLU** (gate + up + down) with the identical weight/FLOP shape as SwiGLU — only the pointwise gate nonlinearity differs (parameterized by `activation_situ_beta` / `activation_situ_linear_beta`). BLIS's memory and compute models depend only on the MLP **matrix count**, not the gate function.

**Actual behavior**
Hard `logrus.Fatalf` from the `swiGLUActivations` allowlist guard.

**Root cause**
`sim/latency/kv_capacity.go:216` guards KV-capacity math against an allowlist:
```go
var swiGLUActivations = map[string]bool{
    "silu": true, "swiglu": true, "geglu": true, "": true,
}
```
`"situ"` is absent, so the guard fatals. This is the **only** behavioral branch on `hidden_act` in the entire codebase — verified exhaustively (6 call sites total):
- `computeModelWeightBytes` hardcodes `3 *` matrices and never reads `hidden_act`.
- `roofline.go` `mlpMatrixCount(hiddenAct)` explicitly discards its argument (`_ = hiddenAct; return 2`).
- `cmd/root.go:613` only tests `HiddenAct == ""` for a log line; `"situ"` takes the same path as `"silu"`.

So adding `"situ"` is both **necessary** (sole blocker) and **complete** (no arithmetic reads the value). The resulting KV block count is byte-identical to what the same config with `"silu"` would produce.

**Proposed fix**
Add `"situ": true` to the `swiGLUActivations` map (`sim/latency/kv_capacity.go:48`) with a comment recording why SiTU-GLU is SwiGLU-family.

None directly — this is a fail-loud guard being over-conservative for a valid 3-matrix gated activation.

**Additional context / caveat (out of scope for this fix)**
Kimi-K3 is a hybrid **linear-attention** model (`KimiLinearForCausalLM`, `linear_attn_config.full_attn_layers` covers only a subset of layers, `kv_lora_rank: 512` MLA-style compressed KV). BLIS's KV-capacity and step-time models assume **uniform full attention on every layer**, so numbers for this model remain approximate even after the activation fix. Tracked separately if we want fidelity here.

susiejojo:
## Summary

Fixes #1526. `blis run` fatally aborted during KV-capacity auto-calculation for any model whose HF config declares `hidden_act: "situ"` (Kimi-K3's SiTU-GLU activation), via an over-conservative allowlist guard in `CalculateKVBlocks`.

SiTU-GLU (Sigmoid Tanh Unit) is Kimi-K3's successor to K2's SwiGLU. It is a **3-matrix gated GLU** (gate + up + down) with the identical weight/FLOP shape as SwiGLU — only the pointwise gate nonlinearity differs (parameterized by `activation_situ_beta` / `activation_situ_linear_beta`). BLIS's KV-capacity and compute models depend only on the MLP **matrix count**, never the gate function, so `"situ"` belongs in the SwiGLU-family allowlist.

Per the issue's exhaustive call-site audit, adding `"situ"` is both **necessary** (sole blocker) and **complete** (no arithmetic reads the activation value) — the resulting KV block count is byte-identical to what `"silu"` produces.

## Changes

- Add `"situ": true` to `swiGLUActivations` (`sim/latency/kv_capacity.go`) with a comment recording why SiTU-GLU is SwiGLU-family (matrix-count equivalence, gate nonlinearity differs).
- Update the unsupported-activation error message to list `situ`.
- Add regression tests:
  - `TestCalculateKVBlocks_SiTU_Accepted` — `"situ"` is admitted.
  - `TestCalculateKVBlocks_SiTU_ByteIdenticalToSwiGLU` — the correctness law: SiTU block count equals SwiGLU count for an otherwise-identical config (activation name must not affect capacity math).

## Verification

- `go build ./...`, `go test ./...`, and `golangci-lint run ./...` (v2.9.0) all pass.
- End-to-end: a Llama-8B config with `hidden_act` flipped to `"situ"` runs to completion and produces **byte-identical stdout** to the same config with `"silu"`.
- Confirmed the pre-fix binary reproduces the exact fatal from the issue, and the fixed binary clears it.

## Out of scope (per the issue)

The exact repro `./blis run --model moonshotai/Kimi-K3 --hardware H100 --tp 16` still fails **after** this fix — but on a *different, downstream* error: `KVBytesPerToken: hidden_dim (7168) must be evenly divisible by num_attention_heads (96)`. This is the MLA / linear-attention fidelity gap the issue author explicitly deferred ("tracked separately if we want fidelity here"). Kimi-K3 uses MLA with decoupled Q/K/V head dims and compressed KV (`kv_lora_rank: 512`), which BLIS's uniform-full-attention head-dim math does not model. Not touched here — deserves its own issue.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


susiejojo:
@claude /blis-pr-review

susiejojo:
Thanks for the review! Addressed the docs nit in 8d677157: `docs/reference/configuration.md:573` now lists `situ` alongside `silu`/`swiglu`/`geglu`, with a one-line note that it's Kimi-K3's SiTU-GLU (a 3-matrix gated GLU sharing SwiGLU's weight/FLOP shape).

Confirming the two deferrals you flagged are intentional:
- **MLA / `hidden_dim` ÷ `num_attention_heads` gap** — out of scope per the issue author's explicit note; deserves its own issue.
- No behavioral/invariant changes — this remains a pure gate relaxation in KV-capacity calc, byte-identical to `silu`.

susiejojo:
@claude /blis-pr-review

namasl (approved):

Issue #1532, opened 2026-08-07
Tracking: GLM-5.2 support — model-shape, MTP, and multi-node deployment gaps


namasl:
## Purpose

Tracking issue for everything BLIS needs in order to **faithfully model GLM-5.2** (and, more broadly, the modern MoE + MLA + speculative-decoding model family it represents). GLM-5.2 surfaced gaps at two levels: the **single-instance model shape** (KV/weight fidelity) and the **multi-node deployment topology** (how the recipe actually runs it). This issue lists all of them and links the sub-issues that track each.

None of these are crashes (except where noted) — BLIS runs and emits numbers. The concern is that the numbers are derived from an inaccurate model or deployment shape, so capacity-planning and throughput conclusions for GLM-5.2 are not trustworthy without these fixes.

## Background: what GLM-5.2 is

From the HF config ([`zai-org/GLM-5.2/config.json`](https://huggingface.co/zai-org/GLM-5.2/blob/main/config.json)) and the [vLLM recipe](https://recipes.vllm.ai/zai-org/GLM-5.2):

- **Architecture:** `model_type: glm_moe_dsa` / `GlmMoeDsaForCausalLM` — a DeepSeek-V3-class MoE that vLLM routes through its DeepSeek path ([vLLM registry](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/registry.py)). Not the older `glm4_moe`.
- **Attention:** Multi-head Latent Attention (MLA) — `kv_lora_rank: 512`, `q_lora_rank: 2048`, explicit `head_dim: 192`, `v_head_dim: 256` — plus DeepSeek Sparse Attention (`index_topk: 2048`).
- **MoE:** 256 routed experts, 8 active, 1 shared, first 3 of 78 layers dense (`first_k_dense_replace: 3`).
- **Speculative decoding:** 5-token MTP (`num_nextn_predict_layers: 1`).
- **Size:** ~743B total / 39B active; FP8 checkpoint `vram_minimum_gb: 893` → **must be multi-node on H100** (an 8×H100 node is 596 GiB usable).
- **Recipe strategies:** `single_node_tp`, `single_node_tep`, `multi_node_tp`, `multi_node_tp_pp`, `multi_node_tep`, `multi_node_dep`, `pd_cluster`.

### A. Model-shape / KV / weight fidelity — tracked by #1527 ✅
Single-instance config extraction and KV/weight models don't represent GLM-5.2's shape:
[...]

Issue #1528, opened 2026-08-07
Speculative decoding / MTP not modeled — decode throughput under-predicted


namasl:
> Part of #1532 (GLM-5.2 support tracking issue).

## Summary

BLIS does not model **speculative decoding / Multi-Token Prediction (MTP)** in any form. Decode is strictly one token per sequence per step, so decode **throughput is systematically under-predicted** for any model that ships MTP or another speculative-decoding scheme.

This is a fidelity gap, not a crash: BLIS runs and emits throughput numbers, but they omit the speedup speculative decoding provides.

## Why this matters for GLM-5.2

GLM-5.2's headline performance feature is **5-token MTP**. The [vLLM recipe](https://huggingface.co/zai-org/GLM-5.2) serves it with `--speculative-config '{"method":"mtp","num_speculative_tokens":5}'`, and the HF config declares `num_nextn_predict_layers: 1` ([config.json](https://huggingface.co/zai-org/GLM-5.2/blob/main/config.json)). With MTP unmodeled, BLIS's decode-throughput estimate for GLM-5.2 misses most of the model's real-world speedup. The same applies to DeepSeek-V3, Kimi-K3, and other MTP/EAGLE/Medusa models.

## Evidence in the code

BLIS's own calibration report acknowledges the decode model is sequential and diverges from speculative-decoding servers:

> "If the real server uses speculative decoding, actual token generation patterns differ from sim's sequential model."

— [`sim/workload/calibrate.go#L469`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/workload/calibrate.go#L469)

There is no `--speculative-config` / draft-token / acceptance-rate flag or config field anywhere in the codebase, and the step model advances one decode token per sequence per step.

## What "modeling MTP" would require (sketch, not a spec)

A faithful model needs at least:
- An **acceptance rate** parameter (fraction of the `num_speculative_tokens` draft tokens accepted per step) — workload- and model-dependent.
- Effective tokens-per-step = `1 + accepted_draft_tokens`, applied to the decode step so throughput scales accordingly.
- The extra draft-model / MTP-layer compute cost per step (so acceptance isn't modeled as free).

Acceptance rate is the hard part — it varies by workload and is low in synthetic benchmarks (the recipe itself warns pure-throughput benchmarks under-report MTP speed). A first cut could expose acceptance as a user-supplied parameter rather than predicting it.

## Relationship to #1527

[...]

PR #1534, opened 2026-08-07
feat(latency): GLM-5.2 model-shape fidelity — head_dim, MLA compressed-KV, first_k_dense_replace (#1527)

### LINKED ISSUE #1527: Bug: GLM-5.2-FP8 (glm_moe_dsa) — head_dim ignored, MLA/compressed-KV, first_k_dense_replace, FP8 block-quant unmodeled
> **Maintainer edit (@namasl, 2026-08-07) — additive, original preserved below.**
> This body was edited to bring the issue to actionable (VALID) status per an issue review. Nothing from the original investigation was removed. Changes: (1) corrected Finding 3's causal claim about the `--tp 8` fatal inline; (2) elevated MTP out of "minor" in Finding 5; (3) added a **Cross-path parity** section; (4) added **Per-finding acceptance criteria & split** at the end (findings are tracked under #1532). Original reporter: @susiejojo.

---

**Summary**
Investigating `zai-org/GLM-5.2-FP8` (config committed at `model_configs/glm-5.2-fp8/config.json`) surfaced several architectural nuances that BLIS's config extraction and KV/weight models do not represent. Unlike Kimi-K3's `situ` activation (#1526, a hard fatal), most of these are **silent** — BLIS runs and emits numbers, but they're derived from an inaccurate model shape. Filing so we can decide which are worth fixing vs documenting as known approximations.

> **Note (maintainer edit):** `model_configs/glm-5.2-fp8/config.json` is **not committed on `main`** (`ls model_configs/ | grep -i glm` → none) — it is local to the reporter. Committing the fixture is a precondition for the regression tests in the acceptance criteria below.

Model facts (from HF `config.json`):
- `architectures: GlmMoeDsaForCausalLM`, `model_type: glm_moe_dsa`
- `hidden_size: 6144`, `num_attention_heads: 64`, `num_key_value_heads: 64`, **explicit `head_dim: 192`**
- MLA attention: `q_lora_rank: 2048`, `kv_lora_rank: 512`, `qk_nope_head_dim: 192`, `qk_rope_head_dim: 64`, `v_head_dim: 256`
- MoE: `n_routed_experts: 256`, `num_experts_per_tok: 8`, `n_shared_experts: 1`, `moe_intermediate_size: 2048`, **`first_k_dense_replace: 3`**
- Quant: `quant_method: fp8`, `weight_block_size: [128,128]`, `activation_scheme: dynamic`, large `modules_to_not_convert` list (layernorms, gates, indexer, `lm_head`, `embed_tokens`, `norm`, MTP modules)
- DSA indexer: `index_head_dim: 128`, `index_n_heads: 32`, `index_topk: 2048`
- MTP: `num_nextn_predict_layers: 1`

---

**Finding 1 — Explicit `head_dim` ignored; KV per-token uses `hidden/heads` (SILENT, wrong)**
`KVBytesPerToken` (`sim/latency/kv_capacity.go`) and `computeModelWeightBytes` derive `headDim = HiddenDim / NumHeads`. For GLM-5.2 that is `6144/64 = 96`, but the config's explicit `head_dim` is **192**. `GetModelConfigFromHF` (`sim/latency/config.go`) never reads the `head_dim` key. Any model where `head_dim != hidden/heads` (common in modern MLA/GQA designs) gets a silently wrong attention weight + KV estimate. This alone is a general BLIS gap, independent of MLA below.

**Finding 2 — MLA / compressed-KV attention not modeled (SILENT, wrong KV footprint)**
GLM-5.2 uses Multi-head Latent Attention (`q_lora_rank`, `kv_lora_rank: 512`, split `qk_nope/qk_rope`, `v_head_dim: 256`). BLIS's KV formula is standard MHA/GQA: `NumLayers × 2 × headDim × numKVHeads × bytes`. MLA stores a compressed latent KV of size `kv_lora_rank + qk_rope_head_dim` per token (≈576), NOT `2 × headDim × num_kv_heads`. BLIS therefore mis-sizes the KV cache dramatically for this architecture (and DeepSeek-V3, Kimi-K3, etc.). Also `head_dim` (192) ≠ `qk_head_dim` (256) ≠ `v_head_dim` (256), so K and V per-head sizes differ — the `2 ×` (K==V) assumption breaks.

**Finding 3 — `first_k_dense_replace: 3` ignored (SILENT, over-counts MoE weight)**
GLM-5.2's first 3 of 78 layers are dense, the rest MoE. `computeModelWeightBytes` treats **all** layers as MoE when `IsMoE` (multiplying by 256 experts across all 78 layers). BLIS reads `interleave_moe_layer_step` and `intermediate_size_mlp` but has no `first_k_dense_replace` handling. This over-counts weights.

> **⚠️ Correction (maintainer edit — the original text below is wrong):** The original claim that this over-count is *"contributing to the `717.60 GiB → minimum 11 GPUs` fatal seen at `--tp 8`"* is **incorrect**. Only **3 of 78 layers** are dense, so fixing it reduces the weight estimate by ~4% (~717 → ~690 GiB) — still far above a single 8×H100 node (**596 GiB** = 8 × 80 GB). **The `--tp 8` fatal is correct behavior, not a bug**: the FP8 variant's own recipe lists [`vram_minimum_gb: 893`](https://huggingface.co/zai-org/GLM-5.2) (~743B params); a single node cannot hold this model. Do not "fix" Finding 3 expecting single-node to fit. Finding 3 is a bounded weight-*accounting* fidelity fix with **no bearing** on the fatal. `first_k_dense_replace` (first K layers dense, remainder MoE) is also **not** the same placement semantics as the existing `interleave_moe_layer_step` (every Nth layer MoE) — the fix must map to dense/MoE layer counts, not reuse the interleave step.
>
> _Original (retained for provenance):_ "This over-counts weights, contributing to the `717.60 GiB weights → minimum 11 GPUs` fatal seen at `--tp 8`."

**Finding 4 — FP8 block-quant treated as flat 1.0 byte/param (approximation)**
`parseQuantizationConfig` (`sim/latency/config.go:199`) returns `1.0` for any `quant_method: "fp8"`. GLM-5.2 is **block-wise** FP8 (`weight_block_size: [128,128]`) with a large `modules_to_not_convert` set (layernorms, MoE gates, indexer, `lm_head`, `embed_tokens`, MTP) that stay bf16. The flat 1.0 ignores both the per-block scale overhead and the un-quantized modules — an optimistic weight estimate. `weight_block_size` and `modules_to_not_convert` are never read.

**Finding 5 — DSA indexer + MTP layer unmodeled**
`GlmMoeDsaForCausalLM` includes a DeepSeek-style sparse-attention indexer (`index_n_heads: 32`, `index_topk: 2048`) and one MTP layer (`num_nextn_predict_layers: 1`). Neither contributes to BLIS's weight/KV/step-time estimates.

> **Correction (maintainer edit):** Calling MTP "minor / second-order" is fair for *weight/KV accounting*, but it obscures a bigger gap: **BLIS models no speculative decoding at all** — decode is strictly one token per step, acknowledged in-code (`sim/workload/calibrate.go:469`: *"actual token generation patterns differ from sim's sequential model"*). GLM-5.2's throughput headline is 5-token MTP ([recipe](https://huggingface.co/zai-org/GLM-5.2)), so for any *throughput* evaluation this is **first-order, not minor**. The DSA indexer contribution remains plausibly second-order for weight/KV. MTP-throughput is split into its own item below.
>
> _Original (retained for provenance):_ "Neither contributes to BLIS's weight/KV/step-time estimates. Likely second-order, noting for completeness."

---

**CLI command (fatals — this is CORRECT behavior, see Finding 3 correction)**
```bash
./blis run --model zai-org/GLM-5.2-FP8 --hardware H100 --tp 8
# fatal: model overhead (730.40 GiB = 717.60 weights ...) exceeds available GPU memory
# ^ correct: the FP8 variant needs ~893 GiB (recipe vram_minimum_gb), > 596 GiB for one 8xH100 node.
```

**Which invariant is relevant?**
INV-4 (KV block conservation) is maintained mechanically, but the *inputs* (KV bytes/token, weight bytes) are computed from a mismatched architecture, so capacity-planning output is inaccurate — not a crash, a fidelity gap.

---

**Cross-path parity (maintainer edit)**
All findings live in `sim/latency/{config.go,kv_capacity.go}` — config extraction and the KV/weight capacity model.
- **`blis run`**: fully affected (auto KV-block sizing computes capacity from the HF config).
- **`blis replay`**: shares the same `sim/latency` capacity path → same wrong inputs. Affected (INV-13 parity with run is preserved — both wrong identically).
- **`blis observe`**: measures a real server and does **not** compute KV capacity from model shape → the capacity-fidelity gap does **not** apply.
- **PD-disaggregated timing**: `KVBytesPerToken` also sizes PD transfer duration (per its docstring), so Findings 1/2 additionally skew PD transfer estimates on run/replay — not just capacity.

---

**Per-finding acceptance criteria & split (maintainer edit)**
These five findings have different types and scopes; each is tracked as its own actionable item under #1532 (GLM-5.2 support tracking). This issue covers the single-instance model-shape / KV / weight fidelity piece.

- **F1 — read explicit `head_dim`** (correctness; small, general). _Done when:_ `GetModelConfigFromHF` reads the `head_dim` key when present; `KVBytesPerToken`/`computeModelWeightBytes` use it instead of `HiddenDim/NumHeads`; a committed GLM-5.2 fixture regression test asserts `headDim == 192` (not 96). Backward-compatible: when `head_dim` is absent, behavior is byte-identical to today (INV-6).
- **F2 — MLA compressed-KV model** (correctness; design doc — spans DeepSeek-V3, Kimi-K3, GLM MLA family). _Done when:_ for MLA models KV bytes/token = `(kv_lora_rank + qk_rope_head_dim) × NumLayers × bytes` (drops the `2×` K==V assumption); a fixture test asserts the compressed footprint; dense/GQA models unchanged (INV-6). Needs a design doc per the design guidelines before micro-planning (new module behavior).
- **F3 — `first_k_dense_replace` weight accounting** (correctness; bounded, **no crash impact**). _Done when:_ `KVCapacityParams` carries a dense-layer count derived from `first_k_dense_replace`; `computeModelWeightBytes` splits dense vs MoE layers (not all-MoE); a fixture test asserts weight ≈ 3 dense + 75 MoE layers. Distinct semantics from `interleave_moe_layer_step`.
- **F4 — FP8 block-quant + `modules_to_not_convert`** (fidelity / hardening — document as a known optimistic approximation unless we target FP8 capacity fidelity). _Done when:_ either a documented "known approximation" note, or `weight_block_size` scale overhead + un-quantized module set are honored in the weight estimate.
- **F5a — DSA indexer** (fidelity; plausibly second-order). Document or model indexer weight/KV.
- **F5b — MTP / speculative-decoding throughput** (feature; **first-order for throughput**, tracked separately, NOT folded under weight-accounting). BLIS currently models one token/step.

**Suggested triage priority:** F1 (quick, general win) → F3 (bounded) → F2 (design doc, highest value) → F4/F5a (document) → F5b (separate feature).

namasl:
## Summary

Fixes #1527 (GLM-5.2 model-shape / KV / weight fidelity, tracked under #1532).

BLIS derives KV-cache block capacity and total model-weight bytes from a HuggingFace `config.json`. For the modern MLA MoE family (DeepSeek-V2/V3, Kimi-K3, GLM-5.2 `glm_moe_dsa`) three shape facts were ignored, producing **silently** wrong capacity-planning numbers (no crash). This PR closes findings **F1–F3** in the KV/weight capacity path, documents **F4/F5a** as known optimistic approximations, and commits a `GLM-5.2-FP8` fixture. **F5b** (MTP / speculative-decode throughput) is out of scope and tracked separately in #1528.

All changes are **no-ops when the new config keys are absent** (INV-6 byte-identity), and are scoped to the **capacity** path — the step-time (latency) model is untouched, so step-time golden datasets and INV-BC-DP1 remain byte-identical.

## Behavioral contracts

- **BC-1/BC-2 (F1 — explicit `head_dim`):** GIVEN a config with `head_dim` (e.g. GLM-5.2 `192`, while `hidden/heads = 6144/64 = 96`), WHEN sizing KV/weight, THEN the explicit `head_dim` is used; GIVEN the key absent, THEN `hidden/heads` (byte-identical to before). `head_dim` deliberately does **not** flow into trained-physics/roofline step time (keeps step-time goldens + INV-BC-DP1 intact).
- **BC-3/BC-4 (F2 — MLA compressed-KV):** GIVEN an MLA config (`kv_lora_rank > 0`), WHEN computing KV bytes/token, THEN it is `(kv_lora_rank + qk_rope_head_dim) × num_layers × compute_bytes`, **independent of TP** (the latent is replicated across TP ranks, matching vLLM's MLA cache), `num_kv_heads`, and `head_dim`; GIVEN a non-MLA config, THEN the standard `2 × head_dim × num_kv_heads / TP` formula is unchanged.
- **BC-5 (F3 — `first_k_dense_replace`):** GIVEN a MoE model with `first_k_dense_replace = K` of `L` layers, WHEN estimating weight bytes, THEN exactly `K` layers use the dense-MLP term and `L−K` use the MoE term (K clamped to `[0,L]`); GIVEN K absent/0, THEN byte-identical to the previous all-MoE accounting. Distinct from the every-Nth `interleave_moe_layer_step`.
- **BC-6:** committed `model_configs/glm-5.2-fp8/config.json` regression test asserts `head_dim == 192`, MLA latent `(512+64)` per layer, and the 3-dense/75-MoE split; `deepseek-v2-lite` (MLA, `K=1`) exercised end-to-end as a second architecture.
- **BC-7:** `docs/reference/models.md` + `CLAUDE.md` document the new shape support and the known optimistic approximations.

## Cross-path parity

`blis run` (auto KV-block sizing) and `blis replay` (per-pool sizing) share `KVBytesPerToken`/`CalculateKVBlocks`, so both inherit the corrected footprint identically (INV-13 preserved). `blis observe` does not derive capacity from model shape and is unaffected. PD KV-transfer duration (`KVBytesPerToken`) picks up the MLA footprint on run/replay.

## Known approximations (documented, F4/F5a)

- Block-wise FP8 (`weight_block_size`, `modules_to_not_convert`) is treated as a flat `1.0` byte/param → slightly **optimistic** weight estimate.
- The DSA indexer (`index_n_heads`, `index_topk`) and MLA attention weight projections keep the standard approximations.
- These are stated honestly in the docs and the design doc, never silently passed as exact.

> Note on the `--tp 8` fatal from the issue: this PR does **not** try to make GLM-5.2 fit a single node. Per the maintainer edit, that fatal is correct (the FP8 checkpoint's own recipe lists `vram_minimum_gb: 893`). F3 is a bounded weight-accounting fix with no crash impact; end-to-end, the weight estimate at TP=8 drops from ~717 GiB to ~697 GiB (the ~3% from correctly counting 3 dense layers) — still above a single 8×H100 node, as expected.

## Testing

- `go build ./...` clean; `go test ./...` all pass; `golangci-lint run` 0 issues on changed packages.
- New behavioral + law tests: MLA TP/kv-head/head-dim invariance and layer-linearity; F3 monotonicity-in-K and degenerate `K ≥ L` clamp; explicit-`head_dim` usage and `HeadDim==0` byte-identity; GLM-5.2-FP8 and deepseek-v2-lite fixture regressions.
- Verified INV-6: existing synthetic KV/weight goldens and the step-time golden datasets (which include a `head_dim`-differing model) are unchanged because the step-time path is untouched.
- End-to-end: `blis run --model deepseek-v2-lite --tp 1` completes; `blis run --model glm-5.2-fp8 --tp 8` fatals on memory as expected (correct behavior).


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

susiejojo:
## `/blis-pr-review` — Verdict: ✅ READY TO MERGE

Reviewed against #1527 (and parent #1532) for correctness, invariant preservation, cross-path parity, standards (R1–R23), test quality, and resolution of the two prior review passes. `go build ./...`, `go test ./sim/...`, and `go test ./cmd/...` pass. No BLOCKING or IMPORTANT findings. This is a well-scoped **capacity-path** fix with an exemplary, behavior-first test suite.

### Does it solve #1527?

Yes — every in-scope finding meets the maintainer's acceptance criteria:

| Finding | Criterion | Status |
|---|---|---|
| **F1** explicit `head_dim` | read the key; feed KV/weight (not `hidden/heads`); INV-6 when absent | ✅ `EffectiveHeadDim()`; fixture asserts 192≠96; byte-identity test |
| **F2** MLA compressed-KV | `(kv_lora_rank + qk_rope_head_dim) × layers × bytes`; drop `2×`; **TP-invariant**; non-MLA unchanged | ✅ correct vs vLLM (latent replicated, not sharded) |
| **F3** `first_k_dense_replace` | split K dense + (L−K) MoE; clamp; distinct from `interleave_moe_layer_step` | ✅ clean `[0,L]` clamp, no off-by-one |
| **F4** block-wise FP8 | document or model | ✅ documented (flat 1.0 byte/param, honestly optimistic) |
| **F5a** DSA indexer | document or model | ✅ documented |
| **F5b** MTP throughput | out of scope, tracked separately | ✅ deferred to #1528 |

### Correctness & invariants — confirmed
- **INV-6 byte-identity**: three no-op-when-absent gates (`HeadDim==0`, `KVLoraRank==0`, `FirstKDenseReplace==0`), each tested; `omitempty` json tags keep output identical.
- **INV-BC-DP1 / step-time untouched**: `trained_physics_model.go` and `roofline.go` reference none of the four new fields; `TestTrainedPhysicsModel_HeadDimDoesNotAffectStepTime` guards it. Verified **empirically** — pinning both pre/post binaries to the same `--total-kv-blocks` yields byte-identical per-request metrics, so the *only* behavioral channel is auto-calculated capacity.
- **Guard hoist** (prior IMPORTANT-2): `NumHeads>0`/`HiddenDim>0` correctly moved above the MLA early-return, protecting `computeModelWeightBytes` (called for MLA too). Only the `HiddenDim%NumHeads` check stays MLA-exempt.
- **Cross-path parity**: `run`/`replay` share `KVBytesPerToken`/`CalculateKVBlocks` (INV-13); `observe` doesn't derive capacity from shape — correctly unaffected.
- **R1/R4**: negative shape fields rejected at parse time, per-field tested; the four fields are set only in `GetModelConfigFromHF` (no bypass, no missed construction site).

All prior-pass items (IMPORTANT-1/2, MINOR-1..5) are present and verified.

### Empirical behavior (main `18c2c926` vs this PR, committed `glm-5.2-fp8` fixture)

**F3 weight accounting at `--tp 8`** (issue's fatal case): weights `717.60 → 696.61 GiB`, min GPUs `11 → 10`. Still correctly fatal — the FP8 checkpoint genuinely needs ~893 GiB, as the PR states.

**F1/F2 KV capacity + runnability across TP:**

```
TP  | main (18c2c926)              | PR #1534
----+-----------------------------+---------------------------
8   | exceeds GPU memory (correct)| exceeds GPU memory (correct)
10  | FATAL: KV heads not ÷ TP     | 7,015 blocks   ← now runs
11  | FATAL: KV heads not ÷ TP     | 60,341 blocks  ← now runs
14  | FATAL: KV heads not ÷ TP     | 220,316 blocks ← now runs
16  | 233,465 blocks              | 326,966 blocks (+40%)
32  | 1,746,732 blocks            | 1,180,166 blocks (−32%)
```

Two effects: (a) **runnability** — main forces MLA onto the GQA path requiring `num_kv_heads=64 ÷ TP`, so TP=10/11/12/14/20 fatal; the PR's TP-invariant MLA branch lifts that. (b) **direction flips with TP** — main divides KV bytes by TP (optimistically), but the MLA latent is replicated, so below ~TP16 the PR gives more blocks and above it fewer. Correct.

**Effect on request metrics:** for GLM-5.2 (743B) itself, TTFT/ITL/E2E/throughput are **byte-identical** at every load tried — compute step-time saturates long before KV binds, so the corrected budget is never the constraint (consistent with the deferred MLA step-time work, which the PR honestly flags via a runtime warning + docs). On a small MLA model where KV *does* bind (`deepseek-v2-lite`, TP=1, 10,165 → 74,312 blocks): `completed_requests +31%`, `preemption_count 1162 → 0`, `tokens_per_sec +30%` — and pinning both to identical `--total-kv-blocks` collapses all deltas to zero, isolating the capacity path as the sole cause.

### MINOR (non-blocking)
1. `TestGLM52FP8_WeightSplit` asserts a directional inequality (split > all-MoE), not the exact `K=3`; the exact count is pinned separately by `TestGLM52FP8_ModelShape` (parse). Combined coverage adequate.
2. F3 landed `FirstKDenseReplace` on `ModelConfig` rather than `KVCapacityParams` as the criterion's wording suggested — behavioral outcome matches; placement is arguably cleaner. Not a defect.
3. Pre-existing gofmt drift in `config.go` (`moeExpertCountFields` alignment) exists identically on `main`, outside this diff — flagging only so it isn't mistaken for a regression.

**Recommendation: merge.** The PR fully closes F1–F3, documents F4/F5a honestly, correctly defers F5b, and is scrupulous about not overstating fidelity (the pessimistic MLA step-time gap is surfaced at runtime and in docs).

<sub>Reviewed with `/blis-pr-review` (automated self-review). Experiment binaries built from `18c2c926` and PR HEAD `0f2a6234`.</sub>

sim/latency/kv_capacity.go
```diff
@@ -83,22 +76,63 @@ var swiGLUActivations = map[string]bool{
...
+	// or headDim) and is REPLICATED across TP ranks rather than sharded, so the
+	// per-GPU value is the full latent width and is NOT divided by TP (design D2;
+	// matches vLLM's MLA cache, e.g. DeepSeek (512+64)×2 bytes/token/layer). This
+	// branch is selected only when KVLoraRank > 0; a standard MHA/GQA config
+	// (KVLoraRank == 0) takes the byte-identical standard path below (INV-6). Only
+	// the hidden%heads divisibility guard is skipped here (the latent path never
+	// uses the hidden/heads quotient); NumHeads>0 and HiddenDim>0 are validated in
+	// the common block above so computeModelWeightBytes stays protected for MLA too.
+	if mc.IsMLA() {
+		if mc.QKRopeHeadDim < 0 {
```
jgchn: Dead check: `QKRopeHeadDim < 0` is unreachable. `GetModelConfigFromHF` already rejects negative values at parse time (the loop at config.go:~357). For direct callers of the public API, `IsMLA()` requires `KVLoraRank > 0` which implies the config was intentionally constructed, a `< 0` guard here is belt-and-suspenders but never fires via the normal path. Consider removing or marking explicitly as a public-API boundary guard in one line, not an error message that suggests this is a realistic failure mode.

sim/latency/kv_capacity.go
```diff
@@ -83,22 +76,63 @@ var swiGLUActivations = map[string]bool{
...
+			return 0, fmt.Errorf("KVBytesPerToken: qk_rope_head_dim must be >= 0, got %d", mc.QKRopeHeadDim)
+		}
+		latentWidth := mc.KVLoraRank + mc.QKRopeHeadDim
+		perTokenKVBytesF := float64(mc.NumLayers) * float64(latentWidth) * mc.BytesPerParam
+		if perTokenKVBytesF <= 0 {
```
jgchn, line 111: Same issue: `perTokenKVBytesF <= 0` cannot fire. At this point `NumLayers > 0` (line 92), `KVLoraRank > 0` (IsMLA gate), `QKRopeHeadDim >= 0` (line 119), `BytesPerParam > 0` (line 101). The product is necessarily positive.

sim/latency/kv_capacity.go
```diff
@@ -83,22 +76,63 @@ var swiGLUActivations = map[string]bool{
...
+	// Validations common to both the MLA and standard paths. NumHeads and HiddenDim
+	// are validated here (not only on the standard path) because CalculateKVBlocks
+	// calls computeModelWeightBytes for ALL configs — including MLA — and that
+	// weight estimate reads NumHeads/HiddenDim (via EffectiveHeadDim and the
+	// attention/embedding terms). A degenerate MLA config that returned a valid KV
+	// value here but then silently under-counted (NumHeads==0 → kvDim 0) or wildly
+	// inflated (HiddenDim==0 → ~0 weight → huge block count) weight would run on
+	// garbage with no error, so these guards must gate the MLA path too. Only the
+	// HiddenDim%NumHeads divisibility check is MLA-exempt (the latent path never
+	// uses that quotient) and stays on the standard path below.
```
jgchn: Is this 10-line comment too verbose?

sim/latency/kv_capacity.go
```diff
@@ -83,22 +76,63 @@ var swiGLUActivations = map[string]bool{
...
+	// (kv_lora_rank + qk_rope_head_dim) — NOT the standard 2 × headDim × numKVHeads.
+	// The latent is shared across all heads (so it does not scale with numKVHeads
+	// or headDim) and is REPLICATED across TP ranks rather than sharded, so the
+	// per-GPU value is the full latent width and is NOT divided by TP (design D2;
+	// matches vLLM's MLA cache, e.g. DeepSeek (512+64)×2 bytes/token/layer). This
+	// branch is selected only when KVLoraRank > 0; a standard MHA/GQA config
+	// (KVLoraRank == 0) takes the byte-identical standard path below (INV-6). Only
+	// the hidden%heads divisibility guard is skipped here (the latent path never
+	// uses the hidden/heads quotient); NumHeads>0 and HiddenDim>0 are validated in
+	// the common block above so computeModelWeightBytes stays protected for MLA too.
```
jgchn: This 13-line comment duplicates the `IsMLA()` docstring, the field-level comments on `ModelConfig`, and `docs/reference/models.md`. The formula + IsMLA() name is self-documenting.

jgchn (commented):
Minor styling issues

namasl:
@claude /blis-pr-review

Follow-up push (`6550d1bc`): addressed the actionable MINOR-1 from the last review — `TestGLM52FP8_WeightSplit` now pins the exact K=3 dense/MoE split via an equal-marginals linearity law over FirstKDenseReplace=0..4 (was a directional inequality). Test-only, no production change. The other two minors (field placement on ModelConfig; pre-existing config.go gofmt drift on main) are intentionally left as-is per the reviewer's own notes.

namasl:
@claude /blis-pr-review

Follow-up push (`a539874d`) addresses @jgchn's inline styling/dead-code comments on `kv_capacity.go`:

- **`:119` / `:124` (dead checks):** removed the standalone `QKRopeHeadDim < 0` error (unreachable — parse-time rejects negatives, MINOR-2) and folded the residual public-API concern into the single `perTokenKVBytesF <= 0` boundary guard, reframed with a one-line comment marking it as a hand-built-config boundary guard (not a realistic failure mode).
- **`:88` / `:117` (verbose/duplicative comments):** trimmed the 10-line common-validation comment and the 13-line MLA-branch comment, which duplicated the `IsMLA()` docstring, the `ModelConfig` field comments, and `docs/reference/models.md`.

No behavioral change; full suite + lint pass. Also strengthened `TestGLM52FP8_WeightSplit` (prior push `6550d1bc`) to pin the exact K=3 split via an equal-marginals linearity law.

susiejojo:
### Intuition: why MLA fits *more* KV blocks at low TP but *fewer* at high TP

A reviewer asked whether the crossover in the TP sweep is physical. It is — it falls straight out of how the two KV layouts behave under tensor parallelism:

- **GQA/MHA KV is *sharded* across TP ranks** → per-GPU bytes/token shrink as **1/TP**.
- **MLA's compressed latent is *replicated* on every rank** (all heads consume the one latent vector, so it can't be split) → per-GPU bytes/token stay **flat** in TP.

So as TP grows, the sharded GQA cache eventually drops below the fixed MLA latent and GQA fits more blocks per GPU. This is stated almost verbatim in **TPLA** ([arXiv:2508.15881](https://arxiv.org/abs/2508.15881)): for LLaMA-3-70B the per-token KV dim is `2×8×128 = 2048` and at TP=4 each device holds `512`, whereas DeepSeek-V3's `64+512 = 576` latent "must be fully replicated on each device regardless of the parallelism degree … eroding the advantage of MLA over GQA." It's also exactly why **DP-attention** exists (SGLang/DeepSeek): under plain TP the MLA KV cache is duplicated on every rank ([SGLang large-scale EP blog](https://lmsys.org/blog/2025-05-05-large-scale-ep/)).

The PR encodes precisely this: the MLA branch is **not divided by TP** (latent replicated), the GQA path is (sharded). ✅

**Two caveats on reading our specific crossover (~TP 21):**
1. It's *fix vs bug*, not architecture vs architecture — both curves model the same GLM-5.2 weights; `main` is the buggy GQA approximation, the PR is the correct MLA formula. The crossover is where the bug's error flips sign.
2. `main` also uses the wrong `head_dim` (96 instead of the explicit 192, i.e. F1), which halves its GQA estimate and pulls the crossover *down*. A GQA model with GLM's true dims (`head_dim=192`) wouldn't cross MLA until **~TP 43** — beyond any realistic single-model TP. So the literature takeaway holds: **at realistic TP (≤16) MLA wins decisively on KV footprint**; the replication penalty only bites at extreme TP (the DP-attention regime).

Block-ratio check: predicted PR/main = `1.33` @TP16 and `0.67` @TP32 (∝ per-GPU bytes/token), observed `1.40` / `0.68` — within ~2% (residual = the PR's ~3% lower weight freeing slightly more KV memory).

susiejojo (approved):
lgtm

namasl:
Marked all four inline review threads resolved — each was addressed in commit `a539874d`:

- **`kv_capacity.go` dead `QKRopeHeadDim < 0` check** → removed; the residual hand-built-config concern is now covered by the single `perTokenKVBytesF <= 0` boundary guard.
- **`perTokenKVBytesF <= 0` "cannot fire"** → kept as the one explicit **public-API boundary guard** (a negative `QKRopeHeadDim` on a hand-built `ModelConfig` can still make `latentWidth <= 0`; normal configs are rejected at parse time), with a one-line comment saying exactly that.
- **10-line common-validation comment** → trimmed to 5 lines.
- **13-line MLA-branch comment duplicating the `IsMLA()` docstring / field comments / docs** → trimmed to reference them.

No behavioral change; `go build`, `go test ./...`, and `golangci-lint run ./...` all pass. Ready to merge.

Issue #1536, opened 2026-08-08
Autoscaler per-variant CostPerReplica ignores node span (mixed span/non-span population within one variant)


namasl:
> Follow-up discovered while implementing multi-node TP placement (PR #1537 for #1529).

## Status / dependency

**This is a latent follow-up gated on #1537 (multi-node TP placement), which is currently OPEN / unmerged.** In current `main`, every placement site sets `inst.CostPerHour = poolCostPerHour` (`cluster.go:375`, `infra_lifecycle_event.go:74`, `direct_actuator.go`) and no instance spans nodes, so **every replica of a given `VariantSpec{GPUType, TPDegree}` has an identical `CostPerHour`, and the current "seed from the first replica" aggregation is exactly correct.** The heterogeneity this issue describes only becomes reachable once #1537 merges (a spanning instance costs `N × cost_per_hour`). The analyzer fix and its unit test can be written independently of #1537 (see Acceptance criteria), but the end-to-end scenario — and any regression — cannot manifest until #1537 lands.

## Summary

Once multi-node TP placement (#1537) lands, two instances of the same `(GPUType, TPDegree)` variant can have **different** per-instance `CostPerHour`: a single-node instance is `1 × cost_per_hour`, a node-spanning instance is `N × cost_per_hour`. #1537 sets this per-instance cost correctly at all three placement sites.

However, the autoscaler's per-variant cost aggregation does **not** carry that heterogeneity through. In `sim/cluster/saturation_analyzer.go`, `variantAgg` is seeded with the **first** replica's `CostPerHour` for a given `VariantSpec{GPUType, TPDegree}`:

```go
agg = &variantAgg{costPerHour: r.CostPerHour}
```

`VariantSpec` has no node-span dimension (`autoscaler.go:21-24`: `{GPUType, TPDegree}`), so a mixed population (some instances of a variant span nodes, others don't) collapses to a single `CostPerReplica` — whichever instance is first in `metrics.Replicas` slice order. Scale decisions in `engine.go` that rank variants by `CostPerReplica` — scale-**up** picks the cheapest (`sortedByAscCost`, `vcs[0]`), scale-**down** picks the most expensive (`sortedByDescCost`) — can then under- or over-price that variant.

## Impact

[...]

Issue #1535, opened 2026-08-08
Pipeline parallelism (PP): new parallelism dimension, bubble/micro-batch latency, stage placement


namasl:
> **Also relevant to Kimi K3** ([#1638](https://github.com/inference-sim/inference-sim/issues/1638)) — shared with the GLM-5.2 family; applies to K3 only if its serving recipe uses pipeline parallelism.

> Split from #1529. #1529 bundled two subsystems — multi-node tensor parallelism and pipeline parallelism — that share the root cause "a model instance cannot cross nodes." Multi-node TP placement is handled by its own PR (span an instance's TP GPUs across nodes of a pool). **This issue tracks the pipeline-parallelism half**, which #1529 itself flagged as "a new subsystem [that] should have a design doc before micro-planning."

## Summary

BLIS has no pipeline parallelism today: no `--pp` / `--pipeline-parallel-size` flag, no `PipelineParallel` config field, and no stage/micro-batch logic. The only parallelism dimensions modeled are TP/DP/EP. This blocks the `multi_node_tp_pp` recipe strategy (e.g. the [GLM-5.2 vLLM recipe](https://recipes.vllm.ai/zai-org/GLM-5.2)).

## What this requires (sketch, not a spec)

- A new parallelism dimension: **stage count** (`--pp` / config field), with validation against layer count and the TP/DP/EP dimensions.
- A **pipeline-bubble / micro-batch latency model** — PP throughput depends on micro-batch count vs stage count (the classic `(S-1)/(M+S-1)` bubble fraction), not just a flat per-layer split.
- **Stage-to-node placement**: map PP stages to nodes, composing with the multi-node TP placement landed in #1529 (a single stage may itself be multi-node TP).
- Interaction with the inter-node network cost model (#1530), since PP boundaries cross nodes.

## Process note

Per CLAUDE.md ("new features that introduce module boundaries … a design doc should exist before micro-planning") and #1529's own triage, **PP should get a design doc first**. This is a new subsystem module (its own contract: what it observes/controls/owns/guarantees), not a policy template behind an existing interface.

## Relationship to sibling issues

- Depends conceptually on the multi-node placement capability from #1529 (an instance/stage spanning nodes).
- Composes with #1530 (inter-node network cost) for cross-stage activation transfer pricing.
- Part of the #1532 GLM-5.2 support epic.

Issue #1530, opened 2026-08-07
Inter-node network cost not modeled (cross-node TP/EP traffic is free)


namasl:
> **Also applies to Kimi K3** ([#1638](https://github.com/inference-sim/inference-sim/issues/1638)) — this is a deployment capability shared by the GLM-5.2 and Kimi K3 MLA-MoE families; landing it here covers both.

> Part of #1532 (GLM-5.2 support tracking issue).

## Summary

BLIS does not model **inter-node network cost**. Placement is topology-blind beyond "which GPUs are free on a node," and the latency model has no term for cross-node bandwidth or latency. As a result, any multi-node MoE deployment modeled via data-parallel-across-nodes (the closest BLIS can get to the `multi_node_dep` recipe strategy) carries **zero interconnect penalty** — the cross-node expert all-to-all and DP-coordination traffic is effectively free in the model.

## Why this matters for GLM-5.2

The recipe's `multi_node_dep` (Wide Expert Parallelism) strategy does an **expert all-to-all every layer across the node boundary** — the dominant cross-node cost, and the part most sensitive to whether the fabric is InfiniBand vs RoCE. On real hardware this is often the throughput bottleneck. A BLIS estimate that omits it will be optimistic in a way that is invisible to the user, and cannot be used to compare interconnect choices or to decide TP-vs-EP-vs-PP tradeoffs (whose whole point is trading compute for different communication patterns).

## Evidence in the code

**The TP all-reduce term assumes intra-node NVLink and is independent of node placement** — it is a function of TP degree and model shape only, never of which nodes an instance's GPUs sit on. See the DP/EP step-time terms in the trained-physics model:

- [`sim/latency/trained_physics_model.go#L222-L279`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/latency/trained_physics_model.go#L222-L279) — the `tpdp = tp·dp` divisor block (no node-topology input). The communication *terms* proper are assembled downstream (`tTpAttention`/`tTpDenseFFN`, `tMoEDispatch`/`tMoEReduce`) from the bases cited next.

Both communication bases divide byte volume by `bwHbmUs` (the on-package HBM/NVLink bandwidth), with the coefficient explicitly documented as absorbing "the NVLink/HBM bandwidth ratio (~0.27 on H100)" — i.e. the model has **only** an intra-node bus in it:

[...]

PR #1537, opened 2026-08-08
feat(cluster): multi-node TP placement — an instance can span nodes within a pool (Closes #1529)

### LINKED ISSUE #1529: Multi-node TP placement: an instance can span nodes within a pool

**Rescoped 2026-08-12:** narrowed to **multi-node tensor-parallelism placement** only — one issue, one PR (PR #1537). The **pipeline-parallelism** half originally bundled here is tracked separately in **#1535** (a new subsystem, design-doc first). This issue no longer covers PP.

## Summary

A BLIS model instance is **confined to a single node**: `PlaceInstance` bin-packs an instance onto one node and skips any node without `tpDegree` free GPUs, so a single TP group can never cross a node boundary. This blocks the `multi_node_tp` and `multi_node_tep` recipe strategies.

GLM-5.2-FP8 (~743B params, [`vram_minimum_gb: 893`](https://huggingface.co/zai-org/GLM-5.2)) does not fit on a single 8×H100 node (596 GiB usable), so **every H100 deployment is multi-node**. The [vLLM recipe](https://recipes.vllm.ai/zai-org/GLM-5.2) lists `multi_node_tp` and `multi_node_tep` as supported strategies; BLIS can model neither without multi-node TP placement. (It can approximate `multi_node_dep` via TP-within-node + DP-across-nodes, but that is a different strategy with its own gap — see #1530 for the inter-node network cost.)

## Evidence in the code

**TP placement is single-node.** `PlaceInstance` bin-packs an instance onto **one** node and skips any node that doesn't have `tpDegree` GPUs free — an instance never spans nodes:

- [`sim/cluster/infra_placement.go#L201-L218`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/cluster/infra_placement.go#L201-L218) — `if node.freeCount() < tpDegree { continue }`, then selects `tpDegree` GPUs from that single node.

Relax placement so one instance can reserve GPUs across >1 node — whole-node occupancy when `tp > gpus_per_node` and `tp % gpus_per_node == 0` — at all placement sites. The cross-node TP all-reduce latency term (today the all-reduce assumes intra-node NVLink) is tracked separately in #1530, which this issue unblocks.

## Delivered by PR #1537

Multi-node TP placement landed as **PR #1537** (linked "Part of #1529"). This issue is fully captured by that PR; close it when #1537 merges.

## Related

- **#1535** — pipeline parallelism (the split-out PP half; new subsystem, design-doc first).
- **#1530** — inter-node network cost (cross-node TP all-reduce pricing; unblocked by this issue).

## Suggested triage

`enhancement`. Multi-node TP placement extends the existing placement path (not a new subsystem — that's PP, in #1535).

namasl:
## Summary

Relax GPU placement so **one model instance can reserve its `--tp` GPUs across more than one node of a single node pool**, enabling multi-node tensor parallelism (e.g. GLM-5.2 at TP=16 on 2×8 H100, which no single node can host).

**Closes #1529.** That issue was rescoped (2026-08-12) to multi-node TP placement only — one issue, one PR — and is fully captured here. The pipeline-parallelism half originally bundled with it is a separate subsystem tracked in **#1535** (design-doc first). This PR **unblocks #1530** (inter-node network cost), which per the #1532 ordering can only price cross-node links once an instance can span nodes.

## What changed

`PlacementManager.PlaceInstance` now runs **two global passes** over the matching pools:

- **Pass 1 — single-node first-fit (unchanged):** the original loop across all matching pools, kept verbatim. If any single Ready node can host the whole TP group, placement is identical to before → **byte-identical stdout (INV-6)**, verified against `upstream/main` for single- and multi-instance-TP runs.
- **Pass 2 — cross-node packing within one pool:** reached **only** when no single Ready node in **any** matching pool can satisfy `tpDegree`. Packs GPUs across Ready nodes of one pool in index order. Single-node placement is therefore strictly preferred to spanning anywhere. Select-then-commit atomicity (R5): a pool that falls short of `tpDegree` mutates nothing. Skips non-Ready nodes; selects only free GPUs. Spanning stays within one pool (cross-pool spanning is not attempted).

Supporting changes:
- **`gpusByID` index + `distinctNodesForGPUs`** — resolve a placed GPU set to its distinct nodes without GPU-ID string parsing; sorted/deterministic (R2). Index populated at the single `&GPU{}` construction site (`newNode`, R4).
- **`ReleaseInstance` fix** — fires the drain-completion callback for **every** node a spanning instance freed (the pre-#1529 code checked only the last node visited during release), iterating touched nodes in sorted order.
- **Cost** — per-instance `CostPerHour = distinct-nodes-spanned × pool cost_per_hour`, applied identically at all three placement sites (startup, deferred `NodeReadyEvent`, autoscaler scale-up) via `InstanceCostPerHour`. Single-node is `1×` (unchanged).
- **Latency honesty** — cross-node interconnect latency is deferred to **#1530**: the trained-physics TP all-reduce term assumes intra-node NVLink and is placement-blind, so a spanning instance's all-reduce is priced optimistically. The first time any instance spans nodes, a one-time `logrus.Warn` (**stderr**, never stdout — INV-6) says so and names #1530.

## Behavioral contracts

- **BC-1 (INV-6):** `gpus_per_node ≥ tp` → placement lands on one node, same node/GPU-ID ordering as before; conservation holds.
- **BC-2:** an instance spans nodes only when no single Ready node in any matching pool fits; single-node placement anywhere is strictly preferred to spanning anywhere.
- **BC-3:** insufficient aggregate capacity → error with **no** partial allocation (atomic).
- **BC-4:** releasing a spanning instance frees every node and fires each fully-freed draining node's callback.
- **BC-5:** spanning instance per-instance cost = nodes-spanned × cost_per_hour (single-node = 1×).
- **BC-6:** spanning emits exactly one stderr warning naming #1530; stdout stays deterministic.

## Testing

`go build ./...` ✓ · `go test ./...` ✓ (all packages) · `golangci-lint run ./...` ✓ (0 issues).

New behavioral tests (INV-A GPU conservation asserted throughout): single-node-unchanged, spans-two-nodes, **spans-three-nodes**, span-needed-but-insufficient, prefers-single-node-over-spanning, **single-node-fit-in-later-pool-beats-spanning-earlier-pool**, **does-not-span-across-pools**, does-not-span-onto-draining-node, release-fires-drain-per-node, cost = nodes × rate, warns-once, single-node-no-warning. Existing placement suite stays green (BC-1 non-perturbation).

**INV-6 verified**: built `upstream/main` and this branch, confirmed byte-identical stdout for a single-instance run and a `--num-instances 3 --tp 2` run.

## Scope & limitations

- **`blis run` only.** `blis replay`/`observe` already `logrus.Fatalf` on node pools, so INV-13 (run/replay parity) is N/A.
- **Cross-pool spanning** (one TP group over mixed GPU types) is out of scope.
- **Pipeline parallelism** is out of scope → **#1535** (design-doc first).
- **Cross-node latency pricing** is out of scope → **#1530**.

## Discovered issues

- **#1535** — pipeline parallelism (the PP half split out of #1529), design-doc first.
- **#1536** — the autoscaler's scale-up headroom is span-unaware (counts free GPUs, not free whole nodes), so it can propose a spanning scale-up that `PlaceInstance` then harmlessly rejects. Deterministic (no INV-6 issue); documented and deferred.
- **#1543** — proper support for the two configs this PR guards with a startup panic: per-role TP overrides under `node_pools` (needs a per-role placement path) and duplicate `gpu_type` across pools (needs pool-identity, not `gpu_type`, cost/capacity lookup).
- **#1545** — pre-existing node-lifecycle/KV gaps this PR makes newly reachable at high TP: possible ~TP× KV-capacity over-sizing, and a not-yet-span-aware `MarkNodeTerminated`.


namasl:
@claude /blis-pr-review

namasl:
Thanks for the thorough review — addressed all items in `2698dcc6`.

**I-1 (semantic correctness) — fixed by switching to whole-node occupancy.** You're right that fragmentation-triggered spanning fabricated asymmetric TP groups. Rather than just gate on `tpDegree > gpus_per_node`, I modeled multi-node TP as **whole-node occupancy**: Pass 2 now engages only when `tpDegree > gpus_per_node` **and** `tpDegree % gpus_per_node == 0`, and allocates exactly `tpDegree/gpus_per_node` **fully-free** Ready nodes (each contributing all its GPUs → uniform per-node rank count, the only physically-realizable multi-node TP shape). The fragmentation case now stays pending, exactly as before the feature. Docblock, CLAUDE.md, and `docs/guide/cluster.md` updated to state the precondition accurately.

**I-2 — fixed.** `distinctNodesForGPUs` now `logrus.Errorf`s an R4 invariant violation on a missing GPU ID instead of silently skipping.

**I-3 — fixed.** `InstanceCostPerHour` now logs and returns `0` for an empty/unresolvable GPU set (fails fast) instead of flooring to `1×`.

**Test gaps — filled:**
- **T-1** `TestDeferredPlacement_SpanningInstanceCost` (deferred `NodeReadyEvent` path, cost == 2× pool)
- **T-2** `TestAutoscalerScaleUp_SpanningInstanceCost` (DirectActuator.scaleUp path, cost == 2× pool)
- **T-3** `VerifyConservation` added to the warns-once test
- **T-4** `TestPlaceInstance_SpanRequiresFullyFreeNodes` (atomicity with a pre-allocated node)
- **T-5** `TestRetryPendingInstances_SpanningPlacement` (pending → NodeReady → retry integration)
- plus `TestPlaceInstance_FragmentationDoesNotSpan` and `TestPlaceInstance_NonDivisibleSpanRejected` guarding the new I-1 precondition.

**D-1/D-2/D-3 — fixed:** file header + struct comment now describe the two-pass algorithm (not "first-fit bin-packing"); inner Pass-1 `select`/`commit` labels renamed to remove the Pass 1/2 collision; the stale `docs/plans/…` reference in the test file replaced with a PR/issue reference (the plan is gitignored per the repo's `.gitignore`).

**Pre-existing items you noted** (pool cost lookup by shared GPUType, `gpusByID` not pruned on termination, `DrainNode` currently test-only, `MarkNodeTerminated` Debug-vs-Warn): left as-is for this PR; the shared-GPUType cost lookup is worth a follow-up issue. The `× span` multiplier does not change the pre-existing first-match behavior.

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` all green. INV-6 re-verified byte-identical to `upstream/main` for single- and multi-instance-TP runs.

@claude /blis-pr-review

namasl:
Round 2 addressed in `11feda15`.

**IMP-1 (diagnosability) — fixed.** A startup placement failure now emits a `logrus.Warnf` (naming the instance, tp, and error) before deferring to pending, explicitly noting that `NodeReady` is not wired in `blis run` so the instance will not be retried. Additionally, `PlaceInstance` now distinguishes a **structurally-unsatisfiable** request (exceeds every matching pool's `gpus_per_node` yet not a whole multiple of it → can never place regardless of capacity) from a transient capacity shortfall, via a new `tpDegreeSatisfiableByShape` helper. The "no matching pool at all" case correctly falls through to the generic capacity error rather than the shape error.

**Minors:**
- **MIN-1:** the 3-node span test now asserts `primary ∈ span` instead of `primary == span[0]` (index order vs lexicographic sort diverge at ≥10 nodes; `primaryNode` is logging-only).
- **MIN-2:** added tests for the I-2 branch (unknown GPU ID → error log, 0 nodes), the I-3 branch (empty set → returns 0 + log), and the new unsatisfiable-shape error message.
- **MIN-3:** stale `selected` → `selectedNodes` in the cross-pool guard comment.
- **MIN-4:** `TestAutoscalerScaleUp_SpanningInstanceCost` now sets `cfg.TP = 16` to match the TP=16 variant (realizable config).
- **MIN-5:** extended the CLAUDE.md #1536 known-limit note — the autoscaler's scale-up headroom gates on free GPU *count*, not free whole *nodes*, so it can propose a spanning scale-up that `PlaceInstance` then rejects (logged, harmless, may repeat per tick).
- **MIN-6:** `VariantSpec.TPDegree` doc now includes 16 (first-class per #1529); the `gpn <= 0` guard is annotated as defensive/unreachable given `NodePoolConfig.IsValid`.
- **MIN-7:** verified — the working tree is clean and the multi-node TP paragraph is present in both `HEAD` and the working tree; the reviewer's `git status` snapshot was a transient artifact from an in-progress edit.

`go build ./...`, `go test ./...`, `golangci-lint run ./...` all green. INV-6 re-verified byte-identical to `upstream/main` (the new warning is stderr; stdout unchanged).

@claude /blis-pr-review

sim/cluster/infra_placement.go
```diff
@@ -227,26 +267,193 @@ func (pm *PlacementManager) PlaceInstance(id InstanceID, model, gpuType string,
...
+	// tpDegree/GPUsPerNode fully-free nodes, each contributing all its GPUs (an equal
+	// rank count per node). This is the only physically-realizable multi-node TP shape
+	// — a real NCCL/vLLM TP group has a uniform per-node rank count, so tpDegree must
+	// be an exact multiple of the pool's GPUsPerNode. This deliberately EXCLUDES the
+	// fragmentation case (tpDegree ≤ GPUsPerNode but no single node momentarily has
+	// room): packing an odd remainder across nodes (e.g. 2+1 for TP=3) would fabricate
+	// an asymmetric TP group that cannot exist, silently converting a visible "pending"
+	// outcome into optimistic capacity. Such an instance stays pending, exactly as
+	// before this feature. A pool is eligible only when tpDegree > GPUsPerNode AND
+	// tpDegree % GPUsPerNode == 0.
```
jgchn: Consider shortening this docstring comment

sim/cluster/cluster.go
```diff
@@ -336,6 +336,15 @@ func NewClusterSimulator(config DeploymentConfig, requestSource RequestSource, o
...
 				// Pass "" as gpuType (any pool) to match AddPending's placement semantics.
+				// R1: surface the failure. NodeReadyEvent (the only path that retries
+				// pending instances) has no production caller in blis run today, so a
+				// startup deferral means this instance is effectively dropped — never
+				// silent. This matters more since #1529's whole-node constraint creates
+				// configs that can NEVER place (e.g. tp not divisible by any pool's
+				// gpus_per_node); PlaceInstance's error names that case.
+				logrus.Warnf("[cluster] instance %s not placed at startup (tp=%d): %v — deferred to "+
+					"pending, but NodeReady is not wired in blis run, so it will not be placed later",
+					id, tpDegree, err)
```
jgchn: Minor: this warning fires for **every** instance that fails startup placement when `InitialNodes=0`
(a valid config for autoscaler-driven clusters). In a 16-instance TP=16 cluster where
all nodes start provisioning (InitialNodes=0), this would emit 16 identical warnings at
startup before any NodeReadyEvent fires. Consider: emitting once with a count, or
gating on whether the request is structurally unsatisfiable.

CLAUDE.md
```diff
@@ -478,6 +478,8 @@ See [`docs/guide/latency-models.md`](docs/guide/latency-models.md) for details.
...
...rained-physics TP all-reduce term (β₄) assumes intra-node NVLink and is placement-blind, so a spanning instance's all-reduce is priced optimistically; the first time any instance spans, a one-time `logrus.Warn` (stderr, never stdout — INV-6) says so and names #1530. `blis replay`/`observe` reject node pools, so this is `blis run` only (INV-13 parity N/A). Scope: multi-node TP placement only; pipeline parallelism (the other half of #1529) is a separate design-doc-first PR (#1535). Cross-*pool* spanning (one TP group over mixed GPU types) is out of scope. Known limits (both tracked as #1536): (1) the autoscaler's per-variant `CostPerReplica` (keyed on `{GPUType, TPDegree}`) still takes the first replica's cost, so a mixed span/non-span population within one variant is a single-value approximation; (2) the autoscaler's scale-up headroom (`GreedyEngine.Optimize`) gates on free GPU *count*, not free whole *nodes*, so for a spanning variant it can propose a scale-up that `PlaceInstance` then rejects (e.g. 16 free GPUs spread as one whole node + two half-free nodes cannot host a whole-node TP=16) — the failed scale-up is logged and returned harmlessly, but may repeat each autoscaler tick.
```
jgchn: This new section in CLAUDE.md is very long. Consider trimming to ~10-15 lines?

sim/cluster/infra_placement.go
```diff
@@ -227,26 +267,193 @@ func (pm *PlacementManager) PlaceInstance(id InstanceID, model, gpuType string,
...
+		if gpn <= 0 {
+			continue
+		}
+		if tpDegree <= gpn || tpDegree%gpn == 0 {
+			return true
+		}
+	}
+	// No matching pool → not a shape problem; let the generic error describe it.
+	return !matched
+}
```
jgchn, line 399: When there is no matching pool at all (`matched == false`), the function returns `true` (meaning "satisfiable by shape"). The comment explains the rationale: "let the generic capacity error describe it." This is correct in the sense that it avoids a misleading shape-error when the real problem is "no pool of that GPU type exists." However, a reader might be confused by the double-negative logic (`return !matched` where `matched=false` → `return true`). A one-line clarifying comment at the call-site in PlaceInstance ("`tpDegreeSatisfiableByShape` returns true when no matching pool exists, we fall through to the generic error") would help.

sim/cluster/infra_placement.go
namasl: Trimmed in 74822581: the inline Pass-2 block is now a short pointer to the method docblock (which already carries the full whole-node rationale).

sim/cluster/cluster.go
namasl: Good catch — fixed in 74822581. Deferrals are now accumulated during the construction loop and emitted as a single summary warning afterward ("N of M instance(s) not placed at startup (first error: …)"), so an InitialNodes=0 cluster gets one line, not N. The first-error text still distinguishes a structurally-unsatisfiable tpDegree from a transient shortfall, and still notes NodeReady isn't wired in blis run.

CLAUDE.md
namasl: Trimmed in 74822581 from one ~450-word paragraph to ~250 words (~12-14 rendered lines), keeping the load-bearing facts (whole-node rule, run-only scope, deferred #1530 latency, #1536 known limits).

sim/cluster/infra_placement.go
namasl: Added a clarifying note at the call site in 74822581: "tpDegreeSatisfiableByShape returns true when NO matching pool exists (a 'no such GPU type' problem, not a shape one), so that case falls through to the generic capacity error below."

namasl:
Thanks @jgchn — all four review comments addressed in `74822581` (replied inline on each thread). Summary:

- **Startup warning noise** (cluster.go): deferrals are now accumulated and emitted as a **single summary warning** after the construction loop (`"N of M instance(s) not placed at startup (first error: …)"`) instead of one per instance — so an `InitialNodes=0` cluster prints one line, not N. The R1 diagnostic is preserved: the first-error text still distinguishes a structurally-unsatisfiable `tpDegree` from a transient shortfall and notes `NodeReady` isn't wired in `blis run`.
- **Pass-2 docstring** (infra_placement.go): trimmed the inline block to a short pointer to the method docblock (which carries the full whole-node rationale).
- **CLAUDE.md section**: trimmed from ~450 words to ~250 (~12-14 rendered lines).
- **`tpDegreeSatisfiableByShape` double-negative**: added a one-line note at the call site clarifying that it returns `true` when no matching pool exists, so that case falls through to the generic capacity error.

No behavior change to placement/cost/tests. `go build ./...`, `go test ./...`, `golangci-lint run ./...` all green; INV-6 re-verified byte-identical (the summary warning is stderr).

@claude /blis-pr-review

namasl:
Fixed in `4f33858b`. You're right — `unplacedFirstErr` was reassigned every iteration, so despite the "first error" label it held the *last* error, masking `instance_0`'s actionable structural message. Changed to keep the first (`if unplacedFirstErr == nil { unplacedFirstErr = err }`).

Also added the test you flagged as missing: `TestStartupPlacement_UnplacedEmitsOneSummaryWarning` — a 3-instance `tp=12` / 8-GPU-node cluster asserting (a) exactly one summary line, (b) the "3 of 3" count, and (c) that the reported error is the structural "unsatisfiable" message. It fails against the old last-error behavior.

`go build ./...`, `go test ./...`, `golangci-lint run ./...` green; INV-6 byte-identical.

@claude /blis-pr-review

namasl:
Round 5 addressed in `aea5bd40`. IMP-1 and IMP-2 are pre-existing conditions that this PR's span-decision and `× span` cost multiplier promote from cosmetic to silently-wrong, so I've **guarded** both (loud, not silently wrong) and filed **#1543** to track the proper fixes.

- **IMP-1** (per-role TP diverges from placement TP): `NewClusterSimulator` now panics when a `--prefill-tp`/`--decode-tp` override differs from the global TP *and* node_pools are configured — placement/span/cost use the global TP, so a differing per-role TP would be simulated at one degree but placed/billed at another. A per-role placement path is future work (#1543). Used `panic` (not `logrus.Fatalf`) to match the library boundary — `NewClusterSimulator` has 83 config-invariant panics; Fatalf belongs in `cmd/`.
- **IMP-2** (two pools sharing `gpu_type` → wrong cost): `NewPlacementManager` now rejects duplicate `gpu_type` across pools. Cost/`gpu_memory_gib` are resolved by first-match on `gpu_type`, so duplicates would mis-bill the `× span` multiplier and mis-size KV. Pool-identity lookup (making duplicate types legal) is future work (#1543).
- **IMP-3** (pre-existing R1): `NodeReadyEvent.Execute` now `logrus.Debugf`s on `MarkNodeReady` error before returning, instead of a silent return.
- **IMP-4**: corrected the `TestStartupPlacement_UnplacedEmitsOneSummaryWarning` docstring — you're right that with a single global `config.TP`, all instances fail with the same error kind, so it can't exercise the first-vs-last ordering; the docstring no longer claims it (and that scenario isn't reachable through `NewClusterSimulator` today).
- **MIN-1**: re-verified — the `#1529` CLAUDE.md paragraph is present in both `HEAD` and the working tree; the `git status` snapshot was a transient artifact (same as round 2's MIN-7).

New tests assert both guard panics (`TestNewPlacementManager_RejectsDuplicateGPUType`, `TestNewClusterSimulator_RejectsPerRoleTPWithNodePools`). `go build ./...`, `go test ./...`, `golangci-lint run ./...` green; INV-6 byte-identical.

@claude /blis-pr-review

namasl:
Thanks — round 6 came back READY TO MERGE. Closed out the two non-blocking notes in `681b415e`:

- **D-1** (undocumented constraints): documented both startup constraints — per-role TP override rejected under `node_pools`, and pools must have distinct `gpu_type` — in the CLAUDE.md #1529 note and a new "Configuration constraints" block in `docs/guide/cluster.md`'s Multi-node TP section, both pointing at #1543.
- **Decode-path coverage**: added `TestNewClusterSimulator_RejectsDecodeTPWithNodePools` so the decode arm of the guard loop is explicitly tested (prefill was already covered).
- **D-2** (TP=0 edge-case message): no change needed — behavior and message are correct as you noted.

Docs + test only, no production logic change. `go build ./...`, `go test ./...`, `golangci-lint run ./...` green; INV-6 byte-identical.

@claude /blis-pr-review

susiejojo:
### BLIS PR Self-Review — PR #1537 (vLLM cross-reference + latency-model focus)

This round targets the two questions in the review request: **(1) validate the multi-node TP model against the actual `vllm` codebase with code proofs**, and **(2) does multi-node TP require latency-model updates now?** All prior rounds (1–7) reached READY TO MERGE; I did not re-litigate those findings. I inspected the vLLM tree at `/Downloads/vllm` (HEAD `12a34a6bc`) and the BLIS trained-physics + roofline latency models.

## Verdict: READY TO MERGE — 0 blocking. 1 important (latency, correctly deferred to #1530), 1 doc-accuracy nit.

No new correctness bug. The placement algebra, R5 atomicity, per-node drain fix, and cost multiplier are sound and match `cost_per_hour` being *per-node* (`infra_config.go:63`). The two things below are about **modeling fidelity vs vLLM**, which is exactly what the request asked me to check.

---

## Finding 1 (doc accuracy, minor) — "the only physically-realizable multi-node TP shape" is overstated

The docblock (`infra_placement.go`) and CLAUDE.md justify the `tpDegree % gpus_per_node == 0` + whole-node constraint by asserting a uniform per-node rank count is *"the only physically-realizable multi-node tensor-parallel shape."* Against vLLM, that is true for one backend and false for the other:

- **Matches vLLM's multiprocessing (`mp`) backend exactly.** It hard-asserts divisibility and forces equal ranks per node:
  ```python
  # vllm/v1/executor/multiproc_executor.py:262
  assert self.world_size % self.parallel_config.nnodes_within_dp == 0, (...)
  # vllm/config/parallel.py:722
  def local_world_size(self) -> int:
      return self.world_size // self.nnodes_within_dp
  ```
- **Stricter than vLLM's Ray backend.** Ray only *warns* on an asymmetric spread and uses `PACK` (best-effort), not `STRICT_PACK`, so a 2+1-style split is permitted:
  ```python
  # vllm/v1/executor/ray_utils.py:317
  if len(bundles) < parallel_config.tensor_parallel_size:
      logger.warning("tensor_parallel_size=%d is bigger than a reserved number of %ss ... "
                     "Tensor parallel workers can be spread out to 2+ nodes which can degrade "
                     "the performance unless you have fast interconnect ... like Infiniband ...")
  # ray_utils.py:664  strategy="PACK"
  ```

**Assessment:** being stricter than the most permissive backend is a *safe* modeling choice, and it matches vLLM's own **recommended** practice (`docs/serving/parallelism_scaling.md:122`: "set the tensor parallel size to the number of GPUs in each node"). So the constraint itself is fine — only the wording "*the only* physically-realizable shape" is too strong. Suggest softening to something like "the standard / recommended multi-node TP shape (and the shape vLLM's `mp` backend enforces); asymmetric spreads are possible under Ray but discouraged." Non-blocking.

---

## Finding 2 (latency, IMPORTANT — but correctly out of scope via #1530) — the optimism claim is verified, and the gap is potentially order-of-magnitude

The PR states the TP all-reduce term "assumes intra-node NVLink and is placement-blind, so a spanning instance's all-reduce is priced optimistically." **This is precisely accurate**, confirmed in code:

- The comm basis divides by **GPU HBM bandwidth**, not any link bandwidth, and folds the NVLink ratio into a *constant* coefficient β₄:
  ```go
  // sim/latency/trained_physics_model.go:495  tpAllReduceBasis
  return units * tokens * float64(m.hiddenDim) * m.activationBPP * 2.0 * tpFactor / m.bwHbmUs
  // :350  "β₄ absorbs the NVLink/HBM ratio (~0.27 on H100)."
  ```
- `HardwareCalib` (`sim/model_hardware_config.go:69`) has **no** interconnect field — only `TFlopsPeak`, `BwPeakTBs` (HBM), `MemoryGiB`. The latency model has **no node-count input**: TP enters only as the degree `m.tp`. So it literally cannot distinguish a 1-node TP=8 from a 2-node TP=16 beyond the ring factor `(tp-1)/tp` (0.875 → 0.9375, ~7%).
- The roofline model is even more optimistic — **zero** all-reduce term (`sim/latency/roofline.go`; TP is a pure divisor).

Two things make this worth flagging louder than a routine deferral:
1. **Direction + magnitude.** Cross-node TP all-reduce runs over InfiniBand/Ethernet (≈ an order of magnitude below NVLink), yet the model keeps charging the NVLink rate. The only degree-sensitivity, `(tp-1)/tp`, nudges the *wrong way* (slightly higher comm for higher TP) while ignoring the bandwidth cliff. On comm-dominated steps the under-pricing can be large.
2. **vLLM actively discourages exactly this config.** `docs/serving/parallelism_scaling.md:8-11,23,122` recommend **PP across nodes, TP within a node**, *because* per-layer TP all-reduce is brutal cross-node. So a capacity-planning user pointed at multi-node TP by this feature is being steered toward vLLM's non-recommended topology while the sim reports optimistic latency/throughput for it.

**Does it block merge?** No. #1530 explicitly owns interconnect pricing, the ordering in #1532 requires *this* PR (an instance able to span nodes) to land first, and the one-time stderr `logrus.Warn` fires on all three placement paths (startup → `PlaceInstance`; deferred → `RetryPendingInstances` → `PlaceInstance`; scale-up → `PlaceInstance`) so the optimism is surfaced. INV-6 is preserved (stderr only). This is the correct scope boundary.

**Suggestions (non-blocking, for #1530 or a doc tweak here):**
- The `cluster.md` "Cross-node interconnect not yet priced" admonition could add one sentence that vLLM recommends PP-across-nodes over TP-across-nodes, so users understand the modeled config is itself the aggressive choice — not just "optimistic numbers."
- #1530 should treat the comm term as bandwidth-parameterized (add an interconnect-bandwidth field to the hardware calib and a node-span input), not merely re-tune β₄ — β₄ is a single constant and cannot express "NVLink within a node, IB across nodes" for a partially-spanning fleet.

---

## Cross-path parity ✅
Correctly `blis run` only — `replay`/`observe` `logrus.Fatalf` on node pools, so INV-13 is N/A (verified). The `× span` cost rule (`InstanceCostPerHour`) is applied identically at all three placement sites (startup `cluster.go`, deferred `infra_lifecycle_event.go`, scale-up `direct_actuator.go`). No preemption/timeout surface touched.

## Bottom line
Ship it. The core mechanism is correct and well-tested; the latency gap is real but correctly fenced to #1530 with a live warning. Recommend the two wording tweaks (Finding 1 "only physically-realizable"; Finding 2 doc note on vLLM's PP-over-TP recommendation) either here or as a fast follow.

namasl:
Thanks @susiejojo — great vLLM cross-reference. Both doc findings fixed in `66bc5cdc`:

- **Finding 1** (wording overstated): reworded the "only physically-realizable multi-node TP shape" claim in the `PlaceInstance` docblock, CLAUDE.md, and `docs/guide/cluster.md`. It now says the whole-node shape is what vLLM's **`mp` executor enforces** (asserts `world_size % nnodes == 0`, equal `local_world_size`) and what its docs recommend, while noting vLLM's **Ray backend tolerates an asymmetric spread but discourages it** — so BLIS is deliberately stricter than the most permissive backend. The constraint itself is unchanged (stricter-than-permissive is a safe modeling choice); only the justification wording.
- **Finding 2** (aggressive topology): the "Cross-node interconnect not yet priced" admonition in `cluster.md` now notes that cross-node TP is itself the aggressive choice — vLLM recommends **PP-across-nodes + TP-within-node** because per-layer TP all-reduce is punishing over inter-node links (~an order of magnitude below NVLink) — so the #1530 optimism can be large, not a rounding gap.

Agreed on your **Finding 2 latency** point being correctly fenced to #1530, and on the suggestion there that the fix should be bandwidth-parameterized (interconnect field + node-span input) rather than a β₄ re-tune — I've noted that on #1530's scope so it isn't lost.

Docs/comments only, no logic change. `go build ./...`, `go vet ./...`, `golangci-lint run ./...` green.

@claude /blis-pr-review

namasl:
Round-8 doc-accuracy findings fixed in `56b64e95`:

- **I-1** (self-contradiction): you're right — citing vLLM as "recommending" the whole-node shape contradicted the PP-over-TP admonition (vLLM's full recommendation prescribes TP=8/PP=2 for a 2×8 box, i.e. *not* the TP=16 shape). Dropped the "docs recommend" clause in the `PlaceInstance` docblock, CLAUDE.md, and `cluster.md`; the justification now rests only on the factual `mp`-executor enforcement (`world_size % nnodes == 0`, equal per-node ranks), and each spot explicitly notes vLLM recommends avoiding multi-node TP entirely (→ #1530). No more arguing both sides.
- **I-2** (doc-vs-code): `docs/reference/configuration.md:579` still documented the pre-#1537 "first-listed pool's memory wins" behavior for shared `gpu_type` — which this PR now panics on. Updated that admonition (duplicate `gpu_type` panics; #1543 tracks lifting it) and added both startup constraints to the `node_pools` schema block at `:441`.
- **I-3** (wrong claim): corrected the CLAUDE.md #1536 note — per-variant `CostPerReplica` **is** span-aware (span-multiplied `inst.CostPerHour` flows through the collector→analyzer; first-replica is correct given the global-TP guard + unique-`gpu_type` requirement). Only the scale-up **headroom** is span-unaware.

Also took the easy round-8 minors: **M-2** (`instance.go` `nodeID` comment now points the #1530 implementer at `distinctNodesForGPUs`), **M-3** (warning tests key on the stable `"spans"` token, not `"#1530"`), **M-4** (`SpansThreeNodes` now asserts the lowest-index primary-node contract), **M-5** (added `TestNewClusterSimulator_PerRoleTPEqualToGlobalOK` — per-role TP equal to global must not panic).

The "file separately" observations (KV-capacity-at-high-TP, `MarkNodeTerminated` not span-aware, tautological `VerifyConservation`, deferral-as-silent-drop) are noted — happy to file them as follow-ups; they're pre-existing / out of this PR's scope.

Docs + tests only, no production logic change. `go build ./...`, `go test ./...`, `golangci-lint run ./...` green; INV-6 byte-identical.

@claude /blis-pr-review

namasl:

Context for this round: the branch was merged with current `upstream/main` (25 commits; head now `eaab5c7a`). Only `CLAUDE.md` conflicted (both sides added feature notes — resolved by keeping all: my #1529 note + upstream's #1589/#1590/#1591/#1581 KV-offload notes); every other file auto-merged. No production logic changed in the merge. Post-merge: `go build`/`go test ./...`/`golangci-lint` all green, and INV-6 re-verified byte-identical to current `upstream/main` (single- and multi-instance-TP).

namasl:

## qa-review — PR #1537: ✅ PASS
_feat(cluster): multi-node TP placement — an instance can span nodes within a pool (Closes #1529) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The rescoped issue requires whole-node multi-node TP placement within a pool. The PR implements a single-node-first, whole-node spanning fallback; preserves atomic allocation and release; charges each occupied node; and wires startup, deferred, and autoscaler placement paths. The additional validation, warnings, and multi-node release handling directly prevent incorrect placement, billing, or lifecycle behavior. Pipeline parallelism and cross-node communication latency remain explicitly out of scope. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. User-facing documentation covers eligibility, single-node preference, fragmentation behavior, per-node billing, configuration restrictions, optimistic inter-node latency, and run-only support. Configuration reference comments and project-level implementation notes were also updated. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | There are a few imprecise comments left. `NewPlacementManager`'s constructor comment says it panics for invalid pools or duplicate names but omits the new duplicate-`gpu_type` panic. `InstanceSimulator.CostPerHour` and `ReplicaMetrics.CostPerHour` still describe the value as directly from `NodePool`, although spanning instances now store the node-span-multiplied per-instance cost. These do not affect runtime behavior. |
| G1 | correctness | ✅ CONFIDENT | Pass 2 in `PlaceInstance` requires `node.freeCount() == node.TotalGPUs` for whole-node occupancy. What happens if a pool has sufficient aggregate free GPUs for `tpDegree` (e.g. `tp = 16`, `gpus_per_node = 8`), but some nodes are partially occupied so fewer than `nodesNeeded` fully-free nodes exist? Does Pass 2 correctly leave the instance pending, and could partial allocations cause an infinite placement retry loop? | Pass 2 correctly rejects that fragmented situation. It collects only fully-free Ready nodes and commits no GPU mutations unless it obtains `nodesNeeded`; otherwise the instance remains pending. Retries are bounded to at most the pending-list size per `RetryPendingInstances` invocation, so partial occupancy cannot create an internal infinite retry loop. The instance may remain pending indefinitely if no future node-ready event supplies enough whole nodes, which is intended. |
| G2 | correctness | ✅ CONFIDENT | In `ReleaseInstance`, `touched` tracks nodes whose GPUs were freed to execute drain callbacks in sorted ID order. If an instance spans multiple nodes and one node is in `NodeStateDraining` while another is in `NodeStateReady`, could clearing `node.drainCallback` on the draining node trigger a state change that leaves the multi-node cluster in an inconsistent drain state? | No inconsistency is introduced. Release first frees all GPUs belonging to the instance, then examines each touched node independently. Only a node that is still Draining, now has zero allocations, and has a callback fires; a Ready node is left unchanged. The callback is cleared before invocation, preventing duplicate firing. In the lifecycle path, the callback schedules that specific node's `NodeDrainedEvent`, which later performs the Draining-to-Terminated transition. |
| G3 | invariants | ✅ CONFIDENT | INV-6 guarantees deterministic simulation output across runs. In Pass 2 of `PlaceInstance`, candidate nodes are selected from `pm.nodesByPool[poolState.config.Name]`. Is the order of nodes in `nodesByPool` strictly deterministic across runs, and how does `PlaceInstance` guarantee that identical multi-node placement decisions are made regardless of map iteration order? | Yes. Pools are stored in declaration-order slices, initial nodes are appended in sequential creation order, and provisioned nodes use deterministic per-pool indices. Passes 1 and 2 iterate those slices rather than `nodesByID`, so map iteration cannot choose placement candidates. The only relevant map-derived node list used during release/cost resolution is sorted before order-sensitive use. |
| G4 | invariants | ✅ CONFIDENT | INV-1 guarantees resource and cost conservation. When a multi-node instance is placed, `InstanceCostPerHour` multiplies `poolCostPerHour` by the distinct node span count. If an instance spans multiple nodes, does `addLiveInstance` and subsequent metrics aggregation in autoscaling policies properly account for this span multiplier without undercounting total cluster operating cost? | Yes for the autoscaling cost signal. All three construction paths calculate the span-adjusted value before assigning `InstanceSimulator.CostPerHour`. Router snapshots copy that value, the collector copies it into `ReplicaMetrics`, and the analyzer exposes it as `VariantCapacity.CostPerReplica`. Thus a TP=16 placement over two $10/hour nodes is represented as a $20/hour replica rather than $10/hour. The accompanying startup, deferred, and autoscaler tests verify those three paths. |
| G5 | rules | ✅ CONFIDENT | Rule R5 requires select-then-commit transactional atomicity. In `PlaceInstance` Pass 2, `selectedNodes` gathers whole free nodes before setting `gpu.AllocatedTo = id`. Is there any code path where partial allocation could occur if `selectedNodes` contains duplicate nodes or if an allocation loop is interrupted? | No realistic partial-allocation path exists. `selectedNodes` is built from one traversal of a node slice, so it cannot contain duplicates under the manager's construction invariant. Capacity and readiness are fully checked before mutation, and the commit loop has no error-returning operations. The simulation is single-threaded, so another allocator cannot interrupt between selection and commit. |
| G6 | rules | ✅ CONFIDENT | Rule R1 forbids silent error handling. In `distinctNodesForGPUs`, if a GPU ID is missing from `pm.gpusByID`, `logrus.Errorf` is logged and the GPU is skipped. Why is an unindexed GPU ID treated as a non-fatal error rather than a panic or hard error, and could skipping a GPU cause `InstanceCostPerHour` to underbill a spanning instance? | Skipping is defensive handling for an invariant-corrupt or hand-supplied GPU list, not a reachable placement result. Every GPU is indexed at the sole node-construction site, and `PlaceInstance` returns IDs from those same GPU objects. Therefore actual placements cannot underbill this way. If corruption did occur, the helper emits an error and `InstanceCostPerHour` also emits an error and returns zero when no node resolves, making the problem visible rather than silently charging a plausible value. |
| G7 | parity | ✅ CONFIDENT | Multi-node TP placement is only enabled for `blis run` (`blis replay`/`observe` reject node pools per docs). How does the system ensure trace headers generated by `blis run` preserve node span information so that downstream analysis tools do not miscalculate per-instance hardware costs? | Run-generated workload trace headers do not preserve node span or deployment cost. They contain workload-oriented metadata, and replay explicitly rejects node-pool configuration. This does not currently cause downstream cost miscalculation because replay/calibration does not reconstruct or report node-pool hardware costs from those traces; live autoscaling receives the effective span-adjusted cost through router snapshots instead. |
| G8 | parity | ✅ CONFIDENT | Pass 1 of `PlaceInstance` preserves single-node first-fit placement verbatim to maintain byte-identical output with pre-#1529 single-node runs (INV-6). Is there any scenario where adding Pass 2 alters the outcome or timing of single-node placements in a multi-pool setup? | No scenario changes an otherwise-successful single-node placement. Pass 1 traverses all matching pools and nodes in the original first-fit order and returns immediately on success; Pass 2 is entered only after every matching Ready node has failed. Consequently a later pool's single-node fit also wins over an earlier pool's spanning opportunity. Failed placements may now succeed via Pass 2 and emit a stderr warning, but that is the intended new behavior and does not alter simulation-clock timing. |
| G9 | tests | ✅ CONFIDENT | The new test file `infra_placement_multinode_test.go` exercises `distinctNodesForGPUs` and `PlaceInstance` spanning. Is there a companion test verifying that `ReleaseInstance` correctly frees all GPUs across a multi-node span and fires all registered drain callbacks on multiple draining nodes? | Yes. `TestReleaseInstance_SpanningFiresDrainPerNode` places a TP=16 instance over two nodes, drains both nodes, releases the instance, verifies both callbacks fired, verifies all eight GPUs on each node are free, and checks conservation. A separate regression test verifies the original single-node callback behavior. |
| G10 | tests | ✅ CONFIDENT | Would the multi-node placement tests in `infra_placement_multinode_test.go` pass if `PlacementManager` were changed to track GPU allocations via node references instead of string GPU IDs in `gpusByID`? | Most placement behavior tests would remain valid if the public/internal placement result continued to expose GPU IDs and `distinctNodesForGPUs` retained equivalent semantics. The helper-specific tests intentionally exercise the current GPU-ID-to-node resolver and would need adaptation if `gpusByID` and string-ID resolution were removed entirely. They are behavioral rather than dependent on parsing the current ID format. |
| G11 | rationale | ✅ CONFIDENT | `PlaceInstance` Pass 2 requires `tpDegree > gpn` and `tpDegree % gpn == 0`, taking whole nodes only and rejecting asymmetric spreads (e.g., 2+1 GPUs across nodes). What alternatives were considered for asymmetric or fragmented multi-node TP placement, and why was whole-node strictness preferred? | The documented alternatives were asymmetric spreading through a permissive Ray-style backend and avoiding cross-node TP by using TP within a node plus pipeline parallelism across nodes. Whole-node strictness was selected because it guarantees equal ranks per node, matches vLLM's multiprocessing executor shape, avoids modeling discouraged fragmented groups, and keeps each span within one homogeneous pool/interconnect domain. Pipeline parallelism and inter-node communication pricing are separate scoped issues. |
| G12 | rationale | ✅ CONFIDENT | `NewPlacementManager` panics at startup if two node pools share the same `gpu_type`. Why was unique `gpu_type` chosen as a hard prerequisite rather than keying hardware calibration, cost, and capacity by pool name (`Name`)? | Unique `gpu_type` is a compatibility guard around the existing metadata lookup architecture: `PlaceInstance` returns only the matched GPU type, and startup, deferred, and autoscaler paths resolve cost, memory, and hardware calibration by first matching that type. Returning and propagating pool identity would require a broader API and construction-path refactor. The PR therefore rejects ambiguity rather than silently selecting the wrong pool; pool-identity lookup is tracked separately. |
| G13 | scope | ✅ CONFIDENT | `NewClusterSimulator` added a startup check that panics if per-role TP overrides (`--prefill-tp` or `--decode-tp`) differ from global `--tp` when `node_pools` are used. Does this PR scope extend beyond multi-node TP placement by enforcing this per-role TP restriction on all node-pool deployments, even single-node ones? | Yes, the restriction applies to every node-pool deployment, including one where the requested TP fits a single node. That is intentional rather than unrelated scope expansion: placement, assigned `TPDegree`, node span, cost, and capacity currently use global `config.TP`, while role-specific simulation configuration can use a different TP. The mismatch is incorrect even without spanning, so the guard prevents silently inconsistent node-pool simulations until role-aware placement exists. |
| G14 | scope | ✅ CONFIDENT | As noted in CLAUDE.md, the autoscaler's headroom calculation (`GreedyEngine.Optimize`) remains span-unaware (#1536). Why was updating `Optimize` to account for whole-node spans excluded from this PR, and how does the cluster prevent autoscaler oscillation when `Optimize` proposes unplaceable spanning instances? | The optimizer change was excluded because the accepted scope is placement, while span-aware autoscaler headroom requires changing inventory/optimization semantics. The cluster safely handles a bad proposal: `PlaceInstance` performs select-then-commit, returns an error without reserving anything, and `DirectActuator` logs and returns that error. It does not, however, suppress the same proposal at a later autoscaler tick; repeated failed attempts are possible until topology or demand changes. |
| G15 | errors | ✅ CONFIDENT | When startup instances cannot be placed in `NewClusterSimulator`, a single `logrus.Warn` summary is printed noting that `NodeReady` is not wired in `blis run`. Why does failing to place an instance at startup result in a warning (leaving unplaced pending instances) rather than halting execution via `logrus.Fatalf`? | Placement failure remains a recoverable library condition because the placement subsystem supports pending instances and retry on future `NodeReadyEvent`s; halting from `sim/` would also violate the library rule against process termination. Since the current `blis run` production path does not wire those retries for startup shortfalls, the PR adds one explicit summary warning so the effective drop is visible while preserving existing deferred-placement semantics and embeddability. |
| G16 | errors | ✅ CONFIDENT | `PlaceInstance` returns a specialized error when `tpDegree` is structurally unsatisfiable (`tpDegreeSatisfiableByShape` is false). However, `tpDegreeSatisfiableByShape` returns `true` when no pool matching `gpuType` exists. Why does a missing GPU type return the generic capacity error instead of an explicit 'no matching pool' error? | The specialized error is deliberately limited to shape incompatibility among pools that actually match the requested GPU type. With no matching pool, divisibility and `gpus_per_node` shape cannot be evaluated, so the helper returns true and lets the generic no-capacity path report the failure. This is safe and non-silent, though less specific than a dedicated missing-pool diagnostic. |

### Items to fix

### Important to consider
- **F1** — The autoscaler inventory heuristic remains whole-node-span-unaware, as explicitly tracked separately in #1536.
  _sim/cluster/infra_placement.go:204-352; sim/cluster/infra_placement.go:432-498; sim/cluster/cluster.go:356-425; sim/cluster/infra_lifecycle_event.go:48-82; sim/cluster/direct_actuator.go:65-120; docs/guide/cluster.md:53-68_
- **F3** — Consider changing the cost comments to say effective per-instance hourly cost and documenting duplicate GPU types in the constructor summary.
  _sim/cluster/infra_placement.go:61-64; sim/cluster/infra_placement.go:74-97; sim/cluster/instance.go:39-44; sim/cluster/autoscaler.go:38-48; sim/cluster/infra_placement.go:432-449_
- **G4** — The analyzer takes one cost value per variant rather than summing replica costs because `CostPerReplica` is a selection metric; the global-TP and unique-GPU-type guards ensure replicas of a variant have the same span cost.
  _sim/cluster/cluster.go:387-411; sim/cluster/infra_lifecycle_event.go:63-80; sim/cluster/direct_actuator.go:91-113; sim/cluster/cluster_event.go:80-106; sim/cluster/default_collector.go:38-54; sim/cluster/saturation_analyzer.go:100-121; sim/cluster/infra_placement_multinode_test.go:637-656; sim/cluster/infra_placement_multinode_test.go:693-755_
- **G6** — Returning an error or panicking would enforce the invariant more strongly, but the current behavior cannot affect IDs produced by the shipped placement path.
  _sim/cluster/infra_placement.go:116-145; sim/cluster/infra_placement.go:264-285; sim/cluster/infra_placement.go:332-352; sim/cluster/infra_placement.go:401-449; sim/cluster/infra_placement_multinode_test.go:105-124_
- **G7** — If a future trace consumer is expected to reproduce infrastructure cost, the trace schema will need explicit deployment/span metadata or a separate cost artifact.
  _cmd/root.go:2254-2275; cmd/replay.go:343-357; sim/cluster/cluster_event.go:80-106; sim/cluster/default_collector.go:38-54_
- **G13** — Supporting per-role TP with node pools is tracked as a broader placement refactor in #1543.
  _sim/cluster/cluster.go:310-329; sim/cluster/cluster.go:335-360; docs/guide/cluster.md:62-65; sim/cluster/infra_placement_multinode_test.go:143-202_
- **G14** — There is no anti-oscillation/backoff mechanism for this case; #1536 is owed follow-up coverage and implementation.
  _sim/cluster/engine.go:51-79; sim/cluster/infra_placement.go:310-330; sim/cluster/direct_actuator.go:65-74; CLAUDE.md:624_
- **G15** — A CLI-level preflight could choose stricter failure semantics for structurally unsatisfiable configurations, but that is a policy choice rather than a placement correctness defect.
  _sim/cluster/cluster.go:342-374; sim/cluster/cluster.go:447-455; sim/cluster/infra_placement.go:580-621; sim/cluster/infra_lifecycle_event.go:32-50; docs/contributing/standards/rules.md:90-98_
- **G16** — An explicit 'no matching pool' error would improve diagnostics but would not change behavior.
  _sim/cluster/infra_placement.go:355-398_

namasl:
@claude /blis-pr-review in addition to all your usual checks, also consider the output of /archon-pr-review above (it already ran on this exact head `eaab5c7a` — verdict `ARCHITECTURAL_CHANGE`, i.e. the new `PlacementManager.InstanceCostPerHour` public method + new test invariants; no architectural regression).

No code change since Round 10 (READY TO MERGE) — this is the same merged head. Re-running to (a) re-confirm the verdict on the current head and (b) clear the stale `Claude Code Action (comment trigger)` required status that a prior non-command comment left pending, which is currently blocking the merge button independent of review approval.

susiejojo (approved):
lgtm

PR #1538, opened 2026-08-08
feat(latency,sim): model speculative decoding / MTP throughput (#1528)

### LINKED ISSUE #1528: Speculative decoding / MTP not modeled — decode throughput under-predicted

namasl:
## Summary

BLIS decode advances **exactly one token per sequence per step**, so decode throughput is systematically under-predicted for any model that ships speculative decoding / Multi-Token Prediction (MTP) — GLM-5.2 (5-token MTP), DeepSeek-V3, Kimi-K3, EAGLE/Medusa. BLIS's own calibration report already admits this (`sim/workload/calibrate.go`: *"If the real server uses speculative decoding, actual token generation patterns differ from sim's sequential model."*).

This PR models the throughput effect of speculative decoding. It is **off by default** — a run without the new flags is byte-identical to a pre-feature build (INV-6) — and `K=0` is a hard, tested gate on every touched code path.

Fixes #1528. Part of #1532 (GLM-5.2 support).

## Model: decouple cost from throughput

A decode step under speculative decoding verifies `K` draft tokens plus 1 always-generated bonus token in a single target forward pass. The PR splits this into two **decoupled** quantities so that accepting drafts is not modeled as free (the failure mode the issue explicitly warns against):

- **Verify width `w = K+1`** drives the per-step **cost**. Both latency backends scale only their decode token-population terms (decode compute FLOPs, KV read/write bytes) by `w`; the once-per-step weight-load, TP/EP communication, and constant-overhead terms are **not** scaled. Cost is therefore *sublinear* in `w` — the physics behind the real speedup, and why verifying is cheap at low occupancy.
- **Accepted tokens `g = 1 + α·K`** (α = user-supplied mean acceptance rate) drives **progress**. A per-request deterministic **fractional carry** advances `ProgressIndex` by `g` tokens/step. No RNG → determinism (INV-6) holds and every metric is the expectation. A peek/commit split debits the carry by the *granted* count (after KV/MaxModelLen capping), so capping defers rather than drops tokens — no long-run drift.

`α` is user-supplied, not predicted — BLIS does not run a real draft model (per the issue's suggested first cut).

## CLI

| Flag | Default | Meaning |
|------|---------|---------|
| `--num-speculative-tokens K` | `0` (off) | draft tokens proposed per decode step |
| `--speculative-acceptance-rate α` | — | mean fraction accepted, `[0,1]`; **required** when `K>0` |
| `--speculative-method` | `""` | informational label: `mtp\|eagle\|medusa\|ngram\|draft` |

```bash
./blis run --model zai-org/GLM-5.2 \
  --num-speculative-tokens 5 --speculative-acceptance-rate 0.7 --speculative-method mtp
```

Requiring `--speculative-acceptance-rate` when `K>0` prevents the footgun where `α` defaulting to 0 would model MTP as pure slowdown (verify-width cost with no throughput gain). Model-level config, supplied by the **shared** flag set to both `blis run` and `blis replay`, so a trace round-trips under identical flags — the same INV-13 mechanism as `--tp` / `--max-model-len` (no TraceV2 schema change).

## Behavioral contracts

- **BC-1 (off byte-identical):** `--num-speculative-tokens 0` (or unset) ⇒ byte-identical stdout vs a build without the feature. *(INV-6, INV-BC-DP1)*
- **BC-2 (throughput):** with `K>0, α>0`, the same output completes in ≈`1+α·K` fewer decode steps; total output tokens unchanged.
- **BC-3 (cost not free):** step time is strictly increasing in `K` for fixed effective `g`, but sublinear in `w`.
- **BC-4 (conservation):** counted output tokens `== len(OutputTokens)` exactly, even when a multi-token final step overshoots `ProgressIndex`. *(INV-1)*
- **BC-6/7 (determinism, parity):** fixed seed ⇒ byte-identical; run→export→replay per-request TTFT/E2E identical. *(INV-6, INV-13)*
- **BC-9 (validation):** `K<0`, `K>1024`, `α∉[0,1]`, non-finite α, dangling knobs (α/method with `K=0`), and unknown method all error at the CLI boundary.

## Testing

- **Config validation** (`sim/config_test.go`): table-driven accept/reject incl. the `MaxSpeculativeTokens=1024` ceiling.
- **Latency cost law** (`sim/latency/...`): both backends — `K=0` byte-identity, monotone in `K`, sublinear in `w`.
- **Simulator** (`sim/simulator_decode_test.go`): throughput reduces step count, output-token conservation incl. heavy overshoot, determinism, ITL-sum conservation, carry no-drift under capping.
- **Parity** (`cmd/`): true run-vs-replay INV-13 for spec-decode (captures the *run* leg's metrics), a PD decode-sub-request completion/conservation test, and a BC-1 stdout byte-identity guard (`--num-speculative-tokens 0` == no flag).

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` are all clean. Empirically: `K=0` is md5-identical to a no-flag run; feature-on conserves output tokens exactly while cutting E2E ~3–4× on a decode-bound workload.

## Scope / deferred

Out of scope (design §9): per-step acceptance *distribution* (this PR uses the mean), per-request heterogeneity, acceptance-rate *prediction*, a separate draft-model engine, and MTP-layer weight/KV *sizing* (that's #1527's Finding 5). The fractional-carry seam is built so a per-step distribution can replace the constant `α` later without reshaping the batch/sim/latency contracts.

**Note on ITL metrics under spec-decode:** raw `AllITLs` percentiles (`itl_p90` etc.) are per-verification-step; use TPOT (`RequestITLs`) for per-token latency. SLO-ITL goodput attainment reads TPOT, so it is unaffected. This is documented in CLAUDE.md and the latency-models guide.


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

Addressed the two minor notes from the previous review (non-blocking): replaced the package-level `validSpeculativeMethods` map with an `isValidSpeculativeMethod` switch (no shared mutable state, R8), and added a comment on the non-LengthCapped TPOT branch explaining why it needs no spec-decode adjustment. No behavior change.

sim/latency/roofline.go
```diff
@@ -322,12 +322,22 @@ func rooflineStepTime(modelConfig sim.ModelConfig, hwConfig sim.HardwareCalib, s
...
+	// tokens per decode request instead of 1. w is carried per-request in
+	// NumNewDecodeTokens (set to K+1 by RooflineLatencyModel.StepTime — the COST
+	// quantity, distinct from the accepted-token count that drives progress). Defaults
+	// to 1 (verify width off / pre-feature callers) ⇒ byte-identical (INV-6).
 	for _, req := range stepConfig.DecodeRequests {
-		f := calculateTransformerFlops(modelConfig, req.ProgressIndex, 1, true, true)
+		w := int64(req.NumNewDecodeTokens)
+		if w < 1 {
+			w = 1 // defensive: never fewer than the 1 bonus token
+		}
```
jgchn, line 342: `NumNewDecodeTokens` is always set to `m.specTokens + 1` (line 125 of latency.go), which is `≥ 1`. The `rooflineStepTime` function is unexported and only called from `RooflineLatencyModel.StepTime`, so no external caller can pass 0. Same for the identical guard at line 357. Not wrong, but unreachable.

sim/simulator.go
```diff
@@ -728,9 +750,24 @@ func (sim *Simulator) recordRequestCompletion(req *Request) {
 			// #588: Use actual decode step count for length-capped requests.
 			// len(req.OutputTokens) is the pre-determined count; len(req.ITL) is actual.
 			// TPOT convention: exclude first generated token → denominator is len(ITL)-1.
-			sim.Metrics.RequestITLs[req.ID] = float64(reqTotalOutput) / float64(max(len(req.ITL)-1, 1))
+			var denom int
+			if sim.specEnabled {
+				// Under speculative decoding / MTP, each ITL entry covers g tokens, so
+				// len(ITL) counts STEPS, not tokens — using it would inflate TPOT by ~g.
+				// Use the token-derived count (ProgressIndex advance + prefill token).
+				tokCount := int(req.ProgressIndex) - int(req.InputLen()) + 1
```
jgchn, line 766: This is correct when `FormBatch`'s `MaxModelLen` cap ensures PI lands exactly at `MaxModelLen-1`. But in the PD path (batch_formation.go:228), the room < 1 flooring case emits `decodeTokens=1` even past the boundary, which can push PI to `MaxModelLen` (one past `MaxModelLen-1`). This would give `tokCount` one higher than the non-PD path for the same MaxModelLen. This asymmetry is pre-existing (PD never had MaxModelLen capping before this PR), so not introduced here, just worth being aware of for the PD TPOT metric.

susiejojo:
## Independent Review: `feat(latency,sim): model speculative decoding / MTP throughput (#1528)`

Reviewed for **run/observe/replay/calibrate parity**, **code quality**, and **vLLM fidelity of the spec-decode flags & model** (verified against a local vLLM checkout via code proofs). This is an independent pass on top of the two prior `/blis-pr-review` rounds; I re-derived the conclusions rather than inheriting them. Build, full `sim/...`+`cmd/...` test suite, and end-to-end binary checks all pass on the checked-out branch.

**Verdict: READY TO MERGE.** No blocking issues. Findings below are all non-blocking (one worth a follow-up note).

---

### vLLM flag & model fidelity (code proofs)

The three flags map onto **real vLLM concepts**, verified against `vllm/config/speculative.py` and `vllm/engine/arg_utils.py`:

| BLIS flag | vLLM source proof |
|---|---|
| `--num-speculative-tokens K` | `speculative.py:89` `num_speculative_tokens: int = Field(default=None, gt=0)`; CLI `--spec-tokens` → `num_speculative_tokens` (`arg_utils.py:1785`). vLLM serves GLM-5.2 with `--speculative-config '{"method":"mtp","num_speculative_tokens":5}'`. ✅ |
| `--speculative-method mtp\|eagle\|medusa\|ngram\|draft` | `SpeculativeMethod`/`MTPModelTypes` literals include `mtp` (`:53`), `eagle` (`:65`), `medusa` (`:69`), `ngram` (`:68`). ✅ (one caveat below) |
| `--speculative-acceptance-rate α` (BLIS-supplied) | vLLM does **not** take a user acceptance rate for real serving — it *measures* it. The PR is explicit that α is user-supplied, not predicted (per the issue's "first cut" suggestion). Correct framing. ✅ |

**Model correctness against vLLM semantics — both key quantities match vLLM's own conventions:**
- **Verify width `w = K+1`.** vLLM processes `num_spec_tokens + 1` positions per target forward pass — proven repeatedly: `ngram_proposer_gpu.py:321` `# [batch_size, num_spec_tokens + 1]`, `:348` `max_new_tokens = ... # num_spec_tokens + 1`, `spec_decode/utils.py:186`. BLIS sizing decode **cost** by `K+1` is faithful.
- **Accepted tokens `g = 1 + α·K`.** vLLM's mean-acceptance-length convention is `metrics.py:114` `mean_acceptance_length = 1 + (num_accepted_tokens / num_drafts)` with the comment `# Conventionally, mean acceptance length includes the bonus token`. BLIS `EffectiveTokensPerStep = 1 + α·K` is exactly this expectation. ✅

**Modeling adequacy.** The cost/throughput **decoupling** is the right call and directly answers the issue's warning ("so acceptance isn't modeled as free"): cost is driven by `w` (sublinear — weight-load/comm/const terms are *not* scaled), progress by `g`. I confirmed both backends scale only the decode token-population terms:
- roofline (`roofline.go:325-336`): decode FLOPs, dynamic KV bytes, and the MoE `nEff` batch term all use `w`; weight bytes stay `/tp`.
- trained-physics (`trained_physics_model.go:245-247`): `totalDecodeTokens += w`, feeding compute-FLOPs and the KV **write** term; per-step weight/comm/const terms untouched.

Empirically (qwen3-14b, seed 42, 100 reqs): K=5/α=0.8 cuts `e2e_mean_ms` 6985 → 1397 (~5×, matching g=5) with `total_output_tokens` **identical** (54117), and `itl_mean_ms` staying ~13ms (the documented per-verification-step artifact).

---

### Cross-path parity (run / replay / observe / calibrate)

- **run: COVERED.** `resolveSpeculativeConfig(cmd)` wired at `cmd/root.go:2070`; flags in `registerSimConfigFlags`.
- **replay: COVERED.** `registerSimConfigFlags(replayCmd)` (`replay.go:743`) + `resolveSpeculativeConfig(cmd)` (`replay.go:480`). Model-level, not persisted in the trace — same mechanism as `--tp`/`--max-model-len`. `TestINV13_RunReplayParity_SpecDecode` captures the **run** leg's per-request TTFT/E2E (not a second replay) and asserts byte-identity. This is a *true* INV-13 test — verified it's non-vacuous (guards `len==0`).
- **observe: N/A by design — CORRECT.** `observe` dispatches to a real server (no DES, no latency model); the server does its own spec-decode internally and the client cannot dictate acceptance rate. `observe_cmd.go` does not call `registerSimConfigFlags`, so the flags are correctly absent. Confirmed no silent misbehavior.
- **calibrate: N/A by design — CORRECT.** `calibrate.go` consumes a pre-computed `SimResult` JSON; it never builds a Simulator. Its sim-side ITL is **token-derived** — `(E2E - TTFT)/(OutputTokens-1)` (`calibrate.go:330`) — so it is inherently spec-decode-safe (reads tokens, not steps). No gap.

---

### Preemption safety

- Carry reset is co-located with progress reset: `preemptedRequest.specDecodeCarry = 0` next to `ProgressIndex = 0`, `ITL = nil`, `TTFTSet = false` (`batch_formation.go:363`). Re-prefill starts a clean decode phase.
- `TotalOutputTokens` is deferred to `recordRequestCompletion` (computed from final `ProgressIndex - InputLen`), not accumulated inline per step — so a preempt-then-rerun cannot double-count.
- Multi-token KV is reserved for the full advance (`AllocateKVBlocks(req, PI, PI+g, …)`), and the completion path clamps counted tokens to `len(OutputTokens)` (`simulator.go:1042`) so overshoot never over-counts (INV-1). Heavy-overshoot case (g=5, output=2) is tested.

---

### TPOT / ITL semantics under spec-decode

The subtle part, and it's handled correctly. Under spec-decode `len(req.ITL)` counts **steps**, not tokens.
- **LengthCapped branch** switches to a token-derived denominator (`ProgressIndex - InputLen + 1`) when `specEnabled` — otherwise TPOT would inflate by ~g. ✅
- **Normal-completion branch** keeps `len(OutputTokens)-1` (already a token count) — correct, and now self-documenting via the added comment. ✅
- **SLO-ITL goodput reads `RequestITLs` (TPOT)**, not the raw per-step `AllITLs` — confirmed at `sim/cluster/metrics.go:428` and `sim/metrics.go:284`. So goodput attainment is unaffected by the per-step artifact, exactly as the PR claims. The raw `itl_p90` etc. remaining per-step is documented in CLAUDE.md and the latency guide.

---

### Non-blocking findings

1. **[low, follow-up] Cross-backend asymmetry in decode-attention scaling under `w`.** The two backends scale the decode **attention** term differently:
   - roofline routes `w` through `calculateTransformerFlops(..., newTokens=w, ...)`, so the attention score ops (`4·nHeads·newT·effectiveCtx·dHead`, `roofline.go:103`) scale with `w`, and `effectiveCtx` even picks up the `(newT-1)/2` prefill-style term.
   - trained-physics keeps decode attention on `sumCtx` (per active sequence, `trained_physics_model.go:280`) and deliberately does **not** scale it by `w` (documented: the w verified positions attend the same shared context, read once).

   Both are *defensible* physical stories (the w draft positions are contiguous and share KV), and the trained-physics choice is the more accurate one for MTP. But the divergence means the two backends' *cost curves in w* differ in shape, not just magnitude. This isn't a correctness bug — both preserve K=0 byte-identity and monotonicity — but a short note in `latency-models.md` documenting that the backends model verify-width attention cost differently would help future calibration work. No code change required.

2. **[cosmetic] `--speculative-method draft` vs vLLM's `draft_model`.** vLLM's method literal is `draft_model` (`speculative.py:71`, `:695`), not bare `draft`. BLIS accepts `draft` as an informational label. Since the label is purely provenance (it does not alter the step math) this is harmless, but if the intent is vLLM-name parity, `draft_model` would be the exact token. Worth a one-word doc clarification that these are BLIS labels, not verbatim vLLM method strings (`ngram`/`eagle`/`medusa`/`mtp` do match verbatim; `draft` is the sole abbreviation).

3. **[verified, no action] `isValidSpeculativeMethod` switch** (replacing the earlier package-level map) is R8-clean — no shared mutable state. Dangling-knob ordering is correct: `--speculative-method mtp` with K=0 errors with the dangling-knob message *before* any method-validity lookup (confirmed at CLI: `method "mtp" set but num-speculative-tokens is 0`).

---

### Checks performed

- `go build ./...` clean; `go test ./sim/... ./cmd/...` all green (incl. the 3 cmd parity tests, 5 sim decode tests, 2 latency cost-law tests — all verified non-vacuous and PASS).
- BC-1 byte-identity **end-to-end on the binary**: `run` vs `run --num-speculative-tokens 0` are md5-identical (`d067d1b5…`).
- Determinism with feature **on**: two identical `--num-speculative-tokens 4 --speculative-acceptance-rate 0.6` runs are md5-identical.
- Validation gates fire at the CLI boundary: missing α when K>0, dangling method/α at K=0, unknown method, α∉[0,1], K>1024 — all `Fatalf`/error as specified (BC-9).
- Roofline backend also honors the feature (output conserved, E2E reduced).

All three governance layers (INV-1/6/8/9/11/12/13, BC-1→9, R1/R4/R8/R23) hold. Ship it.

namasl:
@claude /blis-pr-review

Addressed the independent-review notes from @susiejojo and @jgchn in commit 8e4fa8fb — all non-blocking, documentation/clarity only, no behavior change:

- **roofline.go `w<1` floors** (jgchn): documented as defensive-only against a direct zero-valued `DecodeRequestConfig` literal; unreachable from the production caller.
- **PD length-capped TPOT boundary** (jgchn): noted in `recordRequestCompletion` that a PD decode sub-request can land `ProgressIndex` at `MaxModelLen` (one past the non-PD boundary) — a pre-existing PD asymmetry affecting only the PD length-capped denominator by one token.
- **`--speculative-method` labels** (susiejojo): clarified in config + flag help that these are BLIS labels; `mtp/eagle/medusa/ngram` match vLLM verbatim, `draft` is BLIS shorthand for vLLM's `draft_model`.
- **Cross-backend decode-attention scaling** (susiejojo): documented in `latency-models.md` that trained-physics keeps decode attention on `sumCtx` (unscaled by w) while roofline scales attention FLOPs by w with a `(w-1)/2` `effectiveCtx` term — both defensible, differing only in cost-curve shape.

Build, `go test ./sim/... ./cmd/...`, and `golangci-lint run ./...` all clean.

susiejojo:
### Follow-up: could BLIS use vLLM-style spec-decode CLI args exactly?

Separate from the merge verdict above (which stands — this is a non-blocking design note for a possible follow-up PR). Flagging because the flags this PR adds are BLIS-invented, not verbatim vLLM.

**vLLM's actual CLI surface** (`vllm/engine/arg_utils.py:1562-1570`) is `--speculative-config '<JSON>'` (alias `-sc`), plus the shorthands `--spec-tokens` / `--spec-method` / `--spec-model` that get merged into it (`arg_utils.py:1782-1795`). So the GLM-5.2 recipe is:

```
--speculative-config '{"method":"mtp","num_speculative_tokens":5}'
```

`num_speculative_tokens` is a JSON *key*; the flat flag is `--spec-tokens`. This PR flattened the key into `--num-speculative-tokens` — clear, but not copy-paste-compatible with a vLLM invocation.

**The acceptance rate has a real vLLM home too.** The original review called `--speculative-acceptance-rate` BLIS-supplied / not-in-vLLM; that was incomplete. vLLM has a near-exact equivalent built for exactly this "simulate throughput without a real draft model" case (`vllm/config/speculative.py:217-237`):

- `rejection_sample_method: "synthetic"` — "accepts draft tokens with a decaying probability calibrated to synthetic_acceptance_rate"
- `synthetic_acceptance_length` — "Target mean acceptance length ... in `[1, num_speculative_tokens + 1]`"

That is algebraically identical to BLIS's `EffectiveTokensPerStep = 1 + α·K` (vLLM's mean acceptance length includes the bonus token, `metrics.py:114`), with the invertible mapping:

```
synthetic_acceptance_length = 1 + α·K    ⟺    α = (length − 1) / K
length ∈ [1, K+1]                        ⟺    α ∈ [0, 1]
```

So BLIS's `α` could be expressed in genuine vLLM vocabulary with zero fidelity loss.

**Is exact parity possible? Yes — two shapes:**

- **Option A (true parity):** accept `--speculative-config` JSON; honor the subset BLIS actually models (`method`, `num_speculative_tokens`, `rejection_sample_method:"synthetic"`, `synthetic_acceptance_length`) and **hard-error, naming the field**, on any key BLIS can't act on. Enables paste-a-vLLM-synthetic-config parity.
- **Option B (lighter):** keep flat flags but adopt vLLM names — `--spec-tokens`, `--spec-method`, and replace `--speculative-acceptance-rate α` with `--synthetic-acceptance-length` (float in `[1, K+1]`).

**Why it's a follow-up, not a change to this PR:** BLIS models throughput only, but vLLM's `SpeculativeConfig` has ~30 fields (`draft_model`, eagle head paths, ngram `prompt_lookup_max/min`, `draft_tensor_parallel_size`, …) BLIS cannot honor. Full `--speculative-config` parity forces a per-key decision: silently ignoring unmodeled keys violates R1 and creates a footgun (paste a real recipe, get numbers that quietly drop half the config); rejecting them means it isn't truly copy-paste anyway. Also, bare `--speculative-config` is meaningful in vLLM (it *measures* acceptance from a live draft model), but BLIS has no draft model, so it must require the synthetic length — an irreducible divergence.

**Recommendation:** Option A **restricted to the synthetic-simulation subset** is the sweet spot — real "paste your vLLM synthetic-benchmark config" parity while staying honest about scope (R1-clean rejection of unmodeled keys). Nothing in the current flag design blocks this; the underlying `SpeculativeConfig` maps cleanly onto the vLLM subset. Happy to file it as a cross-repo-parity enhancement issue if useful.

namasl:
Thanks @susiejojo — that's a great catch on the `synthetic_acceptance_length` equivalence, and I agree it belongs in a follow-up rather than this PR (full `--speculative-config` parity would be scope creep here, and the R1 footgun you flagged around silently-dropped keys is exactly why).

Filed as **#1542** (cross-repo enhancement). I verified all the vLLM references against real source at pinned commit `fac808b` before filing — the `synthetic` rejection-sample method + `synthetic_acceptance_length ∈ [1, K+1]` ([speculative.py#L218-L283](https://github.com/vllm-project/vllm/blob/fac808b36f502d0d992509a187dca94c68b0360a/vllm/config/speculative.py#L218-L283)) maps to BLIS's `1 + α·K` exactly via `α = (length−1)/K`, and the bonus-token convention checks out ([metrics.py#L113-L114](https://github.com/vllm-project/vllm/blob/fac808b36f502d0d992509a187dca94c68b0360a/vllm/v1/spec_decode/metrics.py#L113-L114)). The issue captures both candidate shapes (Option A restricted-subset `--speculative-config`, Option B vLLM-named flat flags) and your recommendation.

This PR stays as-is — the merge verdict stands and #1542 tracks the CLI-parity ergonomics separately.

sim/latency/roofline.go
namasl: Agreed — unreachable from the sole production caller. I kept the guards (cheap belt-and-suspenders against a direct zero-valued `DecodeRequestConfig` literal, since the struct is exported and used in tests) but replaced the terse `// defensive` comment with an explicit note of exactly what you observed. Done in 8e4fa8fb — see roofline.go:333-338 (decode loop) and the matching note at :362-363 (weight-load loop). Resolving as documented.

sim/simulator.go
namasl: Good catch on the PD asymmetry — and you're right it's pre-existing (the PD path emitted its final token unconditionally before this PR; the floor-at-1 preserves that exactly rather than stranding the sub-request in WaitQ). Since it's not introduced here and only shifts the length-capped PD TPOT denominator by one token, I documented it in place rather than changing behavior. Done in 8e4fa8fb — see the `Caveat (PD)` note at simulator.go:759-765. Resolving as documented.

susiejojo:
### Note: spec-decode scheduling/batching physics not modeled (optimistic under KV saturation)

Non-blocking, but worth stating as an explicit fidelity boundary. This PR models the *expected throughput* of spec-decode; it does not model two pieces of vLLM's spec-decode batching physics:

1. **Per-step acceptance variance** — deferred by design (§9). BLIS uses the deterministic mean `g=1+α·K` (no RNG, to hold INV-6), so it predicts the expectation of every metric exactly but yields **zero ITL variance**. Tail metrics (p99 ITL, SLO-attainment tails) under spec-decode aren't meaningful.

2. **Speculative KV / token-budget pressure** — *not* on the §9 list, and the more impactful omission. vLLM reserves `num_lookahead_tokens = K` KV slots **per running request per step**, passed unconditionally to `allocate_slots` regardless of acceptance (`vllm/v1/core/sched/scheduler.py:581`, set at `:258`), and a spec step consumes ~`K+1` of `max_num_batched_tokens`. BLIS reserves only `g` slots and debits only `g` from its token budget. So under KV saturation / high occupancy, real vLLM fits **fewer concurrent requests** and **preempts more**; BLIS sees the verify-width *latency* cost but not the batch-*occupancy* contention cost. Net effect: BLIS **over-predicts max batch size and under-predicts preemption** in memory-bound regimes — it shows the MTP speedup without the contention penalty that erodes it (the same effect behind the GLM-5.2 recipe's "pure-throughput benchmarks under-report MTP speed" warning, in reverse).

**Bottom line:** the model is faithful for mean throughput / capacity planning below KV saturation (where MTP is nearly free, as the PR's own test comment notes) — which is the stated use case, so this doesn't block merge. But the saturated-serving regime is silently optimistic. Suggest a one-line "known limitation" note in the docs, and a follow-up to reserve KV / debit token budget by `w=K+1` (speculative footprint) while still advancing progress by `g`, aligning occupancy accounting with vLLM's lookahead reservation. Happy to file that issue with these proofs.

namasl:

namasl:
@susiejojo thanks — you're right that the KV/token-budget *occupancy* reservation (vLLM's unconditional `num_lookahead_tokens = K` per running request per step, independent of acceptance) is the more impactful of the two omissions, and it's not on the §9 list. I verified it against real vLLM source before writing it up: `num_lookahead_tokens` is set on the scheduler ([scheduler.py#L257-L258](https://github.com/vllm-project/vllm/blob/903a02192fca19e4c89705af7017c0f24971ea4f/vllm/v1/core/sched/scheduler.py#L257-L258)) and passed **unconditionally** into `allocate_slots` ([#L653-L658](https://github.com/vllm-project/vllm/blob/903a02192fca19e4c89705af7017c0f24971ea4f/vllm/v1/core/sched/scheduler.py#L653-L658)) — so BLIS reserving/debiting by the accepted count `g` rather than the footprint `K+1` does over-predict max batch size and under-predict preemption under saturation, exactly as you describe.

Done in this round (commit `2f65f2c`):
- Added the **known-limitation note** to CLAUDE.md and `docs/guide/latency-models.md` (faithful below saturation — the stated capacity-planning use case; optimistic above it).
- Filed **#1627** to track the refinement (reserve KV / debit `max_num_batched_tokens` by the `K+1` footprint while still advancing progress by `g`), with the vLLM proofs above.

Kept it out of this PR since it's a distinct occupancy-accounting change (and, as you noted, non-blocking for the mean-throughput use case). The per-step acceptance *variance* stays deferred (§9, deterministic-mean for INV-6). Also brought the branch up to date with main (the earlier conflict is resolved).

susiejojo (approved):

Issue #1540, opened 2026-08-10
Add /archon-pr-review-claude: architecture-aware PR review powered by Archon + Claude


mtoslalibu:
> **Prerequisite:** #1541 (`/archon-pr-review` — plain Archon CI) must land first.

**What problem does this solve?**

Today, PR reviews in BLIS rely on reviewers manually spotting architectural changes buried in large diffs — new package dependencies, widened surfaces, missing contract tests, or unexpected blast radius. This is error-prone because a 500-line diff might contain zero architectural changes (pure internal refactor) or hide a single critical dependency addition among dozens of file moves. Reviewers waste time on internal-only PRs and miss boundary-crossing changes in noisy ones. With #1541 we get raw Archon output, but interpreting architecture deltas, blast radius numbers, and contract gaps still requires expertise. This issue adds a Claude layer on top: Claude reads the structured Archon output and produces a plain-English architectural review with risk assessment and a clear verdict.

**Proposed solution**

A new Claude Code skill (`/archon-pr-review-claude`) that runs Archon commands in CI and feeds their structured output to Claude for an architecture-aware PR review. Single trigger, tiered output: if the architecture didn't change, report that immediately (fast-track signal); if it did, enrich Claude's context with full Archon analysis and produce a reasoned architectural review.

### Flow

```
@claude /archon-pr-review-claude
         │
         ▼
Step 1:  Build or fetch archon-go binary
         │
         ▼
Step 2:  archon-go delta --json <merge-base> <pr-head>
         │
         ├── EMPTY ──→ Post comment:
         │              "✓ No architectural change. Internal-only PR — fast-track eligible."
         │              DONE.
         │
         └── NON-EMPTY ──→ continue
                            │
                            ▼
Step 3:  Gather enriched context:
           • archon-go delta (human-readable summary)
           • archon-go impact <pkg> (for each changed package)
           • archon-go evidence (contract test coverage + pass/fail)
                            │
                            ▼
Step 4:  Feed all outputs into Claude with the prompt template
                            │
                            ▼
Step 5:  Post Claude's architectural review as a PR comment
```

### What lives where

| File | Purpose |
|------|---------|
| `.claude/skills/archon-pr-review-claude/SKILL.md` | The skill definition (prompt + instructions) |
| `scripts/archon-build.sh` (or CI cache) | Builds/fetches the `archon-go` binary (shared with #1541) |
| No new workflow file | Reuses existing `claude.yml` — any `@claude` comment already triggers it |

[...]

PR #1544, opened 2026-08-10
feat(ci): add /archon-pr-review architectural analysis for PRs

### LINKED ISSUE #1541: Add /archon-pr-review CI: plain architectural delta reporting for PRs

Reviewers currently have no automated signal about whether a PR changes the architecture or is purely internal. A 500-line diff that just moves files looks the same as one that introduces a new cross-package dependency. We want a fast, deterministic tool that answers "did the architecture change?" and if yes, reports what changed, what's affected, and whether contracts are covered — all without any LLM involved. This is the foundation for the full Archon + Claude review (#1540), but is independently useful as a lightweight CI check that runs in seconds.

**Proposed solution**

A GitHub Action that triggers on `/archon-pr-review` (PR comment) and posts raw Archon output as a PR comment. No Claude, no LLM — just structured architecture analysis.

### Flow

```
Comment: /archon-pr-review
         │
         ▼
Step 1:  Build or fetch archon-go binary
         │
         ▼
Step 2:  archon-go delta --json <merge-base> <pr-head>
         │
         ├── EMPTY ──→ Post comment:
         │              "✓ No architectural change. Internal-only — fast-track eligible."
         │              DONE.
         │
         └── NON-EMPTY ──→ continue
                            │
                            ▼
Step 3:  Run enrichment commands:
           • archon-go delta (human-readable report)
           • archon-go impact <pkg> (for each changed package)
           • archon-go evidence (contract coverage + pass/fail)
                            │
                            ▼
Step 4:  Post combined Archon output as a PR comment
```

### What lives where

| File | Purpose |
|------|---------|
| `.github/workflows/archon.yml` | New workflow triggered by `/archon-pr-review` comment |
| `scripts/archon-ci.sh` | Orchestration: runs archon commands, formats output, posts comment |
| archon-go binary | Built from source or fetched as release artifact |

### Trigger

```
/archon-pr-review
```

This is a standalone GitHub Action (not Claude Code) — a simple workflow that runs shell commands and posts via `gh pr comment`.

### Example output (non-empty delta)

> **Archon Architectural Delta**
>
> ```
> + arrow scheduler -> metrics [import] (2 witnesses)
> + surface metrics.Histogram
> ```
>
> **Blast Radius**
> ```
> BLAST RADIUS of metrics
>   4 direct dependent(s), 6 total (transitive)
>   direct:   scheduler
>   direct:   api
>   direct:   router
>   indirect: autoscaler
>   indirect: gateway
>   indirect: admission
> ```
>
> **Contract Evidence**
> ```
> Contract: metrics.Collector
>   implementer HistogramCollector — NOT covered (evidence gap)
>   evidence: no contract test binds this interface
> ```

### Example output (empty delta)

> **Archon** ✓ No architectural change detected. Internal-only PR — fast-track eligible.

- [ ] Core simulator (`sim/`)
- [ ] KV cache (`sim/kv/`)
- [x] CLI (`cmd/`) — archon binary availability in CI
- [x] New package needed — archon-go binary (external tool)

**Extension friction check**
- **Files to change:** 2-3 — one workflow file, one shell script, optional Makefile/build step for archon-go
- **New interface?** No — standalone CI action
- **Invariants affected?** None — read-only analysis tool

**Alternatives considered**
1. **Skip standalone, go straight to Archon + Claude (#1540)** — rejected because we want fast iteration and the plain output is already useful for reviewers who can read architecture diffs
2. **Run on every PR automatically** — possible later, but start with manual trigger to validate usefulness first

- Prerequisite for #1540 (Archon + Claude architectural review)
- Archon tool lives in the `repoevolve` research repo — needs to be made available as a CI artifact
- Independent of existing `/blis-pr-review` Claude skill

mtoslalibu:
## Summary

Adds a GitHub Action (`/archon-pr-review`) that runs [Archon](https://github.ibm.com/ai-native-systems/repoevolve/releases/tag/v0.1.0) architectural analysis on pull requests and posts the results as a PR comment.

- **Empty delta** → "No architectural change. Fast-track eligible." (plus any invariant changes)
- **Non-empty delta** → Full report: architectural delta, blast radius per changed package, contract test evidence

## What's included

| File | Purpose |
|------|---------|
| `.github/workflows/archon.yml` | GitHub Action: trigger on `/archon-pr-review`, permission check, build archon, run analysis, post comment |
| `scripts/archon-ci.sh` | Orchestration: runs delta → impact → evidence, formats markdown output |
| `scripts/archon-build.sh` | Builds archon-go from repoevolve v0.1.0 |
| `docs/guide/archon-pr-review.md` | User-facing documentation |
| `mkdocs.yml` | Nav entry for the docs page |

## Behavioral Contracts

- **BC-1:** Triggers on `/archon-pr-review` comment from write-access collaborators
- **BC-2:** Empty delta → fast-track comment (with invariant changes if any)
- **BC-3:** Non-empty delta → full analysis (delta + blast radius + evidence)
- **BC-4:** Errors produce a clear error comment (never silent failure)

## Test Evidence

Tested locally against real BLIS commits:

```
# Empty delta (adjacent commit, no arch change):
## Archon Architectural Review
**No architectural change detected.** Internal-only PR — fast-track eligible.
### Invariants Touched
- + latency.TestCalculateKVBlocks_SiTU_Accepted

# Non-empty delta (multi-commit range with boundary changes):
## Archon Architectural Review
### Architectural Delta
  + arrow saturation -> bytes [import] (1 witness file(s))
  + surface saturation.BuildDetector (func)
  ...
### Blast Radius
  BLAST RADIUS of sim: 7 direct dependent(s), 8 total
### Contract Evidence
  Contract: sim.BatchClassifier — implementer saturation.Bank — NOT covered (evidence gap)

# Error case (bad SHA):
## Archon Error
Failed to compute architectural delta: invalid reference: deadbeef
```

## Relationship

- Fixes #1541
- Prerequisite for #1540 (`/archon-pr-review-claude`)

### DIFF
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -80,8 +80,9 @@ jobs:
           BASE_SHA: ${{ steps.refs.outputs.base_sha }}
           HEAD_SHA: ${{ steps.refs.outputs.head_sha }}
         run: |
+          MERGE_BASE=$(git merge-base "$BASE_SHA" "$HEAD_SHA")
           EXIT_CODE=0
-          OUTPUT=$(scripts/archon-ci.sh . "$BASE_SHA" "$HEAD_SHA") || EXIT_CODE=$?
+          OUTPUT=$(scripts/archon-ci.sh . "$MERGE_BASE" "$HEAD_SHA") || EXIT_CODE=$?
           if [[ $EXIT_CODE -gt 1 ]]; then
             OUTPUT="## Archon Error
 
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -88,7 +88,14 @@ jobs:
           HEAD_SHA: ${{ steps.refs.outputs.head_sha }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
-          MERGE_BASE=$(git merge-base "$BASE_SHA" "$HEAD_SHA")
+          MERGE_BASE=$(git merge-base "$BASE_SHA" "$HEAD_SHA" 2>/dev/null) || {
+            OUTPUT="## Archon Error
+
+          Could not compute merge-base between base ($BASE_SHA) and head ($HEAD_SHA). The PR head may not be reachable — check that the fetch step succeeded."
+            echo "$OUTPUT" >> "$GITHUB_STEP_SUMMARY"
+            echo "$OUTPUT" > /tmp/archon-output.md
+            exit 0
+          }
           EXIT_CODE=0
           OUTPUT=$(scripts/archon-ci.sh . "$MERGE_BASE" "$HEAD_SHA") || EXIT_CODE=$?
           if [[ $EXIT_CODE -gt 1 ]]; then
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -12,7 +12,8 @@ jobs:
   check-permissions:
     if: |
       github.event.issue.pull_request &&
-      contains(github.event.comment.body, '/archon-pr-review')
+      contains(github.event.comment.body, '/archon-pr-review') &&
+      !contains(github.event.comment.body, '/archon-pr-review-claude')
     runs-on: ubuntu-latest
     outputs:
       allowed: ${{ steps.check.outputs.allowed }}
@@ -51,7 +52,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version: '1.24'
 
       - name: Build archon-go
         id: build
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -74,6 +74,12 @@ jobs:
             core.setOutput('base_sha', pr.base.sha);
             core.setOutput('head_sha', pr.head.sha);
 
+      - name: Fetch PR head ref
+        env:
+          PR_NUMBER: ${{ github.event.issue.number }}
+        run: |
+          git fetch origin "pull/$PR_NUMBER/head"
+
       - name: Run Archon analysis
         id: archon
         env:
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -57,6 +57,7 @@ jobs:
         id: build
         run: |
           ARCHON_BIN=$(scripts/archon-build.sh v0.1.0)
+          [[ -n "$ARCHON_BIN" && -x "$ARCHON_BIN" ]] || { echo "::error::archon-build.sh produced no valid binary"; exit 1; }
           echo "archon_bin=$ARCHON_BIN" >> "$GITHUB_OUTPUT"
 
       - name: Resolve PR refs
@@ -76,14 +77,26 @@ jobs:
         id: archon
         env:
           ARCHON_BIN: ${{ steps.build.outputs.archon_bin }}
+          BASE_SHA: ${{ steps.refs.outputs.base_sha }}
+          HEAD_SHA: ${{ steps.refs.outputs.head_sha }}
         run: |
-          OUTPUT=$(scripts/archon-ci.sh . "${{ steps.refs.outputs.base_sha }}" "${{ steps.refs.outputs.head_sha }}" 2>&1) || true
+          EXIT_CODE=0
+          OUTPUT=$(scripts/archon-ci.sh . "$BASE_SHA" "$HEAD_SHA") || EXIT_CODE=$?
+          if [[ $EXIT_CODE -gt 1 ]]; then
+            OUTPUT="## Archon Error
+
+          Script exited with unexpected code $EXIT_CODE. Check the workflow logs."
+          fi
           echo "$OUTPUT" >> "$GITHUB_STEP_SUMMARY"
-          # Write to file for comment posting (handles multiline).
           echo "$OUTPUT" > /tmp/archon-output.md
 
       - name: Post PR comment
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.issue.number }}
         run: |
-          gh pr comment "${{ github.event.issue.number }}" --body-file /tmp/archon-output.md
+          if [[ ! -s /tmp/archon-output.md ]]; then
+            echo "::warning::Archon produced no output. Skipping PR comment."
+            exit 0
+          fi
+          gh pr comment "$PR_NUMBER" --body-file /tmp/archon-output.md
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -52,7 +52,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: '1.26'
 
       - name: Build archon-go
         id: build
@@ -86,6 +86,7 @@ jobs:
           ARCHON_BIN: ${{ steps.build.outputs.archon_bin }}
           BASE_SHA: ${{ steps.refs.outputs.base_sha }}
           HEAD_SHA: ${{ steps.refs.outputs.head_sha }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
           MERGE_BASE=$(git merge-base "$BASE_SHA" "$HEAD_SHA")
           EXIT_CODE=0
@@ -96,6 +97,13 @@ jobs:
           Script exited with unexpected code $EXIT_CODE. Check the workflow logs."
           fi
           echo "$OUTPUT" >> "$GITHUB_STEP_SUMMARY"
+          # GitHub PR comments have a 65536 char limit. Truncate if needed.
+          if [[ ${#OUTPUT} -gt 60000 ]]; then
+            FULL_LEN=${#OUTPUT}
+            OUTPUT="${OUTPUT:0:60000}
+
+          _Output truncated (${FULL_LEN} chars). See the [workflow run](${RUN_URL}) for the full report._"
+          fi
           echo "$OUTPUT" > /tmp/archon-output.md
 
       - name: Post PR comment
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -1,0 +1,89 @@
+name: Archon PR Review
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  check-permissions:
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/archon-pr-review')
+    runs-on: ubuntu-latest
+    outputs:
+      allowed: ${{ steps.check.outputs.allowed }}
+    steps:
+      - name: Check invoker permissions
+        id: check
+        uses: actions/github-script@v8
+        with:
+          script: |
+            try {
+              const { data } = await github.rest.repos.getCollaboratorPermissionLevel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                username: context.actor
+              });
+              const allowed = ['admin', 'write', 'maintain'].includes(data.permission);
+              core.setOutput('allowed', allowed ? 'true' : 'false');
+              if (!allowed) {
+                core.info(`User ${context.actor} has '${data.permission}' permission — skipping.`);
+              }
+            } catch (e) {
+              core.setOutput('allowed', 'false');
+              core.info(`User ${context.actor} is not a collaborator — skipping.`);
+            }
+
+  archon-review:
+    needs: check-permissions
+    if: needs.check-permissions.outputs.allowed == 'true'
+    runs-on: [self-hosted]
+    timeout-minutes: 10
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+
+      - name: Build archon-go
+        id: build
+        run: |
+          ARCHON_BIN=$(scripts/archon-build.sh v0.1.0)
+          echo "archon_bin=$ARCHON_BIN" >> "$GITHUB_OUTPUT"
+
+      - name: Resolve PR refs
+        id: refs
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            core.setOutput('base_sha', pr.base.sha);
+            core.setOutput('head_sha', pr.head.sha);
+
+      - name: Run Archon analysis
+        id: archon
+        env:
+          ARCHON_BIN: ${{ steps.build.outputs.archon_bin }}
+        run: |
+          OUTPUT=$(scripts/archon-ci.sh . "${{ steps.refs.outputs.base_sha }}" "${{ steps.refs.outputs.head_sha }}" 2>&1) || true
+          echo "$OUTPUT" >> "$GITHUB_STEP_SUMMARY"
+          # Write to file for comment posting (handles multiline).
+          echo "$OUTPUT" > /tmp/archon-output.md
+
+      - name: Post PR comment
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr comment "${{ github.event.issue.number }}" --body-file /tmp/archon-output.md
--- a/scripts/archon-build.sh
+++ b/scripts/archon-build.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+#
+# archon-build.sh — Clone and build archon-go from repoevolve at a pinned version.
+#
+# Usage: scripts/archon-build.sh [version]
+#
+# Outputs the path to the built binary on stdout.
+# Requires: git, go (1.22+)
+
+set -euo pipefail
+
+VERSION="${1:-v0.1.0}"
+ARCHON_REPO="git@github.ibm.com:ai-native-systems/repoevolve.git"
+BUILD_DIR="${RUNNER_TEMP:-/tmp}/archon-build"
+
+rm -rf "$BUILD_DIR"
+git clone --depth 1 --branch "$VERSION" "$ARCHON_REPO" "$BUILD_DIR" 2>/dev/null
+
+cd "$BUILD_DIR/code/archon-go"
+go build -o "$BUILD_DIR/archon-go" .
+
+echo "$BUILD_DIR/archon-go"
--- a/scripts/archon-build.sh
+++ b/scripts/archon-build.sh
@@ -5,7 +5,7 @@
 # Usage: scripts/archon-build.sh [version]
 #
 # Outputs the path to the built binary on stdout.
-# Requires: git, go (1.24+)
+# Requires: git, go (1.26+)
 
 set -euo pipefail
 
--- a/scripts/archon-build.sh
+++ b/scripts/archon-build.sh
@@ -10,7 +10,7 @@
 set -euo pipefail
 
 VERSION="${1:-v0.1.0}"
-ARCHON_REPO="git@github.ibm.com:ai-native-systems/repoevolve.git"
+ARCHON_REPO="https://github.com/AI-native-Systems-Research/archon.git"
 BUILD_DIR="${RUNNER_TEMP:-/tmp}/archon-build"
 
 rm -rf "$BUILD_DIR"
--- a/scripts/archon-build.sh
+++ b/scripts/archon-build.sh
@@ -5,7 +5,7 @@
 # Usage: scripts/archon-build.sh [version]
 #
 # Outputs the path to the built binary on stdout.
-# Requires: git, go (1.22+)
+# Requires: git, go (1.24+)
 
 set -euo pipefail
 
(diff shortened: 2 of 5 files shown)


mtoslalibu:
## Self-Review: Code Review Findings & Fixes

Ran code-reviewer and silent-failure-hunter agents. Key findings addressed in `9281a8a`:

### Fixed (Critical/High)

| Issue | Fix |
|-------|-----|
| `git clone 2>/dev/null` hid auth/network errors | Removed — stderr now visible in logs |
| python3 not validated | Added `command -v python3` check upfront |
| No `go.mod` validation — empty module path corrupts blast radius | Added existence check + `awk` parse with validation |
| `${{ }}` interpolation in `run:` blocks (injection risk) | Moved to `env:` vars |
| `|| true` swallows crash exit codes (137, 139) | Now distinguishes exit 0/1 (expected) from >1 (crash) |
| Silent per-package impact failures | Now prints failure message per package |
| Empty output posted as empty PR comment | Added `-s` check, skips with warning |
| Build output not validated | Added `[[ -n && -x ]]` guard |

### Not Fixed (Acceptable)

| Issue | Reason |
|-------|--------|
| `actions/checkout@v5` | v5 exists and is used by our existing `claude.yml` |
| Catch-all in permission check | Same pattern as `claude.yml`, acceptable for this use case |

### Verification

All three cases re-tested locally after fixes:
- ✓ Empty delta → fast-track + invariants
- ✓ Non-empty delta → delta + blast radius + evidence  
- ✓ Error case → clear error message, exit 1

scripts/archon-build.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# Usage: scripts/archon-build.sh [version]
+#
+# Outputs the path to the built binary on stdout.
+# Requires: git, go (1.22+)
+
+set -euo pipefail
+
+VERSION="${1:-v0.1.0}"
+ARCHON_REPO="git@github.ibm.com:ai-native-systems/repoevolve.git"
```
jgchn: Cloning from github.ibm.com might fail and will probably require a github token. Is the repo public?

.github/workflows/archon.yml
```diff
@@ -0,0 +1,89 @@
...
+            });
+            core.setOutput('base_sha', pr.base.sha);
+            core.setOutput('head_sha', pr.head.sha);
+
+      - name: Run Archon analysis
+        id: archon
+        env:
+          ARCHON_BIN: ${{ steps.build.outputs.archon_bin }}
+        run: |
+          OUTPUT=$(scripts/archon-ci.sh . "${{ steps.refs.outputs.base_sha }}" "${{ steps.refs.outputs.head_sha }}" 2>&1) || true
```
jgchn: This OUTPUT contains stderr and stdout. Any git warnings, go build noise, or stray logs from `archon-go` will appear in the comments, which might be too verbose. Consider fetching just the stdout: `OUTPUT=$(scripts/archon-ci.sh . "$base" "$head") || true`.

If archon-ci.sh fails (exit 1), stdout will contain the formatted error markdown. Stderr goes to the GHA log for debugging.

scripts/archon-ci.sh
```diff
@@ -0,0 +1,126 @@
...
+
+ARCHON="${ARCHON_BIN:-archon-go}"
+
+if ! command -v "$ARCHON" &>/dev/null; then
+  echo "Error: archon-go binary not found. Set ARCHON_BIN or add to PATH." >&2
+  exit 1
+fi
+
+# Step 1: Run delta (JSON) to determine if architecture changed.
+DELTA_JSON=$("$ARCHON" delta --json "$REPO_DIR" "$BASE_SHA" "$HEAD_SHA" 2>&1) || {
```
jgchn:
```
DELTA_JSON=$("$ARCHON" delta --json ... 2>&1) || { ... error markdown ... }
EMPTY=$(echo "$DELTA_JSON" | python3 -c "import json,sys; d=json.load(sys.stdin); ...")
```

The `|| { ... }` catches a non-zero exit from archon-go, but if archon-go exits 0 with malformed/empty JSON (e.g., prints a warning to stdout before the JSON), the python3 command crashes with a json.JSONDecodeError. Due to `set -e`, the script exits immediately with no formatted error output, the workflow then posts an empty/partial comment.

Fix: wrap the python3 parse in its own error handler:
```
  EMPTY=$(echo "$DELTA_JSON" | python3 -c "..." 2>/dev/null) || {
    echo "## Archon Error"
    echo ""
    echo "Failed to parse delta JSON output."
    echo '```'
    echo "$DELTA_JSON" | head -20
    echo '```'
    exit 1
  }
```

.github/workflows/archon.yml
```diff
@@ -0,0 +1,102 @@
...
+        id: refs
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            core.setOutput('base_sha', pr.base.sha);
```
jgchn, line 74: `pr.base.sha` is the current tip of main, not the merge-base. On a stale branch this will include architectural changes from other merged PRs as false positives. Consider computing merge-base after checkout: `git merge-base $base $head`.

mtoslalibu:
## Addressing @jgchn's Review Comments

**Comment 2 (stderr in output):** Fixed in `9281a8a` — removed `2>&1` from the workflow step. The script only outputs formatted markdown on stdout; stderr goes to the GHA log for debugging.

**Comment 3 (python3 JSON parse crash):** Fixed in `9281a8a` — the python3 parse is now wrapped with try/except and a `|| { error markdown; exit 1 }` fallback. If archon-go outputs malformed JSON, the user gets a clear error comment instead of an empty one.

**Comment 4 (pr.base.sha is not merge-base):** Good catch — fixed in `c19d675`. Now computes `git merge-base $BASE_SHA $HEAD_SHA` after checkout and passes that to `archon-ci.sh` instead of the raw `pr.base.sha`.

**Comment 1 (github.ibm.com auth):** Discussing with team — the repoevolve repo is currently internal. Options: make it public, add a GHE token secret, or host the binary on github.com. Will update once decided.

mtoslalibu:
## Update: Comment 1 Resolved (github.ibm.com auth)

Moved archon to a **public github.com repo**: https://github.com/AI-native-Systems-Research/archon

`archon-build.sh` now clones via HTTPS from github.com — no SSH keys, no tokens, no GHE auth needed. Pushed in `f71e376`.

## Final Validation (post all fixes)

Built archon-go fresh from the new public repo (`scripts/archon-build.sh v0.1.0`) and ran `archon-ci.sh` end-to-end:

| Test case | Result |
|-----------|--------|
| This PR (scripts/docs only, no sim/ changes) | ✓ "No architectural change detected. Fast-track eligible." exit 0 |
| Real arch change (4c54cf5..18c2c92) | ✓ Full report: delta + blast radius (4 packages) + contract evidence. exit 0 |
| Bad SHA (deadbeef) | ✓ "Archon Error: invalid reference" exit 1 |

All 4 review comments from @jgchn addressed. PR ready to merge.

kalantar:
FWIW, I did a pr review and now that it can see archon it suggests these changes:

- Wrong JSON keys — added/removed packages never reach blast radius (scripts/archon-ci.sh:131-136)

The script reads d.get('boxesAdded', []) / boxesRemoved. Archon emits neither. delta/delta.go:49-50 declares:

PackagesAdded   []PackageRef `json:"packagesAdded,omitempty"`
PackagesRemoved []PackageRef `json:"packagesRemoved,omitempty"`

So both loops are dead code — they always iterate an empty list. And PackageRef is an object ({path, internal}), not a string, so the intended b.startswith(mod) would raise AttributeError once the key is corrected. A PR that adds a package appearing only in packagesAdded gets an empty Blast Radius section with no indication anything was skipped. Note the docs page correctly describes + box pkg — that's the human report wording (delta.go:654); the JSON uses packages*.

Fix also needs a guard: impact ... "$HEAD_SHA" on a package from packagesRemoved cannot resolve at head, so it will always print (impact analysis failed for …).

- 2>&1 on the JSON capture makes a stderr warning corrupt the parse (scripts/archon-ci.sh:49)

DELTA_JSON=$("$ARCHON" delta --json "$REPO_DIR" "$BASE_SHA" "$HEAD_SHA" 2>&1)

Archon writes to stderr while exiting 0 (main.go:421-423):

warning: N package error(s) during extraction of <dir>@<sha> (graph may be partial)

Any partial extraction at either commit — common on a repo this size, and guaranteed if the base commit has a broken package — prepends that line to the captured text, json.load fails, and the whole run degrades to ## Archon Error / Failed to parse architectural delta JSON. Capture stdout only and route stderr to the log.

- Go version mismatch. setup-go: '1.22' (archon.yml:54) vs archon's go.mod: go 1.26.3. This only works via automatic toolchain download; a runner with GOTOOLCHAIN=local or restricted egress fails the build. archon-build.sh:7's "Requires: git, go (1.22+)" is wrong. Also relevant to evidence, which compiles this repo (go 1.24.0).

- Trigger substring matches too much (archon.yml:15). contains(body, '/archon-pr-review') also fires on /archon-pr-review-claude — the explicit follow-up in #1540 — and on any comment quoting the new docs page (which contains the literal string). Anchor the match.

mtoslalibu:
## Fixed: 4 Bugs from @jgchn's Second Review

All addressed in `75b7a3a`:

| # | Bug | Fix |
|---|-----|-----|
| 1 | `2>&1` on delta JSON capture — stderr warnings corrupt JSON parse | Capture stdout only; stderr goes to GHA log |
| 2 | Wrong JSON keys (`boxesAdded`/`boxesRemoved`) — dead code, never picks up added/removed packages | Changed to `packagesAdded`/`packagesRemoved`, access `.path` field; skip removed packages in impact (they don't exist at HEAD) |
| 3 | Go 1.22 in workflow vs archon's go.mod 1.26.3 requirement | Bumped to `go-version: '1.24'` (toolchain download handles archon; 1.24 also covers BLIS's evidence compilation) |
| 4 | `/archon-pr-review` trigger also matches `/archon-pr-review-claude` | Added `!contains(body, '/archon-pr-review-claude')` exclusion |

Re-tested all 3 cases after fix — all pass.

kalantar:
more comments:

1. "Fast-track eligible" hides breaking data-contract changes. isEmpty() consults only packages, edges, and surface — SchemaChanges, Contracts, and ContractViolations are structurally excluded. I built a fixture that drops a serialized field with all exports unchanged. archon-ci.sh posts:

▎ No architectural change detected. Internal-only PR — fast-track eligible.

Archon's own renderer, same commit pair, says:

SCHEMA CHANGED — a serialized data contract (wire/DB payload) changed shape
  - field api.Request.Limit int

The script discards what the tool explicitly surfaces.

2. Fork PRs fail outright — and they're the majority case. actions/checkout fetches +refs/heads/* only; I confirmed no refs/pull/* in a default clone. A fork head isn't in the base repo's heads, so git merge-base exits 128 under bash -e, the job dies, and no comment posts at all. 52 of the last 60 PRs are cross-repo — namasl, atantawi, and susiejojo all work from personal forks. So this would fail on ~87% of this repo's PRs.

.github/workflows/archon.yml
```diff
@@ -0,0 +1,104 @@
...
+  archon-review:
+    needs: check-permissions
+    if: needs.check-permissions.outputs.allowed == 'true'
+    runs-on: [self-hosted]
+    timeout-minutes: 10
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
jgchn, line 51: `fetch-depth: 0` only fetches `refs/heads/*` and `refs/tags/*`. For fork PRs, the head SHA lives under `refs/pull/N/head` which isn't fetched. This line will fail with exit 128 on ~87% of this repo's PRs. Add `git fetch origin "pull/$PR_NUMBER/head"` before this.

mtoslalibu:
## Fixed: 2 Bugs from Latest Review

Both addressed in `491871e`:

| # | Bug | Fix |
|---|-----|-----|
| 1 | "Fast-track eligible" hides schema/contract changes — `emptyAtPackageAltitude` only checks edges/surface/packages, not schema or contracts | Now also checks `schema`, `contracts`, and `contractViolations` — if any are non-empty, falls through to full analysis |
| 2 | Fork PRs fail — `actions/checkout` only fetches `refs/heads/*`, fork heads live under `refs/pull/N/head` | Added explicit `git fetch origin "pull/$PR_NUMBER/head"` step before merge-base computation |

### Validation

Tested on real fork PR #1538 (namasl's speculative decode):
```
git fetch origin pull/1538/head   → OK
git merge-base main <head>        → 1ab4aab
archon-ci.sh . 1ab4aab <head>    → correctly detects SpeculativeConfig surface additions
```

Also tested:
- Empty delta (no schema): fast-track ✓
- Non-empty delta (has schema): full report ✓
- Error case: clean error message ✓
- lora PR1 (new package): `+ box sim/lora` detected ✓
- Single fix commit: edge/surface changes detected ✓

kalantar:
> Bumped to go-version: '1.24' (toolchain download handles archon; 1.24 also covers BLIS's evidence compilation)

If archon requires 1.26; won't this need to be bumped that high?

kalantar:
@mtoslalibu Claude suggests that most of the (git) comment content that is created is boilerplate and that it is large (30K). There is a 64K limit on comment length. It expresses a concern that if a change touches very many more packages it will exceed this limit. I don't know if this is a real concern.

mtoslalibu:
## Addressing @kalantar's Concerns

**Go version (1.24 vs 1.26):** Bumped to `go-version: '1.26'` in `4a10768`. This directly satisfies archon's `go 1.26.3` requirement without relying on automatic toolchain download. Go 1.26 compiles BLIS (go 1.24) with no issues — Go is backward compatible. This only affects the CI runner for this workflow; BLIS's `go.mod` is untouched.

**Comment size limit (64K):** Added truncation at 60K chars in `4a10768`. If output exceeds that, the PR comment is truncated with a link to the full workflow run (which has `GITHUB_STEP_SUMMARY` with no size limit). On typical BLIS PRs the output is 1-5K (empty delta) or 30-45K (full analysis with evidence) — well within the limit. The truncation is a safety net for unusually large diffs.

kalantar:
One more on the schema/contract fix in `491871e` — the verdict flips correctly now, but the blast radius doesn't follow.

The script decides two things independently, from two different sets of JSON fields.

**Is this fast-track eligible?** (`scripts/archon-ci.sh:62-70`) — five axes:

```python
empty = d.get('emptyAtPackageAltitude', False)   # edges + surface + internal packages
has_schema = bool(d.get('schema'))
has_contracts = bool(d.get('contracts')) or bool(d.get('contractViolations'))
print('true' if (empty and not has_schema and not has_contracts) else 'false')
```

**Which packages get a blast radius?** (`scripts/archon-ci.sh:127-150`) — three:

```python
for e in d.get('edgesAdded', []) + d.get('edgesRemoved', []):   # :133
for s in d.get('surface', []):                                  # :137
for p in d.get('packagesAdded', []):                            # :140
```

`schema` and `contracts` are in the first list and not the second. So a delta whose only content is on those two axes flips the verdict to "architecture changed" and then contributes zero packages. `### Blast Radius` is gated on `[[ -n "$CHANGED_PKGS" ]]` at `:154`, so the section is dropped entirely.

I fed both shapes through that exact snippet to check:

```
schema-only delta   → CHANGED_PKGS lines: 0
contract-only delta → CHANGED_PKGS lines: 0
```

Net effect: the case this fix was filed for — a serialized field changing shape — now produces a report that says the architecture changed and then declines to say how far the change reaches. The contract-only case matters too, since a new implementer where the import edge already existed is the thing archon's own comment at `delta/delta.go:83-88` calls out as what the edge-level diff alone would miss.

The fields you need are already in the JSON: `schema[].package` is a package path (`delta/delta.go:21-25`), and `contracts[].interface` is `pkgpath.Interface`, so the package is everything up to the last dot (`delta/delta.go:90`). Roughly:

```python
for s in d.get('schema', []):
    if s['package'].startswith(mod):
        pkgs.add(s['package'])
for c in d.get('contracts', []):
    iface = c.get('interface', '')
    pkg = iface.rsplit('.', 1)[0]
    if pkg.startswith(mod):
        pkgs.add(pkg)
```

Worth noting this is new rather than something missed in the earlier rounds. I diffed the extractor across your fix commits:

```
CHANGED_PKGS extractor: 75b7a3a vs 4a10768 (current head)
IDENTICAL — extractor untouched by 491871e
```

Before `491871e`, a schema-only or contract-only delta was reported as fast-track eligible and the script exited at `:112`, so the blast-radius code was unreachable on those inputs and there was nothing to get wrong. `491871e` opened that path for the first time without extending the extractor to match.

One aside while I was in there: `d.get('contractViolations')` is always absent, so that half of `has_contracts` is inert. `ContractViolations` is only populated by `CheckContract`, which `main.go:195-197` calls only when `--allow` is passed, and the script never passes it. Harmless as written, just not doing anything — if you want that axis live it needs an allow-list file.

jgchn (dismissed):
I think this looks good. We should give it a try!

mtoslalibu:
## Fixed: Blast radius extractor missing schema/contract packages

Addressed in `1cc090f`. The extractor now also reads `schema[].package` and `contracts[].interface` (rsplit to package path) when building the blast radius list. A schema-only or contract-only delta now correctly shows affected packages in the blast radius section.

Re the `contractViolations` note: agreed it's inert (only populated with `--allow`). Leaving it as a harmless forward-compatible check.

mtoslalibu:
## Fixed: Blast radius gap for schema/contract-only changes + major simplification

Addressed in `1cc090f` (extractor fix) and `f0febe6` (refactor).

**Root cause of the recurring triage bugs:** The script was reimplementing archon's own classification logic in python/bash, and every new axis (schema, contracts) required updating both archon AND the script. This divergence produced bugs.

**Fix:** The triage now uses `grep` on archon's own human-readable output markers:
- `"empty at package altitude"` → structurally empty
- `"SCHEMA CHANGED"` → schema changes present  
- `"CONTRACT COVERAGE"` → contract changes present

If structurally empty AND no schema/contract markers → fast-track. Otherwise → full analysis. This can never diverge from archon because it reads archon's own verdict.

**Result:** Script went from 167 lines to 122 lines. Python is no longer needed for the fast-track path (only used for package extraction in the non-empty path). The class of "triage disagrees with archon" bugs is eliminated.

Tested all cases:
- Empty delta (invariants only) → fast-track with invariants shown ✓
- Non-empty (schema + contracts + edges) → full report with all blast radius packages ✓
- Fork PR #1538 → correctly detected ✓
- Error case → clean error ✓

mtoslalibu:
## Summary of Redesign + Final Validation

### What changed (and why)

The original script reimplemented archon's triage logic in python/bash — parsing JSON, checking multiple fields (`emptyAtPackageAltitude`, `schema`, `contracts`, `contractViolations`), and making its own fast-track decision. This caused a chain of bugs where the script's logic diverged from archon's own assessment.

**The fix:** Stop reimplementing archon. Instead, use archon's own human-readable output for triage:
- `archon-go delta` already prints `"empty at package altitude"` or `"boundary changed"` or `"SCHEMA CHANGED"` or `"CONTRACT COVERAGE"`
- The script just greps these markers — it can never disagree with archon because it reads archon's own verdict

Python is now only used in the non-empty path to extract package names for blast radius (the one thing archon doesn't do in a single command).

**Result:** Script went from 167 lines of fragile JSON-parsing logic to 139 lines that mostly just call archon and format the output.

### Final validation against 9 real PRs

Tested with the archon-go binary built from the public repo (https://github.com/AI-native-Systems-Research/archon v0.1.0):

| PR | Author | Type | Verdict | Size |
|----|--------|------|---------|------|
| #1533 | inference-sim | Small fix (SiTU allowlist) | Fast-track | 311 chars |
| #1534 | namasl (fork) | Feature (GLM model-shape) | Full report | 37K |
| #1525 | susiejojo (fork) | Docs (blog post) | Fast-track | 109 chars |
| #1523 | namasl (fork) | Cluster fix (KV capacity) | Full report | 36K |
| #1514 | namasl (fork) | Fix (PD parent E2E) | Fast-track + invariants | 1K |
| #1538 | namasl (fork) | Feature (speculative decode) | Full report | 36K |
| #1537 | namasl (fork) | Feature (multi-node TP) | Full report | 37K |
| #1506 | atantawi (fork) | Docs fix (plans) | Fast-track | 109 chars |
| #1503 | atantawi (fork) | Feature (lora B-6) | Full report | 49K |

- All 3 fork contributors (namasl, susiejojo, atantawi) work correctly
- Docs-only PRs → fast-track (no expensive analysis)
- Small fixes → fast-track (with invariants shown when tests changed)
- Feature PRs → full report (delta + blast radius + contract evidence)
- Largest output (49K) is under the 60K truncation threshold
- Zero failures across all 9 PRs

scripts/archon-ci.sh
```diff
@@ -0,0 +1,138 @@
...
+GO_MODULE=$(awk '/^module /{print $2; exit}' "$REPO_DIR/go.mod")
+if [[ -n "$GO_MODULE" ]] && command -v python3 &>/dev/null; then
+  DELTA_JSON=$("$ARCHON" delta --json "$REPO_DIR" "$BASE_SHA" "$HEAD_SHA") || DELTA_JSON=""
+  if [[ -z "$DELTA_JSON" ]]; then
+    echo "### Blast Radius"
+    echo ""
+    echo "_Unavailable — delta JSON extraction failed. See workflow log._"
+    echo ""
+  elif [[ -n "$DELTA_JSON" ]]; then
+    CHANGED_PKGS=$(echo "$DELTA_JSON" | python3 -c "
```
jgchn: `elif [[ -n "$DELTA_JSON" ]]` is always true after the `-z` check, could just be else. Cosmetic only.

scripts/archon-build.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# archon-build.sh — Clone and build archon-go from repoevolve at a pinned version.
+#
+# Usage: scripts/archon-build.sh [version]
+#
+# Outputs the path to the built binary on stdout.
+# Requires: git, go (1.24+)
```
jgchn: `"Requires: git, go (1.24+)"` but the workflow uses `go-version: '1.26'`. The comment should match

jgchn (approved):

PR #1546, opened 2026-08-12
feat(saturation): uniform last-window plurality reducer for the stdout final label (#1517)

### LINKED ISSUE #1517: Final saturation classification: uniform last-window plurality reducer over detector traces

susiejojo:
Closes #1517.

## What this does

Restores the end-of-run saturation label on **stdout** as a per-detector `map[string]Level`, derived **uniformly from the per-event trace** by one detector-agnostic reducer — the same last-window plurality rule for a single detector (#1516) and the whole bank (#1519), rather than each detector's own `Classify`. This makes cross-detector comparison fair and gives new detectors final labeling for free.

- A run **without** `--detectors` is byte-identical to the historical no-feature output (BC-8; `saturation` stays dropped by `omitempty`).
- A run **with** `--detectors` adds `"saturation": {"composite":"STABLE", ...}`.

## The rule (identical for every detector)

1. Trailing window: `lastT = max(Timestamp)`; keep records with `Timestamp >= lastT - windowUs`.
2. Plurality vote over `Result.Level`; most frequent wins (stays 3-level).
3. Count-tie → more severe (`OVERLOADED > BACKLOGGED > STABLE`) — order-independent (INV-6).
4. Empty group → `STABLE` (R20).

`windowUs` resolution (same for all detectors): `--saturation-final-window` (Go duration) → else `backlog_drift.window_size_sec` → else 30s.

## Changes

- **`sim/saturation/reduce.go`** (new): pure `ReduceOne`/`ReduceAll` (not `Detector` methods).
- **`cmd/saturation.go`**: `--saturation-final-window` flag; `tracer.trace()`→`run()` always replays and returns the final map (no early-return when no report path — the label must still reach stdout).
- **`cmd/root.go` / `replay.go` / `observe_cmd.go`**: splice the label onto stdout via the goodput build-then-mutate-then-emit pattern (`cmd` sets `MetricsOutput.Saturation`; `sim` stays saturation-agnostic).
- **`sim/saturation/replay.go`**: `CombinedReport` gains a `"final"` block → report is `{"final":{...},"trace":[...]}`.
- **Retire `sim.BatchClassifier`**: delete `sim/classifier.go`, drop the detector param from `BuildOutput`/`SaveResults`, rename `Bank.Classify`→`Bank.Run`, fix all call sites.
- Rewrite the BC-8 parity test to the new stdout contract; update `CLAUDE.md`.

## Invariants

- **INV-6**: order-independent reducer + sorted map keys ⇒ deterministic stdout.
- **INV-13**: identical traces ⇒ identical final maps across run/replay (verified end-to-end, byte-identical).

## Verification

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` all pass. Manually verified: plain runs omit the field; `--detectors composite`/`all` emit correct maps; `--saturation-final-window` works and rejects invalid/non-positive values; run→replay saturation maps are byte-identical.

## Additional cleanup

While auditing the retired `BatchClassifier` seam, removed a dead `workload.BacklogClassifier` field (and the unused `NewBacklogDriftDetectorWithClassifier` constructor) from the streaming `BacklogDriftDetector` — it was written but never read. This removes the last *behavioral* coupling (interface/dispatch) from `sim/saturation` onto `sim/workload`. A one-way **data** dependency remains — `sim/saturation` still imports the `workload.BacklogDriftConfig` struct (and its `New`/`Default` constructors) to size the streaming detector's window — so the two packages are **not fully decoupled**. Finishing that (relocating the config type into `sim/saturation`) is Part 2 of #1547. No behavior change. The batch analysis library (`workload.AnalyzeBacklogDrift*`) is left for that separate PR.


susiejojo:
@claude /blis-pr-review

susiejojo:
Addressed all three review items in 5e28e00b:

- **IMPORTANT — stale comment (`cmd/observe_cmd.go`)**: removed the contradictory "always nil as of #1516 … #1517 repopulates this"; now reads `#1517: per-detector final-label map when --detectors is set; nil otherwise (omitempty drops the field)`.
- **MINOR — `resolveFinalWindowUs` error paths untested**: added `TestResolveSaturation_FinalWindowErrors` (unparseable / negative / zero duration) and `TestResolveSaturation_FinalWindowResolutionOrder` (flag → `backlog_drift.window_size_sec` → 30s default), plus a `final-window without --detectors` sub-case in the existing requires-`--detectors` table.
- **MINOR — `integration_test.go` retired shape**: `TestMetricsOutput_SaturationField` now round-trips the actual `map[string]saturation.Level` shape this PR emits (asserting each detector → bare label string), instead of the retired single-detector `*saturation.Result` batch shape.

Left as-is (per the review's own note): the `if err != nil` guard around `Bank.Run` — intentional defensive scaffolding for future fallibility, lint-clean today.

`go build ./... && go test ./... && golangci-lint run ./...` all green.

susiejojo:
@claude /blis-pr-review

jgchn:
@claude /blis-pr-review any correctness bug on this PR that is preventing it from being merged?

sim/saturation/reduce.go
```diff
@@ -0,0 +1,88 @@
...
+		if r.Result.Level >= Stable && r.Result.Level <= Overloaded {
+			counts[r.Result.Level]++
+		}
+	}
+
+	// Most severe level with the maximum count wins. Scanning severity-descending
+	// and using strict `>` means the first (most severe) level holding the max is
+	// selected, so a tie resolves toward the more severe level.
+	best := Stable
+	bestCount := -1
```
jgchn: Minor: 

The `bestCount := -1` initialization means that if all records within the window have a Result.Level outside [0, 2] (filtered by the bounds check on line 47), the first iteration always sets `best = Overloaded` with count 0, returning `Overloaded` instead of the documented Stable degenerate default. This is unreachable in practice (all detectors only emit valid levels), but if you want the defensive guard (line 47-49) to be fully consistent with the "empty → STABLE" contract, bestCount should start at 0 instead of -1:

```go
best := Stable
bestCount := 0
for lvl := Overloaded; lvl >= Stable; lvl-- {
    if counts[lvl] > bestCount {
```

This would make the "all filtered out" case return Stable instead of Overloaded. But again, this path is unreachable given current detector implementations.

susiejojo: Good catch — fixed in c78e82b9. Changed `bestCount := -1` → `0`, so a window containing only out-of-range `Level` values (all dropped by the `[Stable, Overloaded]` bounds check) leaves `counts` all-zero, no level clears the `> 0` bar, and it falls through to the `Stable` default — consistent with the documented "empty group → STABLE" (R20) contract, instead of returning `Overloaded` at count 0.

Added `TestReduceOne_AllLevelsOutOfRange_DefaultsToStable` to lock it in (it fails under the old `-1`). Agreed the path is unreachable with current detectors (they only emit valid levels), but the defensive guard is now internally consistent.

jgchn (approved):

Issue #1343, opened 2026-05-12
PD disaggregation: decode pod should reserve KV blocks during transfer (WAITING_FOR_REMOTE_KVS parity)


mtoslalibu:
**What external behavior are we replicating?**

vllm v1 scheduler's KV block reservation during PD (prefill-decode) disaggregation KV transfer. In the real system, when a request is assigned to a decode pod and KV transfer begins, the decode pod **immediately reserves KV blocks** and places the request in a `WAITING_FOR_REMOTE_KVS` state. BLIS currently does not do this — the decode instance is completely blind to in-flight KV transfers, and blocks are only allocated after transfer completes. This causes BLIS to overcount decode-side drops under KV pressure.

**Reference implementation (MANDATORY)**

**Permalink 1** — Block allocation with `delay_cache_blocks` during PD scheduling (blocks reserved but not cached):
https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/v1/core/sched/scheduler.py#L754-L763

**Permalink 2** — Request enters `WAITING_FOR_REMOTE_KVS` state after block reservation:
https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/v1/core/sched/scheduler.py#L795-L814

**Permalink 3** — `delay_cache_blocks` parameter documentation ("used by P/D when allocating blocks used in a KV transfer which will complete in a future step"):
https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/v1/core/kv_cache_manager.py#L282-L284

**Permalink 4** — Blocks allocated but not cached when `delay_cache_blocks=True`:
https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/v1/core/kv_cache_manager.py#L413-L414

**Permalink 5** — Promotion from `WAITING_FOR_REMOTE_KVS` → `WAITING` → `RUNNING` when transfer completes:
https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/v1/core/sched/scheduler.py#L2070-L2085

**Permalink 6** — `_update_waiting_for_remote_kv` caches blocks after transfer finishes:
[...]

PR #1554, opened 2026-08-12
fix(cluster): exclude no-token PD parents from TTFT/E2E distributions (#1511)

### LINKED ISSUE #1511: Bug: decode-side-dropped PD parents (transfer-start & transfer-complete) enter TTFT/E2E distributions

namasl:
## Summary

- A PD-disaggregated parent that reaches a terminal `CompletionTime` but emits **no token** was treated as `completed` by `projectPDMetrics` and received a phantom prefill-only TTFT and a `CompletionTime`-based E2E — polluting the TTFT/E2E/scheduling-delay distributions and **double-representing** the request (counted as a drop **and** as a latency sample). This violated the documented **INV-PD-6**.
- The bug spans **all three** no-token terminal outcomes, not just the originally-reported transfer-start drop:
  - **drop at transfer start** (`DecodeSubReq == nil`, `#1343`),
  - **drop at transfer complete / late-drop** (`DecodeSubReq == nil`, `TransferCompleteTime > 0`),
  - **decode-timeout-while-queued** (`DecodeSubReq != nil` but empty `ITL`).
- Fix: gate the parent projection on `served := completed && parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0` (the decode sub-request emitted ≥1 token). No-token terminal parents now contribute **no** entry to any of the six per-request metric maps; they remain represented only by the drop/timeout counters. The deleted prefill TTFT is rolled out of the aggregate `TTFTSum` so the reported **mean TTFT** excludes it (BC-3). A parent that emitted a first token and **then** timed out is still served and keeps its entries.

## Issue

Closes #1511

## Behavioral contracts

- **BC-1 (no-token exclusion):** a terminal parent whose decode sub-request emitted no token appears in **none** of `RequestTTFTs`, `RequestE2Es`, `RequestCompletionTimes`, `RequestSchedulingDelays`, `Requests`, `RequestITLs`.
- **BC-2 (served preserved):** a parent that emitted ≥1 token (incl. timeout-after-first-token) keeps its TTFT/E2E entries exactly as before.
- **BC-3 (TTFTSum consistency under drops):** after projection over a mix of served and no-token parents, `TTFTSum == Σ RequestTTFTs`.
- **BC-4 (no double-representation):** a no-token parent stays counted in `DroppedUnservable`/`TimedOutRequests` and is absent from the latency distributions.
- **BC-5 (no-drop no-op):** a workload with no drops is byte-identical to before (INV-6) — `served == completed` for every parent.

## Cross-Path Parity

| `blis run` | yes | yes | fix in `sim/cluster` `projectPDMetrics`, run at finalization |
| `blis replay` | yes | yes (shared code) | same `projectPDMetrics`; deterministic → INV-13 parity by construction |
| `blis observe` | no | N/A | builds metrics from real-server latencies; never calls `projectPDMetrics` |

## Invariants Affected

- **INV-PD-6 (Metric Map Parent Granularity):** restored — the code previously violated its own documented contract for terminal-but-no-token parents. Statement/evidence updated in `invariants.md`, plus INV-PD-6b's fallback description.
- **INV-1 (conservation):** unaffected — only map entries change, never counters.
- **INV-6 (determinism):** preserved — only int64 accumulation + map sets added; no-drop output byte-identical.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (all packages pass)
- [x] `golangci-lint run ./...` (0 issues, v2.9.0)
- [x] New unit test `TestPDParentMetrics_NoTokenExcludedFromLatency` — table over the served/no-token taxonomy across all six maps.
- [x] New `TestPDParentMetrics_TTFTSumConsistentAcrossDrop` — the `TTFTSum == Σ RequestTTFTs` law with a drop present.
- [x] New integration test `TestDisaggregation_DroppedParent_NotInLatencyMaps` — real DES run with tight decode KV; asserts no-token terminal parents are absent from the latency maps while counted in `DroppedUnservable`.
- [x] Updated `TestPDParentE2E_ProjectionBranches` and `TestDisaggregation_TTFT_ProjectionBranches` no-token cases (formerly asserted the phantom entry).

---


namasl:
@claude /blis-pr-review

namasl:

namasl:

namasl:

## qa-review — PR #1554: ✅ PASS
_fix(cluster): exclude no-token PD parents from TTFT/E2E distributions (#1511) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The implementation excludes every terminal parent without a surviving decode-token ITL from TTFT, E2E, scheduling-delay, completion-time, request, and ITL maps, while preserving drop/timeout accounting. It retains served parents, including timeouts after a token. The TTFTSum adjustment is necessary to keep the aggregate consistent with the filtered map. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | No. The canonical invariant documentation was updated, but the PDMetrics field documentation still describes ParentTTFT as the prefill sub-request TTFT, whereas projection now reports decode scheduling delay plus the first decode step. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | Yes. At least one test comment remains stale: the TTFT branch overview still says nil DecodeSubReq or empty ITL takes a prefill-only fallback, although those cases now produce no entry. The PDMetrics ParentTTFT comment is also stale. |
| G1 | correctness | ✅ CONFIDENT | In `projectPDMetrics()`, when `completed` is true and `hasPrefillTTFT` is true but `served` is false, `m.TTFTSum -= int64(prefillTTFT)` rolls back the prefill TTFT. If a decode sub-request was preempted and re-stamped its `FirstTokenTime` into `TTFTSum` before timing out, does rolling back only `prefillTTFT` leave `TTFTSum` corrupted or negative? | For a preempted-then-timed-out request, the premise does not hold in the current implementation: FirstTokenTime is re-stamped during execution, but TTFTSum is only incremented by recordRequestCompletion, which a timed-out request never reaches. Therefore rolling back the completed prefill's TTFT removes the only recorded TTFTSum term and does not make the production value negative. A preempted request that later completes can leave a separate decode TTFTSum residual, but that is the pre-existing #1628 case documented as out of scope. |
| G2 | correctness | ✅ CONFIDENT | What happens if a parent request has `DecodeSubReq != nil` and `len(DecodeSubReq.ITL) > 0`, but all elements in `ITL` are zero or negative due to clock skew or step timing anomalies? Will `servedFirstToken` evaluate to true and project an invalid TTFT into `RequestTTFTs`? | The length check would classify any non-empty hand-constructed ITL as served regardless of its values. A sufficiently negative first element is prevented from entering RequestTTFTs by the newTTFT < 0 guard, which falls back to the prefill TTFT; zero or a negative element that is outweighed by decodeDelay could still be projected. Shipped simulation producers cannot generate non-positive ITLs: each decode ITL is the step advance plus output-token overhead, and the step advance is floored to at least one tick. |
| G3 | invariants | ✅ CONFIDENT | INV-PD-6 requires that incomplete or terminal-but-no-token parents contribute no entry to any of the six per-request metric maps. Does deleting `RequestCompletionTimes` for no-token terminal parents break any invariant or consumer that expects `RequestCompletionTimes` to track all requests that finished execution? | No. RequestCompletionTimes is a completed-latency metric map, not the authoritative lifecycle ledger. No-token drops and timeouts are represented by their terminal counters, while session metrics treat a missing completion-time entry as an uncompleted round. Removing both the sub-request and parent entries is consistent with the revised INV-PD-6 contract. |
| G4 | invariants | ✅ CONFIDENT | Under INV-1 (request conservation), dropped requests are tracked in `DroppedUnservable`. Does excluding no-token parents from `RequestE2Es` and `RequestTTFTs` while keeping them in `DroppedUnservable` create any inconsistency when verifying metric map key counts against total input requests? | No. INV-1 is defined over aggregate lifecycle counters, not the cardinalities of latency maps. Dropped no-token requests remain in DroppedUnservable, and timed-out no-token requests remain in TimedOutRequests; latency maps intentionally contain only served samples. |
| G5 | rules | ✅ CONFIDENT | Rule R1 forbids silent data loss. When `projectPDMetrics()` deletes `pfx` and `dec` sub-request keys from `m.RequestTTFTs` and `m.RequestE2Es` without creating a parent key (when `served` is false), is there any diagnostic log recording which parent IDs were excluded from latency distributions? | There is no per-parent exclusion log in projectPDMetrics. This does not violate R1 because the deletion is intentional projection behavior and the request is not silently lost from lifecycle accounting: decode drops increment DroppedUnservable and timeouts increment TimedOutRequests. The behavior is also explicitly documented in the projection comments. |
| G6 | rules | ✅ CONFIDENT | Does modifying `m.TTFTSum` directly inside `projectPDMetrics()` violate any encapsulation boundaries or canonical metric aggregation patterns established for non-disaggregated paths in `sim/cluster/cluster.go`? | No. projectPDMetrics owns the post-aggregation PD transformation and already adjusts TTFTSum when replacing a prefill TTFT with the user-facing parent TTFT. Direct aggregate adjustment also has an established precedent in aggregateMetrics for warm-up scaling. The non-PD path is unaffected because projection immediately returns when there are no parent requests. |
| G7 | parity | ✅ CONFIDENT | Since `projectPDMetrics()` runs during finalization for both `blis run` and `blis replay` (INV-13), are there any differences in how `DecodeSubReq.ITL` is reconstructed during replay vs simulation that could cause replay to classify a request as `served` when run classified it as no-token? | No replay-specific ITL reconstruction occurs. Replay reconstructs fresh input Request objects from trace token counts and then runs the same ClusterSimulator and Simulator execution code as run; decode ITL is generated during simulated decode steps in both cases. Thus classification uses the same runtime-produced DecodeSubReq.ITL representation. |
| G8 | parity | ✅ CONFIDENT | `blis observe` generates request metrics directly from server trace records without calling `projectPDMetrics()`. Will `observe` exclude no-token or dropped decode requests from its latency distributions in the exact same manner as `run` and `replay` now do? | Observe does not use the same internal predicate because it has no ParentRequest or DecodeSubReq state. It independently excludes failed records and records without valid positive TTFT/E2E, which gives the equivalent outcome for real-server drops or no-token responses: they do not enter the latency distributions. |
| G9 | tests | ✅ CONFIDENT | The new test `TestDisaggregation_DroppedParent_NotInLatencyMaps` tests decode KV allocation failures. Are there companion integration tests that explicitly trigger the late-drop (unroutable decode pod) and decode-timeout-while-queued shapes to verify they are also excluded from latency maps in a full cluster run? | No. The new full-cluster integration test forces decode KV reservation failures at transfer start. Late-drop and timeout-while-queued are covered by direct table-driven projectPDMetrics tests, not by full cluster runs that trigger those lifecycle paths. |
| G10 | tests | ✅ CONFIDENT | If the condition `servedFirstToken := parent.DecodeSubReq != nil && len(parent.DecodeSubReq.ITL) > 0` were accidentally refactored to check `parent.TransferCompleteTime > 0`, which specific test in the test suite would fail? | TestPDParentMetrics_NoTokenExcludedFromLatency would fail, specifically its "drop at transfer complete (late-drop): nil DecodeSubReq" case. That case has TransferCompleteTime=200 but expects every parent metric entry to be absent; using TransferCompleteTime > 0 as served would create TTFT/E2E/completion/request entries. |
| G11 | rationale | ✅ CONFIDENT | Why was `len(DecodeSubReq.ITL) > 0` chosen as the primary indicator that a token was served, rather than checking an explicit lifecycle state or non-zero `FirstTokenTime` on the parent request? | ITL is the direct execution artifact for decode output: the simulator appends one entry only when a decode token is generated. FirstTokenTime is unsuitable because normal PD decode sub-requests start at ProgressIndex == InputLen, so their prefill-transition FirstTokenTime block never fires and FirstTokenTime remains zero even when decode tokens are emitted. ParentRequest has no explicit first-token lifecycle field. |
| G12 | rationale | ✅ CONFIDENT | Why is rolling back `prefillTTFT` from `m.TTFTSum` inside `projectPDMetrics()` preferred over preventing the prefill sub-request from contributing to `TTFTSum` during disaggregated prefill completion? | The no-token disposition is not known when prefill completes: the request may subsequently serve, drop during transfer, or time out. Keeping normal per-instance prefill metric recording avoids making the generic Simulator PD-aware and preserves valid per-instance metrics. projectPDMetrics is already the layer that converts sub-request metrics into parent metrics, so it is the appropriate place to replace or remove the prefill contribution. |
| G13 | scope | ✅ CONFIDENT | The linked issue #1511 was originally reported for drop-at-transfer-start. Does broadening the scope to exclude decode-timeout-while-queued and preempted-then-timed-out requests affect non-disaggregated cluster configurations or metric projections? | No. These additional exclusions only affect PD parents. Non-disaggregated configurations have an empty parentRequests map, causing projectPDMetrics to return without modifying metrics. |
| G14 | scope | ✅ CONFIDENT | The PR notes a diagnostic gap in `PDMetrics.DroppedAtDecodeKV` (`sim/cluster/pd_metrics.go:109`). Does this PR address that reporting gap, or is it explicitly deferred to a separate issue? | The reporting gap is not addressed. PDMetrics.DroppedAtDecodeKV still counts only TransferCompleteTime > 0 with an empty DecodeInstanceID, while current drop paths retain the routed DecodeInstanceID. This PR changes latency projection and leaves that diagnostic for follow-up. |
| G15 | errors | ✅ CONFIDENT | In `projectPDMetrics()`, if `m.TTFTSum` is decremented when `completed` is true and `served` is false, is there any possibility of `TTFTSum` underflowing to a negative value if `prefillTTFT` was incorrect or double-subtracted? | Not in a valid production run. hasPrefillTTFT means the prefill key being removed exists, and its value was included once in the aggregated per-instance TTFTSum; each parent is processed once during finalization. Subtracting that term therefore cannot make the valid aggregate negative. A manually corrupted Metrics object could produce a negative value because there is no defensive lower-bound check. |
| G16 | errors | ✅ CONFIDENT | What happens if `projectPDMetrics()` is called twice on the same `ClusterSimulator` instance? Will the second call panic or subtract `prefillTTFT` from `m.TTFTSum` a second time? | A second call will not panic and will not subtract the no-token prefill TTFT twice, because the first call deletes the prefill key and the rollback requires hasPrefillTTFT. However, the method is not fully idempotent for served parents: decode-side source keys are gone on the second call, so E2E can be recomputed through the CompletionTime fallback and overwrite the previously reconstructed parent E2E. |

### Items to fix

### Important to consider
- **F1** — The explicit acceptance request for separate run and replay drop scenarios is not implemented as two CLI-level tests, but both commands use the same ClusterSimulator.Run finalization path.
  _sim/cluster/cluster.go:1803-1829 defines completed/served; sim/cluster/cluster.go:1890-1938 gates E2E on served; sim/cluster/cluster.go:1977-2030 excludes no-token TTFT and rolls back its sum; sim/cluster/cluster.go:2033-2084 gates the remaining maps; sim/cluster/cluster.go:896-900 preserves DroppedUnservable; sim/cluster/cluster.go:1338-1345 preserves timeout accounting; sim/cluster/pd_e2e_causality_test.go:519-668 covers the complete served/no-token taxonomy._
- **F2** — This is documentation drift, not a runtime defect.
  _docs/contributing/standards/invariants.md:246-264 documents the new served/no-token behavior; sim/cluster/pd_metrics.go:18-24 still says ParentTTFT is the prefill sub-request TTFT; sim/cluster/pd_metrics.go:113-118 later gives the corrected formula._
- **F3** — Comment-only cleanup is advisable.
  _sim/cluster/disaggregation_test.go:1234-1246 describes nil/empty-ITL as fallback; the actual cases at sim/cluster/disaggregation_test.go:1295-1323 expect no entry. sim/cluster/pd_metrics.go:18-24 conflicts with sim/cluster/pd_metrics.go:113-118._
- **G1** — The comments at sim/cluster/cluster.go:2023-2028 and docs/contributing/standards/invariants.md:252 incorrectly imply that a preempted-then-timed-out decode has already added its re-stamped term to TTFTSum.
  _sim/simulator.go:990-1000 re-stamps FirstTokenTime and RequestTTFTs without changing TTFTSum; sim/simulator.go:669-693 increments TTFTSum only on completion; sim/event.go:161-228 times requests out without recording completion; sim/batch_formation.go:308-312 clears ITL and TTFTSet on preemption; sim/cluster/cluster.go:2022-2030 documents and performs the prefill rollback._
- **G2** — Value validation would only harden against corrupt or manually constructed state.
  _sim/cluster/cluster.go:1828 uses only ITL length; sim/cluster/cluster.go:1987-2003 computes and guards projected TTFT; sim/simulator.go:960-987 floors currStepAdvance to at least 1 and appends currStepAdvance plus OutputTokenProcessingTime._
- **G5** — A debug log could aid diagnosis but is not required by R1.
  _docs/contributing/standards/rules.md:24-32 permits a dropped item to be counted or documented; sim/cluster/cluster.go:1780-1790 documents exclusion; sim/cluster/cluster.go:2014-2030 performs it without logging; sim/cluster/cluster.go:896-900 counts decode drops; sim/event.go:161-173 counts timeouts._
- **G7** — The existing run/replay parity test uses successful requests rather than a forced no-token drop.
  _sim/workload/replay.go:51-114 reconstructs only request inputs and metadata; cmd/replay.go:531-553 runs cluster.NewClusterSimulator(...).Run(); sim/simulator.go:981-988 appends decode ITLs during execution; sim/cluster/cluster.go:926-930 calls the shared projection during finalization; cmd/replay_test.go:1956-2007 checks run/replay PD metric parity._
- **G8** — Parity is semantic rather than byte-identical implementation: observe relies on the external trace status and timestamps.
  _cmd/observe_cmd.go:690-701 includes only status="ok" records with TTFT > 0, E2E > 0, and E2E >= TTFT; sim/workload/tracev2.go:692-713 similarly excludes non-ok records and non-positive E2E for saturation metrics._
- **G9** — Full-path coverage for the two additional realistic shapes would be useful, but the shared predicate and table coverage exercise the core new behavior.
  _sim/cluster/disaggregation_test.go:347-408 configures tight decode KV and exercises reservation failure; sim/cluster/pd_e2e_causality_test.go:570-599 models transfer-start drop, transfer-complete drop, and queued timeout as table cases; no corresponding late-drop or queued-timeout integration test exists under sim/cluster._
- **G11** — An explicit parent-level served-token state could be clearer, but the existing ITL signal is the canonical runtime evidence.
  _sim/request.go:50-57 defines FirstTokenTime and ITL; sim/request.go:116-120 documents decode sub-requests starting after prefill; sim/simulator.go:981-1000 appends ITL on decode execution but sets FirstTokenTime only at the prefill boundary; sim/cluster/cluster.go:1853-1868 documents the normal-PD versus re-prefill clock frames; sim/cluster/parent_request.go:7-51 contains no first-token field._
- **G14** — The runtime droppedAtDecodeKV counter used for INV-1 remains correct and is separate from this diagnostic.
  _sim/cluster/pd_metrics.go:48-51 and sim/cluster/pd_metrics.go:107-110 retain the old predicate; sim/cluster/parent_request.go:41-45 says DecodeInstanceID is assigned upfront; sim/cluster/pd_events.go:364-380 records late drops without clearing it._
- **G15** — The direct unit branch test intentionally starts TTFTSum at zero and therefore expects a negative delta, but that fixture is modeling the adjustment rather than a valid full aggregate.
  _sim/simulator.go:686-693 adds each completed request's FirstTokenTime once; sim/cluster/cluster.go:1693 aggregates per-instance TTFTSum; sim/cluster/cluster.go:1977-2030 reads and removes the matching prefill key before subtracting it; sim/cluster/cluster.go:926-930 calls projection once in Run._
- **G16** — The private method is designed for one finalization pass, so repeated-call idempotence is not a production contract.
  _sim/cluster/cluster.go:1890-1937 deletes only pfx/dec E2E keys and may rewrite pid; sim/cluster/cluster.go:1977-2030 deletes the prefill TTFT key and conditions subtraction on hasPrefillTTFT; sim/cluster/cluster.go:2047-2055 only panics for a served parent with nil OriginalRequest; sim/cluster/cluster.go:926-930 has a single production call site._

susiejojo (approved):
lgtm

Issue #1553, opened 2026-08-12
DP-as-placement: support PD disaggregation, the autoscaler, and node pools with --dp > 1


namasl:
## Summary

Follow-up to #1531 (DP-as-real-placement). #1531 delivers DP-as-placement for the plain, non-node-pool `--num-instances` path (MoE, EP off): `--dp N` spawns `N` single-node engine replicas per logical instance, each configured per-rank (`DP=1`). To keep that PR focused, reviewable, and shipping only tested paths, these placement combinations are **guarded with a fail-fast `logrus.Fatalf`**:

- **PD disaggregation** (`--prefill-instances` / `--decode-instances` / `--prefill-decode-instances` / `--encode-instances` > 0): expanding `NumInstances` by `dp` breaks the pool-topology arithmetic (`P+D+S+E == NumInstances`), and the per-pool TP / per-pool KV auto-calc paths thread the global `dataParallelism`. Supporting DP within PD pools (each pool gets `dp` per-rank replicas) needs its own design.
- **Autoscaler** (`--model-autoscaler-interval-us > 0` or a policy-bundle autoscaler): the semantics of dynamically scaling a `dp`-expanded population (add one rank? one whole DP group?) are unclear and need a decision.
- **Node pools** (`node_pools:` in the policy bundle): mechanically the multiplied instance count reuses the tested per-instance node-pool placement path (each `PlaceInstance` reserves `TP` GPUs; `applyPerInstanceKVCapacity` reads `EffectiveDP()==1` for per-rank sizing), but the combination with `N×M` replicas was not audited or tested in #1531, so it is guarded rather than shipped untested. This is also where `--dp N` would *literally* reserve `N × M × TP` GPUs from real inventory.

(EP-on + `--dp > 1` is a separate concern tracked by #1548; replay parity by #1556.)

## Acceptance criteria

- [ ] `--dp N` (MoE) composes with PD disaggregation: each pool spawns `N` per-rank replicas; per-pool TP and per-pool KV auto-calc size per-rank; INV-1 / INV-6 hold; `blis replay` still fails fast.
- [ ] A decision (support or continue to reject with a clear message) for `--dp N` + autoscaler, implemented and tested.
- [ ] `--dp N` + node pools: audited and tested — `N × M` replicas placed, reserving `N × M × TP` GPUs, each sized per-rank from its actual placed GPU.
- [ ] The #1531 guards are lifted for whatever becomes supported.

## Context

[...]

Issue #1548, opened 2026-08-12
EP placement + live EP-mode step-time toggle (split from #1531)


namasl:
> **Also applies to Kimi K3** ([#1638](https://github.com/inference-sim/inference-sim/issues/1638)) — this is a deployment capability shared by the GLM-5.2 and Kimi K3 MLA-MoE families; landing it here covers both.

> Split from #1531 (DP/EP real placement + live EP toggle, part of #1532 GLM-5.2 support). #1531 is now scoped to the **DP-as-real-placement** half (independent, shippable first); this issue tracks the **EP placement + live EP-mode step-time toggle** half.

**Blocked by #1530** (inter-node network cost): a faithful EP-on model needs the all-to-all communication term for *cross-node* traffic, which needs the network model.

## Scope

- Place EP ranks consistently with the `TP·DP` group already used by the latency + KV terms (`EffectiveEP() = TP·DP` **when EP is enabled on an MoE model, else 1**, [`sim/config.go#L179-L207`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/config.go#L179-L207)). EP placement reserves **no** GPUs beyond #1531's `N×TP` (the EP group *is* those same `TP·DP` GPUs); this issue governs how experts map onto that group (EP-on = whole experts per rank / all-to-all vs EP-off = tensor-sharded replicas) and the resulting step-time — **not** additional GPU reservation. This is what cleanly separates it from #1531.
- Make `--enable-expert-parallel` **step-time-live**: EP on vs off produces a different `StepTime` for at least one MoE config, driven by an **EP-mode profile** — **without** adding a second copy of the already-present `DP>1` dispatch/combine comm term (`tMoEDispatch` under `β_EP`, [`trained_physics_model.go#L380-L428`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/latency/trained_physics_model.go#L380-L428)). Framing it as "add comm cost" risks **double-charging** — the existing **network-blind** dispatch/combine term already fires at `DP>1`; the cross-node cost is #1530.
[...]

PR #1555, opened 2026-08-12
feat(cluster): DP-as-real-placement — spawn N engine replicas for MoE --dp N (#1531)

### LINKED ISSUE #1531: DP as real placement — reserve N×TP GPUs, spawn N engine replicas

**Rescoped 2026-08-12:** #1531 now tracks only the **DP-as-real-placement** half — independent and shippable first. The **EP placement + live EP-mode step-time toggle** half (blocked by #1530) moved to #1548.

## Summary

Data Parallelism (DP) exists in BLIS only as **latency-model and KV-capacity terms**, not as a real placement/scheduling construct. `--dp` changes step-time math and (for DP) the auto-calculated KV-block total, but does **not** spawn engines or reserve additional GPUs in the cluster placer. This is a fidelity/coverage gap for MoE deployments (like GLM-5.2's `single_node_tep` / `multi_node_dep` strategies), where DP is a primary throughput lever.

## Evidence in the code

- Step time: DP enters as `tpdp = tp·dp` — [`sim/latency/trained_physics_model.go#L222-L279`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/latency/trained_physics_model.go#L222-L279).
- KV capacity: DP multiplies the aggregate block count for MoE only — [`sim/latency/kv_capacity.go#L155-L196`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/latency/kv_capacity.go#L155-L196).
- Group-size helpers: `EffectiveMoEGroupSize() = TP·DP` — [`sim/config.go#L179-L207`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/config.go#L179-L207).
- Dense `--dp > 1` is rejected (documented equivalent: `--num-instances` router replicas) — [`cmd/root.go#L707-L716`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/cmd/root.go#L707-L716).

## Acceptance criteria

- [ ] `--dp N` on an MoE model reserves `N × TP` GPUs (not `TP`) and spawns `N` engine replicas in the cluster placer. Requests are distributed disjointly across the `N` replicas via the existing cluster router (mirroring vLLM's internal DP load balancer).
- [ ] Each of the `N` replicas is sized with the **per-rank** KV budget — its capacity calc receives `dp=1` — so aggregate KV blocks equal today's single-instance `dp`-multiplied total (`sim/latency/kv_capacity.go`: `if params.IsMoE && dp > 1 { totalBlocks *= int64(dp) }`). **No `dp²` double-count.**
- [ ] `blis replay` `logrus.Fatalf` at startup on `--dp > 1` for MoE — DP-as-placement is **run-only**; no silent single-instance fallback (**INV-13**, never silent degradation).
- [ ] Request conservation (**INV-1**) holds across the DP replicas.
- [ ] **INV-6:** same seed ⇒ byte-identical stdout.

## Parity — INV-13 (must address; the `--dp` path is NOT auto-exempt)

`--dp` is a **shared** flag ([`cmd/root.go#L1104-L1107`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/cmd/root.go#L1104-L1107)) that `blis replay` already threads into the latency model (`dataParallelism` → `sim.NewModelHardwareConfig`) and the KV calc (`CalculateKVBlocks`). So `--dp N` on an MoE model is **today a config supported by both `run` and `replay`**, producing byte-identical metrics (single instance + DP latency-math in both) — it is bound by **INV-13**. The node-pool/autoscaler `Fatalf` guards in `cmd/replay.go` do **not** gate the `--dp` path.

If this issue makes `run` spawn `N` real replicas (changing routing / queueing / batching, hence per-request metrics) while `replay` keeps the single-instance DP-math model, the two paths **diverge for the same flags** — exactly the silent degradation INV-13 forbids. Therefore DP-as-placement must become a **run-only** feature: `blis replay` MUST `logrus.Fatalf` on `--dp > 1` (MoE) at startup, mirroring the existing node-pool/autoscaler guards — never a silent single-instance fallback (see AC above).

Placement itself lives in `sim/cluster` (`blis run`); `observe` uses real-server placement.

## Independence

The DP-placement piece does **not** depend on multi-node TP placement (#1537): DP replicas are separate engines, each single-node (`TP` GPUs apiece), reusing the existing per-instance placement path (as used by `--num-instances`), **not** #1537's multi-node `PlacementManager`. It is independent and shippable first.

## Related

- #1548 — EP placement + live EP-mode toggle (the split-out half, blocked by #1530).
- #1416 / #1419 — DP/EP *latency math* (already landed); this issue is the *placement* slice, not a re-file.
- #1348 — broad "Add EP + DP support for MoE models" umbrella.

namasl:
## Summary

Makes `--dp N` on an MoE model a **real placement construct** on `blis run`, not just a latency/KV divisor. Previously `--dp N` modeled one *lumped* engine (single queue, single batch former, single KV pool) whose step time pretended the batch was spread across `dp` ranks. Real vLLM data parallelism is `N` independent `EngineCore`s with an internal load balancer distributing requests disjointly. This PR spawns those `N` replicas.

On `blis run`, MoE `--dp N` now spawns `N` real single-node engine replicas **per `--num-instances`** (reserving `num_instances × N × TP` GPUs), each configured per-rank (`DP=1`), reusing the existing per-instance placement path. The existing cluster router distributes requests disjointly across them. **The entire change lives in `cmd/` — `sim/cluster` is unchanged** (it already handles `NumInstances` instances correctly; this PR just tells it `numInstances × N` with per-rank config).

Closes #1531. Part of the GLM-5.2 support epic (#1532).

## Why each replica is `DP=1`

The lumped single-instance model divided token-population work by `dp` *precisely because* it held every request. Once the cluster router splits requests across `N` real instances, each instance's batch is already ~`1/N` — so each replica must be `DP=1`, or the `/dp` factor double-counts (≈`N×` throughput inflation). This preserves the per-rank compute time and corrects the routed-expert weight scoping from `moeGroup = TP·DP` (EP-on-like) to `moeGroup = TP` (EP-off: experts replicated per DP rank, tensor-sharded across TP) — the correct physics for the default `--enable-expert-parallel=false`. True EP-on placement (experts sharded across the DP group + inter-node network cost) is deferred to #1548.

## Behavioral contracts

- **BC-1 — DP spawns real replicas.** GIVEN an MoE model, EP off, no PD/autoscaler, `--num-instances M`, `--dp N` (N>1) WHEN `blis run` builds the cluster THEN the cluster contains exactly `M × N` engine replicas and requests distribute across all of them.
- **BC-2 — Per-rank KV, aggregate preserved, no `dp²`.** Each replica's KV-block capacity equals the per-rank (`dp=1`) budget; the sum over the `N` replicas equals the pre-#1531 lumped `dp`-multiplied total — never `dp²·perRank`.
- **BC-3 — Run-only (INV-13).** `blis replay` `logrus.Fatalf`s at startup on MoE `--dp > 1`, before running — never a silent single-instance fallback.
- **BC-4 — Conservation (INV-1).** Request conservation holds across the `M × N` replicas.
- **BC-5 — Determinism (INV-6).** Same flags + seed ⇒ byte-identical stdout.
- **BC-6 — No-op for unaffected configs (INV-6).** `--dp 1` (default) and dense `--dp > 1` (still rejected with its existing message) are byte-identical to a pre-#1531 build.
- **BC-7 — Unsupported placement combos fail fast.** MoE `--dp > 1` with `--enable-expert-parallel` (→ #1548), PD disaggregation, or the autoscaler (→ #1553) `logrus.Fatalf`s with an actionable message rather than being silently mis-modeled.
- **BC-8 — Node pools reserve `N × M × TP` GPUs.** Node pools are supported for free via the multiplied instance count + `DP=1` (the per-instance KV recalc reads `EffectiveDP()==1`); no node-pool-specific code added.

- **run** — implements DP-as-placement.
- **replay** — `logrus.Fatalf` on MoE `--dp>1` (run-only; INV-13 preserved). Dense `--dp>1` and roofline `--dp>1` remain rejected in the shared `resolveLatencyConfig`.
- **observe** — unaffected (uses real-server placement; does not build a cluster from `--dp`).

The `NewModelHardwareConfig` / per-pool `CalculateKVBlocks` wiring and `resolveLatencyConfig` are left untouched, so all existing wiring-guard tests (`TestRunReplay_DPEPFlags_ThreadedIntoConstructor`, `TestPerPoolKVBlocks_ThreadsGlobalDP`, `TestResolveLatencyConfig_DPScalesAutoKVCapacity`) remain valid — the `DP=1` divergence is an explicit post-construction override, not a wiring change.

## Testing

- `go build ./...` — clean
- `go test ./... -count=1` — all pass

New tests (`cmd/dp_placement_test.go`): `TestPlanDPPlacement` (pure decision + EP/PD/autoscaler guards), `TestReplayCmd_MoEDPPlacement_Rejected` (BC-3 subprocess Fatalf), `TestRunCmd_MoEDPPlacement_SpawnsReplicas` (BC-1/4/5/8 — replica count for `--num-instances 1 --dp 2` and `2 --dp 2`, INV-1 conservation, INV-6 byte-identity; offline via the git-tracked `deepseek-v2-lite` MoE fixture), and `TestDPPlacement_PerRankKV_NoDoubleCount` (BC-2 law: per-replica == per-rank and aggregate == lumped).

## Discovered / deferred issues

- **#1553** (filed) — DP-as-placement with PD disaggregation and the autoscaler (guarded here).
- **#1548** (existing) — EP-on placement + live EP-mode step-time toggle (guarded here; blocked by #1530).

## Docs

`--dp` flag help; CLAUDE.md (Current Implementation Focus); glossary + latency-models guide; INV-13 statement synced across `invariants.md` + `CLAUDE.md` + `constitution.md` to list MoE `--dp>1` as a run-only feature.


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:

namasl:

namasl:

## qa-review — PR #1555: ✅ PASS
_feat(cluster): DP-as-real-placement — spawn N engine replicas for MoE --dp N (#1531) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. For supported MoE run configurations, the PR expands each logical instance into DP replicas, configures each replica with DP=1, preserves per-rank KV sizing, routes through the existing cluster, rejects MoE DP replay, and tests conservation and deterministic output. The additional guards, warnings, tests, and documentation are directly related safety work rather than unrelated functionality. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The principal user-facing and invariant documentation is updated: flag help describes run-only placement and unsupported combinations, the latency guide explains per-replica DP=1 behavior, the glossary no longer says BLIS lacks DP, and INV-13/CLAUDE document replay rejection. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | I found no materially stale or false code comment introduced by the change. The new comments consistently describe active placement as M×N instances, per-rank DP=1, auto-KV division, replay rejection, and the guarded combinations. |
| G1 | correctness | ✅ CONFIDENT | In `applyDPPlacement`, when `autoScaled` is true, `perRankKVBlocks` is computed as `totalKVBlocks / int64(dp)`. If `totalKVBlocks` is not evenly divisible by `dp` (e.g. 1000 blocks with `dp=3`), integer division truncates the remainder. Does this loss of blocks cause per-rank capacity drift, and should remainder blocks be handled or rounded up? | No remainder is possible for the shipped auto-capacity path. CalculateKVBlocks first computes an integral per-rank block count and then multiplies that exact integer by dp for MoE. Dividing the resulting aggregate by the same dp therefore recovers the original per-rank count exactly. A hand-constructed call such as total=1000, dp=3 would truncate, but that violates applyDPPlacement's documented autoScaled precondition; rounding up would incorrectly invent capacity. |
| G2 | correctness | ✅ CONFIDENT | In `runCmd` (`cmd/root.go`), `maxModelLen` re-capping is gated on `autoScaled && maxModelLen > 0 && blockSizeTokens > 0`. What happens if `maxModelLen` is explicitly passed on the CLI while `--total-kv-blocks` is ALSO set explicitly (`autoScaled` = false)? Could a user-specified `--total-kv-blocks` combined with `--max-model-len` and `--dp > 1` trigger the per-replica `NewSimulator` panic because `autoScaled` is false and re-capping is skipped? | That combination does not create the proposed panic. resolveLatencyConfig caps maxModelLen against totalKVBlocks regardless of whether the block count was explicit or auto-calculated. Under DP placement, an explicit total is deliberately retained unchanged on every replica, so the original cap remains valid and no second cap is needed. |
| G3 | invariants | ✅ CONFIDENT | INV-13 requires that exported traces replayed with identical flags produce identical metrics. Since `blis replay` fatal-exits when `--dp > 1` on MoE, how does a user replay a trace that was generated from a `blis run --dp 2` simulation, and is there any risk of silently replaying it with `--dp 1` without detecting the instance count mismatch? | A DP=2 run trace currently cannot be faithfully replayed; replaying with --dp 2 fails explicitly and points users back to blis run, with replay support tracked separately. If a user deliberately or accidentally omits the flag and uses the default DP=1, replay cannot detect the mismatch because TraceV2 does not record DP or instance count. That is not an identical-flags INV-13 case, but it is a provenance limitation. |
| G4 | invariants | ✅ CONFIDENT | Under DP-as-placement, `blis run` expands `--num-instances` by factor `N` and relies on the cluster router to distribute requests across `N` replicas. Does the seed passed to `--seed` fully determine the order of request dispatch and batching across all spawned engine replicas to preserve INV-6 (byte-identical stdout for identical seeds)? | Yes. Cluster execution is single-threaded and globally orders events, breaking equal-time ties by instance index. Routing randomness comes from a deterministic, seed-derived router RNG, while each instance has an isolated deterministic RNG partition. The new end-to-end test repeats the same DP-placement run and compares stdout byte-for-byte. |
| G5 | rules | ✅ CONFIDENT | In `cmd/root.go`, `numInstances` and `totalKVBlocks` (which are package-level CLI flag variables) are mutated in place within `runCmd` after `resolveLatencyConfig` completes. Does mutating global CLI flag variables during `runCmd` execution violate clean state isolation or risk side effects if commands are executed sequentially in test suites? | The in-place mutation follows this command package's existing package-level Cobra-state design and does not affect normal CLI use, where one command executes once. Sequential in-process executions without reparsing or resetting globals could retain the expanded values, but the existing resolver already mutates package variables and tests commonly isolate or restore them; the new integration tests use subprocesses. |
| G6 | rules | ✅ CONFIDENT | Rule R23 dictates single-code-path parity between `run` and `replay`. `resolveLatencyConfig` is shared, but `planDPPlacement` and `applyDPPlacement` are called only in `runCmd` while `replay.go` has a separate `logrus.Fatalf` check. Why was `planDPPlacement` / replay gating not integrated directly into `resolveLatencyConfig` to keep flag validation completely unified? | Keeping placement outside resolveLatencyConfig is justified because the resolver is shared latency/model/KV configuration, while placement depends on run-only topology state such as PD pools, policy-bundle autoscaling, and node pools. Putting expansion there would mutate replay's instance count before replay can reject the unsupported feature. Replay instead shares model resolution so it can reliably determine MoE, then performs its command-specific support gate. This does not violate R23 because run and replay intentionally do not both support this transformation. |
| G7 | parity | ✅ CONFIDENT | How does `blis observe` handle `--dp > 1` for MoE models? If `observe` dispatches workload to real endpoints, does it mirror `run`'s placement expansion or does it silently pass `--dp` to the latency model? | observe does not expose the simulation --dp flag and does not construct a latency model or simulated cluster. It sends requests to the single configured real-server URL; any real DP placement/load balancing is therefore owned by that endpoint or its external gateway. Passing --dp to observe is an unknown flag rather than a silent latency-model setting. |
| G8 | parity | ✅ CONFIDENT | When `--dp N` expands `--num-instances M` into `M × N` replicas, the per-instance metric JSON output structure contains `instance_0` through `instance_{M*N-1}`. Are aggregate cluster metrics and per-instance metrics formatted identically to standard `--num-instances M*N` runs so downstream analysis scripts parse them without schema breaking changes? | Yes. DP placement only changes the NumInstances value supplied to the existing DeploymentConfig. Cluster construction, instance naming, per-instance SaveResults calls, and aggregate BuildOutput/EmitOutput paths are exactly the same as for an ordinary --num-instances M×N run, so there is no new output schema. |
| G9 | tests | ✅ CONFIDENT | `TestRunCmd_MoEDPPlacement_SpawnsReplicas` relies on re-executing `blis run` in a subprocess and checking for stdout strings like `"instance_id": "instance_1"`. If the JSON output format or instance naming scheme changes in `sim/cluster`, this test will fail. Is there an in-memory test asserting that the cluster object receives the expected `N` engine replicas directly? | There is no new in-memory test that directly inspects ClusterSimulator's private instance slice after applying the CLI plan. Instead, the pure planning and application functions are unit-tested, and the real command is integration-tested through the public metrics output. That is adequate coverage of the new realistic path; a test tied directly to private cluster internals would be more structural and less refactor-safe. |
| G10 | tests | ✅ CONFIDENT | `TestDPPlacement_PerRankKV_NoDoubleCount` verifies that `perReplica * 2 == lumped` for `dp=2`. If someone refactors `planDPPlacement` to support `dp=3` or non-power-of-two DP degrees, would this unit test detect off-by-one or rounding errors in `applyDPPlacement` when `lumped` is not evenly divisible by `dp`? | The specific no-double-count integration law uses dp=2, while the helper test also exercises dp=4; neither uses dp=3. However, auto-derived totals cannot be non-divisible because the producer computes a per-rank integer and multiplies it by dp before applyDPPlacement divides by that same value. Thus a dp=3 rounding defect is not reachable from the shipped auto-calculation path. |
| G11 | rationale | ✅ CONFIDENT | The PR implements DP placement entirely in the CLI layer (`cmd/root.go`) by multiplying `numInstances` and dividing `totalKVBlocks`, leaving `sim/cluster` unchanged. What alternatives were considered (such as adding native DP topology support to `sim/cluster`), and why was CLI-level instance multiplication preferred over cluster-level DP awareness? | The implementation follows the issue's stated independence strategy: DP ranks are independent, homogeneous EngineCore replicas behind the already-existing router, which is behaviorally equivalent to expanding the existing instance count. Native cluster-level DP topology would add a second replica concept and require changes to routing, metrics, placement, lifecycle, pools, and autoscaling despite those combinations being deferred. CLI expansion therefore reuses the mature cluster semantics while keeping every replica's canonical SimConfig at DP=1. |
| G12 | rationale | ✅ CONFIDENT | Why was `blis replay` support for MoE `--dp > 1` explicitly disabled via `logrus.Fatalf` rather than having `replay` synthesize `N` engine instances from trace request IDs or replay against `N` engine cores? | Replay was disabled because its current path constructs one deployment from replay flags and does not reconstruct DP topology from the trace; request IDs contain no authoritative rank assignment, and synthesizing assignment from them would not necessarily reproduce the run router's queue- and state-dependent decisions. The acceptance criteria explicitly require fail-fast behavior until replay can instantiate and route across the same N replicas. |
| G13 | scope | ✅ CONFIDENT | The PR includes explicit guards and error messages for combination with PD disaggregation, autoscalers, and node pools (referencing issue #1553), as well as warning thresholds when spawning > 512 instances. Were these safety guards part of the original issue #1531 scope, or discovered during implementation? | The acceptance criteria do not individually call out those combinations or the 512-instance warning. They are implementation-time safety boundaries around the scoped design: #1531 explicitly relies on the ordinary fixed per-instance path, whereas PD, autoscaling, and node pools alter topology or lifecycle semantics. Rejecting them prevents silent mis-modeling; the >512 warning is only an operational typo warning and does not change behavior. |
| G14 | scope | ✅ CONFIDENT | The diff adds a specific warning when `--moe-comm-backend` is passed alongside DP-as-placement, informing the user that cross-DP dispatch/combine is inert. Was handling `--moe-comm-backend` part of #1531's acceptance criteria, or does it belong to the EP-on placement scope in #1548? | The acceptance criteria do not require a moe-comm-backend warning. The actual cross-DP dispatch/combine model belongs to DP>1 model configurations and future EP-on placement, while #1531 deliberately configures each real replica at DP=1. The warning is therefore appropriate compatibility/safety handling: it prevents an accepted flag from misleading users while leaving the deferred physics to #1548. |
| G15 | errors | ✅ CONFIDENT | `planDPPlacement` returns a standard Go `error` for unsupported configurations, which `runCmd` catches and turns into a `logrus.Fatalf`. Is `planDPPlacement` ever called outside of CLI commands where returning an error instead of crashing is required, or should it be a library package function in `sim/cluster`? | Production code calls planDPPlacement only from runCmd; its other callers are unit tests. Returning an error is still preferable because it keeps the decision function pure and testable, while the CLI boundary owns process termination. It should not be moved into sim/cluster solely to crash there, since project rule R6 forbids Fatalf in library code and the cluster itself needs no DP-specific awareness for this implementation. |
| G16 | errors | ✅ CONFIDENT | When `--max-model-len` exceeds per-rank KV capacity under DP placement, `runCmd` issues a `logrus.Warnf` and caps `maxModelLen` to `kvFeasibleMax`. Why is this handled with silent capping and a warning rather than failing fast with `logrus.Fatalf` as required by the "never silent degradation / fail fast" principle? | This is not silent degradation: it emits a warning and follows the simulator's established vLLM-style max-length feasibility behavior. maxModelLen is a scheduling/sequence limit constrained by available KV capacity, so lowering it to the largest representable token count produces a valid configuration instead of allowing a constructor panic. Fail-fast is required for unsupported semantics that would otherwise be silently mis-modeled; it does not require every documented auto-adjustment to be fatal. |

### Items to fix

### Important to consider
- **F1** — PD disaggregation, autoscaling, node pools, and EP-on combinations are intentionally rejected and deferred rather than implemented.
  _cmd/root.go:2161-2214 plans and applies DP placement; cmd/root.go:2232-2255 passes the expanded count and per-rank DP into the existing cluster path; sim/cluster/cluster.go:316-335 constructs one TP-sized instance per expanded count; cmd/replay.go:286-301 rejects MoE --dp>1; cmd/dp_placement_test.go:436-459 tests replica expansion, conservation, and byte-identical stdout; go build ./... and go test ./cmd both pass._
- **F2** — docs/guide/cluster.md still describes only generic --num-instances behavior; adding a DP example there could improve discoverability, but the behavior is already documented in the dedicated latency guide and CLI help.
  _cmd/root.go:1231 documents --dp behavior in CLI help; docs/guide/latency-models.md:177-197 documents DP placement and communication semantics; docs/concepts/glossary.md:188 updates the TP/DP/EP overview; docs/contributing/standards/invariants.md:175-181 updates canonical INV-13; CLAUDE.md:441 and CLAUDE.md:477 update the working summary._
- **F3** — The short run-path summary at cmd/root.go:2165 mentions EP, PD, and autoscaler but omits node pools from that parenthetical list; the complete comment and implementation above it include node pools, so this is only a minor wording omission.
  _cmd/root.go:377-417 documents the plan contract and supported combinations; cmd/root.go:446-456 documents explicit versus auto-derived KV semantics; cmd/root.go:2161-2214 matches the implementation; cmd/replay.go:289-300 matches the fatal guard._
- **G1** — A defensive divisibility assertion could make misuse of the pure helper easier to diagnose, but no realistic producer-generated autoScaled value has a remainder.
  _sim/latency/kv_capacity.go:356-375 computes totalBlocks per rank and then performs totalBlocks *= int64(dp); cmd/root.go:452-456 defines autoScaled as that successful global auto-calculation; cmd/root.go:461-464 divides by the same dp. TestDPPlacement_PerRankKV_NoDoubleCount validates recovery of the dp=1 budget in cmd/dp_placement_test.go:626-675._
- **G2** — There is no dedicated end-to-end test combining explicit --total-kv-blocks, an oversized explicit --max-model-len, and DP placement, although the common resolver cap and explicit-KV branch are separately covered.
  _cmd/root.go:717-755 distinguishes explicit and automatic KV counts; cmd/root.go:792-803 performs the initial cap without an autoScaled condition; cmd/root.go:452-456 and 461-464 preserve an explicit per-instance block count; sim/simulator.go:166-178 checks the same ceil(maxModelLen/blockSize) requirement._
- **G3** — Recording resolved topology in trace metadata would allow replay to reject an omitted or conflicting --dp instead of accepting the default DP=1.
  _cmd/replay.go:289-301 fatal-exits for MoE --dp>1; docs/contributing/standards/invariants.md:175-181 declares this configuration unsupported by replay; sim/workload/tracev2.go:17-67 defines the trace header and contains no DP or instance-count field; cmd/root.go:2429-2437 exports that header without deployment topology._
- **G5** — A future command-state refactor could keep resolved deployment values local, which would make repeated rootCmd.Execute calls safer and reduce test setup burden.
  _cmd/root.go:360-361 and 481-484 explicitly document existing package-level mutation; cmd/root.go:2175-2182 adds the expanded assignments; cmd/dp_placement_test.go:423-433 and 487-505 execute run cases in fresh subprocesses._
- **G8** — Downstream tools that assumed exactly the unexpanded CLI --num-instances count will see more standard metric objects, but their JSON schema is unchanged.
  _cmd/root.go:2175-2182 expands numInstances; cmd/root.go:2232-2255 passes it to the standard DeploymentConfig; sim/cluster/cluster.go:323-329 creates standard instance_N instances; cmd/root.go:2444-2452 emits normal per-instance metrics; cmd/root.go:2459-2483 emits the standard cluster aggregate._
- **G9** — The stdout assertion is coupled to the public instance naming/output convention, but it also verifies the actual command wiring that a helper-only test could miss.
  _cmd/dp_placement_test.go:23-117 unit-tests planDPPlacement; cmd/dp_placement_test.go:119-159 unit-tests applyDPPlacement; cmd/dp_placement_test.go:436-459 executes the real command and observes exactly 2 and 4 instance outputs; sim/cluster/cluster.go:20-23 keeps instances private; docs/contributing/standards/principles.md:48-65 prefers observable behavioral assertions over internal-field assertions._
- **G10** — Adding a table row for dp=3 would improve regression coverage of non-power-of-two placement, even though it would still be exactly divisible on the real auto-KV path.
  _cmd/dp_placement_test.go:124-142 tests exact division at dp=4; cmd/dp_placement_test.go:626-675 compares auto-derived dp=1 and dp=2 capacities; sim/latency/kv_capacity.go:356-375 establishes total=(integer per-rank blocks)×dp._
- **G11** — Native topology metadata may become useful when node-pool GPU accounting or EP-aware placement is implemented.
  _cmd/root.go:399-406 states the EngineCore/router equivalence; cmd/root.go:2161-2182 performs the CLI transformation; sim/cluster/deployment.go:12-18 already models a homogeneous cluster through NumInstances; sim/cluster/cluster.go:316-335 constructs and places each instance through the existing path; cmd/root.go:2232-2255 uses the unchanged DeploymentConfig._
- **G12** — Full replay support would naturally reuse the same placement-plan/config construction rather than infer replicas from request IDs.
  _cmd/replay.go:289-300 explains that a single-instance replay would diverge and fatal-exits; sim/workload/tracev2.go:17-67 has no DP topology or request-to-rank field; sim/cluster/cluster.go:422-427 creates the live routing policies from cluster state; docs/contributing/standards/invariants.md:175-181 identifies MoE DP placement as unsupported by replay._
- **G16** — The warning text clearly reports both the requested and effective limits; users wanting strict configuration enforcement must currently inspect diagnostics.
  _cmd/root.go:792-803 already caps maxModelLen against KV capacity with a warning; cmd/root.go:2183-2197 repeats that established cap after converting aggregate auto-KV to per-rank KV; sim/simulator.go:166-178 otherwise returns the KV-capacity error; docs/contributing/standards/principles.md:30-34 defines warnings as stderr diagnostics._

susiejojo (approved):
lgtm

Issue #1590, opened 2026-08-17
S4 · Fill hole H1 kv_tier_chain: N-tier chain, ref_cnt pinning, write-through cascade


sriumcp:
**Sub-issue of #1585.** Fills hole **H1 `kv_tier_chain`** — the central hole.
Blocked by #1589, #1587 and #1588. Verified against BLIS `52161669` and vLLM `63a9a5010a`. Every vLLM/BLIS source claim below carries a **verified git permalink** at those commits (re-read in a local checkout; line ranges confirmed to contain the quoted code).

This replaces the model inside `sim/kv/tiered.go`. Per #1585 §4 the existing box breaks no
architectural rule — its **behavior** is wrong — so this is a rewrite, not an extension.

## Start here: the shape of the thing

```mermaid
graph LR
  GPU["GPU<br/>fastest, smallest"] <--> CPU["CPU tier 0<br/>cache AND waiting room"]
  CPU <--> FS["disk tier 1"]
  CPU <--> OBJ["object store tier 2"]
  FS -. NEVER directly .-x GPU

  classDef gpu fill:#dbe7f8,stroke:#3b5f8f,stroke-width:3px,color:#0d1b2e
  classDef stage fill:#ffeccc,stroke:#b45309,stroke-width:3px,color:#3b2200
  classDef cold fill:#e8eaed,stroke:#6b7280,color:#111827
  class GPU gpu
  class CPU stage
  class FS,OBJ cold
```

Only tier 0 (CPU) talks to the GPU. A disk hit is therefore **two hops** and occupies a CPU slot
for the whole journey. That is why `cpu_bytes_to_use` is a **throughput** knob, not just a
capacity knob.

## The hole

```
hole kv_tier_chain {
  surface:
    // read/write = CPU<->secondary; load/store = CPU<->GPU.
    // vLLM aliases these (tiering/manager.py:109-115):
    //   prepare_read  = prepare_load     complete_read  = complete_load
    //   prepare_write = prepare_store    complete_write = complete_store
    Lookup(BlockKey, ReqCtx) -> HIT | HIT_PENDING | RETRY | MISS
    PrepareStore(keys, ReqCtx) -> StoreGrant | nil   // allocates, ref_cnt = -1
    PrepareLoad(keys, ReqCtx)  -> LoadGrant          // pins, ref_cnt++
    CompleteStore(keys, ok)                          // -1 -> 0, or remove+free
    CompleteLoad(keys)                               // ref_cnt--
[...]

PR #1592, opened 2026-08-18
feat(kv-offload): fill hole H2 kv_transfer_station (#1588)

### LINKED ISSUE #1588: S3 · Fill hole H2 kv_transfer_station: bounded priority queueing station
**Sub-issue of #1585.** Fills hole **H2 `kv_transfer_station`**. **Unblocked — startable today.**
Disjoint from #1589 and #1587, so all three can proceed in parallel.
Verified against BLIS `52161669` and vLLM `63a9a5010a`. Every vLLM/BLIS source claim below carries a **verified git permalink** at those commits (re-read in a local checkout; line ranges confirmed to contain the quoted code).

## Start here: what this box is

When vLLM reads KV blocks off disk (or writes them there), the work goes to a **fixed pool of
I/O threads**. This box models that pool. Getting its *shape* right matters more than getting its
bandwidth constant right.

## The hole

```
hole kv_transfer_station {
  surface:
    Submit(TransferJob) -> JobId | Rejected
    Poll(now) -> []JobId                       // completed since last poll
    ActiveJobs(TierIndex, Direction) -> int
  allow:
    import sim                      // clock, event scheduling
    import sim/internal/util
  contract:
    BC-S1 at most n_read + n_write jobs in service at once, per tier
    BC-S2 read-priority servers take reads first, write-priority servers take
          writes first; a server idles only when BOTH queues are empty
    BC-S3 service_time = base + bytes/bandwidth, per (tier, direction)
    BC-S4 completion order is a deterministic function of
          (submit tick, tier, direction, job id) — never of wall clock
    BC-S5 a job spans >= 1 block; JOBS are the unit of service, not blocks
  evidence:
    property_test    (BC-S1, BC-S3, BC-S4)
    metamorphic_test (BC-S2)
    fuzz             (Submit under adversarial sizes/ordering)
    differential_test(BC-S2 vs tiering/fs/thread_pool.py)
    benchmark        (P: O(1) amortized Submit/Poll)
}
```

## BC-S2 — the correction that flips a sign

`vllm/v1/kv_offload/tiering/fs/thread_pool.py:62`, from its own docstring:

> Two queues (load, store) and two sets of threads: **Load-priority threads** drain the load
> queue first, then the store queue. **Store-priority threads** do the reverse.

So it is **priority-ordered with fallback**, not symmetric work-stealing. Reads keep
`n_read_threads` dedicated first-choice servers and only spill into write threads when those go
idle.

```mermaid
graph LR
  LQ["load queue<br/>(disk reads)"] --> RT["n_read_threads<br/>read-priority"]
  SQ["store queue<br/>(disk writes)"] --> WT["n_write_threads<br/>write-priority"]
  LQ -. fallback only .-> WT
  SQ -. fallback only .-> RT

  classDef q fill:#dbe7f8,stroke:#3b5f8f,color:#0d1b2e
  classDef srv fill:#d6f0dd,stroke:#1a7f37,stroke-width:3px,color:#0b2b16
  class LQ,SQ q
  class RT,WT srv
```

Two consequences:

- **Right abstraction:** a **bounded queue with N servers** (N = `n_read` + `n_write`) and
  priority classes. Disks and object stores saturate on **concurrency**, not on divided bandwidth.
- **Wrong abstraction:** fair-share bandwidth — the shape of BLIS's existing
  `PDTransferContention` ([`sim/cluster/pd_events.go:224-255`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/cluster/pd_events.go#L224-L255)). **Do not reuse it here.**

The two models disagree in a way you can test:

| Add more write threads, hold everything else | This model predicts | Fair-share predicts |
|---|---|---|
| read-path latency | **non-increasing** (more fallback capacity) | **worse** (bandwidth split more ways) |

Opposite signs. That is why BC-S2's metamorphic relation **discriminates** rather than confirms.

## BC-S4 — where fidelity and determinism genuinely conflict

vLLM's pool is **truly nondeterministic**: real OS threads, GIL scheduling. BLIS must produce
**byte-identical output for a fixed seed** (INV-6). Both cannot hold literally.

**Resolution, decided here rather than discovered mid-implementation:**

> Match the **distribution**; fix the **sequence**.

Service order is a deterministic function of `(submit tick, tier, direction, job id)`, with any
random component drawn from the seeded partitioned RNG. This is the discipline the engine already
uses for event ties — `(timestamp, priority, seqID)`.

## BC-S5 — jobs, not blocks

`_flush_pending_promotions()` submits **one batched job per (tier, request)**, and vLLM's
`active_promotion_jobs` gauge counts **jobs**. Modeling one event per block would be **slower and
less faithful**. The station serves jobs; blocks ride inside them.

## BC-S3 — read and write bandwidth must be separate

Measured NVMe KV-offload work shows read/write asymmetry far too large for a single bandwidth
number (see the prior-art survey in #1569). Per-`(tier, direction)` is **required**, not optional.

## What to reuse

`eventflow . ./sim/...` shows the engine already has this pattern **twice**:

```
KVTransferCompletedEvent   -. via helper .-> StepEvent
AdapterLoadCompletionEvent -. via helper .-> StepEvent
```

`AdapterLoadCompletionEvent` (from the LoRA work) is the closest analog. **Reuse the event shape.
Do not reuse the fair-share bandwidth math.**

## Evidence

|---|---|---|
| **BC-S2** | **metamorphic** | raise `n_write_threads`, hold the rest -> read latency **non-increasing**. A fair-share model fails with the opposite sign. |
| **BC-S2** | **differential** | drain order vs `thread_pool.py`'s worker loop on a scripted job sequence |
| BC-S1 | property | generated arrival storms: in-service count never exceeds N, per tier |
| BC-S4 | property | same seed -> identical completion sequence, across repeated runs **and** across `GOMAXPROCS` values |
| BC-S3 | property | service time linear in bytes, per direction; read != write representable |
| — | **fuzz** | `Submit` with 0-byte, 1-block and very large jobs; interleaved directions; full queues |
| P | **benchmark** | amortized O(1) `Submit`/`Poll`; no per-block allocation in steady state |

## Done when

- BC-S1 through BC-S5 hold with the evidence above.
- **The BC-S2 metamorphic test provably fails against a fair-share implementation** — write that
  implementation once as a throwaway and confirm the test rejects it. An assertion that cannot
  fail is not evidence.
- `archon-go delta`: new box plus declared arrows only; `reflexion` still 0 divergent.

namasl:
## Summary

Fills hole **H2 `kv_transfer_station`** of the multi-tier KV-offload design (tracking issue #1585). Adds `sim/kvtransfer`: a **bounded priority queueing station with N servers per tier** that models vLLM's KV-offload I/O thread pool (`DualQueueThreadPool`, `vllm/v1/kv_offload/tiering/fs/thread_pool.py`).

The right abstraction is a **queue with N = NRead+NWrite servers and priority classes** — disks/object stores saturate on *concurrency*, not divided bandwidth. This is deliberately **NOT** the fair-share bandwidth shape of `sim/cluster`'s `PDTransferContention`; the two predict opposite signs for "add write threads, watch read latency" (see BC-S2), which is exactly what makes the difference testable.

The station is a **standalone, deterministic, leaf module** (stdlib-only). It is **not wired into the simulator** — H1 (`kv_tier_chain`, #1590) is the consumer. Until then it has no effect on any run, so output stays byte-identical (INV-6).

## Issue

Closes #1588

## Behavioral contracts & evidence

| Contract | What it guarantees | Evidence |
|---|---|---|
| **BC-S1** | ≤ `NRead+NWrite` jobs in service per tier | property (arrival storm) + fuzz invariant |
| **BC-S2** | read-priority servers take reads first, write-priority writes first; either falls back; a server idles only when **both** queues empty | **differential** test vs `thread_pool.py`'s worker loop (priority + fallback) **and** a **metamorphic** test (read latency non-increasing as `NWrite` grows) that **provably fails against a throwaway fair-share model** — proving the relation discriminates rather than merely confirms |
| **BC-S3** | `service_time = base + ⌊bytes/bandwidth⌋`, per `(tier, direction)`, read ≠ write representable | formula table + linearity + completion-time test (+ int64 overflow guard) |
| **BC-S4** | completion order is a total deterministic function of `(completeAt, submitTick, tier, direction, jobID)` | determinism across repeated runs **and** across `GOMAXPROCS`; fuzz determinism. No RNG, no wall clock. |
| **BC-S5** | one `TransferJob` = one unit of service over many blocks; never split (`enqueue_store(job_id, 1, [task])`, `fs/manager.py:229`) | differential test (big multi-block job occupies exactly one server for its whole duration) |
| **BC-S6** | match the distribution, fix the sequence *(attested)* | `doc.go` attestation |
| **P** | `Submit`/`Poll` amortized O(1), 1 alloc/job | `BenchmarkSubmitPoll` — flat ns/op across 50k→5M jobs, 1 alloc/op |

The **"done when"** requirement — *"the BC-S2 metamorphic test provably fails against a fair-share implementation; write that implementation once as a throwaway and confirm the test rejects it"* — is `TestBCS2_Metamorphic_ReadLatencyNonIncreasing`: it runs the same manipulation against the real station (non-increasing ✓) and the `fairShareStation` throwaway (opposite sign, would fail the relation), asserting the fair-share leg exhibits the wrong sign so the test has discriminating power.

## Fuzzing

`FuzzStation` drives adversarial-but-legal scenarios (0-byte, 1-block, ~GiB jobs; interleaved directions; bounded queues forcing rejection) and asserts BC-S1, conservation (every accepted job completes exactly once), and BC-S4 determinism on every input. 7.6M+ executions, zero failures.

## Cross-Path Parity

| `blis run` | no | N/A | station unused until #1590 (H1) |
| `blis replay` | no | N/A | — |
| `blis observe` | no | N/A | — |

## Invariants Affected

- **INV-3** (clock monotonicity): the station never moves its clock backward; a submit is guarded so a job never enters service in the past.
- **INV-6** (determinism): pure deterministic model — no RNG, no map iteration for ordered output; and nothing imports the package yet, so stdout is byte-identical to a pre-feature build.
- **INV-13** (run/replay parity): N/A this PR (not wired into run/replay/observe).

## Notes / deviations

- **`Submit → Rejected`**: vLLM's queues are unbounded `deque()`; the station mirrors this with `MaxQueueDepth` defaulting to `0` (unbounded). A positive bound is a BLIS-extra that enables the `Rejected` path (exercised by the fuzz "full queues" obligation).
- **CI**: `./sim/kvtransfer` was added to the test matrix's "core packages" entry — `./sim/kv` is non-recursive and would not cover the new package.
- Docs (CLAUDE.md / project-structure) land with the wiring PR (#1590), where the module first affects observable behavior.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./...` — all 12 packages pass
- [x] `golangci-lint run ./sim/kvtransfer/...` — 0 issues
- [x] `FuzzStation` 45s (7.6M execs, 0 failures)
- [x] `BenchmarkSubmitPoll` — O(1) amortized, 1 alloc/job

---


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

Pushed a follow-up (3c18f399) adding the one test suggested in the prior review — `Poll` with a backward tick as a safe no-op (INV-3). No production code changed. Please re-confirm.

namasl:

sriumcp (dismissed):
lgtm

namasl:

sriumcp (approved):

PR #1593, opened 2026-08-18
fix(kv): consult offload tiers on every request, not only on GPU alloc failure (#1586)

### LINKED ISSUE #1586: S0 · D3: consult offload tiers on every request, not only on GPU alloc failure
**Sub-issue of #1585.** Not a hole fill — a **bug fix** on the existing code. Verified against
BLIS `52161669` and vLLM `63a9a5010a`. Every vLLM/BLIS source claim below carries a **verified git permalink** at those commits (re-read in a local checkout; line ranges confirmed to contain the quoted code).

## Start here: what is wrong

BLIS checks the CPU cache tier **only after GPU allocation has already failed**.
[`sim/kv/tiered.go:195-201`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/tiered.go#L195-L201):

```go
func (t *TieredKVCache) AllocateKVBlocks(req *sim.Request, startIndex, endIndex int64, cachedBlocks []int64) bool {
	ok := t.gpu.AllocateKVBlocks(req, startIndex, endIndex, cachedBlocks)
	if ok {
		return true          // <-- CPU tier never consulted on this path
	}
	// GPU allocation failed — try targeted CPU reload for this request's prefix.
	reloaded := t.reloadPrefixFromCPU(req.FullInputTokens())
```

vLLM does the opposite. Its offload connector is a **prefix-cache extension consulted on every
request**, before any allocation is attempted — `get_num_new_matched_tokens` is called on the
normal waiting-queue path at `vllm/v1/core/sched/scheduler.py:820-833`, with no reference to GPU
pressure.

## Why this matters more than it looks

The error is **load-shaped**, not a constant offset:

```mermaid
graph LR
  A["GPU has free space<br/>(light load)"] --> B["BLIS: skips CPU tier<br/>reports MISS"]
  C["GPU is full<br/>(heavy load)"] --> D["BLIS: checks CPU tier<br/>reports HIT"]
  B --> E["hit rate too LOW<br/>error is MAXIMAL"]
  D --> F["hit rate about right<br/>error near zero"]

  classDef bad fill:#f8d7da,stroke:#a4272f,stroke-width:3px,color:#3d0d11
  classDef good fill:#d6f0dd,stroke:#1a7f37,color:#0b2b16
  class E bad
  class F good
  class A,B,C,D plain
```

So the bias **varies with load** rather than cancelling out. Any latency coefficient fitted while
this bug is live absorbs a load-dependent error permanently. **That is why this lands before any
measurement or modelling work.**

## Why it is first

Unblocked. Needs no unreleased vLLM metrics, no new events, no new boxes, and no ARCHON changes.

## Scope

Interior change to `sim/kv.TieredKVCache`. **No** change to the `sim.KVStore` surface (still 12
entities), **no** new package import. `archon-go delta` must therefore stay clean at package
altitude, and `archon-go reflexion` must stay at **0 divergent edges**.

`sim/kv`'s allowed imports remain exactly `sim`, `sim/internal/hash`, `sim/internal/util`
(`archon-go contract`).

## Contract

Stated only in terms of the existing public surface (the paper requires contracts to mention
nothing but the surface):

| ID | Guarantee |
|---|---|
| **BC-D3.1** | `GetCachedBlocks(tokens)` reflects blocks resident in **any** tier, not only GPU |
| **BC-D3.2** | `CacheHitRate()` never decreases when a matching block is added to a lower tier |
| **BC-D3.3** | the tier lookup happens for **every** request, whatever the GPU occupancy |
| **BC-D3.4** | INV-4 holds: `allocated + free = total`, per tier |
| **BC-D3.5** | INV-6 holds: byte-identical stdout for a fixed seed |

## Evidence

|---|---|---|
| **BC-D3.3** | **metamorphic** | Sweep GPU capacity upward, tiers and workload fixed. Hit rate must be **flat**. Under today's code it **rises** — that is the bug's signature, and this is the discriminating test. |
| BC-D3.2 | property | generated prefix families; hit rate monotone in residency |
| BC-D3.4 | property | conservation after every allocate/release |
| BC-D3.5 | golden | golden dataset **re-baselined** — see below |
| P | benchmark | lookup now runs per request; prove no material regression |

## Two things that will surprise you

**1. Golden tests will change, and that is correct.** Hit rate rises, so simulated latencies
move. The re-baseline **must** ship together with the BC-D3.3 metamorphic test. A golden update
alone answers *"did the output change?"* but not *"is it now right?"* — see CLAUDE.md BDD/TDD
rule 4 and the "test laws, not just values" requirement.

**2. This is now a hot path.** `reloadPrefixFromCPU` currently runs only on the alloc-failure
path; afterwards it runs on **every** request. The benchmark obligation is real: if it regresses
materially, this waits for the key-interning work in #1589 rather than shipping a slowdown.

## Done when

- BC-D3.1 through BC-D3.5 hold, with the evidence above.
- `archon-go delta` shows no package-altitude change; `reflexion` still 0 divergent.
- `archon-go evidence` still reports `gapKind: none` for `sim.KVStore`.
- PR reviewed by commenting **`/archon-pr-review`** (`.github/workflows/archon.yml`).

namasl:
## Summary

`TieredKVCache` consulted the CPU offload tier **only after GPU allocation failed**
([`sim/kv/tiered.go:195`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/tiered.go#L195-L201)).
Under light GPU load (space available) a prefix resident on CPU-but-evicted-from-GPU was
silently recomputed and counted as a **MISS**. This made cache-hit accounting **load-shaped**:
sweeping GPU capacity upward changed the hit rate, so any latency coefficient fitted across a
capacity sweep absorbed a load-dependent bias.

This PR moves the CPU-tier consultation to the **front** of `AllocateKVBlocks` so it runs on
**every** request, before GPU allocation — mirroring vLLM's offload connector
(`get_num_new_matched_tokens` on the normal waiting-queue path, with no reference to GPU
pressure). The existing reload + commit machinery is reused verbatim; only the call site moves.

**Closes #1586** (sub-issue of the multi-tier KV offload epic #1585 — the prerequisite bug fix).

## What changed

- `AllocateKVBlocks` consults the CPU tier on every request (before GPU allocation) instead of
  only on the alloc-failure path.
- `reloadPrefixFromCPU` gained a `startBlock`/`prevHash` parameter so the scan resumes just past
  the prefix the request already owns/matched (hot-path mitigation, evidence P) and never reloads
  a block position the request already holds — in particular a partially-filled last block, whose
  empty hash would otherwise place a duplicate on the free list that the running-request path never
  claims but the GPU pre-check still budgets for.
- `cpuMissCount` keeps its pre-#1586 semantic exactly: incremented only when the CPU tier cannot
  help **and** the GPU allocation fails.

## Behavioral contracts (GIVEN / WHEN / THEN)

| ID | Guarantee | Test |
|---|---|---|
| **BC-D3.1** | GIVEN a prefix resident only on CPU, WHEN a request with that prefix is allocated (even with free GPU space), THEN the tier is consulted and `GetCachedBlocks` reflects the materialized prefix | `TestTieredKVCache_GetCachedBlocks_ReflectsCPUTierAfterAllocate` |
| **BC-D3.2** | `CacheHitRate` does not decrease when a matching block is resident in the CPU tier | `TestTieredKVCache_HitRate_MonotoneInCPUResidency` |
| **BC-D3.3** (discriminating) | With tier + workload fixed, `CacheHitRate` is **flat** as GPU capacity is swept (it *rose* under the bug) | `TestTieredKVCache_HitRate_FlatAcrossGPUCapacity` |
| **BC-D3.4** | INV-4 (`allocated + free = total`) holds across light-load reloads | existing conservation tests + light-load path |
| **BC-D3.5** | INV-6: byte-identical stdout | non-tiered runs (exact) + committed golden datasets (unchanged) |

`BC-D3.3` is the law-not-value test required by the BDD/TDD rules: it fails on the unmodified
code (hit rate `[0, 0, 0.7, 0.7]` across capacities) and passes after the fix (flat).

## Why `GetCachedBlocks` stays a pure GPU query (BC-D3.1 interpretation)

`GetCachedBlocks` is called **live** on the routing/scoring path
(`InstanceSimulator.GetCachedBlockCount` → `KVCache.GetCachedBlocks`) in Immediate snapshot mode
(INV-7). Making it reload CPU→GPU would mutate KV state on non-selected candidate instances during
scoring and pollute transfer latency — a worse bug than the one fixed. So `GetCachedBlocks` and
`SnapshotCachedBlocksFn` are left untouched (pure, GPU-only), and BC-D3.1 is satisfied at the
**allocation boundary**: `AllocateKVBlocks` consults all tiers on every request and materializes
CPU-resident prefix blocks onto the GPU, after which `GetCachedBlocks` reflects them.

## Evidence

**Unit (metamorphic + property):** the three BC-D3 tests above; the discriminating BC-D3.3 test was
confirmed to fail pre-fix and pass post-fix.

**End-to-end** (roofline, `meta-llama/llama-3.1-8b-instruct`, shared-prefix workload, 40 unique
system prompts, low arrival rate so prefixes evict to CPU-only with free space at re-admission),
sweeping `--total-kv-blocks`:

| `--total-kv-blocks` | OLD `Cache Hit Rate` | this PR |
|---|---|---|
| 1200 | 0.8003 | 0.8050 |
| 8000 | 0.8047 | 0.8047 |

OLD **rises** with capacity (undercounts CPU-resident prefixes at small capacity); this PR is
**flat** — the load-shaped bias is removed. At large capacity both agree, because prefixes stay on
GPU and the buggy condition never fires.

**Golden datasets are byte-identical** (`go test ./sim/cluster -run Golden` passes unmodified). The
committed roofline tiered cases use large GPUs relative to their working set, so they never evict a
prefix to CPU-only and never hit the buggy condition. INV-6 therefore holds even for tiered configs
here; the metamorphic + end-to-end evidence above stands in for a golden re-baseline. (The issue
anticipated a golden change; empirically the committed workloads don't trigger the bug.)

**Performance (evidence P — reload now runs per request):**

```
BenchmarkTieredAllocate_GPUCachedPrefix-24        58631 ns/op   10281 B/op   224 allocs/op
BenchmarkSingleTierAllocate_GPUCachedPrefix-24    53287 ns/op   10074 B/op   220 allocs/op
```

~10% on the KV-allocate micro-op for a 33-block request (negligible end-to-end, where allocation is
a small fraction of step time). The `startBlock` mitigation bounds this to ~one hash + one CPU
lookup for a GPU-cached prefix; without it, reload would re-hash the entire prefix on every request.

## Scope

Interior change to `sim/kv.TieredKVCache`. `sim.KVStore` surface unchanged (12 methods); no new
`sim/kv` import; `GetCachedBlocks`/`SnapshotCachedBlocksFn` untouched.

## Verification

- `go build ./...` — clean
- `go test ./...` — all 11 packages pass
- `golangci-lint run ./...` (v2.9.0) — 0 issues

### DIFF
--- a/sim/kv/tiered.go
+++ b/sim/kv/tiered.go
@@ -193,14 +193,34 @@ func NewTieredKVCache(gpu *KVCacheState, cpuBlocks int64, threshold, bandwidth f
 }
 
 func (t *TieredKVCache) AllocateKVBlocks(req *sim.Request, startIndex, endIndex int64, cachedBlocks []int64) bool {
-	ok := t.gpu.AllocateKVBlocks(req, startIndex, endIndex, cachedBlocks)
-	if ok {
-		return true
+	// BC-D3.3 (#1586): consult the CPU offload tier on EVERY request, before GPU
+	// allocation — mirroring vLLM's offload connector (get_num_new_matched_tokens on
+	// the normal waiting-queue path, with no reference to GPU pressure). Previously the
+	// reload ran only after GPU allocation failed, so under light GPU load a prefix
+	// resident on CPU-but-evicted-from-GPU was silently recomputed and counted as a
+	// miss — making cache-hit accounting load-shaped rather than load-independent.
+	//
+	// Resume the prefix hash chain just past the GPU-resident prefix the caller already
+	// matched (cachedBlocks), using its last block's hash as the chain seed. This keeps
+	// the common fully-GPU-cached case at one hash + one CPU lookup instead of
+	// re-hashing the whole prefix from block 0 (hot-path mitigation, evidence P).
+	reloadStartBlock := int64(len(cachedBlocks))
+	reloadPrevHash := ""
+	if reloadStartBlock > 0 {
+		reloadPrevHash = t.gpu.Blocks[cachedBlocks[reloadStartBlock-1]].Hash
 	}
-	// GPU allocation failed — try targeted CPU reload for this request's prefix.
-	reloaded := t.reloadPrefixFromCPU(req.FullInputTokens())
-	if reloaded {
-		// Re-compute cached blocks now that CPU content is back on GPU
+	// A continuing (running) request already owns its prefix blocks (cachedBlocks is empty
+	// on that path); resume past what it owns so we never reload a block position it
+	// already holds — in particular its partially-filled last block, whose empty hash
+	// would also break the chain seed. Reloading such a position would place a duplicate,
+	// full copy on the free list that the running-request path never claims but the GPU
+	// pre-check still budgets for, spuriously failing the allocation.
+	if owned, ok := t.gpu.RequestMap[req.ID]; ok && int64(len(owned)) > reloadStartBlock {
+		reloadStartBlock = int64(len(owned))
+		reloadPrevHash = t.gpu.Blocks[owned[len(owned)-1]].Hash
+	}
+	if t.reloadPrefixFromCPU(req.FullInputTokens(), reloadStartBlock, reloadPrevHash) {
+		// Re-compute cached blocks now that CPU content is back on the GPU free list.
 		newCached := t.gpu.GetCachedBlocks(req.FullInputTokens())
 		newStart := int64(len(newCached)) * t.gpu.BlockSize()
 		if newStart > startIndex {
@@ -250,11 +270,18 @@ func (t *TieredKVCache) AllocateKVBlocks(req *sim.Request, startIndex, endIndex 
 			}
 			return t.gpu.AllocateKVBlocks(req, newStart, endIndex, newCached)
 		}
-		// No new cache hits — retry with original params (reload freed up space)
+		// Reload produced no prefix hit beyond startIndex — allocate with the
+		// caller's original params (any space freed by the reload still helps).
 		return t.gpu.AllocateKVBlocks(req, startIndex, endIndex, cachedBlocks)
 	}
-	t.cpuMissCount++
-	return false
+	// No CPU-resident continuation: allocate on GPU exactly as the single-tier path.
+	// cpuMissCount++ preserves the pre-#1586 semantic — incremented only when the CPU
+	// tier could not help AND the GPU allocation fails.
+	ok := t.gpu.AllocateKVBlocks(req, startIndex, endIndex, cachedBlocks)
+	if !ok {
+		t.cpuMissCount++
+	}
+	return ok
 }
 
 // reloadPrefixFromCPU attempts to reload prefix-matching blocks from CPU to GPU.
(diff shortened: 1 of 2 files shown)


namasl:
@claude /blis-pr-review

sriumcp (approved):
lgtm

PR #1594, opened 2026-08-18
feat(kv-offload): fill hole H5 — capture vLLM's KV-offload config surface (#1587)

### LINKED ISSUE #1587: S2 · Fill hole H5 kv_offload_config: capture vLLM's offload CLI/config surface
**Sub-issue of #1585.** Fills hole **H5 `kv_offload_config`**. **Unblocked — startable today.**
Disjoint from #1589 and #1588, so all three can proceed in parallel.
Verified against BLIS `52161669` and vLLM `63a9a5010a` — every vLLM claim below carries a **verified
git permalink** at that commit (re-read in a local checkout, line ranges confirmed).

> **Corrected in this revision:** the `store_threshold` row and trap 3. It is **not** a knob on the
> multi-tier path — `TieringOffloadingSpec` rejects `>=2` outright — so it has been removed from the
> YAML schema and moved to the reject-loudly branch of BC-G1. Recorded as **A4** in #1585 §12.

## Start here: why config is a fidelity requirement

vLLM's offload behavior **changes with its flags**. Same code, different physics. BLIS cannot
claim to model a mechanism it cannot be *configured into*, so the knobs are part of the fidelity
surface — not documentation.

ARCHON already treats this as first-class: `config` is one of its five typed arrow kinds, and
`archon-go render` emits `-. config .->` arrows for flags and env vars. These knobs belong **in
the graph**.

## The hole

```
hole kv_offload_config {
  surface:
    TierChainSpec -> ordered []TierSpec
    PrimarySpec   -> capacity, policy, storeThreshold
    ChunkSpec     -> tokensPerHash, blocksPerChunk, tokensPerBlock
    OffloadScope  -> promptOnly | promptAndDecode
  allow:
    import sim                      // KVCacheConfig sub-config only
  contract:
    BC-G1 every knob below is representable; anything vLLM accepts either maps
          to a BLIS config or fails LOUDLY at startup — never silently ignored
    BC-G2 defaults equal vLLM's defaults, knob for knob
    BC-G3 validated at construction; no tier is half-configured
    BC-G4 read-only after construction (never fitted)
    BC-G5 byte-identical output when offload is disabled
    BC-G6 the resolved config is recorded in the trace header; replay
          rejects a config it cannot reproduce (INV-13, never silent)
  evidence:
    property_test    (BC-G3)
    metamorphic_test (BC-G5)
    differential_test(BC-G2)
    static_analysis  (BC-G4)
    fuzz             (BC-G3)
    replay_trace     (BC-G6)
}
arrow kv_tier_chain -> kv_offload_config : config
```

## The knobs, read from vLLM source

| vLLM knob | Default | Source | What it changes |
|---|---|---|---|
| `cpu_bytes_to_use` | *required* | [`tiering/spec.py:10`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L10) | CPU capacity -> lock pressure -> recompute path |
| `block_size` | GPU block size | [`tiering/spec.py:11`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L11) | bytes per block -> transfer time |
| `eviction_policy` | `"lru"` | [`tiering/spec.py:12-15`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L12-L15) | `lru` / `arc` — which block is dropped |
| `secondary_tiers[]` | `[]` | [`tiering/spec.py:19`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L19) | tier count + order -> lookup order, fan-out width |
| `secondary_tiers[].type` | *required* | [`tiering/spec.py:21`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L21) | `fs` / `obj` / `p2p` / `example` |
| `offload_prompt_only` | **`true`** | [`base.py:597-598`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/base.py#L597-L598); enforced [`scheduler.py:595-596`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py#L595-L596) | decode KV **not** offloaded by default (truncate-then-floor-divide; see #1584) |
| `self_describing_kv_events` | `false` | [`base.py:588-590`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/base.py#L588-L590) | event payload (observability only) |
| `tokens_per_hash` | — | [`config.py:30`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/config.py#L30) | key stride (#1589 BC-K4) |
| `blocks_per_chunk` | — | [`config.py:32`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/config.py#L32) | job batching width (#1588 BC-S5) |
| `store_threshold` | `1` (single-tier ctor); **`>=2` REJECTED for tiering** | [`cpu/manager.py:48`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/cpu/manager.py#L48), [`tiering/spec.py:289-293`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L289-L293) | **NOT a knob on this path** — fixed at 1, reject `>=2` loudly (see corrected trap 3) |
| `n_read_threads` | **16** | [`fs/manager.py:115`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L115) | read servers (#1588 BC-S1/S2) |
| `n_write_threads` | **16** | [`fs/manager.py:116`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L116) | write servers (#1588 BC-S1/S2) |
| `root_dir` | *required* | [`fs/manager.py:114`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L114) | which device -> bandwidth/latency class |
| `locality` | `None` | [`fs/manager.py:118`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L118) | `LOCAL` / `REMOTE` -> latency class |
| `enable_kv_events` | `false` | [`fs/manager.py:117`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L117) | per-tier event emission |
| O_DIRECT vs buffered | **probed at runtime** | [`fs/manager.py:185-195`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L185-L195), [`fs/io.py:41`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/io.py#L41) | page cache on/off — very different physics |

## How the knobs reach BLIS

**This is a design decision, not an implementation detail, so it is settled here.** vLLM takes
these knobs as a nested JSON blob inside `--kv-transfer-config`. BLIS must not mirror that
literally, and must not turn 16 knobs into 16 flags.

### The problem with flags

`blis run` already registers **112 flags** (21 of them on `runCmd` directly). Adding sixteen
more — several of them nested per-tier, and `secondary_tiers[]` being a *list* of heterogeneous
objects — is not expressible as flat flags at all. A repeated
`--secondary-tier type=fs,root_dir=/mnt,n_read_threads=16` mini-language would be a parser we
maintain by hand, which R10 (strict parsing) exists to prevent.

### The pattern BLIS already uses three times

```mermaid
graph TB
  subgraph T1["tier 1 — a few scalar flags"]
    F["--kv-cpu-blocks<br/>--kv-offload-config"]
  end
  subgraph T2["tier 2 — strict-YAML file, subsystem inert when absent"]
    Y["kv_offload: block<br/>full tier chain, per-tier params"]
  end
  subgraph T3["tier 3 — defaults.yaml, shipped constants"]
    D["per-device bandwidth/latency<br/>classes (nvme, sata, s3)"]
  end

  F -->|"overrides"| M["resolved KVOffloadConfig<br/>validated once, read-only"]
  Y -->|"overrides"| M
  D -->|"seeds"| M

  classDef pick fill:#d6f0dd,stroke:#1a7f37,stroke-width:3px,color:#0b2b16
  class M pick
  class F,Y,D plain
```

**Precedent, in the repo today — follow it rather than inventing:**

| Mechanism | Existing example | Use here for |
|---|---|---|
| Strict-YAML file, one top-level block, subsystem **inert when absent** | `--lora-config` (`cmd/root.go:1225`), `--saturation-config` (`cmd/saturation.go:35`) | **the tier chain and all per-tier params** |
| A *few* scalar flags for the common case, file for the rest | `--lora-load-base-latency-us` alongside `--lora-config` | `--kv-cpu-blocks` (already exists) |
| `defaults.yaml`, strict-parsed, shipped with the binary | `lora:` block (`defaults.yaml:192`), `trained_physics_coefficients:` | **device bandwidth/latency classes** |

`--saturation-config` is the closest structural match: optional per-detector blocks, `*T` pointer
fields so "absent" is distinguishable from "zero" (R9), and `decoder.KnownFields(true)` so an
unknown or misplaced key is a **hard error** (`sim/saturation/config.go:72`). Copy that shape.

### The decision

**One new flag: `--kv-offload-config <path>`.** A strict-YAML file with a single top-level
`kv_offload:` block. Absent ⇒ the offload subsystem is **inert** and output is byte-identical to
today (BC-G5). That is exactly `--lora-config`'s contract.

```yaml
kv_offload:
  cpu_bytes_to_use: 17179869184     # required when the block is present
  block_size: 16
  eviction_policy: lru              # lru | arc
  offload_prompt_only: true         # vLLM default — stated, not implied
  # store_threshold is deliberately ABSENT: TieringOffloadingSpec rejects >=2,
  # so it is fixed at 1 on this path. Supplying it must fail loudly (BC-G1).
  tokens_per_hash: 16
  blocks_per_chunk: 4
  secondary_tiers:
    - type: fs
      root_dir: /mnt/kv-cache
      n_read_threads: 16
      n_write_threads: 16
      locality: LOCAL
      direct_io: true               # BLIS makes vLLM's runtime probe EXPLICIT
      device_class: nvme_gen4       # resolves bandwidth/latency from defaults.yaml
```

**No new scalar flags.** `--kv-cpu-blocks` already exists and keeps working as the single-tier
knob. Everything multi-tier is file-only, because it is nested and list-valued.

**Three things to notice, because they are the whole design:**

1. **`direct_io` is explicit here although vLLM probes for it.** vLLM discovers O_DIRECT support
   at startup and silently falls back to buffered I/O. A simulator cannot probe the operator's
   disk, and the two modes have substantially different physics, so BLIS makes it a declared
   field. **Faithful simulation of an implicit runtime property requires making it explicit** —
   and it must be *recorded* in the run's config snapshot, because a trace taken under buffered
   I/O is not comparable to one taken under direct I/O.
2. **`device_class` is BLIS-only and is not a vLLM knob.** vLLM gets its bandwidth from the actual
   device; BLIS needs a number. This resolves against a new `kv_offload_devices:` block in
   `defaults.yaml` (alongside `lora:` and `trained_physics_coefficients:`), so per-device physics
   ships with the binary and is not retyped per run. An explicit `read_bandwidth` /
   `write_bandwidth` / `base_latency` triple overrides the class — same override shape as
   `--lora-load-base-latency-us` over the `lora:` block.
3. **Per-`(tier, direction)` bandwidth is required, not optional** (#1588 BC-S3). Hence separate
   `read_bandwidth` and `write_bandwidth`, never a single number.

### Is this consistent with "what is fixed"?

Checked against #1585 §5, because a new flag is a surface change and the fixed list must not be
quietly violated:

| Claim in §5 | Does this delivery design honor it? |
|---|---|
| `sim.KVStore` keeps 12 entities | **Yes.** Config is constructor input, not an interface method. |
| `sim/kv` acquires no import outside `sim`, `internal/hash`, `internal/util` | **Yes.** Parsing lives in `cmd/`; `sim/kv` receives an already-validated struct. |
| Layering stays 0 divergent | **Yes.** `cmd → kernel` is a downward edge; no new upward edge. |
| Config grouped by module (R16) | **Yes.** A `KVOffloadConfig` sub-config of `KVCacheConfig`; factories take the narrowest type. |
| Canonical constructor (R4) | **Requires care.** `NewKVCacheConfig` (`sim/config.go:21`) is positional with 6 params and is the single construction site. Adding a 7th positional param is hostile; grep **all** call sites first and prefer a nested `KVOffloadConfig` value over more positional args. |
| Never fitted (BC-G4) | **Yes**, and the file-based route reinforces it: a config file is operator input, structurally separate from `trained_physics_coefficients:`. |

**One genuine addition to §5's "what changes" list:** exactly **one** new CLI flag
(`--kv-offload-config`), one new `defaults.yaml` block (`kv_offload_devices:`), and one new
sub-config type. The flag count goes 112 → **113**, not 112 → 128.

### INV-13 (run/replay parity) — an obligation this creates

The resolved offload config must be **recorded in the exported trace header**, and `blis replay`
must reject a trace whose recorded config it cannot reproduce (`logrus.Fatalf`, never silent
degradation). Otherwise a trace captured with a 3-tier chain silently replays as single-tier and
the parity invariant is violated without any error. **This is a new BC-G6.**

---

## Three traps

**1. `offload_prompt_only` defaults to `true`.** Decode KV is **not** offloaded unless you ask for
it. Any test or experiment that assumes generated tokens reach the tiers is silently exercising a
**non-default** configuration. Getting this default wrong changes what the whole simulation is
about.

**2. O_DIRECT is not a flag.** vLLM *probes* for it and silently falls back to buffered I/O with a
startup warning:

```mermaid
graph LR
  S["vLLM starts FS tier"] --> P{"O_DIRECT<br/>supported here?"}
  P -->|yes| D["direct I/O<br/>page cache BYPASSED"]
  P -->|no| B["buffered I/O<br/>page cache ACTIVE<br/>(warning logged)"]
  D --> X["very different<br/>storage physics"]
  B --> X

  classDef q fill:#ffeccc,stroke:#b45309,stroke-width:3px,color:#3b2200
  classDef out fill:#f8d7da,stroke:#a4272f,stroke-width:3px,color:#3d0d11
  class P q
  class X out
  class S,D,B plain
```

BLIS must model this as an **explicit config axis** even though vLLM discovers it at runtime.
Faithful simulation sometimes requires making an implicit property explicit.

**3. `store_threshold` is NOT a knob on the multi-tier path — CORRECTED (see #1585 §12 A4).**
An earlier version of this issue claimed `>=2` gates stores on reuse and therefore drives write
volume -> lock pressure -> the #1590 recompute path. **That causal chain does not exist on this
spec.** The `>=2` machinery is real but **single-tier only**: `CPUOffloadingManager` allocates a
counting map only when `store_threshold >= 2` ([`cpu/manager.py:72`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/cpu/manager.py#L72)) and filters keys by
reuse count ([`cpu/manager.py:173`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/cpu/manager.py#L173)) — while **`TieringOffloadingSpec` rejects `>=2`
outright**:

```python
if int(self.extra_config.get("store_threshold", 0)) >= 2:
    raise ValueError(
        "store_threshold is not supported for TieringOffloadingSpec"
    )
```
([`tiering/spec.py:289-293`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/spec.py#L289-L293); the docs table agrees — *"single-tier"*, *"Values ≥ 2
are rejected by `TieringOffloadingSpec`"*,
[`docs/features/kv_offloading_usage.md:75`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/docs/features/kv_offloading_usage.md#L75)).

**What H5 must therefore do:** treat `store_threshold` as **fixed at 1**, omit it from the YAML
schema, and **reject `>=2` loudly** if supplied — this is precisely BC-G1's "fails LOUDLY at
startup, never ignored" branch. Offering it as a live knob would let BLIS be configured into a state
vLLM refuses to enter: a fidelity violation in the direction hardest to notice, because the sim
would run and only the mechanism would be wrong. It is also **not** an input to mechanism (c); the
lock pressure there comes from write fan-out on `complete_store`, not from store gating.

## House rules that apply

- **R16** group config by module: this is a `KVCacheConfig` sub-config, independently validatable;
  factories take the narrowest sub-config.
- **R9 / R10** YAML pointer types and strict parsing — hence the fuzz obligation on BC-G3.
- **R8** no exported mutable maps — relevant to `secondary_tiers`.
- **R4** canonical constructor: one struct-literal site. Grep **all** construction sites before
  adding fields.
- **BC-G4** encodes the #1569 conclusion that offload config is *configuration, never fitted*.
  Fitting a knob whose value we already know exactly would be fitting to known truth.

## Evidence

|---|---|---|
| **BC-G2** | **differential** | one test **per row** against vLLM's default — not one aggregate test |
| **BC-G5** | **metamorphic** | offload disabled -> byte-identical stdout vs `52161669` (INV-6, INV-13) |
| BC-G3 | property + **fuzz** | invalid configs fail with actionable messages; strict parsing rejects unknown keys |
| BC-G1 | property | a representable-but-unimplemented config must `logrus.Fatalf`, never degrade silently |
| BC-G4 | static analysis | no write to config fields outside construction |
| **BC-G6** | **replay trace** | export a 3-tier trace, replay it: config must round-trip. Replay of an unreproducible config must **fail loudly**, never degrade to single-tier. |

## Done when

- All 16 rows representable or explicitly rejected; defaults match vLLM knob for knob.
- **BC-G5 holds** — this is the regression guard every later sub-issue depends on.
- **BC-G6 holds** — config round-trips through the trace header (INV-13).
- **Exactly one new flag** (`--kv-offload-config`): flag count 112 -> 113. If the PR adds more
  than one, the delivery design was not followed.
- `NewKVCacheConfig` call sites all audited (R4) — a nested sub-config value, not a 7th
  positional parameter.
- `archon-go delta` shows the intended `config` arrows and nothing else; `reflexion` still 0.

namasl:
## Summary

Fills hole **H5 `kv_offload_config`** (sub-issue of #1585): captures vLLM's KV-cache **offload CLI/config surface** as a first-class, validated, round-trippable BLIS config — so BLIS can be *configured into* the offload mechanisms the sibling holes (#1588/#1589 and later) will implement.

This PR is **config-capture only** — the offload subsystem is **inert**: no `sim/` code reads the new config yet, so a run without `--kv-offload-config` is byte-identical to a build without the feature (BC-G5). The consuming multi-tier transfer/keying physics land separately.

**Exactly one new flag** (`--kv-offload-config`, flag count 112 → 113), one new `defaults.yaml` block (`kv_offload_devices:`), one new sub-config type (`sim.KVOffloadConfig`).

## Design

Follows the `--lora-config` / `--saturation-config` precedent: a strict-YAML file with a single top-level `kv_offload:` block, registered on `run` and `replay` via the shared `registerSimConfigFlags`. Absent ⇒ subsystem inert.

- **`sim.KVOffloadConfig`** — a sub-config of `KVCacheConfig` (R16), resolved/validated/read-only (BC-G4). Threaded through the canonical `NewKVCacheConfig` via a **variadic `WithKVOffload` option** rather than a 7th positional param (which would break all 161 existing call sites); the single struct-literal site is preserved (R4).
- **cmd side** — pointer-typed YAML structs (R9), strict `KnownFields(true)` parsing (R10), a **pure error-returning `resolveKVOffload`** (unit-testable/fuzzable) + a thin CLI wrapper that `logrus.Fatalf`s. `device_class` names resolve against the shipped `kv_offload_devices:` block; an explicit `read_bandwidth`/`write_bandwidth`/`base_latency` triple overrides the class.
- **Trace round-trip (INV-13)** — the *resolved* config (bandwidth/latency numbers, not `device_class`) is recorded in `TraceHeader.KVOffload`; on `replay` the header is authoritative — a config the binary cannot reconstruct/validate fails loudly, a matching `--kv-offload-config` is accepted, a genuine mismatch or adding offload to a no-offload trace is a hard error.

## Behavioral Contracts

- **BC-G1 (representable-or-loud)** — GIVEN a config naming an unrepresentable knob (`store_threshold >= 2`; tier `type` obj/p2p/example) WHEN resolved THEN startup fails with an actionable error — never silent.
- **BC-G2 (defaults match vLLM)** — omitted knobs resolve to vLLM's defaults knob-for-knob: `offload_prompt_only=true` (trap 1), `eviction_policy=lru`, `n_read_threads=n_write_threads=16`, `enable_kv_events=self_describing_kv_events=false`, `blocks_per_chunk=1`, `block_size=`GPU block size, `secondary_tiers=[]`. Verified against vLLM `63a9a5010a`.
- **BC-G3 (validated at construction)** — every enabled tier is complete/consistent (type/root_dir/direct_io/finite per-direction bandwidth), `block_size` XOR `blocks_per_chunk`; NaN/Inf guarded.
- **BC-G4 (read-only, never fitted)** — AST-scan guard: no `sim/` write to the sub-config outside its type definition + constructor.
- **BC-G5 (byte-identical when disabled)** — no flag ⇒ inert zero value, `kv_offload` header key omitted, stdout unchanged (INV-6). Guaranteed by existing run/replay byte-identity parity tests still passing.
- **BC-G6 (round-trip + loud replay)** — resolved config round-trips run→replay through the header; an unreproducible config `logrus.Fatalf`s, never a silent degrade to single-tier (INV-13).

## Traps handled (from the issue)

1. `offload_prompt_only` **defaults to `true`** — decode KV not offloaded unless disabled.
2. **O_DIRECT is explicit**: `direct_io` is a required per-fs-tier field (vLLM probes it at runtime; a simulator can't, and the two regimes are different physics), and it is recorded in the trace.
3. **`store_threshold`**: `TieringOffloadingSpec` rejects `>= 2` — BLIS rejects `>= 2` with vLLM's exact message; `0`/`1` are accepted as a no-op on the multi-tier path (vLLM parity), negatives rejected.

## Testing

`go build ./...`, `go test ./...`, `golangci-lint run ./...` all clean. New coverage:
- Property tests for `Validate()` (BC-G3) and the resolver reject table (BC-G1).
- Per-row default assertions vs vLLM (BC-G2).
- Metamorphic fuzz through the real trace-header YAML serialization path (INV-13 round-trip idempotence, ~285k execs clean).
- BC-G4 static-analysis AST scan.
- End-to-end `run --kv-offload-config` → header records resolved config → `replay` reproduces it; plus header strict-parse + omit-when-absent tests.
- Existing run/replay byte-identity parity tests confirm BC-G5.

Closes #1587

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,17 @@ go build -o blis main.go
 # Run and export workload as TraceV2 (prefix auto-appends .yaml/.csv)
 ./blis run --model qwen/qwen3-14b --trace-output traces/run1
 
+# Run with a multi-tier KV-offload config (#1587). One strict-YAML flag captures vLLM's
+# offload config surface (kv_offload: block — cpu_bytes_to_use, block_size/blocks_per_chunk,
+# eviction_policy, offload_prompt_only [vLLM default TRUE], secondary_tiers[] with per-tier
+# device_class/direct_io/bandwidth). Absent => offload subsystem inert, byte-identical output
+# (BC-G5). Defaults match vLLM knob-for-knob; store_threshold>=2 and non-fs tiers fail loudly.
+# device_class names resolve against defaults.yaml kv_offload_devices. The resolved config is
+# recorded in the trace header and round-trips through replay (INV-13); on replay the header is
+# authoritative (a config it cannot reproduce is a hard error). Config-capture only today — the
+# consuming multi-tier mechanisms land separately.
+./blis run --model qwen/qwen3-14b --kv-offload-config offload.yaml --trace-output traces/run1
+
 # Replay a captured TraceV2 file through the DES (fixed timing from trace)
 ./blis replay --trace-header t.yaml --trace-data d.csv --model qwen/qwen3-14b
 
--- a/cmd/default_config.go
+++ b/cmd/default_config.go
@@ -30,6 +30,21 @@ type Config struct {
 	Workloads              map[string]Workload      `yaml:"workloads"`
 	TrainedPhysicsDefaults *TrainedPhysicsDefaults  `yaml:"trained_physics_coefficients,omitempty"`
 	LoRADefaults           *LoRADefaults            `yaml:"lora,omitempty"`
+	// KVOffloadDevices maps a device_class name to its bandwidth/latency physics for
+	// the KV-offload config surface (H5, #1587). Shipped constants (operator input),
+	// never fitted (BC-G4). Inert: only consulted when --kv-offload-config names a
+	// device_class. A struct-field map (decode target, read-only after load) — not an
+	// exported package var (R8-compatible, like Defaults/Workloads above).
+	KVOffloadDevices map[string]KVOffloadDeviceDefaults `yaml:"kv_offload_devices,omitempty"`
+}
+
+// KVOffloadDeviceDefaults is one device_class's resolved physics for KV offload
+// (H5, #1587). Bandwidths are bytes per microsecond; base_latency is microseconds.
+// Fields are plain float64 (always present within a device block).
+type KVOffloadDeviceDefaults struct {
+	ReadBandwidth  float64 `yaml:"read_bandwidth"`
+	WriteBandwidth float64 `yaml:"write_bandwidth"`
+	BaseLatency    float64 `yaml:"base_latency"`
 }
 
 // LoRADefaults holds inert defaults for the LoRA control-plane subsystem's cost
--- a/cmd/kv_offload.go
+++ b/cmd/kv_offload.go
@@ -1,0 +1,371 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/inference-sim/inference-sim/sim"
+	"github.com/inference-sim/inference-sim/sim/workload"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+// kvOffloadConfigPath backs the single new --kv-offload-config flag (H5, #1587).
+// Absent ("") ⇒ the offload subsystem is inert and output is byte-identical to a
+// build without the feature (BC-G5). Registered on run and replay via the shared
+// registerSimConfigFlags (INV-13 parity).
+var kvOffloadConfigPath string
+
+// kvOffloadFile is the on-disk shape of a --kv-offload-config YAML file: a single
+// top-level kv_offload: block. Strict-parsed (R10). Mirrors the --lora-config /
+// --saturation-config precedent.
+type kvOffloadFile struct {
+	KVOffload *kvOffloadBlock `yaml:"kv_offload"`
+}
+
+// kvOffloadBlock is the user-facing kv_offload: block. All fields are pointers so an
+// absent key is distinguishable from an explicit zero (R9): the resolver applies
+// vLLM's default only when the pointer is nil.
+//
+// NOTE store_threshold: it is captured (as *int) but is NOT a live knob on the
+// multi-tier path — vLLM's TieringOffloadingSpec rejects values >= 2 (it is
+// single-tier-only). The resolver rejects >= 2 loudly (BC-G1) and treats nil/0/1 as a
+// no-op (matching vLLM, which accepts 0/1 and ignores them on this path).
+type kvOffloadBlock struct {
+	CPUBytesToUse          *int64               `yaml:"cpu_bytes_to_use"`
+	BlockSize              *int64               `yaml:"block_size"`
+	BlocksPerChunk         *int64               `yaml:"blocks_per_chunk"`
+	TokensPerHash          *int64               `yaml:"tokens_per_hash"`
+	EvictionPolicy         *string              `yaml:"eviction_policy"`
+	OffloadPromptOnly      *bool                `yaml:"offload_prompt_only"`
+	SelfDescribingKVEvents *bool                `yaml:"self_describing_kv_events"`
+	StoreThreshold         *int                 `yaml:"store_threshold"`
+	SecondaryTiers         []kvOffloadTierBlock `yaml:"secondary_tiers"`
+}
+
+// kvOffloadTierBlock is one user-facing secondary tier. Pointer fields (R9).
+type kvOffloadTierBlock struct {
+	Type           *string  `yaml:"type"`
+	RootDir        *string  `yaml:"root_dir"`
+	NReadThreads   *int64   `yaml:"n_read_threads"`
+	NWriteThreads  *int64   `yaml:"n_write_threads"`
+	Locality       *string  `yaml:"locality"`
+	EnableKVEvents *bool    `yaml:"enable_kv_events"`
+	DirectIO       *bool    `yaml:"direct_io"`
+	DeviceClass    *string  `yaml:"device_class"`
+	ReadBandwidth  *float64 `yaml:"read_bandwidth"`
+	WriteBandwidth *float64 `yaml:"write_bandwidth"`
+	BaseLatency    *float64 `yaml:"base_latency"`
+}
+
+// loadKVOffloadConfigFile reads and strictly parses a --kv-offload-config file.
+// Returns an error (never Fatalf — so it is unit-testable/fuzzable, Deviation #9);
+// the CLI wrapper turns errors into logrus.Fatalf. Unknown keys error via
+// KnownFields(true) (R10).
+func loadKVOffloadConfigFile(path string) (kvOffloadFile, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return kvOffloadFile{}, fmt.Errorf("read --kv-offload-config file %q: %w", path, err)
+	}
+	f, err := parseKVOffloadBytes(data)
+	if err != nil {
+		return f, fmt.Errorf("parse --kv-offload-config file %q: %w", path, err)
+	}
+	return f, nil
+}
+
+// parseKVOffloadBytes strictly decodes kv_offload YAML bytes (R10). Split out so it is
+// directly fuzzable without touching disk.
+func parseKVOffloadBytes(data []byte) (kvOffloadFile, error) {
+	var f kvOffloadFile
+	decoder := yaml.NewDecoder(bytes.NewReader(data))
+	decoder.KnownFields(true)
+	if err := decoder.Decode(&f); err != nil {
+		return f, err
+	}
+	return f, nil
+}
+
+// resolveKVOffload turns a parsed kv_offload block into a resolved, validated
+// sim.KVOffloadConfig. It is PURE and error-returning (Deviation #9) so every reject
+// branch is unit-testable and fuzzable. It applies vLLM's defaults knob-for-knob
+// (BC-G2), resolves device_class against the shipped device physics (explicit
+// read/write/base override the class), enforces the block_size XOR blocks_per_chunk
+// user-input rule then derives the canonical pair, rejects store_threshold>=2 and
+// non-fs tiers loudly (BC-G1), and finally runs cfg.Validate() (BC-G3).
+//
+// gpuBlockSizeTokens is the GPU block size (--block-size-in-tokens); vLLM's block_size
+// default equals it, and it converts between the block_size and blocks_per_chunk
+// encodings of the same quantity.
+func resolveKVOffload(block *kvOffloadBlock, devices map[string]KVOffloadDeviceDefaults, gpuBlockSizeTokens int64) (sim.KVOffloadConfig, error) {
+	if block == nil {
+		return sim.KVOffloadConfig{}, fmt.Errorf("kv_offload: the --kv-offload-config file has no top-level kv_offload: block")
+	}
+	if gpuBlockSizeTokens <= 0 {
+		return sim.KVOffloadConfig{}, fmt.Errorf("kv_offload: internal error — GPU block size must be > 0, got %d", gpuBlockSizeTokens)
+	}
+	cfg := sim.KVOffloadConfig{Enabled: true}
+
+	// cpu_bytes_to_use: required when the block is present (vLLM marks it required).
+	if block.CPUBytesToUse == nil {
+		return cfg, fmt.Errorf("kv_offload: cpu_bytes_to_use is required")
+	}
+	cfg.CPUBytesToUse = *block.CPUBytesToUse
+
+	// store_threshold: reject >= 2 (vLLM TieringOffloadingSpec); nil/0/1 are a no-op.
+	if block.StoreThreshold != nil && *block.StoreThreshold >= 2 {
+		return cfg, fmt.Errorf("kv_offload: store_threshold=%d: store_threshold is not supported for TieringOffloadingSpec (values >= 2 are rejected; it is single-tier-only and fixed at 1 on the multi-tier offload path)", *block.StoreThreshold)
+	}
+
+	// block_size XOR blocks_per_chunk (mutually exclusive alternate encodings of the
+	// same chunk-granularity quantity). Derive one canonical pair.
+	if block.BlockSize != nil && block.BlocksPerChunk != nil {
+		return cfg, fmt.Errorf("kv_offload: block_size and blocks_per_chunk are mutually exclusive (they encode the same quantity: block_size = blocks_per_chunk × gpu_block_size); set at most one")
+	}
+	switch {
+	case block.BlockSize != nil:
+		bs := *block.BlockSize
+		if bs <= 0 || bs%gpuBlockSizeTokens != 0 {
+			return cfg, fmt.Errorf("kv_offload: block_size (%d) must be a positive multiple of the GPU block size (%d)", bs, gpuBlockSizeTokens)
+		}
+		cfg.BlockSize = bs
+		cfg.BlocksPerChunk = bs / gpuBlockSizeTokens
+	case block.BlocksPerChunk != nil:
+		bpc := *block.BlocksPerChunk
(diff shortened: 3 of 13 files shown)


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

sriumcp (approved):
lgtm

PR #1595, opened 2026-08-18
feat(kv): fill hole H4 kv_block_key — chunk-stride keys + interning (#1589)

### LINKED ISSUE #1589: S1 · Fill hole H4 kv_block_key: chunk-stride keys + interning
**Sub-issue of #1585.** Fills hole **H4 `kv_block_key`**. Blocked by #1586.
Verified against BLIS `52161669` and vLLM `63a9a5010a`. Every vLLM/BLIS source claim below carries a **verified git permalink** at those commits (re-read in a local checkout; line ranges confirmed to contain the quoted code).

Serves both project objectives at once: **fidelity** via chunk-stride keys (BC-K4), **performance**
via interning (BC-K3).

## Start here: what a "block key" is

BLIS decides whether a prompt prefix is already cached by hashing its tokens into a **key** and
looking that key up. Every tier — GPU, CPU, disk — and the router's prefix-affinity scorer must
compute **the same key for the same tokens**, or they will disagree about what is cached.

## The hole

```
hole kv_block_key {
  surface:
    DeriveChunkKeys(prevKey, tokens []tokenid.TokenID, tokensPerChunk) -> []BlockKey
    Intern(BlockKey) -> KeyId
  allow:
    import sim/internal/hash        // the ONLY hash source
    import sim/internal/tokenid
  contract:
    BC-K1 one hash function serves GPU, CPU, every secondary tier, AND
          router-side prefix affinity — identical bytes for identical tokens
    BC-K2 chunk keys chain hierarchically: a match at chunk i implies
          matches at every j < i
    BC-K3 Intern is injective and stable within a run
    BC-K4 offload keys are derived at CHUNK stride, not block stride
  evidence:
    property_test    (BC-K1, BC-K2, BC-K3)
    metamorphic_test (BC-K1)
    differential_test(BC-K4 vs offloading/scheduler.py)
    static_analysis  (BC-K1: no other hash import inside sim/kv)
    benchmark        (P)
}
```

## What already exists — do not rebuild it

Most of BC-K1 is already true. `sim/internal/hash` says so itself:

> *"These functions are shared between sim/ (routing) and sim/kv/ (cache) to ensure hash
> consistency (BC-3)."*

It is used by `sim/kv/cache.go`, `sim/kv/tiered.go`, `sim/prefix_cache_index.go` and
`sim/routing_prefix_scorer.go`, and `TestComputeBlockHashes_MatchesManualChaining` already guards
its two derivation paths (`HashBlock` sequential vs `ComputeBlockHashes` hasher-reuse) against
divergence.

**Two facts that are new on current main and change this issue:**

1. The signatures now take **`[]tokenid.TokenID`**, not `[]int`. `TokenID` is a distinct `int32`
   type, *not* an alias, so mixing with `int` is a **compile error** — the check is the point.
2. `sim/internal/hash --import--> sim/internal/tokenid` is now a real graph edge, and `tokenid` is
   the **highest-blast package in the repo (10)** per `archon-go health`.

```mermaid
graph LR
  gpu["GPU tier<br/>sim/kv/cache.go"] --> H["sim/internal/hash<br/>blast 9"]
  cpu["CPU tier<br/>sim/kv/tiered.go"] --> H
  idx["router index<br/>sim/prefix_cache_index.go"] --> H
  sc["prefix scorer<br/>sim/routing_prefix_scorer.go"] --> H
  new["NEW: disk tier + chunk keys"] -. must also use .-> H
  H --> T["sim/internal/tokenid<br/>blast 10"]

  classDef hub fill:#dbe7f8,stroke:#3b5f8f,stroke-width:3px,color:#0d1b2e
  classDef add fill:#ffeccc,stroke:#b45309,stroke-width:3px,stroke-dasharray:6 3,color:#3b2200
  class H,T hub
  class new add
  class gpu,cpu,idx,sc plain
```

So **BC-K1 promotes an existing convention into a gated contract.** The genuinely new work is
BC-K4 and BC-K3.

## BC-K4 — chunk stride (fidelity)

vLLM does **not** derive offload keys per block. It derives them per **chunk**:

- `tokens_per_hash` and `blocks_per_chunk` (`vllm/v1/kv_offload/config.py:30-32`) are separate
  values from `tokens_per_block`.
- [`offloading/scheduler.py:1021-1055`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py#L1021-L1055) walks `offload_keys[start_chunk_idx:end_chunk_idx]`.
- `_flush_pending_promotions()` then submits **one batched job per (tier, request)**.

Consequences for the design:

- Transfer-chunk keys are a **second keyspace**, derived from the same chain at a different
  stride. Both keyspaces must come from this box.
- The transfer unit is a **job** covering many blocks (see #1588 BC-S5). vLLM's
  `active_promotion_jobs` counts **jobs**, not blocks.

Modeling per-block would be **both slower and less faithful.** Fidelity and performance agree.

## BC-K3 — interning (performance)

Keys today are 64-byte hex strings used directly as map keys:

```go
HashToBlock    map[string]int64      // sim/kv/cache.go
cpuTier.blocks map[string]*cpuBlock  // sim/kv/tiered.go
```

Adding a disk tier plus per-block `ref_cnt` plus in-flight job tracking multiplies the number of
`map[string]` probes per block per step, and each probe hashes 64 bytes.

**Fix:** intern each key once to an integer `KeyId`; key the hot maps on that; keep the string
only where content identity crosses a boundary (KV events, traces, the router index).

This is the real constant-factor lever in the whole project, and it is why #1586's benchmark
obligation may end up waiting on this issue.

## Evidence

|---|---|---|
| **BC-K1** | **metamorphic** | the same token sequence entering via GPU, CPU or a disk tier must produce **byte-identical** keys. Catches a tier that grows its own hash. |
| BC-K2 | property | generated token streams; prefix-match closure downward |
| BC-K3 | property | injectivity and stability over generated key sets — collisions must be **impossible**, not merely unlikely |
| **BC-K4** | **differential** | key sequence vs vLLM's chunk walk for the same tokens and the same `tokens_per_hash` / `blocks_per_chunk` |
| BC-K1 | **static analysis** | gate: no `crypto/*`, `hash/*` or `fnv` import inside `sim/kv` outside this box |
| P | **benchmark** | hot-path probes keyed on `KeyId`; report ns/op against baseline |

## Done when

- BC-K1 through BC-K4 hold; `sim/internal/hash` remains the sole hash source.
- All new signatures use `tokenid.TokenID` — no `int` token parameters.
- `archon-go contract` shows no import outside `sim/kv`'s baseline
  (`sim`, `sim/internal/hash`, `sim/internal/util`).
- `reflexion` still 0 divergent; `evidence` still `gapKind: none` for `sim.KVStore`.
- Benchmark shows no regression vs `52161669` — ideally an improvement.

namasl:
## Summary

Fills **hole H4 `kv_block_key`** of the multi-tier KV-offload epic (#1585): a single, gated home for KV block-key derivation and interning, in a new leaf package `sim/internal/kvkey`. It serves both epic objectives at once — **fidelity** (chunk-stride keys matching vLLM's transfer granularity) and **performance** (interning, the one real constant-factor lever for the coming disk-tier probes).

Two new primitives:

1. **`DeriveChunkKeys(prevKey, tokens, tokensPerChunk) []BlockKey`** — hierarchical content keys at an arbitrary stride, seeded by `prevKey`. A strict generalization of `hash.ComputeBlockHashes`: with `tokensPerChunk = tokensPerBlock × blocks_per_chunk` it yields one key per **chunk** (a group of `blocks_per_chunk` blocks), matching vLLM's per-chunk offload keys — **not** one key per block.
2. **`Interner`** with `Intern(BlockKey) KeyID` — an injective, idempotent, dense `string → int` interner, plus a reverse `Key(KeyID)` accessor for the boundaries where content identity must stay a string (KV events, traces, the router index).

Plus the `BlockKey` / `KeyID` defined types (the same compile-time discipline the codebase already applies to `tokenid.TokenID`), and a **static-analysis gate** that promotes "one hash source" from convention to an enforced test.

### Scope: primitives-only (no rewire)

This PR adds the primitives with full evidence and does **not** rewire `sim/kv`'s hot maps (`HashToBlock`, `cpuTier.blocks`) onto `KeyID`. The probe pressures BC-K3's rationale cites — disk tier, per-block ref-counts, in-flight job tracking — don't exist yet; the rewire lands with the future hole that adds them. Because no existing code is modified and nothing on the hot path calls the new primitives, **stdout is byte-identical (INV-6)** and there is **zero benchmark regression** by construction — matching the build-then-wire posture of sibling epic PRs #1588 (standalone) and #1587 (inert).

## Behavioral Contracts

- **BC-K1 (one hash source):** GIVEN identical tokens and stride, WHEN keys are derived via `kvkey.DeriveChunkKeys("", tokens, blockSize)` and via `hash.ComputeBlockHashes(blockSize, tokens)`, THEN the two sequences are byte-identical, AND no production file in `sim/kv` or `sim/internal/kvkey` imports `crypto/*`, stdlib `hash`/`hash/*`, or `fnv`.
- **BC-K2 (hierarchical chaining):** GIVEN two sequences sharing their first *k* complete chunks, WHEN chunk keys are derived at the same stride, THEN the first *k* keys are identical and a length-(*m*·stride) prefix's keys equal the first *m* keys of any longer sequence (a match at chunk *i* implies a match at every *j* < *i*).
- **BC-K3 (interning):** GIVEN any key stream (with repeats) into one `Interner`, THEN interning the same key always returns the same `KeyID` (idempotent), distinct keys always return distinct dense `KeyID`s (injective — collisions impossible), `Key(Intern(k))` round-trips, and two Interners fed the identical sequence assign identical ids (deterministic).
- **BC-K4 (chunk stride, not block stride):** GIVEN tokens and `(B, C)`, WHEN keys are derived at chunk stride `B·C` and block stride `B`, THEN chunk-stride yields `⌊numBlocks / C⌋` keys — C× fewer than block stride — and chunk key *i* is a function of exactly the first `(i+1)·B·C` tokens, matching vLLM's per-chunk offload-key walk.
- **BC-K5 (no behavioral change):** stdout byte-identical to before (INV-6); no existing file modified.
- **BC-K6 (degenerate inputs):** `tokensPerChunk ≤ 0` panics; a short/empty token slice returns nil; an out-of-range/negative `KeyID` returns `("", false)`.

## Fidelity note (BC-K4)

vLLM (commit `63a9a5010a`) anchors each chunk's offload key to the chunk's *trailing block-hash*, so a chunk key equals a value in the block-hash chain (a subsequence). The frozen surface here provides only the chunk stride, so `DeriveChunkKeys` chain-hashes each whole chunk with BLIS's own SHA256 — a **disjoint** keyspace. This is functionally sufficient (a stable per-chunk content key) and loses no capability: a consumer with the request tokens derives *both* keyspaces from this one hash source. The differential test therefore verifies the chunk **stride** (chunk count, C×-ratio, per-chunk coverage) — which discriminates a block-stride bug — not byte-fidelity to vLLM's anchoring.

## Testing

- `go test ./sim/internal/kvkey/...` — 14 tests + 2 fuzz targets pass (property, metamorphic, static gate, differential, degenerate/panic, determinism).
- `FuzzDeriveChunkKeys` and `FuzzInterner` — 6s each, no crashes.
- Benchmarks (evidence P): `KeyID`-keyed probe **14.7 ns/op** vs 64-byte-string probe **22.1 ns/op** (~34% faster), zero allocs — the interning lever, demonstrated without rewiring.
- Full suite `go test ./... -count=1` — green; `golangci-lint run ./...` (v2.9.0) — 0 issues.

Naming note: the issue writes `KeyId`; this PR uses `KeyID` to satisfy Go initialism convention (ST1003) and match `tokenid.TokenID`.

Closes #1589.


namasl:

namasl:

namasl:

sriumcp (approved):
lgtm

Issue #1591, opened 2026-08-17
S5 · Fill hole H3 kv_deferral: step-boundary re-poll (the dominant TTFT effect)


sriumcp:
**Sub-issue of #1585.** Fills hole **H3 `kv_deferral`** — **the dominant TTFT effect.**
Blocked by #1590. Verified against BLIS `52161669` and vLLM `63a9a5010a`. Every vLLM/BLIS source claim below carries a **verified git permalink** at those commits (re-read in a local checkout; line ranges confirmed to contain the quoted code).

## Start here: the single most important fact in this project

> The delay a request suffers waiting for a disk fetch is set by **how often the scheduler runs**,
> not by **how fast the disk is**.

`vllm/v1/core/sched/scheduler.py:835-841`:

```python
if ext_tokens is None:
    # The request cannot be scheduled because
    # the KVConnector couldn't determine
    # the number of matched tokens.
    request_queue.pop_request()
    step_skipped_waiting.prepend_request(request)
    continue
```

vLLM does **not** block on the disk. It starts the read in the background, sets the request aside,
and **checks again on the next scheduler step**.

```mermaid
graph LR
  A["request arrives<br/>needs disk blocks"] --> B["fetch starts<br/>request set aside"]
  B --> C["step boundary<br/>check again"]
  C -->|"not ready"| C
  C -->|"ready"| D["request runs"]

  classDef key fill:#ffeccc,stroke:#b45309,stroke-width:3px,color:#3b2200
  classDef plain fill:#e8eaed,stroke:#6b7280,color:#111827
  class C key
  class A,B,D plain
```

So if a disk read takes **2 ms** and steps run every **40 ms**, the request waits **~40 ms**, not
2 ms.

**Why this is the whole ballgame:** a bandwidth model is wrong by the ratio between those numbers
— and wrong in the **wrong direction** as batch size changes, because step time grows with batch
size while disk latency does not. Fitting to average TTFT would pick the wrong theory and then
extrapolate badly.

## It costs at least TWO rounds, not one

`get_num_new_matched_tokens` (`vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py:936-991`)
returns `None` on **two separate paths**:

```python
if req_status.transfer_jobs:                     # path 1 — lines 962-967
    return None, False                           # ANY in-flight transfer defers too
...
num_hit_tokens = self._lookup(req_status)        # path 2 — unresolved lookup
if num_hit_tokens is None: ...
```

[...]

PR #1597, opened 2026-08-19
feat(kv): fill hole H1 kv_tier_chain — N-tier chain, ref_cnt pinning, write-through cascade (#1590)

### LINKED ISSUE #1590: S4 · Fill hole H1 kv_tier_chain: N-tier chain, ref_cnt pinning, write-through cascade

namasl:
## Summary

Fills the central hole **H1 `kv_tier_chain`** of the multi-tier KV-offload epic (#1585): replaces BLIS's GPU+single-CPU-mirror offload model with a faithful **N-tier chain** — GPU ↔ a `ref_cnt`-managed **CPU staging tier** ↔ ordered secondary **`fs` tiers** — with a bounded **transfer station** (#1588) for per-job promotion/cascade service and block keying through **`sim/internal/kvkey`** (#1589). This models vLLM's `tiering/manager.py` + `cpu/manager.py` *mechanism*, not merely its outputs, so hit-rate, block-locking, and the **recompute-under-lock-saturation** degradation mode are now represented (epic §4 divergences **D2/D4/D5**).

New type `sim/kv.OffloadCache` implements `sim.KVStore` (still **12 entities**) plus the optional `SnapshotCachedBlocksFn`. It is **additive and inert unless `cfg.Offload.IsEnabled()`** (`--kv-offload-config`, #1587): the legacy `--kv-cpu-blocks` `TieredKVCache` is untouched, so all existing runs are byte-identical (INV-6). The two offload models are mutually exclusive (loud error if both set).

**Design (user-approved):** driven through the existing **synchronous KVStore seam** — `SetClock(now)` advances the station and applies completions; no new `sim.Event` types, no interface growth. Timing flows through **token counts** (recompute raises prefill tokens via the existing `StepTime` model); the dominant step-boundary deferral / per-request latency is **H3 (#1591)**, which is blocked by this PR.

## Behavioral Contracts

- **BC-C1 (two-hop):** a secondary block reaches GPU strictly via secondary→CPU→GPU, never directly.
- **BC-C2 (INV-4 multi-tier):** CPU tier `used + free == capacity` always.
- **BC-C3 (`ref_cnt` state machine):** `-1` allocated-not-ready (HIT_PENDING, unreadable+unevictable) / `0` ready-evictable (HIT) / `n>0` pinned (HIT, non-evictable); only `completeStore(ok)` does `-1→0`; a failed store removes+frees.
- **BC-C4:** a block with `ref_cnt != 0` is never evicted.
- **BC-C5 (the trap):** a promotion of `k` blocks succeeds iff `k ≤ free + evictable` — gating on the **evictable** count, never the free count; a lock-saturated CPU tier degrades promotions to recompute.
- **BC-C6:** secondary lookup order is deterministic, independent of arrival order.
- **BC-C7a/b:** writes fan out to **every** secondary tier (eager replicas, no spill-chain); tier-order permutation / a never-hit tier do not change the promotion outcome.
- **BC-C8:** the evictable counter is O(1) (maintained on zero-crossings, never scanned).
- **BC-N1 (INV-6):** offload disabled ⇒ byte-identical stdout.
- **BC-C13 (INV-13):** the resolved config (incl. derived `per_block_bytes`) round-trips through the trace header; run vs replay per-request metrics identical.

## Testing

- `go test ./...` green; `golangci-lint run ./...` (v2.9.0) 0 issues; `go test -race ./sim/kv/` green.
- Evidence suite: **property** (conservation, state machine, evictable==scan), **metamorphic** (BC-C5 lock-saturation with a `promotions_fired>0` guard, BC-C7b order/unused invariance), a **direct correct-vs-mutant** unit assertion for the evictable gate (`free=2/evictable=1/k=4 → refuse`), a **differential** two-hop staging test, **INV-4/C4** conservation (mid-cascade completion never leaks a pin), **INV-13** run/replay parity with offload active, and a **benchmark** (offload allocate ≈ single-tier; CPU-tier ops O(1)).
- INV-6 verified: offload-on and offload-off both byte-identical across seeded runs.

## Scope trims (rejected loudly, follow-ups)

- `blocks_per_chunk == 1` only (block-granular; `>1` chunk coalescing deferred — the block/chunk keyspace reconciliation is the highest-risk leg and orthogonal to BC-C1..C8).
- `eviction_policy: lru` only (`arc` deferred).

Offload tiers are intentionally invisible to the `precise-prefix-cache` router scorer (`GetCachedBlocks`/`SnapshotCachedBlocksFn` stay GPU-only) — a routing-fidelity boundary noted for a later hole.

`archon-go delta` is expected to report `ARCHITECTURAL_CHANGE` (`sim/kv` gains downward/lateral imports of `sim/kvtransfer` + `sim/internal/kvkey`, both acyclic leaves) with reflexion still 0-divergent — as the issue's Done-when anticipates.

Fixes #1590.


namasl:

namasl:

sriumcp:
@claude Does the archon review above indicate adherence to the architecture/design plans mentioned in the parent issue?

namasl:
Yes — the `ARCHITECTURAL_CHANGE` verdict is exactly what the parent issue's acceptance criteria call for, not a deviation.

Issue #1590's **"Done when"** states verbatim:

> `archon-go delta`: `ARCHITECTURAL_CHANGE` is **expected here** and must be reviewed; `reflexion` still 0 divergent; `evidence` still `gapKind: none`.

The archon report confirms each of those three, and nothing beyond them:

- **Expected architectural change, and only it.** The witness delta's only ADDED edges are the two intended wirings this hole was designed around: `sim/kv → sim/internal/kvkey` (block-key derivation, #1589) and `sim/kv → sim/kvtransfer` (the transfer station, #1588). Every other edge is `STRENGTHENED`, none `REMOVED`.
- **Reflexion 0-divergent.** No upward/lateral layering violation is introduced — both new imports are one-way into acyclic leaves (`sim/kvtransfer` is stdlib-only; `sim/internal/kvkey` imports only `sim/internal/{hash,tokenid}`), consistent with the `cmd/ → sim/cluster/ → sim/` arrow. This upholds epic #1585 §5's "layering stays at 0 divergent edges" commitment.
- **Evidence `gapKind: none`.** `sim.KVStore` gains `kv.OffloadCache` as an implementer with **no uncovered (evidence-gap) row** — matching the epic's "KVStore evidence stays at `gapKind: none`" constraint.

The other epic §5 "Fixed" invariants also hold: `sim.KVStore` keeps its **12 entities** (no 13th — `OffloadCache` implements all 12 with no interface growth; the `SnapshotCachedBlocksFn` capability is a separate optional method, not a `KVStore` entity); one hash source module-wide (all keying routes through `sim/internal/kvkey`, enforced by the existing `static_test.go`); and INV-3/4/6/8/13 are each covered by a named test (see the blis-pr-review table above). The only wire-schema addition is the resolved `TraceKVOffloadConfig.PerBlockBytes`, which round-trips for INV-13.

So the archon verdict indicates adherence: it's the anticipated import-graph growth for filling H1, with the frozen surfaces (interface entity count, layering, evidence coverage) all preserved.

sriumcp (approved):

PR #1598, opened 2026-08-19
feat(calibrate): KV cache hit-rate observe/replay/calibrate loop (#1583)

### LINKED ISSUE #1583: [Tiered KV Cache] PR-CAL — calibrate and validate tier parameters against cluster experiments
**Part of #1585** — multi-tier KV offload (primary tracking). Slotted as **S6** in #1585 §9.
**Depends on:** H1 tier chain (#1590) for a working multi-tier model; re-run/extended as H2 (#1588)
and H1 (#1590) add parameters. **Fills the calibration/validation gap that #1585 §11 scopes out** —
per-tier metrics and coefficient fitting are declared outside the mechanism holes, but validating
BLIS against the cluster experiments is the objective that motivates them.

### What
Close the loop against the cluster experiments: **derive** each tier's parameters (bandwidth, base
latency, I/O parallelism) and **validate** BLIS's outputs against real vLLM runs, using the existing
**`blis observe → replay → calibrate`** pipeline. Keep offload cost in the analytical/event layer
(not the trained coefficients) unless validation shows otherwise.

### Why it matters
Mechanism without calibration is not a model; the cluster group's runs are the ground truth.
Parameters can come from published numbers, small microbenchmarks, and vLLM's per-tier Prometheus
metrics — see the [calibration writeup](https://github.com/inference-sim/inference-sim/discussions/1569#discussioncomment-18017438)
and the [prior-art survey](https://github.com/inference-sim/inference-sim/discussions/1569#discussioncomment-18020495) in #1569.

### Gaps this PR must address (grounded in current `calibrate`)

**1. Hit-rate is not compared today.** `calibrate` compares **TTFT, E2E, and ITL — and nothing
else**. Proof on the BLIS side: the compared set is exactly three fields on
[`CalibrationPairs`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/workload/calibrate.go#L138-L141) (`sim/workload/calibrate.go:138-141`:
`TTFT`, `E2E`, `ITL`), and the reported metrics are those same three
([`cmd/calibrate.go:277-287`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/cmd/calibrate.go#L277-L287)). There is **no** hit-rate field in
either. This PR must add a **KV cache hit-rate** metric to the compared set, mapping the sim-side
[`CacheHitRate`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv_store.go#L12) (`sim/kv_store.go:12`) to vLLM's
[`block_hits`/`block_queries`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/base.py#L48-L49)
(`kv_offload/tiering/base.py:48-49`).

**2. NEW — and this is a hard prerequisite the original issue missed: there is no real-side
hit-rate capture path at all.** Adding a metric to `calibrate` is not sufficient, because
`calibrate` only ever sees what `observe` wrote into the trace.
[`TraceRecord`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/workload/tracev2.go#L58-L90) (`sim/workload/tracev2.go:58-90`) has **no**
cache-hit field — its 31 fields are request identity, token counts, and four timestamps — and
nothing in `observe` scrapes vLLM's metrics endpoint (there is no Prometheus client in `go.mod` at all). So the work is:

   a. **`observe` must scrape** the per-tier counters from the server's `/metrics` endpoint
      (`block_hits`/`block_queries` above; per-tier read/write time from
      [`tiering/base.py:42-45`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/base.py#L42-L45)). These are
      *engine-level gauges/counters*, not per-request fields, so they belong in the **trace header**
      (a resolved-observation block), not as a per-record column — a per-record column would imply
      per-request attribution the engine does not expose.
   b. **The trace header must carry them**, which puts this change on the INV-13 surface: #1585 §5
      already commits the header to gaining the resolved offload config, and this is the same
      seam. `replay` must not silently ignore a header hit-rate it cannot reproduce.
   c. **Only then** can `calibrate` compare. Sim side already has the number
      ([`CacheHitRate`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv_store.go#L12)); real side does not exist yet.

   **This reorders the work:** (a) and (b) are trace/observe changes and are *startable before*
   H1 (#1590) lands, because they do not depend on the tier chain. Only the actual comparison and
   tolerance check need H1. Splitting this PR at that seam is recommended.

**3. Dependency on unreleased vLLM.** #1585 §11 deliberately keeps every *mechanism* hole free of
the per-tier metrics because they are **in no release tag** (`vllm:kv_offload_tiering_*`, PR #48798,
merged 2026-08-10). This issue is the one place that dependency is load-bearing: the real-side
hit-rate **cannot be observed** on a released vLLM. Record the pinned commit the cluster runs use,
and treat a released-vLLM fallback (GPU-only `gpu_prefix_cache_hit_rate`) as a *separate, weaker*
validation, not as a substitute.

**4. Concrete tolerance (initial targets, using `calibrate`'s existing quality bands):** TTFT
**MAPE ≤ 15%** and hit-rate **absolute error ≤ 5 pp** on the minimal experiments; also compare
per-tier [read/write time](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/base.py#L42-L45) against the derived
bandwidths. Note that vLLM's read/write time is **device service time excluding queue wait** — it
is measured after dequeue ([`fs/thread_pool.py:176-181`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/thread_pool.py#L176-L181):
`popleft()` at :176, `start_time` at :178) — so it calibrates H2's *service* term (#1588 BC-S3),
**not** its queueing term. Do not fit queue wait against it.

**5. Minimal experiment set (named):** the four canonical cache paths (GPU hit / CPU hit / storage
hit / full miss) × a small prefix-length sweep, single replica, TP=1 — the same minimal set
unblocked by H1 (#1590). Document the method in a short `docs/` calibration note.

### Cross-path parity
Parts (2a)/(2b) **do** touch the trace surface and therefore INV-13: a header field that `replay`
cannot reproduce must fail loudly, never degrade silently (#1585 §5). The comparison itself
(`calibrate`) consumes the DES kernel rather than modifying it. External tools:
[`fio`](https://github.com/axboe/fio), [Prometheus](https://prometheus.io/docs/introduction/overview/).

### Done when
- `observe` records the real-side per-tier hit-rate into the trace header; `replay` rejects (never
  ignores) a header it cannot reproduce; parity test green.
- `calibrate` compares TTFT, E2E, **and** KV hit-rate; sim-side hit-rate maps to the vLLM counters.
- On the named minimal experiments, BLIS meets the tolerance above, with a documented method and
  the pinned vLLM commit recorded.

namasl:
## Summary

Closes the KV-cache hit-rate validation gap for the tiered KV-offload epic (#1585, slot **S6**): adds a **real-side hit-rate capture path** to `blis observe` and a **sim-vs-real hit-rate comparison** to `blis calibrate`, so the multi-tier offload model can be validated against real vLLM runs through the existing `observe → replay → calibrate` pipeline.

Today `calibrate` compares TTFT, E2E, and ITL — and nothing else — and there is **no real-side hit-rate source at all** (`TraceHeader` had no cache-hit field; `observe` never scraped `/metrics`). The sim already computes the number (`sim.KVStore.CacheHitRate()`) but it only reached stdout as human text. This PR wires the whole loop.

## What this adds

1. **`observe --scrape-kv-metrics`** scrapes the server's Prometheus `/metrics` at the start and end of the measured window and records the observed hit-rate in a new resolved-observation header block (`observed_kv_metrics`). Primary signal: `vllm:kv_offload_tiering_block_hits`/`_block_queries` (vLLM PR #48798, **unreleased** — pin it with `--vllm-commit`). Released fallback: `vllm:gpu_prefix_cache_*`, tagged `source: gpu-prefix-cache-fallback` (a distinct, weaker signal, never conflated with a tiered number).
2. **`replay --metrics-path`** writes the sim aggregate `cache_hit_rate` into the MetricsOutput file (symmetric with `blis run`).
3. **`calibrate --sim-metrics`** now compares TTFT, E2E, **and** KV hit-rate — a `hit_rate` report block with `abs_error_pp` and a within-tolerance verdict (default **≤ 5 pp**), plus a TTFT-MAPE verdict (`--ttft-mape-threshold`, default **0.15**).
4. **Docs**: `docs/guide/kv-offload-calibration.md` — the minimal-experiment method (four canonical cache paths × prefix sweep, single replica, TP=1), the pinned-vLLM dependency, and the released-vLLM fallback caveat.

## Behavioral contracts

- **BC-1/2/13** — observe records `(block_hits Δ)/(block_queries Δ)` over the measured window, tagged by source; degenerate cases (zero queries, counter reset, no recognized counters, unreachable `/metrics`) **warn and omit** the block, never fatal, never a bogus value.
- **BC-3/4** — the header block round-trips and is strict-parsed (unknown sub-key = error); `replay --trace-output` preserves it verbatim.
- **BC-5/8/9** — `replay --metrics-path` file carries `cache_hit_rate`; **stdout stays byte-identical** to a pre-feature build (the field is `*float64,omitempty`, populated only in the file branch of `EmitOutput`, alongside the existing file-only `requests[]`) — INV-6. `run` and `replay --metrics-path` of the same trace produce **identical** `cache_hit_rate` — INV-13.
- **BC-6/7/12** — calibrate emits the `hit_rate` block only when both sides are present; otherwise it **skips with a warning** (TTFT/E2E rows still computed).
- **BC-10** — a `source: tiered` observation with **no reproducible `kv_offload` config** on replay is a **hard error** (never a silent GPU-only degradation) — INV-13, mirroring the #1587 `reconcileReplayKVOffload` seam.

## Scope decisions (called out for review)

- **Overall hit-rate, not per-tier.** `OffloadCache.CacheHitRate()` exposes a single aggregate (no per-tier `block_hits`), so the automated comparison is on the overall hit-rate. Per-tier `read_time`/`write_time` deltas are recorded in the header as *informational* data for the documented manual bandwidth cross-check.
- **Empirical cluster tolerance-pass is an operator step.** BLIS is CPU-only and the tiering counters require an unreleased vLLM on a GPU cluster, so the "BLIS meets ≤15% TTFT / ≤5 pp hit-rate on the named experiments" result cannot run in CI. This PR ships the full mechanism + the released-vLLM fallback + the documented method + pinned-commit recording — reported honestly, **not** claimed as passed.
- **`replay` gains `--metrics-path`.** The existing `TestReplayCmd_HasResultsPathFlag` asserted replay must *not* have `--metrics-path`. That boundary is now stale — the codebase already references "run/replay --metrics-path" (adapter-reference `--sim-metrics` help). The test is updated to assert the flag is present; `--results-path` (per-request) is unchanged.
- **`cache_hit_rate` is file-only, not on stdout.** Deliberate, to keep stdout INV-6-frozen without a broad golden regeneration; the human-readable rate is still printed by `printKVCacheMetrics`. The `--metrics-path` file is already a superset of stdout (`requests[]`).
- No new `go.mod` dependency — a small dependency-free Prometheus-text parser (`sim/workload/prom_hitrate.go`).

## Testing

`go build ./...`, `go test ./...` (all packages pass), and `golangci-lint run ./...` (0 issues) are green. New tests cover: Prometheus parse + hit-rate derivation (tiered/fallback/gauge/reset/zero-queries), header round-trip + strict-parse + zero-hit-rate preservation, `EmitOutput` file-only + stdout-absence, the BC-10 guard, INV-13 run/replay `cache_hit_rate` parity, and the calibrate present/skip paths. A CLI smoke confirms the file-only field and stdout byte-identity for both `run` and `replay`.

Fixes #1583.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,9 +117,11 @@ go build -o blis main.go
 # the observed hit-rate (vllm:kv_offload_tiering_block_hits/_block_queries, or the GPU
 # prefix-cache fallback) in the trace header for downstream calibrate. --vllm-commit
 # records the pinned (unreleased, PR #48798) vLLM the tiering counters require. A scrape
-# miss warns and omits the block (never fatal). See docs/guide/kv-offload-calibration.md.
+# miss warns and omits the block (never fatal). observe is a black-box dispatcher — it
+# does NOT take --kv-offload-config; the offload config is supplied on the replay step
+# (below). See docs/guide/kv-offload-calibration.md.
 ./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
-  --kv-offload-config offload.yaml --workload chatbot --rate 10 --num-requests 100 \
+  --workload chatbot --rate 10 --num-requests 100 \
   --scrape-kv-metrics --vllm-commit 63a9a5010a \
   --trace-header trace.yaml --trace-data trace.csv
 
@@ -131,8 +133,13 @@ go build -o blis main.go
 # MetricsOutput from `blis replay --metrics-path`, which carries cache_hit_rate). The
 # report gains a hit_rate block with abs_error_pp and a within-tolerance verdict
 # (default 5 pp); a TTFT-MAPE verdict uses --ttft-mape-threshold (default 0.15).
-# Skipped with a warning when --sim-metrics is absent or lacks cache_hit_rate.
+# Skipped with a warning when --sim-metrics is absent or lacks cache_hit_rate. For a
+# tiered observed hit-rate, supply --kv-offload-config on replay to model the observed
+# deployment (observe traces are exempt from the "cannot add offload on replay" rule;
+# sim-generated traces stay header-authoritative). Replaying a tiered observation with
+# no offload config is a hard error (never a silent GPU-only value).
 ./blis replay --trace-header t.yaml --trace-data d.csv --model qwen/qwen3-14b \
+  --kv-offload-config offload.yaml \
   --results-path results.json --metrics-path simagg.json
 ./blis calibrate --trace-header t.yaml --trace-data d.csv --sim-results results.json \
   --sim-metrics simagg.json --hit-rate-tolerance-pp 5 --ttft-mape-threshold 0.15 \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,8 +112,31 @@ go build -o blis main.go
   --record-itl --itl-output trace.itl.csv \
   --trace-header trace.yaml --trace-data trace.csv
 
+# Observe with KV cache hit-rate scraping (#1583). --scrape-kv-metrics scrapes the
+# server's Prometheus /metrics at the start and end of the measured window and records
+# the observed hit-rate (vllm:kv_offload_tiering_block_hits/_block_queries, or the GPU
+# prefix-cache fallback) in the trace header for downstream calibrate. --vllm-commit
+# records the pinned (unreleased, PR #48798) vLLM the tiering counters require. A scrape
+# miss warns and omits the block (never fatal). See docs/guide/kv-offload-calibration.md.
+./blis observe --server-url http://localhost:8000 --model qwen/qwen3-14b \
+  --kv-offload-config offload.yaml --workload chatbot --rate 10 --num-requests 100 \
+  --scrape-kv-metrics --vllm-commit 63a9a5010a \
+  --trace-header trace.yaml --trace-data trace.csv
+
 # Compare real observed latencies against simulator predictions
 ./blis calibrate --trace-header t.yaml --trace-data d.csv --sim-results results.json --report calibration.json
+
+# Compare KV cache hit-rate too (#1583). The real hit-rate comes from the trace header
+# (observe --scrape-kv-metrics); the sim hit-rate comes from --sim-metrics (a
+# MetricsOutput from `blis replay --metrics-path`, which carries cache_hit_rate). The
+# report gains a hit_rate block with abs_error_pp and a within-tolerance verdict
+# (default 5 pp); a TTFT-MAPE verdict uses --ttft-mape-threshold (default 0.15).
+# Skipped with a warning when --sim-metrics is absent or lacks cache_hit_rate.
+./blis replay --trace-header t.yaml --trace-data d.csv --model qwen/qwen3-14b \
+  --results-path results.json --metrics-path simagg.json
+./blis calibrate --trace-header t.yaml --trace-data d.csv --sim-results results.json \
+  --sim-metrics simagg.json --hit-rate-tolerance-pp 5 --ttft-mape-threshold 0.15 \
+  --report calibration.json
 
 # Compare with ITL metric included (requires observe --record-itl)
 ./blis calibrate --trace-header t.yaml --trace-data d.csv --sim-results results.json \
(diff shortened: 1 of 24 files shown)


namasl:

namasl:

namasl:

namasl:

sriumcp (approved):
lgtm

PR #1599, opened 2026-08-19
feat(kv-offload): non-linear FS device model — queue-depth ramp, O_DIRECT/buffered regime, seeded jitter (#1581)

### LINKED ISSUE #1581: [Tiered KV Cache] PR4 — model lower-level filesystem device dynamics
**Part of #1585** — multi-tier KV offload (primary tracking). Slotted as **S7** in #1585 §9.
**Depends on:** H2 transfer station (#1588). **Extends H2:** its BC-S3 fixes
`service_time = base + bytes/bandwidth` (linear); this adds the non-linear device regime —
queue-depth-dependent bandwidth saturation and a latency distribution. Read/write asymmetry is
exercised once H1 (#1590) adds write cost.

### What
Refine the storage tier's simple `base_latency + bytes/bandwidth` cost into a **device model**:
effective bandwidth that depends on I/O parallelism / queue depth, **read vs. write asymmetry**,
direct (`O_DIRECT`) vs. buffered I/O, and an optional latency **distribution** rather than a single
mean.

### Why it matters — and why it is genuinely additive to H2, not a duplicate

The QD=1 linear cost is only the uncontended case. Real filesystems saturate bandwidth only at some
queue depth. Two source facts make this a separate concern from H2's queue:

1. **vLLM's own transfer-time metric excludes queue wait.** In the worker loop, the task is
   dequeued *first* and the clock starts *after* —
   [`fs/thread_pool.py:176-181`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/thread_pool.py#L176-L181):

   ```python
   task, state = primary.popleft() if primary else secondary.popleft()   # L176 — dequeue
   try:
       start_time = time.monotonic()                                     # L178 — clock starts here
       task()
       transfer_time = time.monotonic() - start_time                     # L180
       job_finished, success, total_time = state.task_done(True, transfer_time)
   ```

   So `vllm:kv_offload_tiering_read_time`/`write_time`
   ([`tiering/base.py:42-45`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/base.py#L42-L45)) is **device service
   time**, not end-to-end latency. H2's queue supplies the wait; this issue supplies the service
   curve. Together they reproduce observed latency; neither alone does.

2. **`O_DIRECT` is probed at runtime with a warned buffered fallback**, so a BLIS run cannot know
   which physics applied unless it is an explicit axis —
   [`fs/manager.py:185-195`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L185-L195):

   ```python
   # Prefer O_DIRECT to bypass the page cache, but fall back to buffered
   # I/O on filesystems that reject it (e.g. overlayfs, some NFS mounts)
   # rather than failing every block.
   self._use_o_direct = probe_o_direct(os.path.dirname(config_path))
   if not self._use_o_direct:
       logger.warning("O_DIRECT is not supported at '%s'; falling back to buffered I/O ...")
   ```

   This is #1585 §6 "trap 2" — vLLM *discovers* it, so BLIS must **declare** it.

Concurrency is bounded by the pool's thread counts, both defaulting to **16**
([`fs/manager.py:115-116`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/tiering/fs/manager.py#L115-L116); documented in
[`docs/features/kv_offloading_usage.md:128-129`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/docs/features/kv_offloading_usage.md#L128-L129)),
which is the queue depth this device curve is a function of.

### Model shape (concrete, so tests can assert numbers)
- **Bandwidth vs. queue depth:** linear ramp from a single-transfer bandwidth up to a
  **configurable saturation queue depth** `Qsat`, flat thereafter. Assert: effective per-transfer
  bandwidth at concurrency `q ≤ Qsat` scales as modeled, and is flat for `q > Qsat`. The relevant
  `q` is H2's in-service job count, which is capped per tier/direction by the 16/16 thread defaults
  above.
- **Read/write asymmetry:** separate read and write bandwidths (config from H5 #1587); the write
  side is only *exercised* once write cost lands in H1 #1590 — until then it is config/plumbing.
  This is already required by #1588's BC-S3 ("read != write representable"); this issue makes the
  two curves non-linear, it does not introduce the split.
- **`O_DIRECT` vs. buffered:** an explicit config regime selecting a different
  `(bandwidth, base_latency)` pair. Per #1585 §5 the flag budget is **one** new flag
  (`--kv-offload-config`), so this is a field in that YAML's device class — **not** a new flag.
- **Latency distribution:** optional jitter around the mean, **sampled from the existing seeded
  RNG** ([`sim/rng.go`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/rng.go#L17)) so `blis run`/`blis replay` stay deterministic and
  byte-identical (INV-6, INV-13).

### Scope note (from #1585 §6 / BC-S3)
#1588's BC-S3 is deliberately linear — its evidence line reads *"service time linear in bytes, per
direction; read != write representable."* Absorbing the non-linear regime into calibration (#1583)
instead of modeling it is a **legitimate** alternative, and #1585 §11 takes that position for
per-tier metrics generally. This issue exists so that choice is explicit rather than accidental. If
calibration (#1583) shows the linear term suffices at the queue depths the cluster runs use, close
this as won't-fix and record the evidence.

### Cross-path parity
`sim/kv/` DES kernel → both `blis run` and `blis replay` via
[`kv.NewKVStore`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/cluster/instance.go#L56) (`sim/cluster/instance.go:56`);
[parity test](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/cmd/parity_run_replay_test.go#L360) must stay green — the distribution **must**
draw from the seeded RNG for this to hold. `observe` — **not affected**. External tools for
characterizing these curves: [`fio`](https://github.com/axboe/fio),
[`O_DIRECT` (open(2))](https://man7.org/linux/man-pages/man2/open.2.html).

### Done when
- Storage bandwidth scales with modeled queue depth up to `Qsat`, then is flat (asserted with
  concrete numbers).
- Read and write use separate bandwidths; `O_DIRECT` vs. buffered is a selectable regime with
  distinct parameters, delivered as YAML fields under the single `--kv-offload-config` flag.
- If the latency distribution is enabled, sampling is seeded and run/replay remain byte-identical
  (parity test green).

namasl:
## Summary

Refines the KV-offload **storage tier**'s linear `base + bytes/bandwidth` service cost (#1588 BC-S3) into a non-linear **filesystem device model** (S7 of the multi-tier KV-offload epic #1585). Three additive, **opt-in** layers, all gated so the default (no device-model config) is byte-identical to pre-#1581 (INV-6):

1. **Queue-depth bandwidth ramp** (deterministic, in the transfer station): effective per-transfer bandwidth ramps linearly from `f₁·bw` at the tier-direction's in-service depth `q=1` up to the peak `bw` at `q=Qsat`, flat beyond. `q` is fixed at service-start so `completeAt`/BC-S4 stay stable.
2. **O_DIRECT vs buffered regime**: each `kv_offload_devices` device class may declare a buffered `(bandwidth, base_latency[, Qsat, f₁, σ])` set alongside its O_DIRECT set; the per-tier `direct_io` bool (captured by #1587) now **selects** the regime, buffered falling back per-field to the O_DIRECT value. Delivered as YAML fields under the single `--kv-offload-config` flag — **no new flag** (epic §5).
3. **Optional seeded latency jitter**: a per-device relative stddev `σ` makes `OffloadCache` draw a multiplicative factor `max(0.05, 1+N(0,σ))` from a dedicated seeded RNG partition (`SubsystemKVOffload`, derived from the run seed) and pass it on the `TransferJob`. The transfer station **draws no randomness itself** — its no-RNG determinism guarantee (BC-S4) is preserved; it consumes a caller-supplied scalar. `σ=0` (default) draws nothing.

The three resolved fields round-trip through the trace header so `run`→`replay` under the same seed is byte-identical (INV-13). The committed device classes are unchanged (ramp/jitter/buffered all absent) — tuned curves are calibration's job (#1583, per the issue's scope note).

### Why the station stays RNG-free
`sim/kvtransfer` imports only `container/heap`+`fmt` and advertises "no RNG" as its determinism guarantee. The **queue-depth ramp** needs `q` (known only at service-start inside the station), so it lives there and stays deterministic. The **jitter** needs an RNG; putting it in the station would break that load-bearing invariant, so the draw happens in `OffloadCache` (on the seeded side of the system) and the station receives only a scalar `JitterFactor`.

## Behavioral contracts

- **BC-D1** — queue-depth ramp: `effBW(q) = bw·(f₁ + (1−f₁)·(min(q,Qsat)−1)/(Qsat−1))` for `q≤Qsat`, `bw` for `q≥Qsat`; asserted with concrete numbers (`bw=10, base=100, Qsat=4, f₁=0.5, bytes=1000` → q=1:300, q=2:250, q=3:220, q=4:200, q=8:200) + a monotonicity law.
- **BC-D2** — ramp/jitter off (`Qsat≤1`, `f₁≥1`, or `σ=0`, the default) ⇒ byte-identical to the pre-#1581 formula (INV-6); exact integer path retained (overflow-clamp edge preserved).
- **BC-D3** — `direct_io` selects O_DIRECT vs buffered; buffered falls back per-field to O_DIRECT; distinct params ⇒ distinct service times.
- **BC-D4** — read/write asymmetry preserved (ramp indexed by each direction's own in-service count); #1588 discriminating metamorphic test stays green.
- **BC-D5** — seeded jitter: `OffloadCache` draws `max(0.05, 1+N(0,σ))`; σ=0 draws nothing (RNG stream untouched); a σ>0 tier without an RNG is a construction error.
- **BC-D6** — jitter run/replay + run-twice parity (INV-13/INV-6): same seed ⇒ byte-identical per-request TTFT/E2E.
- **BC-D7** — config surface: new fields validated (`f₁∈(0,1]` when `Qsat≥2`; `Qsat≥1`; `σ≥0`), round-trip losslessly through the trace header (BC-G6), strict-YAML; BC-G4 static guard preserved.

## Testing

- `sim/kvtransfer`: ramp numbers, monotonicity, ramp-off identity, jitter-factor application, no-jitter byte-identity, BC-S4 determinism with factors; fuzz now exercises the ramp+jitter path (1M+ execs, BC-S1/conservation/BC-S4 hold).
- `sim`: `KVOffloadTier` validation table (reject + positive cases), BC-G4 static guard.
- `cmd`: device-class regime selection (O_DIRECT/buffered/fallback), reject table, header round-trip (both regimes), end-to-end `defaults.yaml`→resolver parse, `FuzzKVOffloadRoundTrip`.
- INV-13: `TestINV13_RunReplayParity_Offload_Jitter` (ramp+jitter active, run==replay). INV-6: `TestINV6_OffloadJitter_RunTwiceIdentical`.
- Full suite `go test ./... -count=1` green; `golangci-lint run ./...` 0 issues; `-race` on `sim/kvtransfer`+`sim/kv` green.

## Notes
- `NewKVStore(cfg)` gains a `seed` param (single prod call site; derives the jitter partition). `NewOffloadCache` gains a `WithOffloadRNG` functional option (no test-caller churn).
- Observability of the model is via completion-timing → hit/miss/recompute → token counts (H1 still charges no explicit per-request transfer latency; that is H3 #1591).

Closes #1581.


namasl:

namasl:

namasl:

sriumcp (approved):
lgtm

PR #1600, opened 2026-08-19
feat(kv): model offload_prompt_only as an explicit prompt-only vs decode-block offload policy (#1584)

### LINKED ISSUE #1584: [Tiered KV Cache] PR-POLICY — model prompt-only vs decode-block offload (offload_prompt_only)
**Part of #1585** — multi-tier KV offload (primary tracking). Slotted as **S8** in #1585 §9.
**Depends on:** H1 tier chain (#1590) and H4 keys (#1589).
**Realizes the behavior behind H5's `OffloadScope: promptAndDecode`** — #1587 declares it as a config
knob; this is the BLIS work that makes the non-default actually change hit-rate (decode blocks are
unhashed today, so it touches H4 and H1). Scope: whatever tiers exist after H1 (#1590).

### What
Model vLLM's `offload_prompt_only` behavior — prompt KV is offloaded to lower tiers, decode KV is
not (unless configured) — as an explicit, configurable policy in BLIS.

### Important: current-state correction
BLIS on `main` **already effectively offloads prompt-only, incidentally**: pure decode blocks never
receive a hash (a block is only hashed when `req.ProgressIndex < req.InputLen()`, because decode
blocks "should not participate in prefix caching" — [`sim/kv/cache.go:352-359`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/cache.go#L352-L359)),
and [`MirrorToCPU` skips hashless blocks](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/tiered.go#L407-L411) (`sim/kv/tiered.go:407-411`).
So the **default is already met** — but by accident, not by an explicit policy, and
[`KVBlock`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/cache.go#L24-L34) carries no prompt/decode marker. This PR replaces "gate
`MirrorToCPU`" (its original, incorrect framing) with the three real deliverables below.

### Why it matters
`offload_prompt_only` defaults to `True` — [`kv_offload/base.py:597-598`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/v1/kv_offload/base.py#L597-L598) —
and determines *what is even in the tiers*, which drives hit-rate and tier traffic; central to the
agentic multi-turn workloads in #1569. Documented default is also `true`
([`docs/features/kv_offloading_usage.md:78`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/docs/features/kv_offloading_usage.md#L78)).
None of the #1569 writeups covered it.

### RESOLVED: vLLM's boundary rule is truncate-then-floor-divide, not per-block classification

The original deliverable 2 asked to "confirm against vLLM's block-level prompt-range rule."
**There is no block-level classification rule.** Read at `63a9a5010a`, enforcement is a two-step
token-count truncation followed by floor division into chunks:

1. **Truncate the offloadable token count to the prompt length** —
   [`offloading/scheduler.py:588-597`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py#L588-L597):

   ```python
   def _calc_num_offloadable_tokens(
       self, req_status: RequestOffloadState, num_computed_tokens: int
   ) -> int:
       num = min(num_computed_tokens, req_status.req.num_tokens)
       max_offload_tokens = req_status.max_offload_tokens
       if max_offload_tokens is not None:
           num = min(num, max_offload_tokens)
       if self.config.offload_prompt_only:
           num = min(num, req_status.req.num_prompt_tokens)   # <-- the whole mechanism
       return num
   ```

2. **Floor-divide that count into whole chunks** —
   [`offloading/scheduler.py:401`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py#L401)
   inside [`storable_chunks`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py#L382-L407):

   ```python
   num_chunks = num_offloadable_tokens // group_config.tokens_per_chunk
   ```

**Consequence:** a chunk that would contain even one decode token is **never formed** under
prompt-only. It is not classified as decode — it is truncated away by integer division. A prompt
whose length is not a whole multiple of `tokens_per_chunk` simply leaves its partial tail
un-offloaded.

vLLM's own test asserts exactly this shape — a 2-chunk prompt plus 4 chunks' worth of decode
tokens offloads GPU block offsets `(0..5)` only, and `prepare_store` is never offered more than
`num_prompt_blocks` distinct keys
([`tests/.../test_scheduler.py:1827-1874`](https://github.com/vllm-project/vllm/blob/63a9a5010a6d1539c52957646ef9d6bbcf7a4deb/tests/v1/kv_connector/unit/offloading_connector/test_scheduler.py#L1827-L1874)).

**So deliverable 2 changes:** the proposed default ("any decode token ⇒ classified decode") is
*behaviorally equivalent to vLLM at chunk granularity*, but it is the wrong **mechanism** to
implement, and #1585 §4 is explicit that mechanism fidelity is the point (BC-C9 / D-series). Model
the truncation directly: compute an offloadable token count, clamp it to `InputLen()` when
prompt-only, then floor-divide by the chunk stride from H4 (#1589). A per-block prompt/decode
predicate would coincide with vLLM only while chunk stride == block size, and diverge silently as
soon as H4 lands `blocks_per_chunk > 1`.

### Deliverables
1. **Make prompt-only explicit and asserted**, not incidental: add the offloadable-token clamp
   above as the single decision point, and a config knob (default **true**, mirroring vLLM) wired
   through H5's `OffloadScope` (#1587) rather than as a new standalone flag — #1585 §5 fixes the
   flag budget at **one** new flag (`--kv-offload-config`), so this must be a field in that YAML,
   **not** a `--kv-offload-prompt-only` flag as this issue originally proposed.
2. **Implement the truncate-then-floor-divide rule** (replaces the old "classify the boundary
   block" deliverable, now resolved above). Test the partial-tail case explicitly: a prompt of
   `1.5 × tokens_per_chunk` offloads exactly 1 chunk, and the half-chunk tail is never offered for
   store. Note the interaction with BLIS's [partial-fill hash path](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/cache.go#L296-L306)
   (`sim/kv/cache.go:296-306`), which *does* hash a prompt block completed by decode tokens — under
   the truncation rule such a block is simply outside the offloadable range, so no new
   classification is needed, but the test must prove that.
3. **Support the non-default (offload decode KV).** This is the larger part and **reverses a
   deliberate design decision**: decode blocks currently get no hash and reload only queries
   input-prefix hashes ([`reloadPrefixFromCPU`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/kv/tiered.go#L268), `sim/kv/tiered.go:268`).
   Enabling decode offload that actually changes hit-rate requires assigning prefix-consistent
   hashes to decode blocks and permitting their reuse. Scope this explicitly (it may warrant its
   own follow-up PR).

### Cross-path parity
`sim/kv/` DES kernel → both `blis run` and `blis replay` via
[`kv.NewKVStore`](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/sim/cluster/instance.go#L56) (`sim/cluster/instance.go:56`);
[parity test](https://github.com/inference-sim/inference-sim/blob/52161669e133afc100f0d99223dde933c8b1b924/cmd/parity_run_replay_test.go#L360) must stay green. `observe` — **not affected**.

### Done when
- The offloadable-token clamp is the single decision point; the scope knob (default prompt-only)
  is honored from the H5 YAML; a test asserts decode blocks are not offloaded under the default.
- **Partial-tail test:** a prompt of `1.5 × tokens_per_chunk` offloads exactly 1 chunk.
- **Testable scenario:** a multi-turn workload where turn N+1's input contains turn N's output —
  with prompt-only (default), those tokens produce **no** CPU hit; with decode-offload enabled,
  they **do** (hit-rate reflects the policy).
- Parity test green.

namasl:
## Summary

Fixes #1584 (S8 of the multi-tier KV-offload epic #1585). Makes vLLM's `offload_prompt_only` an **explicit, tested, and documented** policy in the KV-offload chain (`sim/kv.OffloadCache`), replacing the incidental per-block predicate that H1 (#1590) left in `MirrorToCPU`.

The offloadable-block decision is now a single **offloadable-token clamp** that models vLLM's actual mechanism (`_calc_num_offloadable_tokens` + `storable_chunks`), rather than a per-block prompt/decode classification:

```
numOffloadableTokens = len(ownedFullBlocks) * blockSize
if offload_prompt_only { numOffloadableTokens = min(_, InputLen()) }   // vLLM prompt-only clamp
maxOffloadBlocks = (numOffloadableTokens / tokensPerChunk) * blocksPerChunk   // storable_chunks floor-divide
```

At `blocks_per_chunk == 1` (the only value H1 supports) this is **byte-identical to the prior code for both policy values, in all phases** (mid-prefill and post-prefill), so it is a pure mechanism-fidelity refactor — existing offload tests and parity stay green (INV-6 / INV-13).

### Notable finding (recorded honestly)
The issue's premise that "decode blocks currently get no hash" is **incorrect for `block_size > 1`**. The `cache.go:352` guard lives in the *new-block* path, which never forms a full decode block (decode advances one token per step); decode blocks are completed — and hashed prefix-consistently — via the **partial-fill path** (`cache.go:288-305`), which has no such guard. Verified empirically. Consequences:

- **No new decode-hashing code is required.** Once the clamp admits decode blocks under `offload_prompt_only: false`, they are already hashed, so `MirrorToCPU` offloads them and a later same-instance request whose *input* contains those output tokens reloads them via the existing `consultAndReload`.
- `MirrorToCPU` remains a **pure consumer** of `gpu.HashToBlock` (it never writes it), so switching the policy cannot perturb GPU-tier prefix-cache behavior — default and prompt-only runs stay byte-identical (INV-6). This drops the originally-planned `assignDecodeHashes` helper (the highest-risk change) entirely.
- Edge: at `block_size == 1` decode blocks take the guarded new-block path and are not hashed, so decode offload is inert — a degenerate offload config (real offload block sizes track the GPU block size, typically 16). Documented, not handled.

The config knob already exists (`KVOffloadConfig.OffloadPromptOnly`, default `true`) and round-trips through the trace header, so **no new flag** is added (epic §5's one-flag budget). Naming note: the issue/epic call it `OffloadScope: promptAndDecode`; H5 (#1587) landed a boolean `offload_prompt_only` (`true` = prompt-only, `false` = promptAndDecode).

## Behavioral Contracts

- **BC-1 (single decision point):** the offloadable amount is a single token-count clamp floor-divided into whole chunks — not a per-block classification.
- **BC-2 (prompt-only default honored):** under the default, no decode-generated block is offloaded; byte-identical to the pre-PR binary (INV-6).
- **BC-3 (partial tail):** a prompt of `1.5 × tokens_per_chunk` offloads exactly 1 chunk; the tail block completed and hashed by decode is outside the offloadable range.
- **BC-4 (decode offload changes hit-rate):** with `offload_prompt_only: false`, a later same-instance request whose input contains earlier output tokens gets a tier hit for that region; under `true` it does not. Reuse is single-instance (offload tiers are per-instance, invisible to the router).
- **BC-5 (isolation / INV-6):** `MirrorToCPU` never writes `gpu.HashToBlock` under either policy.
- **BC-6 (run/replay parity, INV-13):** parity holds for `offload_prompt_only: false`.

## Testing

- `go build ./...`, `go test ./... -count=1`, `golangci-lint run ./...` — all green.
- New tests (`sim/kv/offload_decode_reuse_test.go`, `sim/kv/offload_mirror_test.go`):
  - `TestOffload_OffloadableTokenClamp_SingleDecisionPoint` (BC-1)
  - `TestOffload_PromptOnly_PartialTail_OffloadsExactlyOneChunk` (BC-3)
  - `TestOffload_DecodeBlockGetsPrefixConsistentHash` (pins the partial-fill hashing the reuse depends on)
  - `TestOffload_DecodeOffload_ResidencyByPolicy` (BC-2/BC-4 offload side)
  - `TestOffload_MultiTurn_OutputReuse_HitByPolicy` (BC-4 end-to-end: 1 reload under `true`, 2 under `false`)
  - `TestOffload_MirrorToCPU_DoesNotMutateGPUHashes` (BC-5)
  - `TestINV13_RunReplayParity_Offload_DecodeOffload` (BC-6)
- Existing offload/parity tests unchanged and green.

## Done-when (issue #1584)
- [x] Offloadable-token clamp is the single decision point; scope knob honored; a test asserts decode blocks are not offloaded under the default.
- [x] Partial-tail test: `1.5 × tokens_per_chunk` → exactly 1 chunk.
- [x] Multi-turn scenario: prompt-only → no CPU hit for reused output; decode-offload → hit.
- [x] Parity test green.


namasl:

namasl:
Thanks — addressed all three findings in `3455bd3a`:

- **Important (stale godoc):** removed the `assignDecodeHashes` reference from the `MirrorToCPU` godoc; it now describes the actual mechanism (decode blocks are already hashed prefix-consistently by the GPU tier's partial-fill path; `MirrorToCPU` never writes `gpu.HashToBlock`).
- **Low (test hashes):** synthetic decode-block hashes are now 64-hex (SHA-256 shape) and distinct.
- **Low (bs==1 edge):** added `TestOffload_DecodeOffload_InertAtBlockSizeOne`, pinning that decode-KV offload is inert at `block_size == 1` (guarded new-block path leaves decode blocks unhashed) — a regression guard on the `cache.go:352` interaction with BC-4.

`go build ./... && go test ./... -count=1 && golangci-lint run ./...` all green.

namasl:

namasl:
Rebased onto `upstream/main` (now includes #1601 / H3 kv_deferral). Clean rebase — H3's deferral refactor of `offload_chain.go` and this PR's offloadable-token clamp coexist without conflict (H3 left `MirrorToCPU` untouched; the struct/constructor additions were in disjoint regions). Full suite + lint green post-rebase; offload, H3 deferral, and both-policy parity tests all pass together.

/archon-pr-review

namasl:

namasl:
Rebased onto `upstream/main` again (now includes #1598 / S6 KV hit-rate calibrate loop). Clean rebase — S6 touches the observe/replay/calibrate + metrics surface, disjoint from this PR's `sim/kv` offload change; only CLAUDE.md overlapped and auto-merged (different sections). Full suite + lint green post-rebase.

/archon-pr-review

namasl:
(archon's previous run hit a transient `HttpError: Unexpected end of JSON input` at the *Resolve PR refs* step — a GitHub-API hiccup right after the force-push, unrelated to the change. Re-triggering.)

/archon-pr-review

namasl:
Rebased again onto `upstream/main` (now includes #1599 / S7 non-linear FS device model). One semantic conflict surfaced and was resolved: #1599's new `offload_jitter_parity_test.go` called the shared `makeSharedPrefixRequests()` helper that this PR parameterized — its two call sites now pass the original `outputLen` (4). S7's jitter fields and this PR's `blocksPerChunk` auto-merged cleanly in the `OffloadCache` struct; `MirrorToCPU` untouched by S7. Full suite + lint green post-rebase (13 packages).

/archon-pr-review

namasl:

namasl:
Addressed the four round-4 low-severity items (comment/doc-only, no behavioral change; head `925c9bfc`): tightened the `blocksPerChunk` godoc (attaches to the field; dropped the over-strong blocks_per_chunk>1 forward-promise), corrected the decode-offload parity test comment (points to `TestKVOffload_EndToEnd_RunReplayRoundTrip` for the header round-trip), and documented the `block_size == 1` inert edge in kv-cache.md. Full suite + lint green.

/archon-pr-review

namasl:

sriumcp (approved):
lgtm

PR #1601, opened 2026-08-19
feat(kv): fill hole H3 kv_deferral — step-boundary re-poll (the dominant TTFT effect) (#1591)

### LINKED ISSUE #1591: S5 · Fill hole H3 kv_deferral: step-boundary re-poll (the dominant TTFT effect)

namasl:
## Summary

Fills hole **H3 `kv_deferral`** of the multi-tier KV-offload epic (#1585) — **the dominant TTFT effect**. Blocked by #1590 (H1, merged as #1597), now unblocked.

vLLM does **not** block on the disk when a request needs KV blocks resident only on a secondary tier (disk/object store). It sets the request aside and re-examines it on the **next scheduler step** (`vllm/v1/core/sched/scheduler.py:835-841` `continue`). So the offload-attributable TTFT delay is a **whole multiple of step time, not disk latency** — and because step time grows with batch size while disk latency does not, a bandwidth model drifts in the wrong direction.

This PR ports that mechanism. A new prefill admission whose prefix is secondary-resident is **deferred** (left in the WaitQ, re-polled each step) instead of promoted-and-recomputed immediately (the H1 behaviour). The delay is realized by **later admission**, not by any per-step latency term, and there is **no new `sim.Event`** — the existing step loop does the re-poll.

Closes #1591.

## Behavioral contracts

- **BC-T1** (differential): an unresolved request is re-examined at the **next `StepEvent`**, not at transfer completion — no re-poll/completion event is scheduled for it.
- **BC-T2** (metamorphic + discrimination): offload-attributable delay = `(k-1) × step`. ~3× longer steps ⇒ ~3× delay (slope ≈ 1); halving disk bandwidth ⇒ delay ≪ 2× (unchanged while `service ≪ step`). A bandwidth-only surrogate fails **both** relations — encoded as an explicit discrimination assertion.
- **BC-T3** (property): no request defers forever — a single fetch attempt bounded by the transfer station's completion guarantee, then admit-or-recompute (plus a load-independent defensive backstop, R19). Adversarial locked-tier configs terminate.
- **BC-T4** (round count): a **COLD** secondary hit costs `k ≥ 3` rounds (RETRY → promote → in-flight → HIT); a **WARM** hit `k ≥ 2` (existence resolved). Both modeled and **distinguished** (`cold − warm == 1`) via a persistent per-key existence cache.
- **BC-T5** (INV-8): a deferred request never leaves the simulator idle while work is waiting — deferred requests stay in the WaitQ, so `scheduleNextStep` keeps a `StepEvent` scheduled (or an `AdapterLoadCompletionEvent` is pending under a co-active LoRA load).
- **BC-T6** (differential): within a step, completions are polled **before** lookups (`SetClock` runs at the top of `scheduleBatch`); a promotion submitted this step is visible only next step (`SubmitTick = now`).
- **BC-T7** (design): step-quantization is the primary model — no per-request disk-bandwidth latency term (`ConsumePendingTransferLatency` stays 0).
- **P** (benchmark): re-poll cost is **O(deferred)**, not O(waitq) — measured ~linear in the deferred-set size.

## Design

- New narrow capability `sim.DeferrableKVStore` (`PollDeferred` / `IsDeferred` / `ClearDeferred`), implemented **only** by `kv.OffloadCache` and type-asserted by `FormBatch`. No new package-altitude arrow (the interface lives in `sim`, implemented in `sim/kv`).
- `FormBatch` Phase 2 becomes a **non-blocking skip-scan**: a still-deferred request is skipped and admittable requests behind it are admitted (vLLM's `step_skipped_waiting` + `continue`), enabling **concurrent** promotions through the #1588 N-server transfer station; GPU pressure still breaks head-of-line. The scan preserves the O(1) `DequeueBatch` head path so it is **byte-identical when offload is off (INV-6)**.
- `PollDeferred` advances the state machines in **sorted `Request.ID` order for its side effects** (station `Submit` assigns JobIDs by call order; `prepareStore` evicts CPU-LRU by call order) — the real determinism guard.
- Deferral is gated to **new prefill admissions only**: running continuations and PD decode-KV reservations keep the H1 background-promote path.
- `ClearDeferred` is wired into every non-admit WaitQ removal (timeout, gateway eviction, drain-redirect) so the deferred map never leaks.

## Invariants

- **INV-6** determinism: no RNG; sorted side-effect order; offload-off byte-identical (verified by the full existing suite).
- **INV-8** work-conserving: deferred requests remain in the WaitQ; verified end-to-end (the offload run fully drains).
- **INV-1** conservation: deferred = `still_queued` (in WaitQ), never lost or double-counted.
- **INV-13** run/replay parity: no new config field; the deferral is deterministic. Covered by the existing `cmd/offload_parity_test.go::TestINV13_RunReplayParity_Offload` (secondary tiers active), which passes.
- **INV-4** GPU conservation: GPU allocation only on admit/recompute; a cleared deferral's `-1` CPU slot completes normally.

## Testing

- `go test ./...` green; `go test ./sim/... -race` green; `golangci-lint run ./...` clean (v2.9.0).
- kv-level: cold/warm round count + delta (BC-T4), phase sequence, metamorphic step-vs-bandwidth with discrimination (BC-T2), bounded-then-recompute (BC-T3), re-poll at step boundary + completions-before-lookups (BC-T1/BC-T6), determinism under concurrent deferrals (INV-6), concurrent-cold convoy dedup (BC-C2), `ClearDeferred`, running-request and PD-decode-sub-request gating, `PollDeferred` benchmark (P).
- `sim`-level: `FormBatch` skip-scan (fake `DeferrableKVStore` double — skip / resolve / GPU-pressure-break / skip-then-break); `WaitQueue.PeekAt`.
- cluster-level end-to-end: a cycling-prefix churn workload evicts re-used prefixes to the secondary tier so later requests take the deferral path; asserts the run fully drains (INV-8/INV-1), is byte-identical across two same-seed runs (INV-6), and that deferrals actually fired (non-vacuous).

## Review notes

Self-review (multi-perspective plan review + code review + self-audit) caught and fixed two issues before opening: a concurrent-cold convoy miss (same-prefix requests now share one promotion) and a PD decode-KV reservation regression (decode sub-requests are excluded from the deferral gate so they are never dropped).


namasl:

sriumcp:

sriumcp (approved):
lgtm

Issue #1487, opened 2026-07-22
PR-D: OTel session-pool replay on blis observe (--corpus-header/--corpus-data + --concurrent-sessions)


namasl:
**Part of #1477 — PR-D of the stack.** Builds on **PR-C (#1480)**: reuses its `SessionPoolDriver` verbatim. Extends the OTel session-pool replay concept from `blis replay` (simulator) to `blis observe` (live server).

## What external behavior are we replicating?

The observe-side twin of PR-C: drive a captured OTel agentic corpus as a **fixed pool of N concurrent closed-loop sessions against a real inference server**, recording the observed timing. This is exactly what inference-perf does with `concurrent_sessions` — PR-C brought it to the simulator; PR-D brings it to the real-server observe path so `blis calibrate` can compare real vs simulated timing over the *identical* session set.

## Problem this PR solves

`blis observe` today can only build its workload from a generated `WorkloadSpec` (`GenerateWorkload`); it has **no path to accept a trace corpus as input**. So an OTel corpus converted via `blis convert otel` (PR-B) can be simulated (`blis replay --concurrent-sessions`, PR-C) but cannot be driven against a real server. Without PR-D, the calibration loop is one-sided: there is a simulated result but no matching observed result produced from the same corpus under the same pool semantics.

## Key finding (why this is small)

`blis observe` already runs closed-loop dispatch: a single-threaded serializer goroutine calls `SessionManager.OnComplete(req, tick) []*sim.Request` and routes follow-ups back through an internal channel (BC-8 pattern). `SessionPoolDriver.OnComplete` has the **identical signature**. So the pool driver slots in behind the same seam — no new event plumbing.

- Serializer seam: https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/cmd/observe_cmd.go#L959-L1006
- Active-session drain accounting (the one real interaction, below): https://github.com/inference-sim/inference-sim/blob/0e213340e353c8b0019805b367fb084087db3b68/cmd/observe_cmd.go#L1077-L1129
- `SessionPoolDriver.OnComplete` (reused; file is new in PR-C, so this links the fork branch): https://github.com/namasl/inference-sim/blob/1460de5b3b9703007cd7f5728f8f082ba4a28d1d/sim/workload/session_pool.go#L145-L166
[...]

PR #1605, opened 2026-08-24
feat(cmd): blis convert weka — Weka CC agentic trace JSONL to TraceV2 (PR-F of #1477)

### LINKED ISSUE #1604: PR-F: blis convert weka — Weka CC agentic trace JSONL to TraceV2 (linear)
**Part of #1477.** ⚠️ **Depends on PR-A (#1478)** (the `think_time_us` trace column + `session_context_growth` header field) **and PR-B (#1479)** (the shared session→TraceV2 encoder core). This PR adds a second `blis convert` front-end alongside `otel`; it does not touch the replay engine.

## What this PR adds

`blis convert weka` — a reader for the **SemiAnalysis WekaTrace** JSONL schema that produces the same normalized session the shared encoder (PR-B) turns into a TraceV2 corpus. The engine (accumulate fidelity #1478, session pool #1480, observe pool #1487) is source-agnostic and unchanged.

Datasets (Hugging Face, `apache-2.0`):
- [`semianalysisai/cc-traces-weka-062126`](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-062126) — v7, 393 traces, 98,827 model requests.
- [`semianalysisai/cc-traces-weka-with-subagents-051926`](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-with-subagents-051926) — v5, 219 traces, 78,099 model requests.

```bash
blis convert weka --input traces.jsonl --trace-output corpus \
  --context-growth accumulate --max-think-time 0
# writes corpus.yaml + corpus.csv (header: session_context_growth: accumulate)

  --model qwen/qwen3-14b --session-mode closed-loop \
  --max-model-len 1000000        # Weka ISL is huge — see caveat below
```

**Replay requirement (same as PR-B's OTel corpus).** The output is delta-encoded, so it MUST be replayed with `--session-mode closed-loop` — only that path runs the accumulate buffer that reconstructs each round's absolute input from the deltas. `blis replay` defaults to `--session-mode fixed`, which reads `InputTokens` as absolute per-round inputs and would misinterpret the deltas silently. PR-C's `--concurrent-sessions` auto-promotes to closed-loop.

## WekaTrace schema (input)

Each JSONL line is one session:

| field | type | notes |
|---|---|---|
| `id` | str | proxy session id → `TraceRecord.SessionID` (feeds session-affinity, cf. #1505) |
| `models` | list[str] | recorded models (metadata only) |
| `block_size` | int | 64 |
| `hash_id_scope` | str | `local` (per-trace block hashes) |
| `requests` | list | ordered main-agent requests **and** `type:"subagent"` groups |

Main-agent request (`type:"n"` non-streaming / `"s"` streaming):

| field | type | use |
|---|---|---|
| `t` | float | seconds since trace start (arrival) |
| `model` | str | dropped (routing safety — see below) |
| `in` | int | effective o200k input tokens. **v7 (062126):** `= len(hash_ids)×64` (block-quantized). **v5 (051926):** exact effective count, `≤ len(hash_ids)×64` (`len(hash_ids)=ceil(in/64)`). **Read `in` directly; never recompute from `hash_ids`.** |
| `out` | int | output tokens (Anthropic-reported) |
| `hash_ids` | list[int] | per-64-token-block hashes (**not consumed** this PR — future work) |
| `api_time` | float | real end-to-end server time (for calibration, not replay spacing) |
| `think_time` | float? | wall-clock gap from previous request's **end** (pure think) |
| `ttft` | float? | streaming only (calibration) |

Sub-agent group (`type:"subagent"`): `agent_id`, `subagent_type`, `duration_ms`, nested `requests[]`, … — **skipped by this converter; deferred to PR-E.**

## Proposed change

New `cmd/convert_weka.go` (subcommand) + `sim/workload/weka_convert.go` (reader). The reader parses one session, filters to the linear main-agent stream, and hands a normalized session to the PR-B encoder. Per session:

- **`SessionID`** = `id`; **`RoundIndex`** = 0..N over main-agent requests in stream order.
- **`OutputTokens`** = `out`.
- **`InputTokens`** = per-round **delta** via the shared encoder: round 0 = `in_0`; round N+1 = `max(0, in_{N+1} − in_N − out_N)`. Reuses PR-B's delta law verbatim (operates on `in` as-is under either dataset encoding; non-monotone rounds from context compaction clamp to 0, same documented deviation as OTel).
- **`Model`** = **left empty** (`""`). Same routing-safety contract as PR-B: `TraceRecord.Model` filters routable instances at replay, and Weka records `claude-*` which would never match `--model` → every request silently dropped. The recorded model name is dropped during conversion.
- **`ThinkTimeUs`** (the PR-A column) = **recomputed between consecutive main turns**: `think_i = max(0, round(t_i − t_{i-1} − api_time_{i-1}))·1e6`. Round 0 has no think. This is the pure post-response gap the user model wants ("think starts when the response is received"); it was verified to equal the recorded `think_time` to the microsecond on subagent-free sessions. Recomputing (rather than trusting the recorded `think_time` directly) keeps the main chain clean when sub-agent entries are removed, since a recorded `think_time` may reference a chronologically-preceding sub-agent inner request.
  - **Documented deviation:** in linear-only mode the wall-clock time a sub-agent group ran is absorbed into the *following* main turn's think gap (the main agent really was blocked on the Task tool). PR-E models the fan-out explicitly.
- **`ArrivalTimeUs`** = round-0 injection time = the **first main turn's `t`** (may be `>0` when the session's first `requests[]` entry is a sub-agent group, which this converter skips); follow-up spacing comes from `ThinkTimeUs` in closed-loop replay.
- **Header:** `Version: 3`, `Mode: "generated"`, `SessionContextGrowth: "accumulate"` (or `""` when `--context-growth independent`).

**Flags:** `--input` (JSONL), `--trace-output`, `--context-growth accumulate|independent` (default `accumulate`), `--max-think-time` (cap on the recomputed gap; default `0` = uncapped — Weka's gaps are genuine away-from-keyboard times, so uncapped is fidelity-preserving; a bounded-horizon capacity run may want a cap so a single multi-minute/hour away-gap doesn't idle a pooled session for that long of sim time — differs from OTel's 15s default), `--min-rounds` (default 1). (No `--include-errors`: Weka errors are filtered at source.)

## Caveats to document

- **Token scale + `in` encoding (differs by dataset version).** `in` is o200k-tokenizer (~1.67× Anthropic's own count); the sim runs under one `--model` with its own tokenizer, so absolute counts are approximate — same class of caveat as OTel's recorded-model ≠ sim-model. **`in` is the effective prompt-token count, encoded differently across the two datasets:** v7 (062126) block-quantizes to ×64 (`in = len(hash_ids)×64`; verified 131/131 on sample), v5 (051926) reports the exact effective count (`in ≤ len(hash_ids)×64`; verified 71/72 strictly less; `len(hash_ids)=ceil(in/64)`). **The converter must read `in` directly and never derive it from `len(hash_ids)`** — recomputing would over-count up to 63 tokens/request on v5.
- **Huge ISL.** p50 ≈ 110K, p90 ≈ 395K, p99 ≈ 800K input tokens. Default `--max-model-len` (~41K) drops every request as unservable (`completed_requests: 0`). Callers must raise `--max-model-len` and scale `--total-kv-blocks`. Add a doc admonition (mirrors the existing Exgentic note).
- **Sub-agents skipped.** `type:"subagent"` entries are dropped; their token/latency content is not replayed (only their wall-clock is absorbed into the following main turn's think). Deferred to PR-E.

- [x] Workload generation (`sim/workload/`) — new `weka_convert.go`
- [x] CLI (`cmd/`) — new `convert_weka.go`

**Cross-path parity:** convert-only — emits TraceV2 consumed unchanged by `blis replay` (and `blis observe` corpus-mode via PR-D). No `blis run` equivalent is needed (converters target replay, exactly like `convert otel`); the replay engine and all other command paths are untouched.

## Extension friction check

- **New files:** `sim/workload/weka_convert.go`, `cmd/convert_weka.go` (+ tests). **Modified:** none beyond registering the subcommand and (in PR-B) exporting the shared encoder.
- **Invariants:** INV-3 (think clamped ≥ 0). INV-6 determinism: JSONL sessions consumed in file order; global request IDs assigned in order; `think_time_us` populated deterministically.

- [ ] `ConvertWekaSession` unit test: delta reconstruction law holds (monotone case); `RoundIndex` 0..N over main turns only; `type:"subagent"` entries excluded from the emitted records.
- [ ] **Think-time test:** recomputed `ThinkTimeUs` equals `max(0, t_i − t_{i-1} − api_time_{i-1})` (µs) and equals the recorded `think_time` on a subagent-free fixture; round 0 has no think.
- [ ] **Non-monotone clamp test:** `in_{N+1} < in_N + out_N` yields `delta == 0` (never negative).
- [ ] **Model-field routing-safety test:** emitted `TraceRecord.Model` is empty → replay under a differing `--model` produces completions (`routingRejections == 0`).
- [ ] **Sub-agent skip test:** a session with `type:"subagent"` groups converts using only main-agent turns; the following main turn's think gap absorbs the group's wall-clock (asserted, documented).
- [ ] **All-subagent / 0-main-turn session:** a session whose `requests[]` holds only `type:"subagent"` entries (0 usable main turns) is **skipped via `--min-rounds`** — the converter emits no records and never panics or writes an empty session. (Won't occur in these datasets given the ≥20-main-turn source filter, but the reader must handle it.)
- [ ] **Negative-gap think-clamp test:** overlapping main turns (`t_i < t_{i-1} + api_time_{i-1}`, a negative raw gap — real in these traces) yield `ThinkTimeUs == 0`, never negative (INV-3), mirroring the input-delta clamp test.
- [ ] End-to-end: written TraceV2 loads via `LoadTraceV2`; header carries `session_context_growth: accumulate`; **closed-loop replay** reconstructs each monotone session's recorded per-round `in` totals via the accumulate buffer, and round N+1 arrives at `completion_sim[N] + ThinkTimeUs` (INV-10).
- [ ] Manual smoke: a byte-range slice of a real `traces.jsonl` → `blis convert weka` → replay `--session-mode closed-loop --max-model-len 1000000` → confirm completions and per-round think spacing.

## Out of scope (this PR)

- **Sub-agent fan-out** (`type:"subagent"`) → PR-E (arrival-offset / DAG session admission).
- **`hash_ids`-driven prefix sharing** (exact block reuse, incl. cross-session parent→subagent) → future KV-model feature; this PR ignores `hash_ids` and relies on the accumulate buffer's approximate growing prefix.

Tracked by #1477. Blocked by PR-A (#1478) and PR-B (#1479). Produces a TraceV2 corpus consumed unchanged by PR-C (#1480) / PR-D (#1487). Bonus: Weka's real `api_time`/`ttft` give `blis calibrate` a ground-truth latency target the OTel corpus could not (fake `end_time`).

namasl:
## Summary

Adds **`blis convert weka`** — a second `blis convert` front-end (alongside `otel`) that reads the **SemiAnalysis WekaTrace** JSONL schema and produces the same normalized session the shared PR-B encoder (`EncodeSessionToTraceRecords`, #1479) turns into a TraceV2 corpus. The replay engine (accumulate fidelity #1478, session pool #1480, observe pool #1487) is source-agnostic and **unchanged** — this is a convert-only PR.

Depends on (both merged into `main`): **PR-A #1478** (`think_time_us` column + `session_context_growth` header) and **PR-B #1479** (shared session→TraceV2 encoder). Part of #1477.

```bash
blis convert weka --input traces.jsonl --trace-output corpus \
  --context-growth accumulate --max-think-time 0
  --model qwen/qwen3-14b --session-mode closed-loop --max-model-len 1000000
```

## Behavioral contracts

- **BC-1 (delta reconstruction).** GIVEN a monotone Weka session, WHEN converted, THEN round 0's `InputTokens == in_0` and round `k>0` = `max(0, in_k − in_{k-1} − out_{k-1})`; the accumulate buffer reconstructs each round's absolute input exactly. (Reuses the merged encoder's delta law verbatim.)
- **BC-2 (sub-agent exclusion).** `requests[]` is filtered to the linear main-agent stream; `type:"subagent"` groups are skipped (deferred to PR-E). `RoundIndex` is 0..N over main turns in stream order; a skipped group's wall-clock is absorbed into the *following* main turn's think gap.
- **BC-3 (think recompute).** `ThinkTimeUs[k] = round((t_k − t_{k-1} − api_time_{k-1})·1e6)` for `k>0`, `0` for round 0; on a subagent-free chain this equals `round(think_time_k·1e6)`. Recomputing (not trusting the recorded `think_time`) keeps the main chain clean when sub-agent entries are removed.
- **BC-4 (clamp, INV-3).** Overlapping main turns (negative raw gap — real in these traces) → `ThinkTimeUs == 0`, never negative.
- **BC-5 (cap).** `--max-think-time D>0` caps the recomputed gap; `0` (default) leaves it uncapped (Weka gaps are genuine away-from-keyboard times).
- **BC-6 (`in` read directly).** `in` is read directly, never derived from `len(hash_ids)×64` (would over-count ≤63 tok/req on the v5 encoding).
- **BC-7 (Model routing-safety).** Every emitted `TraceRecord.Model == ""` so requests inherit `--model` at replay instead of being dropped at routing (Weka records `claude-*`).
- **BC-8 (min-rounds skip).** Below `--min-rounds` (incl. all-`subagent` / 0-main-turn) → `(nil, nil)`, no records, no panic.
- **BC-9/BC-10 (corpus).** `independent`→empty header; sessions consumed in file order with global request IDs assigned in order (INV-6).

## Testing

- `sim/workload/weka_convert_test.go` — 8 unit tests: delta reconstruction + sub-agent skip, think recompute (== formula AND == recorded `think_time`, with sub-agent wall-clock absorption), negative-gap clamp, max-think cap/uncapped, non-monotone input clamp, `in`-read-directly (v5), min-rounds/all-subagent skip, no-session-id error.
- `cmd/convert_weka_test.go` — end-to-end convert → `LoadTraceV2` → `SessionManager` accumulate reconstruction of absolute per-round inputs + `ThinkTimeSampler` yields recorded think (INV-10 spacing); plus `.jsonl` multi-session determinism, `independent`→empty header, warn+skip of malformed/below-min-rounds lines, no-usable-sessions hard error (writes no files).
- `go build ./...`, full `go test ./...`, and `golangci-lint run ./...` all green.
- **Manual smoke (operator step, needs HF LFS dataset):** a byte-range slice of a real `traces.jsonl` → `convert weka` → `replay --session-mode closed-loop --max-model-len 1000000`. Not automated (dataset not in CI), mirroring how PR-B/PR-C handled real-data validation.

## Caveats (documented in CLAUDE.md)

- **Huge ISL** (p50 ≈ 110K, p90 ≈ 395K): the ~41K default `--max-model-len` drops every request unservable — raise it and scale `--total-kv-blocks` at replay.
- **Delta-encoded output** MUST be replayed with `--session-mode closed-loop` (or `--concurrent-sessions`, which auto-promotes); the default `fixed` mode would misread deltas as absolutes.
- **Sub-agents skipped** — their token/latency content is not replayed (only their wall-clock is absorbed into the following main turn's think). Deferred to PR-E.

## Notes / deviations

- The issue's "Add a doc admonition (mirrors the existing Exgentic note)" references a note not present in `main` (it lived only in unmerged fork branches; merged PR-B added no guide docs and didn't list `convert otel` in CLAUDE.md). This PR documents the Weka caveats in **CLAUDE.md** — the canonical always-loaded doc — alongside a brief `convert otel`/`convert weka` note, keeping scope minimal (no new guide-doc section) to match PR-B's precedent.

## Out of scope

- Sub-agent fan-out (`type:"subagent"`) → PR-E.
- `hash_ids`-driven prefix sharing → future KV-model feature (this PR relies on the accumulate buffer's approximate growing prefix).

Closes #1604.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,7 +234,8 @@ go build -o blis main.go
 # traffic (+312% on this dataset). Replayed input length / KV pressure / hit-rate is therefore
 # a substantial UPPER BOUND — do NOT read it as a faithful reproduction of the recorded ISL.
 # (Property of the PR-A/PR-B accumulate delta law, not this converter; the conversion is exact
-# per that law. A non-lossy think encoding is tracked in #1608.)
+# per that law. Faithful compaction support is tracked in #1609; the separate think-time
+# lossy-0 sentinel is #1608.)
 ./blis convert weka --input traces.jsonl --trace-output corpus \
   --context-growth accumulate --max-think-time 0
 #
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -227,6 +227,10 @@ go build -o blis main.go
 # = uncapped, since Weka gaps are genuine away-from-keyboard times). Reads `in` directly
 # (never len(hash_ids)×64). Weka ISL is huge (p50 ≈ 110K, p90 ≈ 395K), so replay MUST raise
 # --max-model-len (the ~41K default drops every request unservable) and scale --total-kv-blocks.
+# Fidelity note: real agentic traces compact/trim context often (~28% of rounds on the
+# 051926 dataset have in_N < in_{N-1}+out_{N-1}); such non-monotone rounds clamp their input
+# delta to 0, so the accumulate buffer OVER-counts the true per-round ISL (it cannot shrink).
+# Replayed KV pressure / hit-rate is therefore an UPPER BOUND, not an exact reproduction.
 ./blis convert weka --input traces.jsonl --trace-output corpus \
   --context-growth accumulate --max-think-time 0
 #
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -227,10 +227,14 @@ go build -o blis main.go
 # = uncapped, since Weka gaps are genuine away-from-keyboard times). Reads `in` directly
 # (never len(hash_ids)×64). Weka ISL is huge (p50 ≈ 110K, p90 ≈ 395K), so replay MUST raise
 # --max-model-len (the ~41K default drops every request unservable) and scale --total-kv-blocks.
-# Fidelity note: real agentic traces compact/trim context often (~28% of rounds on the
-# 051926 dataset have in_N < in_{N-1}+out_{N-1}); such non-monotone rounds clamp their input
-# delta to 0, so the accumulate buffer OVER-counts the true per-round ISL (it cannot shrink).
-# Replayed KV pressure / hit-rate is therefore an UPPER BOUND, not an exact reproduction.
+# Fidelity note (important): real agentic traces compact/trim context heavily — ~30% of
+# rounds on the full 051926 dataset (219 sessions, 37.7K rounds) have in_N < in_{N-1}+out_{N-1}.
+# Such non-monotone rounds clamp their input delta to 0 (the accumulate buffer can only grow,
+# never shrink), so it OVER-counts the true cumulative input by ≈3–4× on real Claude Code
+# traffic (+312% on this dataset). Replayed input length / KV pressure / hit-rate is therefore
+# a substantial UPPER BOUND — do NOT read it as a faithful reproduction of the recorded ISL.
+# (Property of the PR-A/PR-B accumulate delta law, not this converter; the conversion is exact
+# per that law. A non-lossy think encoding is tracked in #1608.)
 ./blis convert weka --input traces.jsonl --trace-output corpus \
   --context-growth accumulate --max-think-time 0
 #
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -207,17 +207,37 @@ go build -o blis main.go
 ./blis convert inference-perf --spec spec.yaml
 ./blis compose --from spec1.yaml --from spec2.yaml
 
-# Convert OTel agentic trace JSON to a TraceV2 corpus (accumulate = strict growing prefix)
-./blis convert otel --input otel_json/ --trace-output corpus \
-  --context-growth accumulate --max-think-time 15s
-
-# Replay a fixed pool of N concurrent closed-loop sessions from the corpus
-# (--total-sessions duplicates the corpus with cache-busting to fill the target)
-# Real agentic prompts can be huge (tens of thousands to >100K tokens) and grow
-# each round — raise --max-model-len (default ~41K) and scale --total-kv-blocks
-# accordingly, or every request drops as unservable (completed_requests: 0).
+# Convert captured agentic traces to a TraceV2 corpus for closed-loop replay (#1477).
+# Both readers feed the shared session→TraceV2 encoder (#1479): each session becomes
+# rounds 0..N with per-round input DELTAS, TraceRecord.Model left empty (routing safety —
+# the recorded cross-model name would drop every request under --model), and closed-loop
+# replay's accumulate buffer reconstructs the growing prompt. Output is delta-encoded, so
+# it MUST be replayed with --session-mode closed-loop (or --concurrent-sessions, which
+# auto-promotes); the default --session-mode fixed would misread the deltas as absolutes.
+#
+# convert otel — OpenTelemetry agentic trace JSON (file / dir of *.json / *.jsonl); one
+# LLM chat span per round; think derived from arrival gaps (--max-think-time default 15s).
+./blis convert otel --input traces.jsonl --trace-output corpus \
+  --context-growth accumulate
+#
+# convert weka — SemiAnalysis WekaTrace JSONL (#1604, PR-F; one proxy session per line).
+# Filters requests[] to the linear main-agent stream (type:"subagent" groups skipped,
+# deferred to PR-E); recomputes pure client think as max(0, t_i − t_{i-1} − api_time_{i-1})
+# between consecutive main turns into the think_time_us column (--max-think-time default 0
+# = uncapped, since Weka gaps are genuine away-from-keyboard times). Reads `in` directly
+# (never len(hash_ids)×64). Weka ISL is huge (p50 ≈ 110K, p90 ≈ 395K), so replay MUST raise
+# --max-model-len (the ~41K default drops every request unservable) and scale --total-kv-blocks.
+./blis convert weka --input traces.jsonl --trace-output corpus \
+  --context-growth accumulate --max-think-time 0
+#
+# Replay a corpus (from either converter) closed-loop — reconstructs each session's growing prompt.
 ./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
-  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200
+  --model qwen/qwen3-14b --session-mode closed-loop --max-model-len 1000000
+# Or replay a fixed pool of N concurrent closed-loop sessions (#1486, PR-C); --total-sessions
+# duplicates the corpus with cache-busting to fill the target (--concurrent-sessions auto-promotes
+# to closed-loop). Same huge-ISL caveat — raise --max-model-len and scale --total-kv-blocks.
+./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
+  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200 --max-model-len 1000000
 
 # Run with gateway queue flow control (utilization-based saturation gating)
 ./blis run --model qwen/qwen3-14b --flow-control --saturation-detector utilization \
--- a/cmd/convert_weka.go
+++ b/cmd/convert_weka.go
@@ -25,10 +25,9 @@ var convertWekaCmd = &cobra.Command{
 TraceV2 pair (<prefix>.yaml + <prefix>.csv) suitable for 'blis replay --session-mode closed-loop'
 (or 'blis replay --concurrent-sessions', which auto-promotes to closed-loop).
 
-Input is a .jsonl file (one session per line; a directory of *.json or a single JSON
-session file also work). Each session's requests[] is filtered to the linear main-agent
-stream — type:"subagent" groups are skipped (deferred to PR-E) — and each main turn
-becomes one round. Per-round input token counts are stored as deltas so accumulate replay
+Input is a .jsonl file (one session per line) — the native Weka shape. Each session's
+requests[] is filtered to the linear main-agent stream — type:"subagent" groups are
+skipped (deferred to PR-E) — and each main turn becomes one round. Per-round input token counts are stored as deltas so accumulate replay
 reconstructs the exact growing prompt with a strictly-identical shared prefix. Per-round
 pure client think time is recomputed as max(0, t_i − t_{i-1} − api_time_{i-1}) between
 consecutive main turns and carried in the think_time_us column.
--- a/cmd/convert_weka.go
+++ b/cmd/convert_weka.go
@@ -1,0 +1,139 @@
+package cmd
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/inference-sim/inference-sim/sim/workload"
+)
+
+var (
+	wekaInputPath     string
+	wekaTraceOutput   string
+	wekaContextGrowth string
+	wekaMaxThinkTime  time.Duration
+	wekaMinRounds     int
+)
+
+var convertWekaCmd = &cobra.Command{
+	Use:   "weka",
+	Short: "Convert SemiAnalysis WekaTrace agentic JSONL to a TraceV2 file for closed-loop replay",
+	Long: `Convert SemiAnalysis WekaTrace agentic JSONL (one proxy session per line) into a
+TraceV2 pair (<prefix>.yaml + <prefix>.csv) suitable for 'blis replay --session-mode closed-loop'
+(or 'blis replay --concurrent-sessions', which auto-promotes to closed-loop).
+
+Input is a .jsonl file (one session per line; a directory of *.json or a single JSON
+session file also work). Each session's requests[] is filtered to the linear main-agent
+stream — type:"subagent" groups are skipped (deferred to PR-E) — and each main turn
+becomes one round. Per-round input token counts are stored as deltas so accumulate replay
+reconstructs the exact growing prompt with a strictly-identical shared prefix. Per-round
+pure client think time is recomputed as max(0, t_i − t_{i-1} − api_time_{i-1}) between
+consecutive main turns and carried in the think_time_us column.
+
+Caveats: Weka input token counts are huge (p50 ≈ 110K, p90 ≈ 395K); raise --max-model-len
+(the ~41K default drops every request as unservable) and scale --total-kv-blocks at replay.
+The recorded model name (claude-*) is dropped so requests inherit --model (routing safety).
+
+Note: unlike other 'blis convert' subcommands (which emit a WorkloadSpec to stdout),
+'convert weka' writes TraceV2 files, because it targets 'blis replay' (fixed per-call
+token counts) rather than 'blis run' (distribution sampling).`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if wekaContextGrowth != "accumulate" && wekaContextGrowth != "independent" {
+			logrus.Fatalf("--context-growth must be \"accumulate\" or \"independent\", got %q", wekaContextGrowth)
+		}
+		if wekaMinRounds < 1 {
+			logrus.Fatalf("--min-rounds must be >= 1, got %d", wekaMinRounds)
+		}
+		if wekaMaxThinkTime < 0 {
+			logrus.Fatalf("--max-think-time must be >= 0, got %s", wekaMaxThinkTime)
+		}
+		opts := workload.WekaConvertOptions{
+			ContextGrowth:  wekaContextGrowth,
+			MaxThinkTimeUs: wekaMaxThinkTime.Microseconds(),
+			MinRounds:      wekaMinRounds,
+		}
+		if err := runConvertWeka(wekaInputPath, wekaTraceOutput, opts); err != nil {
+			logrus.Fatalf("Weka conversion failed: %v", err)
+		}
+	},
+}
+
+// runConvertWeka reads all Weka sessions at inputPath, converts each to session
+// records, assigns global request IDs, and writes a TraceV2 pair to
+// <outPrefix>.yaml / <outPrefix>.csv. Reuses collectTraceInputs (defined in
+// convert_otel.go — format-agnostic: it splits .jsonl by line, reads *.json in a
+// directory, or treats any other file as a single payload).
+func runConvertWeka(inputPath, outPrefix string, opts workload.WekaConvertOptions) error {
+	// Return errors (do not Fatalf): the cobra Run wrapper owns the CLI exit
+	// boundary (R6 — library-style helpers surface errors, the CLI layer
+	// terminates). --input/--trace-output are also MarkFlagRequired'd, so these
+	// guards are defense-in-depth.
+	if inputPath == "" {
+		return fmt.Errorf("--input is required")
+	}
+	if outPrefix == "" {
+		return fmt.Errorf("--trace-output is required")
+	}
+	inputs, err := collectTraceInputs(inputPath)
+	if err != nil {
+		return err
+	}
+
+	var allRecords []workload.TraceRecord
+	nextID := 0
+	skipped := 0
+	for _, in := range inputs {
+		recs, err := workload.ConvertWekaSession(in.raw, opts)
+		if err != nil {
+			logrus.Warnf("skipping unparseable weka session %s: %v", in.name, err)
+			skipped++
+			continue
+		}
+		if recs == nil {
+			logrus.Debugf("skipping weka session %s: below --min-rounds", in.name)
+			skipped++
+			continue // below MinRounds
+		}
+		for i := range recs {
+			recs[i].RequestID = nextID
+			nextID++
+		}
+		allRecords = append(allRecords, recs...)
+	}
+
+	if len(allRecords) == 0 {
+		return fmt.Errorf("no usable sessions found in %q (skipped %d)", inputPath, skipped)
+	}
+
+	growth := opts.ContextGrowth
+	if growth == "independent" {
+		growth = "" // empty header value = per-round-independent inputs
+	}
+	header := &workload.TraceHeader{
+		Version:              3,
+		TimeUnit:             "microseconds",
+		Mode:                 "generated",
+		SessionContextGrowth: growth,
+	}
+	if err := workload.ExportTraceV2(header, allRecords, outPrefix+".yaml", outPrefix+".csv"); err != nil {
+		return err
+	}
+	logrus.Infof("Wrote %d records from %d sessions to %s.{yaml,csv} (skipped %d)",
+		len(allRecords), len(inputs)-skipped, outPrefix, skipped)
+	return nil
+}
+
+func init() {
+	convertWekaCmd.Flags().StringVar(&wekaInputPath, "input", "", "Path to WekaTrace JSONL file (one session per line), directory of *.json, or single JSON session (required)")
+	convertWekaCmd.Flags().StringVar(&wekaTraceOutput, "trace-output", "", "Output TraceV2 prefix; writes <prefix>.yaml + <prefix>.csv (required)")
+	convertWekaCmd.Flags().StringVar(&wekaContextGrowth, "context-growth", "accumulate", "Prefix model: \"accumulate\" (strict growing shared prefix) or \"independent\"")
+	convertWekaCmd.Flags().DurationVar(&wekaMaxThinkTime, "max-think-time", 0, "Cap on the recomputed per-round think gap; 0 = no cap (Weka gaps are genuine away-from-keyboard times)")
+	convertWekaCmd.Flags().IntVar(&wekaMinRounds, "min-rounds", 1, "Skip sessions with fewer than N usable main-agent turns")
+	_ = convertWekaCmd.MarkFlagRequired("input")
+	_ = convertWekaCmd.MarkFlagRequired("trace-output")
+
+	convertCmd.AddCommand(convertWekaCmd)
+}
--- a/cmd/convert_weka_test.go
+++ b/cmd/convert_weka_test.go
@@ -1,0 +1,190 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/inference-sim/inference-sim/sim"
+	"github.com/inference-sim/inference-sim/sim/workload"
+)
+
+// T9 — end-to-end: runConvertWeka → LoadTraceV2 (header accumulate, global IDs
+// 0..N) → LoadTraceV2SessionBlueprints + SessionManager reconstructs each
+// monotone session's recorded absolute per-round inputs via the accumulate
+// buffer, and the ThinkTimeSampler yields the recomputed pure-think values
+// (which drive follow-up arrival = completion + think, INV-10).
+func TestConvertWeka_EndToEndAndAccumulateReconstruction(t *testing.T) {
+	dir := t.TempDir()
+	inPath := filepath.Join(dir, "traces.jsonl")
+	// One monotone session, 3 main turns (in 100/150/215, out 10/20/5),
+	// think gaps: t1−t0−api0 = 8−0−1 = 7s; t2−t1−api1 = 20−8−2 = 10s.
+	session := `{"id":"sess-1","models":["claude-sonnet-4"],"requests":[` +
+		`{"type":"n","t":0.0,"in":100,"out":10,"api_time":1.0},` +
+		`{"type":"s","t":8.0,"in":150,"out":20,"api_time":2.0},` +
+		`{"type":"n","t":20.0,"in":215,"out":5,"api_time":1.0}` +
+		`]}`
+	if err := os.WriteFile(inPath, []byte(session+"\n"), 0o644); err != nil {
+		t.Fatalf("write input: %v", err)
+	}
+	outPrefix := filepath.Join(dir, "out")
+
+	if err := runConvertWeka(inPath, outPrefix, workload.WekaConvertOptions{ContextGrowth: "accumulate", MinRounds: 1}); err != nil {
+		t.Fatalf("runConvertWeka: %v", err)
+	}
+
+	trace, err := workload.LoadTraceV2(outPrefix+".yaml", outPrefix+".csv")
+	if err != nil {
+		t.Fatalf("load exported trace: %v", err)
+	}
+	if trace.Header.SessionContextGrowth != "accumulate" {
+		t.Errorf("header growth = %q, want accumulate", trace.Header.SessionContextGrowth)
+	}
+	if len(trace.Records) != 3 {
+		t.Fatalf("records = %d, want 3", len(trace.Records))
+	}
+	for i, r := range trace.Records {
+		if r.RequestID != i {
+			t.Errorf("record %d RequestID = %d, want %d (assigned in stream order)", i, r.RequestID, i)
+		}
+		if r.Model != "" {
+			t.Errorf("record %d Model = %q, want empty (routing safety)", i, r.Model)
+		}
+	}
+	// think_time_us column round-trips: recorded 7s / 10s on rounds 1, 2.
+	if trace.Records[1].ThinkTimeUs != 7_000_000 || trace.Records[2].ThinkTimeUs != 10_000_000 {
+		t.Errorf("think_time_us = [_,%d,%d], want [_,7e6,10e6]", trace.Records[1].ThinkTimeUs, trace.Records[2].ThinkTimeUs)
+	}
+
+	// Closed-loop reconstruction through the SessionManager (accumulate buffer).
+	r0, bps, err := workload.LoadTraceV2SessionBlueprints(trace, 42, nil, 0)
+	if err != nil {
+		t.Fatalf("LoadTraceV2SessionBlueprints: %v", err)
+	}
+	if len(bps) != 1 {
+		t.Fatalf("blueprints = %d, want 1", len(bps))
+	}
+	// ThinkTimeSampler yields the recorded pure-think sequence (INV-10 spacing):
+	// round 1 arrives at completion[0]+7s, round 2 at completion[1]+10s.
+	ts := bps[0].ThinkTimeSampler
+	if ts == nil {
+		t.Fatal("expected a ThinkTimeSampler (recorded think present)")
+	}
+	if g1, g2 := ts.Sample(nil), ts.Sample(nil); g1 != 7_000_000 || g2 != 10_000_000 {
+		t.Errorf("think samples = [%d, %d], want recorded [7e6, 10e6]", g1, g2)
+	}
+
+	// Reconstruct absolute per-round inputs via the accumulate buffer.
+	sm := workload.NewSessionManager(bps)
+	round0 := r0[0]
+	if round0.InputLen() != 100 {
+		t.Fatalf("round0 input len = %d, want 100", round0.InputLen())
+	}
+	round0.State = sim.StateCompleted
+	round0.ProgressIndex = int64(round0.InputLen()) + 10 // 10 output tokens generated
+	fu := sm.OnComplete(round0, 8_000_000)
+	if len(fu) != 1 {
+		t.Fatalf("round0 follow-ups = %d, want 1", len(fu))
+	}
+	round1 := fu[0]
+	// Round 1 absolute input = 100 + 10 + delta(40) = 150 (the recorded in_1).
+	if round1.InputLen() != 150 {
+		t.Fatalf("round1 reconstructed input len = %d, want 150 (recorded in_1)", round1.InputLen())
+	}
+	round1.State = sim.StateCompleted
+	round1.ProgressIndex = int64(round1.InputLen()) + 20 // 20 output tokens
+	fu = sm.OnComplete(round1, 30_000_000)
+	if len(fu) != 1 {
+		t.Fatalf("round1 follow-ups = %d, want 1", len(fu))
+	}
+	round2 := fu[0]
+	// Round 2 absolute input = 150 + 20 + delta(45) = 215 (the recorded in_2).
+	if round2.InputLen() != 215 {
+		t.Fatalf("round2 reconstructed input len = %d, want 215 (recorded in_2)", round2.InputLen())
+	}
+}
+
+// T10 — CLI corpus paths: .jsonl multi-session with deterministic global IDs in
+// file order, "independent"→empty-header mapping, warn+skip of a malformed line
+// and a below-min-rounds session, and the no-usable-sessions hard error (writes
+// no output files).
+func TestConvertWeka_JSONLIndependentAndSkips(t *testing.T) {
+	t.Run("jsonl multi-session, independent growth, deterministic IDs", func(t *testing.T) {
+		dir := t.TempDir()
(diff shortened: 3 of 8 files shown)


namasl:

namasl:
Thanks — blis self-review came back **READY TO MERGE** (no blocking issues), and archon is `ARCHITECTURAL_CHANGE` (benign: the only delta is the pre-existing `cmd → sim/workload` edge *STRENGTHENED* by the new subcommand — no new edge, no cycle; the "invariants touched" are the new tests). CI shards are green (build, lint, `workload & cmd`, core, etc.).

Disposition of the three **non-blocking** observations from the self-review — no changes made, with reasoning:

1. **`nil In/Out` main turn silently dropped (suggest a debug log).** Kept as-is for parity with the merged `blis convert otel`, which drops nil-token spans with the same bare `continue` (no log) at `sim/workload/otel_convert.go:140`. Adding a log to the Weka reader alone would create the exact otel/weka asymmetry that R23 (parallel-path parity) guards against. Real Weka main turns always carry `in`/`out`; the drop is defense-in-depth.
2. **All-zero `think_time_us` → arrival-gap fallback.** Already documented at `sim/workload/tracev2.go:188` (the lossy-sentinel note that explicitly names this converter), and the non-lossy encoding is out-of-scope per #1604. In practice it cannot arise for a real ≥20-main-turn session (some round always has a positive gap), so the recorded column is used.
3. **No `--include-errors` (asymmetric with `otel`).** Intentional — Weka errors are filtered at source, so every main turn is `Status: "ok"`; there is nothing to include.

Ready for maintainer review/merge.

namasl:

namasl:

## qa-review — PR #1605: ⛔ BLOCK
_blis convert weka — Weka CC agentic trace JSONL to TraceV2 (PR-F of #1477) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer Claude Opus_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR implements exactly what #1604 (PR-F) scopes: a new sim/workload/weka_convert.go reader and cmd/convert_weka.go subcommand that filter to the main-agent stream, reuse the shared PR-B encoder for input deltas, recompute pure think, drop Model, and skip sub-agents. It reuses collectTraceInputs and EncodeSessionToTraceRecords rather than duplicating logic, and adds no engine changes. Nearly all acceptance-criteria tests are present (delta reconstruction, subagent-skip, think recompute, non-monotone clamp, negative-gap clamp, min-rounds/all-subagent skip, read-in-directly, end-to-end accumulate reconstruction). |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Only partially. CLAUDE.md was updated with a full convert weka example including the huge-ISL admonition the issue requested. However the dedicated replay guide docs/guide/observe-replay-calibrate.md has a whole 'Replaying OTel agentic traces' section (242-278) and was NOT extended to mention convert weka — grep finds 'convert weka' only in CLAUDE.md. So documentation is not fully updated: the guide covers only the OTel converter. |
| F3 | fixed | ❌ FLAW_FOUND | Are there any stale code comments left by this change? | The PR's own new files contain no stale comments. But this PR (which IS #1604) leaves two pre-existing comments stale/misleading: both state that a 'non-lossy encoding (distinguishing recorded-0 from absent) is deferred to the Weka converter work (#1604).' This PR does not implement that non-lossy encoding — it keeps the lossy 0-sentinel, and Weka's overlap-clamped rounds legitimately emit ThinkTimeUs==0, which reads as 'not recorded.' A reader following those comments to #1604 will not find the promised work. |
| G1 | correctness | ✅ CONFIDENT | In `ConvertWekaSession` (`sim/workload/weka_convert.go`), how are non-finite or extreme values (such as `NaN`, `Inf`, or large values) in floating-point fields `t` and `api_time` handled, and could `math.Round((tn.t - prev.t - prev.api) * 1e6)` cause integer overflow when cast to `int64`? | Non-finite values cannot reach the arithmetic: JSON has no NaN/Inf literals, so json.Unmarshal rejects t/api_time of NaN, Infinity, or out-of-range magnitudes (e.g. 1e400) with an error, and the session is skipped with a warning. Overflow of int64(math.Round((t-prev.t-prev.api)*1e6)) is only reachable with absurd magnitudes (~>9.2e12 seconds), far beyond realistic Weka t values. The think-gap conversion is additionally protected by the gapUs<0 clamp. |
| G2 | correctness | ✅ CONFIDENT | In `ConvertWekaSession`, if a main-agent request carries negative values for `In` or `Out` (`*r.In < 0` or `*r.Out < 0`), how does `EncodeSessionToTraceRecords` handle them, and could this break the delta accumulation logic? | It does not break delta accumulation or panic. For round 0 the encoder writes InputTokens=InputTokensAbs directly, so a negative in would produce a negative round-0 InputTokens; for rounds>0 the delta is max(0, in-prevIn-prevOut), always clamped >=0. A negative round-0 value does not silently corrupt anything — it surfaces as a loud error when the exported CSV is loaded (LoadTraceV2 rejects negative input_tokens/output_tokens), before it could reach GenerateRandomTokenIDs (which would otherwise panic). |
| G3 | invariants | ✅ CONFIDENT | In `runConvertWeka` (`cmd/convert_weka.go`), `collectTraceInputs` loads sessions from files or directories. When given a directory input, is the file traversal order guaranteed to be deterministic across different platforms to maintain INV-6 determinism? | Deterministic. For a directory input, collectTraceInputs gathers *.json filenames and sorts them with sort.Strings (commented 'deterministic corpus order (INV-6)'). For a .jsonl file it iterates lines in file order. Global RequestIDs are assigned sequentially in that deterministic order, so IDs are stable across platforms/runs. |
| G4 | invariants | ✅ CONFIDENT | When a session's initial `requests[]` entries are skipped `subagent` groups, round 0's `ArrivalUs` is set to `tn.t * 1e6` (> 0). How does closed-loop replay handle a non-zero arrival time for round 0 of a pooled session without causing unexpected delays in session admission? | A non-zero round-0 ArrivalTimeUs is just an absolute arrival timestamp; the DES schedules an ArrivalEvent at that time and advances the clock — it never busy-waits or stalls other sessions. In pool replay only the INITIAL concurrent sessions keep their recorded round-0 arrival (a bounded one-time offset equal to the first main turn's t); every refill rebases arrival to the completion tick, so wave-2+ sessions incur no delay. In plain closed-loop replay each session starts at its recorded first-main-turn t, which is fidelity-preserving. |
| G5 | rules | ✅ CONFIDENT | In `runConvertWeka` (`cmd/convert_weka.go`), malformed JSON lines in a `.jsonl` file log a warning and continue (`logrus.Warnf(...)`). Does skipping malformed input lines silently without a strict failure flag violate BLIS error handling guidelines against silent error swallowing? | No R1 violation. Malformed lines are not silently swallowed: runConvertWeka logs logrus.Warnf('skipping unparseable weka session ...'), increments a skipped counter, and reports the skip count; if every session is skipped it returns a hard 'no usable sessions found' error. R1 explicitly permits an error path that increments a counter — this does both (log + count). |
| G6 | rules | ✅ CONFIDENT | Does `WekaConvertOptions` or `wekaSession` violate any BLIS design rules regarding missing canonical option validators or exporting unvalidated fields? | No rule violation. WekaConvertOptions is a plain options struct with no maps (R8 N/A), no YAML tags and never YAML-parsed (R9/R10 N/A). Fields are validated at the CLI boundary (context-growth in {accumulate,independent}, min-rounds>=1, max-think-time>=0), and ConvertWekaSession defensively re-clamps MinRounds. wekaSession/wekaRequest use JSON tags and deliberately ignore unknown fields (DisallowUnknownFields intentionally NOT used, correct for real traces). Mirrors OTelConvertOptions/otelSpan. |
| G7 | parity | ✅ CONFIDENT | The default for `--max-think-time` in `convert weka` is `0` (uncapped), whereas `convert otel` defaults to `15s`. If a Weka trace contains an inter-turn gap of several hours, how does closed-loop replay handle this without stalling pool execution compared to OTel traces? | A multi-hour gap becomes a multi-hour ThinkTimeUs, and in closed-loop replay the follow-up ArrivalEvent is scheduled that far ahead in sim time — it does NOT stall the run: the DES advances to the next event and other sessions proceed. In pool mode that one session holds its slot idle (lowering effective concurrency), or the follow-up lands past --horizon and the session is horizon-interrupted (INV-11). This is the documented trade-off, and --max-think-time caps it. Not a defect. |
| G8 | parity | ✅ CONFIDENT | When `--context-growth independent` is set, `runConvertWeka` maps `growth` to `""` for `TraceHeader.SessionContextGrowth`. Does `convert otel` use identical header value mapping, and does `LoadTraceV2` parse empty header fields consistently across all converters? | Identical. Both converters map ContextGrowth 'independent' → '' before building the header. At replay, LoadTraceV2SessionBlueprints validates the header value up front: only '' or 'accumulate' is accepted, any other value is a hard error (R1) — so empty header fields parse consistently regardless of which converter produced them. |
| G9 | tests | ✅ CONFIDENT | While `TestConvertWeka_JSONLIndependentAndSkips` tests basic line skipping, is there a test that verifies `collectTraceInputs` with directory inputs produces deterministic global `RequestID` assignments across multiple runs? | No such test exists. TestConvertWeka_JSONLIndependentAndSkips verifies deterministic global RequestID assignment for a single multi-session .jsonl file (file order), but no weka test exercises a DIRECTORY input, and none runs conversion twice to assert cross-run byte-identity. Directory determinism is guaranteed structurally by sort.Strings in the shared collectTraceInputs (and the OTel suite exercises the same helper), so behavior is correct; just not directly pinned for weka. |
| G10 | tests | ✅ CONFIDENT | The think-time test asserts recomputed values, but would `weka_convert_test.go` fail if `ThinkTimeUs` inadvertently omitted the subtraction of `prev.api` and calculated gap purely as `tn.t - prev.t`? | Yes — the test would fail. If the code computed the gap as t_i - t_{i-1} (dropping - prev.api), T2 would fail on round 1 (asserts 5e6 for 6-0-1; wrong gives 6e6) and round 2 (asserts 2e6 for 10-6-2; wrong gives 4e6). T4's uncapped case would also fail (asserts 100e6 for t=101,api=1; wrong gives 101e6). The api_time subtraction is firmly pinned. |
| G11 | rationale | ✅ CONFIDENT | Why was the decision made to unconditionally clear the model name (`TraceRecord.Model = ""`) rather than providing a CLI flag to optionally retain or override recorded models for multi-model replay routing? | Rationale holds. TraceRecord.Model is routing-significant at replay (buildRouterState filters routable instances by it), and Weka records claude-* names that would never match --model, so retaining them would silently drop every request at routing. Clearing it makes each request inherit --model. Same contract PR-B established for OTel via the shared encoder; these corpora are replayed under a single --model, so a retain/override flag is unneeded and would reintroduce the drop-everything footgun. |
| G12 | rationale | ✅ CONFIDENT | Why recompute think time as `max(0, t_i - t_{i-1} - api_time_{i-1})` on subagent-free chains instead of directly using the recorded `think_time` field when present in the input JSON? | Because the recorded think_time is measured against the chronologically-preceding request — which may be a sub-agent inner request that this converter filters out. Once sub-agent entries are removed from the linear main chain, trusting the recorded field would misalign think to the wrong predecessor. Recomputing max(0, t_i - t_{i-1} - api_{i-1}) strictly over consecutive MAIN turns keeps the chain self-consistent and correctly absorbs a skipped sub-agent group's wall-clock into the following main turn's think. Verified to equal recorded think_time on subagent-free chains. |
| G13 | scope | ✅ CONFIDENT | In `wekaRequest` (`sim/workload/weka_convert.go`), are there any fields parsed or defined that are unused in this PR and solely intended for sub-agent handling in PR-E? | No. wekaRequest defines exactly five fields — Type, T, In, Out, API — and all are consumed: Type gates the sub-agent skip, and T/In/Out/API drive turn construction and think recompute. There are no sub-agent-only fields (agent_id, subagent_type, duration_ms, nested requests[]) defined or parsed; sub-agent payloads are discarded via encoding/json's standard unknown-field drop. Scope is clean — no dead PR-E fields. |
| G14 | scope | ✅ CONFIDENT | The changes in `CLAUDE.md` update example commands for both `convert otel` and `convert weka`. Do the documentation updates for `convert otel` stay within the boundaries of this PR? | The convert-otel example was edited (input changed from otel_json/ to traces.jsonl; explicit --max-think-time 15s dropped in favor of the default; the two examples restructured under a shared intro). The edits are accurate — the otel default max-think-time is indeed 15s, and no otel behavior changed — so this is a coherent documentation consolidation presenting both converters together, not a functional change. |
| G15 | errors | ✅ CONFIDENT | In `runConvertWeka`, if all sessions in an input file are skipped due to `--min-rounds` or missing fields, `runConvertWeka` returns an error (`no usable sessions found`). Does Cobra handle this error cleanly without double-printing error messages or exiting with an improper status code? | Clean handling. The subcommand uses Run: (not RunE:), and the returned error is terminated inside the Run wrapper via logrus.Fatalf('Weka conversion failed: %v', err), which prints once to stderr and calls os.Exit(1). Because Run returns nothing to cobra, cobra does not additionally print the error or usage — no double-printing — and the exit code is a proper non-zero. Mirrors convert otel. |
| G16 | errors | ✅ CONFIDENT | In `ConvertWekaSession`, if `json.Unmarshal` succeeds but `Requests` is empty or lacks session `ID`, it returns an error vs returning `(nil, nil)`. How does `runConvertWeka` differentiate between a corrupt/unparseable session and an intentionally filtered session? | By log level/message and return shape: a corrupt/unidentifiable session (unmarshalable JSON, or missing 'id') makes ConvertWekaSession return a non-nil error → runConvertWeka logs Warnf('skipping unparseable weka session ...') and skipped++; an intentionally-filtered session (fewer than --min-rounds usable main turns) returns (nil,nil) → Debugf('... below --min-rounds') and skipped++. The question's premise is slightly off: an EMPTY Requests list with a valid id is NOT an error — it yields 0 turns < minRounds → (nil,nil), treated as filtered, not corrupt. Only a missing id (or unparseable JSON) errors. |

### Items to fix
- **F3 · FLAW_FOUND** — The PR's own new files contain no stale comments. But this PR (which IS #1604) leaves two pre-existing comments stale/misleading: both state that a 'non-lossy encoding (distinguishing recorded-0 from absent) is deferred to the Weka converter work (#1604).' This PR does not implement that non-lossy encoding — it keeps the lossy 0-sentinel, and Weka's overlap-clamped rounds legitimately emit ThinkTimeUs==0, which reads as 'not recorded.' A reader following those comments to #1604 will not find the promised work.
  _sim/workload/tracev2.go:188 and sim/workload/replay.go:122-128 both attribute the deferred non-lossy think-time encoding to '#1604'; the merged weka converter (weka_convert.go:517-525) still clamps overlapping think gaps to 0 with no non-lossy signaling._

### Important to consider
- **F1** — One AC only partially met: the 'model-field routing-safety' criterion says replay under a differing --model should yield routingRejections==0, but no test runs a full replay asserting that — tests only assert TraceRecord.Model=='' at the field level (T1, T9). The field-level assertion is the correct mechanism, but the end-to-end routing assertion the issue lists is not automated.
  _sim/workload/weka_convert.go:468-541; cmd/convert_weka.go:79-197; tests in weka_convert_test.go (T1-T8) and cmd/convert_weka_test.go (T9-T10)_
- **F2** — The requested huge-ISL admonition IS present in CLAUDE.md, so the hard requirement is met; the gap is completeness — the standalone guide should gain a parallel weka subsection. project-structure.md lists neither otel_convert.go nor weka_convert.go, so omitting weka there is consistent with existing practice.
  _CLAUDE.md:223-230 (weka added); docs/guide/observe-replay-calibrate.md:242-278 (OTel-only section, no weka); grep 'convert weka' hits CLAUDE.md only; issue.md:70 asks to add a doc admonition_
- **G1** — The ArrivalUs conversion (weka_convert.go:535 int64(math.Round(tn.t*1e6))) has no negative/overflow clamp, unlike the think gap — harmless for realistic inputs but there is no explicit finite/range guard; the code relies on the JSON parser plus realistic magnitudes.
  _Go probe: {t:NaN}/{t:Infinity}/{t:1e400} all error at json.Unmarshal; {t:1e18} overflows to MinInt64. weka_convert.go:475-477 (unmarshal error → skip), 518-521 (negative clamp)._
- **G2** — The converter itself does not pre-validate in/out >= 0; a bad negative round-0 count is only caught at load time, not conversion time. Not realistic for Weka token counts, but a defensive guard in ConvertWekaSession would fail earlier.
  _trace_encode.go:37-56 (round-0 uses abs; rounds>0 clamp to 0); tracev2.go:527-537 (LoadTraceV2 rejects negatives); request.go:188-194 (make would panic on negative)._
- **G3** — The directory branch only collects files ending in '.json', so a directory of Weka '.jsonl' files would be silently ignored (collect nothing → 'no usable sessions'). Inherited from the shared OTel helper; Weka input is normally a single .jsonl file, but the Long help's claim that 'a directory of *.json also work[s]' is the only supported directory form.
  _cmd/convert_otel.go:74-116 (sort.Strings at :90; jsonl line order :106-113); cmd/convert_weka.go:146-163 (sequential nextID in input order)._
- **G5** — There is no --strict flag to fail-fast on the first malformed line (parity with convert otel). A corpus could lose many sessions with only WARN logs; acceptable under R1, but a strict mode could be future hardening.
  _cmd/convert_weka.go:147-166 (Warnf + skipped++ + 'no usable sessions' error); rules.md:26 (R1 allows counter increment); mirrors runConvertOtel._
- **G7** — OTel defaults to a 15s cap; Weka's uncapped default is a deliberate fidelity choice (genuine away-from-keyboard gaps), documented in the flag help and the issue.
  _weka_convert.go:415-419 and cmd/convert_weka.go:191 (uncapped default 0, documented); session_pool.go:198-201; replay.go:234-268._
- **G9** — Owed test: a directory-input determinism test (and/or a run-twice byte-identity check) would pin INV-6 for the weka path directly. Compounded by the .jsonl-in-directory collection gap noted in G3.
  _cmd/convert_weka_test.go:314-348 (.jsonl file-order only, no directory case, no repeat-run); determinism source cmd/convert_otel.go:90._
- **G11** — Multi-model routable replay would be genuine future work; a CLI opt-in to retain the recorded model could be added then, but is out of scope here.
  _trace_encode.go:28-33 and weka_convert.go:440-447 (Model empty, routing rationale); replay.go:103,306; issue.md:59._
- **G14** — Mild scope expansion: an 'add weka converter' PR also rewrote the otel example lines. Harmless and correct, but strictly the otel example edit is adjacent to the PR's core purpose.
  _CLAUDE.md diff (patch 9-49); otel default confirmed cmd/convert_otel.go:185 (15*time.Second)._
- **G16** — Both cases increment the same skipped counter and the aggregate 'no usable sessions (skipped N)' error does not separate corrupt from filtered — only the per-session log level distinguishes them (WARN vs DEBUG, the latter invisible at default log level).
  _weka_convert.go:474-505 (error only on unmarshal failure or empty id; empty/too-few turns → nil,nil at :503-505); cmd/convert_weka.go:147-157 (Warnf vs Debugf)._

susiejojo:
## BLIS PR Review — #1605 `feat(cmd): blis convert weka` (PR-F of #1477)

Reviewed against the child issue (#1604), the tracking plan (#1477), the archon architectural verdict, and a **real-dataset end-to-end smoke test** on `semianalysisai/cc-traces-weka-with-subagents-051926`. Verdict up front: **ready to merge** — one findings section below is worth reading before merge, but nothing blocking.

---

### ✅ Real-dataset validation (the explicitly-requested check)

I byte-range-sliced the live HF LFS `traces.jsonl` (1.07 GB), extracted the first **10 complete sessions** (988 request entries, **13 of them `type:"subagent"`**), and ran the full pipeline from the repo root:

```
blis convert weka --input weka_real.jsonl --trace-output corpus --context-growth accumulate --max-think-time 0
  → Wrote 975 records from 10 sessions (skipped 0)     # 988 − 13 subagents = 975 ✔ BC-2 confirmed on real data
blis replay --model qwen/qwen3-14b --session-mode closed-loop --max-model-len 1000000 --total-kv-blocks 200000
  → completed 147, still_running 5, timed_out 2, injected 154   # INV-1 holds: 147+5+2 = 154 ✔
  → Cache Hit Rate: 0.8981                                       # accumulate growing-prefix reconstruction works ✔
```

- The real schema matches the reader exactly (`id`, `models`, `block_size`, `hash_id_scope`, `requests[]` with `t/in/out/api_time/type/ttft/hash_ids`). Unknown fields ignored cleanly.
- **Huge-ISL caveat verified**: same corpus at the default `--max-model-len` → `completed 7, dropped_unservable 10` — i.e. the documented "raise `--max-model-len`" instruction is real and necessary, not theoretical.
- **INV-6 determinism verified empirically**: two converts → byte-identical `.yaml`+`.csv`; two replays → byte-identical stdout.
- **Real drop path is dead code in practice**: 0/975 main turns had `nil in`/`out`, so the silent `continue` at `weka_convert.go:99` never fires on this dataset (see finding F2).

Full suite green (`go test ./...`), `go vet` + `gofmt` clean on all four changed files.

---

### Correctness & behavioral contracts

All 10 tests pass and the contracts hold on real data. Reader is a thin, correct adapter over the merged PR-B encoder:
- **BC-1/BC-6/BC-7** (delta law, read `in` directly, empty `Model`) — delegated to the shared encoder + direct `*int` parse; confirmed on real records (`Model` empty in all 975).
- **BC-3/BC-4/BC-5** (think recompute, `≥0` clamp, cap) — `max(0, t_i − t_{i-1} − api_{i-1})`, correct; on real data **40/965 gaps clamp negative→0** (overlapping turns are genuinely present).
- **BC-8** (min-rounds skip returns `(nil,nil)`) and **BC-9/BC-10** (independent→empty header, global IDs in file order) — verified.

Design/separation is clean: `sim/workload/weka_convert.go` is a pure library fn (no `Fatalf`, R6-correct); `cmd/convert_weka.go` owns the CLI boundary; `collectTraceInputs`/`ExportTraceV2`/`EncodeSessionToTraceRecords` reused without duplication. The `SequenceSampler` backing the think replay is stateful+ordered+wrapping, so T9's two `Sample(nil)` calls are a valid INV-10 assertion (this closes the prior self-review's "minor observation" — it was a non-issue).

**Cross-path parity (run/replay/observe):** correctly N/A. This is convert-only; it emits standard TraceV2 consumed unchanged by `blis replay` (and observe corpus-mode via the merged PR-D). No new flag touches run/replay/observe. `--max-think-time` default differs from `otel`'s (0 vs 15s) but is justified in-code and in the issue (Weka gaps are genuine away-time). Preemption/timeout safety: no simulation logic added — N/A.

**Archon `ARCHITECTURAL_CHANGE` verdict — benign.** It fired only on a *strengthened* `cmd → sim/workload` edge (new exported `ConvertWekaSession`/`WekaConvertOptions`). No package boundary moved, no invariant removed, no cycle. This is a new leaf front-end on an existing edge — exactly the `preset`/`servegen`/`otel` multi-front-end pattern. No architectural objection.

**Relevance to #1477:** squarely on-plan. PR-F sits below the merged PR-A (#1478 think column) and PR-B (#1479 encoder), reuses both verbatim, defers subagents to PR-E as the plan specifies, and closes #1604. Scope discipline is good.

---

### Findings (non-blocking)

**F1 — [fidelity, worth flagging in docs] Non-monotone context compaction is pervasive on real Weka data, and the delta-clamp over-count is large.** The encoder doc and PR both describe non-monotone rounds as "over-count by the clamped deficit — an accepted, documented deviation." That framing understates the magnitude on this dataset:
- **270 / 975 rounds (28%)** are non-monotone (`in_N < in_{N-1} + out_{N-1}` — Claude Code compacts/trims context constantly).
- Reconstructed absolute inputs **over-count recorded totals by 54.2%** cumulatively (101.4M vs 186.9M recorded tokens), worst single round +486K tokens, concentrated in 2 long sessions.

This is a property of the accumulate model (a monotone growing buffer cannot represent compaction), not a bug in this PR — and it's mostly PR-A/PR-B's encoder. But since PR-F is the one that makes *real* Weka data flow through it, and the 0.90 hit-rate / ISL numbers people will quote come from here, I'd suggest one sentence in the CLAUDE.md Weka caveat block noting that context compaction (common in real traces) makes accumulate over-count the true per-round ISL, so replayed KV pressure is an upper bound. Purely documentation; no code change needed.

**F2 — [R1 nit] Silent drop of a main turn with `nil in`/`out`** (`weka_convert.go:98-100`). Confirmed 0 occurrences on the real dataset, so it's defensive-only, but it's an un-counted silent `continue`. A `logrus.Debugf("weka session %s: dropping main turn %d with missing in/out", s.ID, i)` would make a future data-shape change debuggable without changing behavior. Matches the self-review's own observation; agree it's non-blocking.

**F3 — [doc consistency, trivial] `convert weka --help` long-text says "a directory of *.json or a single JSON session file also work."** True (it reuses `collectTraceInputs`), but the Weka source format is JSONL-per-session; a stray directory of single-session `*.json` is an unusual shape for this format. Not wrong, just slightly surprising phrasing carried over from the shared helper. Leave as-is or trim — cosmetic.

---

### Verdict: **READY TO MERGE**

The PR does exactly what #1604 specifies, reuses the merged encoder/replay spine without touching it, passes the full suite + lint, holds INV-1/INV-3/INV-6/INV-10, and — most importantly — **works end-to-end on the real SemiAnalysis Weka dataset** including the subagent-skip and huge-ISL paths. F1 (document the compaction over-count magnitude) is the only item I'd recommend addressing before merge, and it's a one-line doc addition, not a code change. F2/F3 are optional polish.

<sub>🤖 Reviewed with Claude Code — real-data smoke test on a 20MB byte-range slice of the HF LFS traces.jsonl (10 complete sessions).</sub>

namasl:
Thanks @susiejojo — especially for the real-dataset end-to-end smoke test (the one check I couldn't run here); great to see BC-2 (975 = 988−13 subagents), the huge-ISL drop path, and INV-1/3/6/10 all confirmed on live data. Addressed all three findings in `3ba4ad8c` (docs/test only — converter output unchanged):

- **F1 (recommended):** Added a fidelity note to the CLAUDE.md Weka block — real traces compact/trim context often (your ~28% non-monotone / +54% over-count numbers), so the accumulate clamped-delta over-counts true per-round ISL and replayed KV pressure/hit-rate is an **upper bound**, not exact.
- **F2:** The nil-`in`/`out` main-turn drop now logs `logrus.Debugf` instead of a bare `continue`, so a future data-shape change is debuggable. Kept it weka-local (records are unchanged, so R23 transformation-parity with `convert otel` holds; not touching merged OTel code). Also added `TestConvertWekaSession_DropsMainTurnMissingTokens` pinning the drop + contiguous renumber + the dropped turn's wall-clock being absorbed into the next think gap.
- **F3:** Trimmed the `--help` long text to lead with the native JSONL-per-session shape.

Re-running archon + blis on the new commit.

namasl:

susiejojo:
### Follow-up: full-dataset validation (all 219 sessions, complete 1.07 GB file)

I re-ran the smoke test on the **entire** `traces.jsonl` (not the 10-session head-slice) to remove the sampling caveat from my earlier review. Downloaded the full 1,076,644,069-byte file (219 sessions, last line intact), converted, and replayed.

**Convert — clean at scale:**
```
blis convert weka --input traces.jsonl --trace-output corpus --context-growth accumulate --max-think-time 0
  → Wrote 37732 records from 219 sessions (skipped 0), 4.3s
  → 0 malformed lines, 0 sessions missing id, 0 main turns dropped for nil in/out
  → 1994 subagent entries skipped (BC-2) across the corpus
  → INV-6: re-convert byte-identical ✔
```

**Replay — INV-1 holds, no errors:**
```
blis replay --session-mode closed-loop --max-model-len 2000000 --total-kv-blocks 2000000
  → completed 147, still_queued 96, still_running 13, timed_out 109, injected 365
  → 147+96+13+109 = 365 = injected  ✔ (conservation exact)
  → horizon-bounded: only 365/37732 injected before the 600s sim horizon (expected;
     accumulate-inflated ISL is slow on one instance — not a defect)
  → stderr: only benign default-hardware warnings, no errors/fatals
```

**⚠️ Correction to finding F1 — the over-count is much larger than my head-slice indicated.** My first-10-sessions sample measured a 54% reconstruction over-count; the **full 219-session population is 312%**:

| metric | head-slice (10 sess) | **full dataset (219 sess)** |
|---|---|---|
| rounds | 975 | **37,732** |
| non-monotone rounds (delta clamped→0) | 28.0% | **30.3%** (11,444) |
| negative think gaps (clamped→0) | ~4% | **2.2%** (819) |
| recorded input tokens | 186.9M | **9.60B** |
| accumulate reconstruction over-count | +54.2% | **+312.2%** (+29.97B tokens) |
| worst single-round over-count | +486K | **+9.25M tokens** |

So on real Claude Code traffic the accumulate buffer reconstructs **~4× the recorded cumulative input tokens**, because context compaction (30% of rounds) can't be represented by a monotonically-growing prefix and each deficit clamps to a 0 delta rather than shrinking the buffer. This is a property of the PR-A/PR-B accumulate encoder, **not a bug in this PR** — but PR-F is what puts real Weka data through it, and the ISL/KV-pressure numbers people quote will come from here.

This **strengthens F1 from "worth a doc sentence" toward "should be documented before merge"**: the CLAUDE.md Weka caveat block should state that replayed input length / KV pressure is a substantial *upper bound* (≈3–4× recorded on this dataset) due to compaction, so nobody reads the replay's ISL or hit-rate as a faithful reproduction of the recorded workload. Still a documentation change, not a code fix — the conversion itself is correct per the (documented) delta law.

Everything else in my original review stands. Verdict unchanged: **ready to merge**, with the upgraded recommendation to document the compaction over-count magnitude.

<sub>🤖 Full-dataset re-run — complete 219-session file, convert + closed-loop replay + population statistics.</sub>

namasl:
Addressed the qa-review's `FLAW_FOUND` and @susiejojo's full-dataset follow-up in `3d985bf1` (comment/doc only — converter behavior unchanged):

**qa-review F3 (⛔ the only FLAW) — fixed.** The stale `#1604` attributions in `sim/workload/tracev2.go:188` and `replay.go:128` (which promised the non-lossy `think_time_us` encoding "in #1604", work this PR intentionally does *not* do) are now retargeted to a new tracking issue **#1608**, and reworded to note that #1604 deliberately kept the lossy `0`-sentinel. (`trace_encode.go`'s `#1604` is the converter's own issue reference and is correct — left as-is.)

**qa-review F2 (guide completeness) — done.** Added a parallel **"Weka CC traces"** subsection to `docs/guide/observe-replay-calibrate.md` alongside the OTel section (JSONL input, subagent skip, think recompute, and the compaction upper-bound warning).

**@susiejojo F1 (upgraded to "document before merge") — done.** Updated the CLAUDE.md caveat with the authoritative **full 219-session** numbers: ~30% non-monotone rounds → **+312% (≈3–4×)** accumulate over-count of recorded cumulative input; replayed input length / KV pressure / hit-rate is a substantial **upper bound**, not a faithful reproduction. Thanks for the full-dataset re-run.

On the qa-review's non-blocking "important to consider" items: the `G1–G16` answers were all `CONFIDENT`/correct-as-is; F1's note that the model-routing-safety AC lacks an *end-to-end* `routingRejections==0` assertion is satisfied by the field-level `Model==""` tests **plus** @susiejojo's real-data replay under `--model qwen/qwen3-14b` (the recorded `claude-*` sessions completed rather than being dropped at routing). Nothing there blocks merge.

Re-running archon + blis on `3d985bf1`.

namasl:

namasl:
Per maintainer request: the compaction **fix** now has its own tracking issue — **#1609** (represent context compaction in accumulate replay; a schema + behavioral-contract change to the PR-A/PR-B accumulate path, needs a design doc). The compaction *documentation* (the ~4× upper-bound caveat) already landed in `3d985bf1`; `f1b34905` just retargets the CLAUDE.md + guide caveats to point the over-count at #1609 and keeps #1608 for the separate think-time lossy-0 corner (the two were previously conflated). Doc-only, no code change.

Re-running archon + blis on `f1b34905`.

namasl:

susiejojo (approved):
lgtm

PR #1610, opened 2026-08-24
fix(workload): stop dropping the last request of every per-window epoch (#1607)

### LINKED ISSUE #1607: Bug: per-window workload generation drops the last request of every window (short windows produce zero traffic)
**Describe the bug**

In per-window (time-varying) workload generation, **the last request of every window is
dropped, for every window and every client** — unconditionally, not as an edge case.

`generateRequestsForWindow` samples `numRequests` inter-arrival times and rescales them so
their sum is *exactly* the window duration (`rescaleIATsToMatchDuration`, `sim/workload/generator.go:1202`),
then emits with:

```go
// sim/workload/generator.go:1100-1106  (single-shot branch)
for i := 0; i < numRequests; i++ {
    currentTime += iats[i]
    if currentTime >= window.EndUs {
        break
    }
```

Because the rescale makes the IATs sum to exactly `window.EndUs - window.StartUs`, the
cumulative time after the final IAT lands **exactly** on `window.EndUs`, so the guard
discards it every time. A window therefore emits `numRequests - 1` requests, not
`numRequests`.

Since `numRequests := int(math.Ceil(expectedRequests))` (`generator.go:977`), a window
expecting fewer than one request computes exactly 1 and then loses it — so **short windows
produce no traffic at all**, while still reporting the declared `trace_rate`.

The realized aggregate rate is therefore `rate - (clients x windows)/horizon` rather than the
declared rate, and the error grows as windows get shorter. Anything that sweeps window length
confounds the swept axis with the workload's own fidelity.

**Scope — three call sites, two of them defective:**

| location | status |
|---|---|
| `generator.go:1104` single-shot branch | **defective** |
| `generator.go:1060` multi-session reasoning branch (`single_session: false`) | **defective** — identical `currentTime >= window.EndUs` guard over the same `iats`, loses the last *session* per window |
| `generator.go:1020` / `:1046` / `:1086` single-session + round filters | correct — these compare a request's own arrival, not a rescaled cumulative sum |

Both the eager (`generator.go:914`) and lazy (`stream.go:543`) generation paths call this
function, so `--lazy-generation` is affected identically (measured below). The static
(non-per-window) generation path does not rescale to an exact sum and is unaffected.

**CLI command to reproduce**

`win-1s.yaml` — 10 windows of 1 s at 4 req/s aggregate. Expect `10 x ceil(4x1) = 40`:

```yaml
version: "2"
seed: 42
category: language
clients:
    - id: c0
      rate_fraction: 1
      arrival:
        process: constant
      input_distribution:
        type: gaussian
        params: {max: 800, mean: 256, min: 2, std_dev: 100}
      output_distribution:
        type: gaussian
        params: {max: 1024, mean: 256, min: 1, std_dev: 100}
      streaming: false
      lifecycle:
        windows:
        - {start_us: 0,       end_us: 1000000,  trace_rate: 1.0}
        - {start_us: 1000000, end_us: 2000000,  trace_rate: 1.0}
        - {start_us: 2000000, end_us: 3000000,  trace_rate: 1.0}
        - {start_us: 3000000, end_us: 4000000,  trace_rate: 1.0}
        - {start_us: 4000000, end_us: 5000000,  trace_rate: 1.0}
        - {start_us: 5000000, end_us: 6000000,  trace_rate: 1.0}
        - {start_us: 6000000, end_us: 7000000,  trace_rate: 1.0}
        - {start_us: 7000000, end_us: 8000000,  trace_rate: 1.0}
        - {start_us: 8000000, end_us: 9000000,  trace_rate: 1.0}
        - {start_us: 9000000, end_us: 10000000, trace_rate: 1.0}
aggregate_rate: 4
```

`win-0.2s.yaml` — the same 10 windows shortened to 0.2 s each (`end_us` at 200000, 400000, …,
2000000). Expect `10 x ceil(4x0.2) = 10`.

```bash
./blis run --model qwen/qwen3-14b --workload-spec win-1s.yaml   --horizon 20000000
./blis run --model qwen/qwen3-14b --workload-spec win-0.2s.yaml --horizon 20000000
```

**Expected behavior**

`injected_requests` equals `sum over windows of ceil(rate x duration)` — 40 and 10
respectively — with every arrival strictly inside its window.

No invariant from the list below is violated: conservation and causality both hold on the
requests that *are* generated (30 injected, 30 completed). The defect is upstream of the
simulator, in workload generation: the realized rate silently disagrees with the declared
one, so every result computed from a per-window spec is measured at the wrong offered load.

**Actual behavior**

| spec | expected | actual (`63bd024f`) | actual, `--lazy-generation` |
|---|---|---|---|
| 10 x 1 s windows @ 4 req/s | 40 | **30** (−1 per window) | **30** |
| 10 x 0.2 s windows @ 4 req/s | 10 | **0** | **0** |

The 0.2 s case:

```
level=warning msg="[cluster] no requests provided — simulation will produce zero results"
level=warning msg="[cluster] no requests completed — horizon may be too short or workload too small"
  "completed_requests": 0,
  "injected_requests": 0,
```

Note the warning attributes an empty run to "horizon too short or workload too small". Both
are false here — the horizon is 20 s for a 2 s workload and the spec declares 4 req/s — so
the diagnostic points away from the actual cause. The run exits 0.

Duration-driven, not count-driven: 0.2 s windows yield zero at any window count.

- [ ] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [ ] Work-conserving
- [x] None / unknown

(Workload-generation fidelity: realized rate != declared rate. See "Expected behavior".)

**Environment**
- OS: macOS 15 (darwin/arm64)
- Go version: go1.24.4
- Commit hash: 63bd024f

**Suggested fix**

Sample `numRequests + 1` IATs and rescale across all of them, then emit only the first
`numRequests`. The trailing gap absorbs the exact-sum property instead of one of the emitted
gaps, so every arrival lands strictly inside `(StartUs, EndUs)` with jitter at both ends and
the count is exactly `numRequests`. The effective mean gap inside a window becomes
`duration/(numRequests+1)` rather than `duration/numRequests`, which is the correct reading of
"N arrivals inside a window of length D". The reasoning branch needs its loop bound moved from
`len(iats)` to `numRequests` so it does not consume the new trailing gap as an extra session
attempt. The `>= EndUs` guards are worth keeping as defence in depth, documented as
unreachable in normal operation.

Two alternatives considered and rejected:

- **`>` instead of `>=`** — admits an arrival at exactly `window.EndUs`, which is the next
  window's start instant. Any consumer bucketing by `arrival_time / window_length` would
  attribute it to the wrong window, with the wrong window's parameters.
- **Moving `currentTime += iats[i]` to the end of the loop** — puts an arrival at exactly
  `window.StartUs` for every co-active client simultaneously: a thundering herd at every
  window boundary.

I have this fix and its regression test running locally and can open a PR.

Regression test shape: assert `len(requests) == ceil(rate x duration)` and
`StartUs < arrival < EndUs` for every request, across several window lengths **including one
where `ceil` gives 1** — the case that currently yields zero. Against `63bd024f` that test
fails off-by-one-low in every case and reports 0 for the short window; both branches need a
case, since the reasoning branch has its own copy of the defect.

**Additional context**

Discovered while running a capacity-planning sweep in which per-window epoch length was the
swept axis, so the per-window count was small enough for the missing request to dominate. The
mechanism above predicted every observed count exactly before the code was read.

**Adjacent, not part of this report:** `numRequests := int(math.Ceil(expectedRequests))`
(`generator.go:977`) also means every client with any traffic gets at least one request per
window, so a skewed popularity distribution across clients flattens toward uniform as windows
shorten, independent of this bug. That is a design question about per-window count sampling
(e.g. Poisson with mean `rate x duration`, or one continuous per-client arrival process
carrying the fractional remainder across window boundaries), not a defect fix, and is better
tracked separately. Fixing the drop reported here does not address it.

atantawi:
Fixes #1607.

## What was wrong

`generateRequestsForWindow` rescales its sampled inter-arrival times so they sum to *exactly*
the window duration, then emits with `currentTime += iats[i]; if currentTime >= window.EndUs { break }`.
The cumulative time after the final gap therefore lands exactly on `window.EndUs`, and the
boundary guard discarded it — **every window, every client, unconditionally**. Since
`numRequests = ceil(rate x duration)`, a window expecting fewer than one request computed 1 and
lost it, so short windows produced no traffic at all while still reporting the declared
`trace_rate`.

## The fix

Sample `numRequests + 1` gaps and rescale across all of them, then emit only the first
`numRequests`. The trailing gap absorbs the exact-sum property instead of one of the emitted
gaps, so every arrival lands strictly inside `(StartUs, EndUs)` with jitter at both ends and the
count is exactly `numRequests`. The effective mean gap inside a window becomes
`duration/(numRequests+1)` rather than `duration/numRequests` — the correct reading of "N
arrivals inside a window of length D", since N interior points have N+1 surrounding gaps.

Two alternatives were rejected, and the reasoning is in the commit message and code comments:
`>` instead of `>=` admits an arrival at exactly `window.EndUs`, which any consumer bucketing by
`arrival_time / window_length` would attribute to the wrong window with the wrong window's
parameters; moving the `currentTime` increment to the end of the loop puts an arrival at exactly
`window.StartUs` for every co-active client simultaneously, a thundering herd at every boundary.

The `>= EndUs` guards are kept as defence in depth and now documented as unreachable in normal
operation.

## Three branches, all three now pinned by tests

| branch | code edited? | was defective? |
|---|---|---|
| single-shot (`generator.go:1104`) | yes — loop bound + `numRequests+1` sampling | yes — lost the last request per window |
| multi-session reasoning (`:1060`) | yes — loop bound `len(iats)` -> `numRequests` | yes — lost the last *session* per window |
| single-session reasoning (`:1044`) | **no** | **yes** — at `ceil == 1`, `iats[0]` equalled the whole duration, so `startTime == EndUs` and the branch returned `(nil, nil)`: the window's only session, silently dropped |

The third row is worth calling out. That branch has no emission loop — it derives its single
session start from `iats[0]` alone — so it needed no edit and the first commit's message calls
it "untouched". That is true of the code and misleading about the behavior: it was defective for
the same root cause and is repaired as a side effect of the shared rescale. The third commit adds
its regression test, confirmed RED against the pre-fix generator (0 requests at all six seeds
for the `ceil == 1` case, 1 after).

None of the four pre-existing reasoning tests could have caught any of this: they assert
`len(sessionsFound) < 2` or a zero/nonzero session count, both of which a dropped session
satisfies.

## Verification

End-to-end, with the 10-window specs from #1607 (`--horizon 20000000`):

| spec | expected | before | after | after, `--lazy-generation` |
|---|---|---|---|---|
| 10 x 1 s windows @ 4 req/s | 40 | 30 | **40** | **40** |
| 10 x 0.2 s windows @ 4 req/s | 10 | 0 | **10** | **10** |

Both the eager (`generator.go:914`) and lazy (`stream.go:543`) paths call this function, so both
were affected and both are fixed. The static (non-per-window) generation path does not rescale to
an exact sum and is untouched.

- `go build ./...` — clean
- `go test ./...` — all 16 packages pass, 0 failures
- INV-6: two runs at a fixed seed produce byte-identical stdout (verified by hash)

## Behavioral change to disclose

At a fixed seed, per-window workloads now produce **different arrival times and different sampled
token lengths** than before. The arrival change is the fix itself. The token change is a
second-order consequence: the extra `SampleIAT` draw per window shares the same `*rand.Rand` as
`inputSampler.Sample(rng)` / `outputSampler.Sample(rng)` (`generator.go:1020` vs `:1147-1148`),
so the stream shifts by one draw per window.

This is unavoidable without partitioning the RNG further, which would be a larger change than a
defect fix warrants. No committed golden depends on it — the full suite is green — but anyone
holding local per-window baselines will see them move, and they were measured at the wrong
offered load.

## Not addressed here

`numRequests := int(math.Ceil(expectedRequests))` (`generator.go:977`) also means every client
with any traffic gets at least one request per window, so a skewed popularity distribution across
clients flattens toward uniform as windows shorten. That is a separate design question about
per-window count sampling (Poisson with mean `rate x duration`, or one continuous per-client
arrival process carrying the fractional remainder across boundaries), independent of this bug and
not fixed by it. Flagged in #1607 under "Adjacent, not part of this report" so it can be tracked
on its own.

## Tier

Two files, behavioral change, so **Medium** per the "when in doubt, tier up" rule in
`pr-workflow.md`. Self-audit run across all 10 dimensions — the single-session branch defect and
the RNG-stream shift above are both findings from that pass, not from the original fix.


atantawi:
## Self-review verdict: READY TO MERGE — 0 blocking, 3 non-blocking observations

Reviewed directly (single reviewer, no agent fan-out) against `docs/contributing/standards/{invariants,rules,principles}.md` and the cross-path / preemption / timeout checklists.

### Correctness and invariants

- **Root cause addressed, not symptom.** The exact-sum rescale is what put the final arrival on `window.EndUs`; adding a trailing gap removes the coincidence rather than loosening the guard. The two tempting alternatives (`>` instead of `>=`, or moving the increment) are both rejected with reasons in the code and commit message, and both would have introduced worse artifacts (cross-window misattribution; boundary thundering herd).
- **A second, unstated benefit of the shape chosen.** `rescaleIATsToMatchDuration` adds its rounding residual to `rescaled[len-1]`. Post-fix that index is the *unused* trailing gap, so the residual no longer perturbs any emitted gap. The residual is always `>= 0` (flooring guarantees `rescaledSum <= targetDuration`), so the trailing gap can only grow — which is what makes the strict-interior property hold unconditionally rather than probabilistically.
- **INV-6 determinism:** two runs at a fixed seed produce byte-identical stdout (hash-verified). Cross-version output does change for per-window specs; disclosed in the PR body with the mechanism (shared `*rand.Rand`, one extra draw per window).
- **INV-5 causality / INV-1 conservation:** unaffected. Arrivals remain ordered and strictly inside the window; deadlines derive from arrivals via `computeDeadline`, so they shift consistently.
- **INV-13 run/replay parity:** N/A by construction — replay is trace-driven and never calls the generator.

### Cross-path parity — checked, and satisfied without per-path work

| path | uses `generateRequestsForWindow`? | status |
|---|---|---|
| `blis run` | yes — `cmd/root.go:1786` / `:1798` | affected, fixed, measured (40/40, 10/10 eager and lazy) |
| `blis observe` | yes — `cmd/observe_cmd.go:444` / `:453`, same `GenerateWorkload`/`GenerateWorkloadLazy` entry points | affected, fixed by the shared library change |
| `blis replay` | no — trace-driven ingestion | not applicable |

Parity here is structural rather than duplicated: the fix lands in `sim/workload`, below all three commands, so `observe` inherits it. No CLI flag, spec field, default, or metric is introduced, so there is no surface that could drift between paths. **No follow-up parity issue needed.**

### Preemption safety, timeouts

Both N/A, verified rather than assumed: the diff touches only pre-simulation workload construction. No per-request metric, state transition, aggregate counter, or `TimeoutEvent` interaction. Neither timeout mechanism (DES `TimeoutEvent` vs HTTP client) is touched; the 300s defaults are unchanged.

### Tests

Behavior-based and refactor-surviving — they assert counts and interval membership, not internal structure. All three affected branches are now pinned, each confirmed RED against the pre-fix generator. Worth noting the third test exists because the first commit's "the single-session branch is untouched" was true of the code and misleading about the behavior; that branch was silently dropping its only session at `ceil == 1`.

### Performance

One extra `int64` and one extra `SampleIAT` per window. Negligible even for the many-small-windows layouts this path targets.

---

## Non-blocking observations

**1. The `>= EndUs` guards are now unreachable but still fail silently.** They are documented as defence in depth, which is right, but their failure mode is a silent truncation of the window. Pre-fix that was normal operation; post-fix, if one ever fires it means the `numRequests+1` invariant has been broken by a later change — exactly the kind of thing R1 says should not pass quietly. Consider a `logrus.Warnf` in the guard body so a future regression surfaces instead of silently shortening windows. Not done here because it widens a defect fix into a logging change.

**2. No test pins the interior-gap semantics or strict ordering.** The tests assert count and `StartUs < arrival < EndUs`, but not that arrivals are strictly increasing, nor that the effective mean gap is `duration/(numRequests+1)`. The latter is the deliberate semantic change and is currently documented only in a code comment. A tiny table-driven addition would pin it as a law rather than a comment.

**3. The mean-gap semantics are undocumented outside the code.** `duration/(numRequests+1)` is the correct reading of "N arrivals inside a window of length D", but a user reading the workloads guide would still expect `duration/numRequests`. A one-line note in the per-window section of the workload docs would close that gap.

None of the three changes the correctness of this PR; all three are better handled as separate small changes than by widening this one.

## Explicitly out of scope

The `ceil`-based per-window count (`generator.go:977`) flattens skewed client popularity as windows shorten. Independent of this defect, unfixed by it, and flagged in #1607 for separate tracking.

susiejojo:
## Independent review verdict: ✅ READY TO MERGE — 0 blocking, 3 non-blocking (all pre-noted by author)

Reviewed against `docs/contributing/standards/{invariants,rules,principles}.md` and the run/replay/observe parity, preemption-safety, and timeout checklists. I re-derived the fix from the code, ran the suite, and reproduced the end-to-end numbers rather than taking the PR body on trust.

### Correctness — verified, and the fix is *provably* (not just empirically) robust

The root cause is real and the fix is the right one: `rescaleIATsToMatchDuration` forces the emitted gaps to sum to *exactly* `windowDurationUs`, so the cumulative time after the final gap landed on `window.EndUs` and the `>= EndUs` guard dropped it — every window, every client. Sampling `numRequests+1` gaps and emitting only the first `numRequests` moves the exact-sum residual into an unused trailing gap.

The load-bearing claim is *"the trailing gap is strictly positive, so emitted arrivals fall strictly inside the window."* I checked the branch that guarantees it:
- Every arrival sampler reachable via `NewArrivalSampler` — `constant`, `poisson`, `gamma`, `weibull`, and the `default`→Poisson fallback (`sim/workload/arrival.go:214-288`) — floors its IAT to **≥ 1µs**.
- The one sampler that can return 0 (`NormalizedExponentialSampler`, the exhaustion sentinel) has **no switch case** and is unconstructible on this path.
- `rescaleIATsToMatchDuration` (`generator.go:1229-1230`) adds the residual `targetDuration - rescaledSum` to `rescaled[len-1]`; flooring guarantees `rescaledSum <= targetDuration`, so the residual is ≥ 0 and the trailing gap can only grow.

⟹ the sum of the emitted `numRequests` gaps is **strictly < windowDurationUs**, unconditionally. The `>= EndUs` guards are genuinely unreachable in normal operation, exactly as documented. Nice property; it's what makes the interior bound hold for *any* draw rather than probabilistically.

### Invariants
- **INV-6 determinism:** reproduced — two fixed-seed runs of the 0.2s spec produce byte-identical stdout (`shasum` match). The cross-version output change for per-window specs (one extra `SampleIAT` draw per window shifts the shared `*rand.Rand`) is correctly disclosed with its mechanism.
- **INV-1 conservation / INV-5 causality:** unaffected — diff is entirely pre-simulation workload construction; arrivals stay ordered and interior, deadlines derive from arrivals via `computeDeadline`.
- **INV-13 run/replay parity:** N/A by construction — confirmed `cmd/replay*.go` contains **no** call to `GenerateWorkload*`/`generateRequestsForWindow`; replay is trace-driven.

### Cross-path parity — satisfied structurally, no per-path work needed
The fix lives in `sim/workload`, below all three commands, so it's inherited rather than duplicated:

| path | reaches `generateRequestsForWindow`? | status |
|---|---|---|
| `blis run` | yes (`generator.go:914` eager, `stream.go:543` lazy) | fixed — reproduced **10/10** (was 0) for both eager and lazy on the #1607 0.2s spec |
| `blis observe` | yes — same `GenerateWorkload`/`GenerateWorkloadLazy` entrypoints (`cmd/observe_cmd.go:444,453`) | fixed by the shared library change |
| `blis replay` | no — trace-driven | N/A |

No CLI flag, spec field, default, or metric is added, so there is no surface that could drift between paths. **No follow-up parity issue required.** Preemption-safety and timeout checklists are both N/A (no per-request metrics, state transitions, aggregate counters, `TimeoutEvent`, or HTTP-timeout interaction touched) — verified, not assumed.

### Architecture (requested "Archon" review)
There is no architectural change to review: this is a two-file, in-place defect fix in an existing function behind an unchanged signature — no new module boundary, interface, config, or extension point. Archon-style architectural review is **N/A** here.

### Tests
Behavior-based and refactor-surviving — they assert exact counts (`ceil(rate*duration)`), distinct session counts, and strict interior membership (`StartUs < arrival < EndUs`), not internal structure. All three affected branches are pinned, each confirmed RED against the pre-fix generator, including the `ceil==1` case where an off-by-one is a 100% miss rather than a rounding blur. The single-session-reasoning test is a genuinely good catch — that branch was defective for the same root cause despite needing no code edit, and the four pre-existing reasoning tests (`len<2` / zero-vs-nonzero) provably could not have caught it. Full `go test ./sim/workload/` passes on the branch; `go build ./...` clean.

### Non-blocking observations (all already flagged by the author — none block merge)
1. **Unreachable `>= EndUs` guards still fail silently.** Now that they're unreachable in normal operation, a firing means the `numRequests+1` invariant was broken by a later change — the kind of thing R1 says shouldn't pass quietly. A `logrus.Warnf` in the guard body would surface a future regression. Reasonably deferred to avoid widening a defect fix into a logging change.
2. **No test pins the `duration/(numRequests+1)` mean-gap semantics or strict monotonicity of arrivals.** The deliberate semantic change lives only in a code comment. A small table-driven addition would pin it as a law.
3. **Mean-gap semantics undocumented outside the code.** `docs/guide/workloads.md` has no per-window mean-gap note; a reader would expect `duration/numRequests`. A one-line addition to the workloads guide would close it.

### Out of scope (correctly deferred)
The `ceil`-based per-window count (`generator.go:977`) flattens skewed client popularity as windows shorten — an independent design question about per-window count sampling, flagged in #1607 for separate tracking.

**Verdict: ready to merge.** The fix is correct, minimal, provably robust, deterministic, parity-clean, and well-tested. The three observations are quality-of-life follow-ups the author already identified, none affecting correctness.

atantawi:
## Review response — all 3 non-blocking observations addressed

Both reviews converged on the same three observations. All three are now fixed in three focused commits, with one documented deviation from what was suggested (observation 2).

| # | Observation | Commit | Outcome |
|---|---|---|---|
| 1 | Unreachable `>= EndUs` guards fail silently | `d430e180` | Implemented as suggested |
| 2 | No test pins mean-gap semantics / ordering | `0d91ee5c` | Implemented, with a **correction** to the "strictly increasing" part |
| 3 | Mean-gap semantics undocumented outside the code | `7bf0cf68` | Implemented as suggested |

### 1. `logrus.Warnf` on the unreachable guards (`d430e180`)

Each of the three start-time guards now names the client, window, offending arrival, and the broken invariant before truncating. Warnings go to stderr, so stdout stays byte-identical (INV-6).

Scoped deliberately to the **three start-time guards only**. The two inner reasoning-round filters (`req.ArrivalTime >= window.EndUs`, generator.go:1071/1119) are left silent on purpose: a multi-turn session's later rounds legitimately arrive past the window end, so those fire in normal operation and are not invariant violations. Warning there would be noise on every reasoning run.

Verified silent in normal operation — the #1607 specs plus single-session and multi-session reasoning variants emit **zero** warnings, eager and lazy.

### 2. Tests pinning the semantics — with a correction on "strictly increasing"

Two tests added.

**`TestGenerateRequestsForWindow_InteriorGapSemantics`** pins the `D/(N+1)` gap law using the `constant` arrival process, which makes it deterministic and seed-independent rather than statistical: every sampled IAT is identical, so the rescale divides the window evenly. It asserts gap uniformity *exactly*, the gap width as `D/(N+1)`, and a trailing slack of at least one gap — pre-fix that slack was identically zero, which is precisely why the final arrival collided with `EndUs`.

Both axes are discriminating, confirmed by perturbation rather than assumed:

- reverted to `numRequests` gaps → **count fails** (0 vs 1, 4 vs 5, 39 vs 40, …)
- perturbed to `numRequests+2` → count passes, **spacing fails** by 581µs at `numRequests=40`, against a 2µs tolerance

One deliberate imprecision: the gap width is asserted within 2µs rather than exactly. `ConstantArrivalSampler` derives its IAT as `int64(1.0/(rate/1e6))`, and that float division truncates — at 5 req/s it yields `199999µs`, not `200000µs`, shifting the rescaled gap by 1µs. That artifact belongs to the sampler, not to the gap semantics under test, and 2µs is orders of magnitude tighter than the `D/N` vs `D/(N+1)` difference the test exists to distinguish (2× at `N == 1`).

**`TestGenerateRequestsForWindow_ArrivalsOrderedAndInterior`** covers poisson/gamma/weibull × 6 seeds for exact count, strict interiority, and ordering.

> **Correction on the "strict ordering" suggestion.** Ordering is pinned as **non-decreasing**, not strictly increasing. The reasoning in the review — that every sampler reachable via `NewArrivalSampler` floors its raw IAT to ≥ 1µs — is correct, but the floor happens *before* the rescale. `rescaleIATsToMatchDuration` then multiplies by a scale factor that can be < 1 and truncates toward zero, so a zero-width gap is representable:
>
> ```
> rescaleIATsToMatchDuration([1, 1, 1, 1000], 1000) == [0, 0, 0, 1000]
> ```
>
> Two arrivals sharing one microsecond is legal and harmless (the downstream merge sorts by `ArrivalTime`). Asserting strict monotonicity would encode a property the rescale does not guarantee, and could fail on an unlucky future draw. Non-decreasing is the true law, and the test says so with the reason inline.
>
> This does **not** weaken the unreachability argument for the guards in observation 1 — that argument needs the *trailing* gap to be strictly positive, which holds independently: `rescaled[last] = floor(iat_last x scale) + residual`, and for the trailing gap to be 0 both terms must vanish, which requires `iat_last == 0` — impossible given the ≥ 1µs sampler floor. Zero-width gaps can occur in the *interior* without ever putting an emitted arrival on `EndUs`.

### 3. Docs (`7bf0cf68`)

Added a note to the Phased Workloads section of `docs/guide/workloads.md`: the exact per-window count, strict interiority, the `D/(N+1)` mean gap and why it is the correct reading, and the `ceil` rounding that gives any non-zero-rate window at least one arrival.

### Verification

- `go build ./...` — clean
- `go test ./...` — all 16 packages pass, 0 failures
- `golangci-lint run ./...` — **0 issues**
- INV-6: two fixed-seed runs produce byte-identical stdout (`shasum` match)
- End-to-end #1607 numbers unchanged: **40/40** and **10/10**, eager and lazy
- New guard warnings confirmed silent across all three branches (single-shot, single-session reasoning, multi-session reasoning)

The out-of-scope `ceil`-based per-window count (`generator.go:977`) remains deferred to #1607's separate-tracking note, unchanged.

atantawi:
@claude /blis-pr-review

**Scope note for this pass (and for human reviewers):** both prior verdicts — [mine](#issuecomment) and @susiejojo's — were against the diff *before* the review-response commits. Everything up to that point was already agreed READY TO MERGE. What has not been reviewed by anyone is the delta `d430e180..7bf0cf68`:

| Commit | Change |
|---|---|
| `d430e180` | `fix(workload)`: window guards now `Warnf` instead of silently truncating |
| `0d91ee5c` | `test(workload)`: pin interior-gap semantics and arrival ordering |
| `7bf0cf68` | `docs(workloads)`: document per-window arrival spacing |

Please weight the pass toward that delta rather than re-deriving the original fix. Two specific asks:

1. **Observation 2 was resolved with a documented deviation** from what both reviewers suggested. Please evaluate the deviation on its merits and say plainly whether you agree with it or think the originally-suggested approach was better — do not treat "author documented it" as sufficient.
2. `d430e180` changes a previously-unreachable guard from silent truncation to a warning. Confirm the guard is genuinely unreachable in the new code (if it is reachable, a warning is the wrong severity per R1), and that the new tests in `0d91ee5c` assert observable arrival behavior rather than internal window structure.

CI is green on `7bf0cf68` (build, lint, all 6 test shards).

Please post the verdict as a comment. Do not push commits to this branch — if you find something that needs changing, describe it and I will make the change locally.

atantawi:
## Follow-up on the automated delta review: its interiority nit is correct, and it understates the case

The automated pass returned READY TO MERGE / 0 blocking on `d430e180..7bf0cf68`, and flagged one non-blocking item — that the doc comment on `TestGenerateRequestsForWindow_ArrivalsOrderedAndInterior` claims "Strict interiority IS unconditional whenever the mean gap is at least 1us." I measured this instead of accepting or dismissing it: a temporary test driving `generateRequestsForWindow` over **200,000 seeds per case** (since removed, not committed).

| Case | N | first arrival == `StartUs` | runs containing a zero *interior* gap |
|---|---|---|---|
| poisson, 100 req/s, 1s | 100 | 15 (0.0075%) | **1881 (0.94%)** |
| gamma, 100 req/s, 1s | 100 | 15 (0.0075%) | 1834 (0.92%) |
| weibull, 100 req/s, 1s | 100 | 20 (0.0100%) | 1900 (0.95%) |
| poisson, 10 req/s, 1s | 10 | 1 (0.0005%) | 21 |
| poisson, 5 req/s, 0.1s | 1 | 0 | 0 |

**Mechanism.** `rescaleIATsToMatchDuration` computes `rescaled[i] = int64(float64(iat) * scaleFactor)` and then adds the rounding residual **only to the last element**. So the trailing gap is floored up to `>= 1us` — which is exactly why the `>= EndUs` guards are unreachable and why `< EndUs` is a genuine law — but *nothing floors the leading gap*. When `scaleFactor < 1`, `rescaled[0]` can truncate to 0 and the first arrival lands on `window.StartUs`.

Two corrections to the automated review, in both directions:

1. It estimated ~11 per million. The real rate is **~75 per million** at the bursty parameters — off by ~7x, though the mechanism it described is exactly right.
2. It classified this as comment-accuracy only. It is more than that: `assert.Greater(req.ArrivalTime, window.StartUs)` **encodes the same non-law that observation 2's deviation correctly rejected for interior gaps** — same truncation, same absence of a floor. An arrival exactly at `StartUs` is legal (the window is half-open, and the sweep confirms those requests are not dropped by any guard), so that assertion is strictly stronger than the contract and would fail if the pinned seed set ever changed.

**Unplanned upside for @susiejojo's and my own strict-monotonicity suggestion:** zero *interior* gaps occur in **~0.94% of runs** at the bursty parameters. The strictly-increasing assertion we both asked for would not have been theoretically flaky — it would fail roughly 1 run in 100. The deviation was right for a substantially stronger reason than its commit message claims.

### Change I am about to push

- Relax the lower bound to `assert.GreaterOrEqual(req.ArrivalTime, window.StartUs)`.
- **Keep `assert.Less(req.ArrivalTime, window.EndUs)` strict** — that bound *is* proven unconditional by the trailing residual, and it is the actual subject of #1607. This is deliberately asymmetric.
- Rewrite the doc paragraph to state the asymmetry and cite the measured rate, replacing the "unconditional" claim.

To be explicit that this is not a weakened test: the regression #1607 guards against is the final arrival colliding with `EndUs`, and that assertion stays strict. The bound being relaxed asserted something the rescaler never guaranteed and that the half-open window does not require.

atantawi:
@claude /blis-pr-review

atantawi:
## Review response — the one blocking finding is fixed in `c8be2f2b`

The automated delta pass was right, and it found the one thing the three earlier verdicts (mine, @susiejojo's, and the first automated pass) all missed: `7bf0cf68` and `0f6b4068` contradict each other on the lower interiority bound.

- `docs/guide/workloads.md:484` said arrivals are placed **"strictly inside the window — never on either boundary instant."**
- `0f6b4068` had already relaxed the test to `assert.GreaterOrEqual(req.ArrivalTime, window.StartUs)` after measuring that an arrival *can* land on the window start.

A reader who trusted the guide and then observed an arrival at `start_us` would have concluded they'd found a bug. Docs-only fix; the docs claim was the stale half, not the test.

### Verified before accepting the finding

I re-derived it from the code rather than taking the review's word:

| site | fact |
|---|---|
| `generator.go:1292` | `rescaled[len-1] += targetDuration - rescaledSum` — residual lands **only** on the trailing gap, so it is floored up and `< end_us` is a genuine unconditional law |
| `generator.go:1287` | `rescaled[i] = int64(float64(iat) * scaleFactor)` — **no floor** on `rescaled[0]`; a scale factor below 1 truncates the leading gap to 0 |
| `generator.go:1044`, `:1152` | first arrival is `window.StartUs + iats[0]`, so it can land exactly on the start instant |

The asymmetry is real and it is structural, not a draw-dependent accident.

### Re-measured the rate rather than copying it

The paragraph now cites a number in user-facing prose, and the first automated pass's own estimate had been off by ~7×, so I re-ran the sweep instead of reusing `0f6b4068`'s figure (temporary harness over 200,000 seeds per arrival process, driving `generateRequestsForWindow` directly; not committed):

| process | window | first arrival == `start_us` | runs with a zero interior gap |
|---|---|---|---|
| poisson | 100 req/s, 1 s (N=100) | 15/200000 (7.5e-05) | 1895/200000 (0.95%) |
| gamma | 100 req/s, 1 s (N=100) | 15/200000 (7.5e-05) | 1848/200000 (0.92%) |
| weibull | 100 req/s, 1 s (N=100) | 20/200000 (1.0e-04) | 1920/200000 (0.96%) |

Roughly one draw in 13,300 — `0f6b4068`'s figure confirmed independently. (My zero-interior-gap counts differ trivially from that commit's, 1895 vs 1881, because I count each run once on first occurrence; immaterial to the conclusion.)

### Three corrections, not one

The review's suggested replacement sentence fixed the bound but left two adjacent claims stale, so the rewrite goes slightly further than quoted:

1. **The bound itself** — now "strictly before `end_us`" (unconditional) and "at or after `start_us`" (leading gap may vanish), with the reason: a window is half-open, `[start_us, end_us)`.
2. **"The window opens and closes with a gap rather than with an arrival"** was half-wrong for the identical reason — the review noted this in passing but its proposed sentence kept the phrase. Only the *closing* gap is guaranteed, and the paragraph now says only that.
3. **The thundering-herd rationale needed qualifying, not deleting.** That artifact came from placing *every* co-active client's first arrival on the start instant *deterministically* — which one client occasionally landing there does not reintroduce. Dropping the rationale would have lost the reason interior placement was chosen; leaving it unqualified would have contradicted the new bound.

Also switched `window.EndUs` / `window.StartUs` to the YAML spellings `end_us` / `start_us`, matching lines 464–476 of the same guide — the Go field names were mine and were out of register for a user-facing document.

### On the earlier misclassification

Worth recording, since it is the one process miss in this PR's review history: the first automated pass identified the exact mechanism and then filed it as *"a comment-accuracy nit, not a test correctness issue."* It was neither — the same non-law was encoded in a live `assert.Greater`, and once `0f6b4068` relaxed that assertion the defect moved into the docs. The second pass caught both the reclassification and the doc contradiction. The lesson generalizes: a finding about a *comment* deserves a check of whether the assertion beneath it encodes the same claim.

### Verification

- `go build ./...` — clean
- `go test ./...` — 13 packages ok, **0 FAIL**
- `golangci-lint run ./...` — **0 issues**
- Docs-only diff (1 line changed), so no behavior change: INV-6 byte-identity, INV-1, INV-5 and cross-path parity are untouched by construction
- End-to-end #1607 numbers unaffected (40/40 and 10/10, eager and lazy) — no generator code in this commit

Nothing else from the delta review remains open: asks 1 and 2 were both answered affirmatively (guards genuinely unreachable, `Warnf` the right severity, non-decreasing deviation correct and now empirically vindicated at ~1 run in 100), and the non-blocking list was empty.

The out-of-scope `ceil`-based per-window count (`generator.go:977`) stays deferred to #1607's separate-tracking note, unchanged.

atantawi:
@claude /blis-pr-review

atantawi:
## Review response — blocking finding fixed in `5b9cebe0`, plus both comment observations

The pass was right, and it caught something the previous four verdicts (including my own two) all missed: `0f6b4068` fixed the strict lower bound in **one** of four places. Three earlier tests still asserted the law the PR itself had disproved.

### Blocking: confirmed as real latent flakiness, not theoretical

I did not take this on probability. Sweeping 3,000,000 seeds per case at these tests' own parameters (rate 5 req/s, poisson) produced concrete witnesses:

| case | first arrival == `start_us` | rate | first witness |
|---|---|---|---|
| `dur=100000us` (N=1) | 37/3,000,000 | 1.2e-05 | **seed 276753** |
| `dur=500000us` (N=3) | 20/3,000,000 | 6.7e-06 | seed 276753 |
| `dur=1000000us` (N=5) | 17/3,000,000 | 5.7e-06 | seed 276753 |
| `dur=5000000us` (N=25) | 14/3,000,000 | 4.7e-06 | seed 276753 |

Adding seed 276753 to `ExactCountAndInteriorArrivals` made **all four subtests fail** on `assert.Greater`, for behavior this PR documents as legal:

```
--- FAIL: .../short_window,_ceil_gives_1
    Messages: [seed %d request %d: arrival must be strictly > window start 276753 0]
--- FAIL: .../sub-second_window,_ceil_gives_3
--- FAIL: .../one-second_window
--- FAIL: .../multi-second_window
```

**Mechanism** (probed directly rather than inferred): at seed 276753 the sampler draws a raw first IAT of exactly **1µs — its own floor** — and `rescaleIATsToMatchDuration` multiplies by a scale factor below 1 and truncates, so the leading gap becomes 0. That holds at `N = 1, 2, 3, 5, 25`, i.e. at every count all three tests use, which is why a single seed covers all of them.

### What changed, and one thing beyond the required fix

The required three one-liners, plus a step I think earns its place:

- `assert.Greater` → `assert.GreaterOrEqual` at the three sites; messages `"strictly >"` → `"at or after"`. The `assert.Less` upper bounds **stay strict** — that bound *is* unconditional and is the actual #1607 regression.
- **Seed 276753 is now pinned in all three tests**, with a comment stating what it exercises. Changing the assertion alone would have left the inclusive bound *asserted* but never *exercised* — still passing by luck, just with a weaker predicate. Pinning the witness converts it into a covered boundary case. Verified discriminating: reverting the three assertions to `Greater` fails **all 8 subtests deterministically**, so a future regression to the strict form cannot slip through.

This cannot introduce flakiness in the other direction: with an inclusive bound, an arrival at `start_us` is legal, so if a future RNG-stream shift moves seed 276753 off the boundary the test simply stops exercising that case — it never fails for legal behavior.

### Both non-blocking comment findings also fixed

**Observation 1** — `generator.go:1010` still claimed arrivals "fall strictly inside `(window.StartUs, window.EndUs)`". Same stale symmetric claim `c8be2f2b` fixed in the guide; the code comment now states the asymmetry and points at the tests.

**Observation 2** — the three guard comments credited unreachability to the rescale's exact-sum property alone. Correct that this is incomplete: trailing-gap positivity needs **two** facts, the residual landing on the last element **and** the ≥ 1µs raw IAT floor every `NewArrivalSampler` sampler applies. Without the floor a zero raw IAT would rescale to a zero trailing gap. All three comments now name both.

> One correction to that observation's supporting example. It offered `rescaleIATsToMatchDuration([1,1,1,1000], 1000)` → `[0,0,0,1000]` as showing that exact-sum alone does not guarantee a positive trailing gap. That example's trailing gap is **1000** — it demonstrates a zero *leading* gap, which is the separate (and now well-documented) asymmetry. The precondition point stands on its own: trailing gap `= floor(iat_last × scale) + residual` is zero only if both terms vanish, and `residual = 0` forces every product to be an exact integer, so `floor(iat_last × scale) = 0` then requires `iat_last = 0` — which only the sampler floor rules out. That is the reasoning the comments now carry.

**Observation 3** (lazy path / proportional-allocation coverage) — declining, as the review itself classified it "not a correctness gap." `stream.go:543` and the eager path call the *same* function; a duplicate law test through the lazy entrypoint would assert the identical contract against the identical code. The lazy path's distinct behavior (per-window batch merging) is covered by its own tests. Happy to revisit if you disagree.

### Verification

- `go build ./...` — clean
- `go test ./...` — 13 packages ok, **0 FAIL** (`-count=1`, cache bypassed)
- `golangci-lint run ./...` — **0 issues**
- Revert-to-strict check: 8/8 subtests fail deterministically, confirming the pinned seed enforces rather than decorates
- No behavior change: comments and test assertions only, so INV-6 byte-identity holds and the #1607 end-to-end numbers (40/40, 10/10, eager and lazy) are unaffected

Pre-existing `gofmt` comment-alignment drift at test-file lines 13/33/72 is present on `HEAD` too, is not lint-enforced, and is unrelated to this change — left untouched rather than adding formatting churn to a fix PR.

### One process note

This is the second time a finding in this PR was first filed as comment-accuracy and turned out to sit under a live assertion encoding the same false law — first the `ArrivalsOrderedAndInterior` comment, now these three. The generalizable rule: when a *comment* is found to overstate a bound, grep the file for assertions on that same bound before closing it out. I'd have caught these three at `0f6b4068` time by doing that.

susiejojo (approved):
lgtm

PR #1304, opened 2026-05-08
fix(replay): use send_time_us as DES injection time for observed traces

sriumcp:
## Summary

- Fixes systematic TTFT pessimism in calibration for concurrency-mode traces
- Adds `injectionTime()` helper in `sim/workload/replay.go` that returns `SendTimeUs` when `> 0`, else falls back to `ArrivalTimeUs`
- Applies at all three `sim.Request` construction sites in `LoadTraceV2Requests` and `LoadTraceV2SessionBlueprints`
- Think-time gap derivation for sessions (`ArrivalTimeUs` differences) is deliberately unchanged per issue specification

## Root cause

When `blis observe --concurrency N` runs, requests that wait for a free slot have `send_time_us > arrival_time_us`. `blis calibrate` correctly measures real TTFT as `first_chunk_time_us - send_time_us`. But `blis replay` was injecting at `arrival_time_us` — up to 50ms+ too early — causing the sim to accumulate phantom queue time that never existed on the server. This made the simulator look consistently pessimistic for all concurrency-mode traces.

For rate-mode traces and generated traces, `send_time_us == arrival_time_us` (or `send_time_us == 0`), so those are unaffected.

## Behavioral Contracts

- **BC-1**: GIVEN a trace record where `SendTimeUs > 0`, WHEN `LoadTraceV2Requests` builds a request, THEN `req.ArrivalTime == rec.SendTimeUs`
- **BC-2**: GIVEN `SendTimeUs == 0`, THEN `req.ArrivalTime == rec.ArrivalTimeUs` (backward-compatible)
- **BC-3**: Session round-0 requests also use `SendTimeUs` when set
- **BC-4**: Non-session records in `LoadTraceV2SessionBlueprints` also use `SendTimeUs`
- **BC-5**: Inter-round think-time gaps still derived from `ArrivalTimeUs` differences (unchanged)

## Test plan

- [ ] `TestLoadTraceV2Requests_ConcurrencyModeUseSendTime` — table-driven: non-zero SendTimeUs overrides, equal values, zero fallback
- [ ] `TestLoadTraceV2SessionBlueprints_ConcurrencyModeInjection` — BC-3 (session round-0), BC-4 (non-session), BC-5 (think-time gap unchanged)
- [ ] All existing tests pass (`go test ./... -count=1`)
- [ ] Lint clean (`golangci-lint run ./...`)

## Discovered Issues

[...]

PR #1611, opened 2026-08-24
fix(replay): normalize injection origin so real observe traces replay (#1606)

### LINKED ISSUE #1606: Bug: replay injects 0 completions on real observe traces (send_time_us epoch vs arrival_time_us relative origin mismatch)
**Describe the bug**

`blis replay` on a real `blis observe` trace injects **0 completions** — every request is
immediately marked "timed out" — because it uses the trace's `send_time_us` column as the
absolute DES injection tick, but `send_time_us` is written in **Unix epoch microseconds**
while `arrival_time_us` / `deadline_us` are written on a **run-relative origin** (t≈0 at
observe start). The two clocks differ by ~15 orders of magnitude, so each request is injected
at tick ≈1.79e15 while carrying a relative `deadline_us` ≈3.4e8 → deadline precedes injection
→ instant timeout. Replay reports success (exit 0) but writes an empty `sim_result.json` (`[]`),
which then makes `blis calibrate` fatal with "No sim results found ... cannot calibrate with
empty data".

This was introduced by #1304 (`fix(replay): use send_time_us as DES injection time`), whose
tests used toy values (`arrival=40000, send=180000`) where both columns share an origin and
differ only by a concurrency-slot wait. On a real observe trace they do **not** share an
origin, so `injectionTime()` returns an epoch value used as an absolute sim tick.

**Root cause (two origins, one bug + one latent bug)**

`cmd/observe.go` writes the columns on different clocks:
- `arrival_time_us` = `req.ArrivalTime` (workload generator, starts ~0) → **relative**
- `deadline_us`     = `arrival + timeout` → **relative**
- `send_time_us`    = `time.Now().UnixMicro()` (observe.go:228) → **epoch**
- `first_chunk_time_us` / `last_chunk_time_us` = `time.Now().UnixMicro()` → **epoch**

`sim/workload/replay.go:44 injectionTime()` returns `SendTimeUs` when `> 0` (always true for a
real HTTP trace), and this becomes `sim.Request.ArrivalTime` — an **absolute** injection tick.
An absolute value is only meaningful on the sim clock whose origin is 0; the epoch value is not.

Why `blis calibrate` is unaffected: it only takes **differences** on the epoch clock
(`realTTFT = FirstChunkTimeUs - SendTimeUs`, calibrate.go:255) and normalizes with
`firstSend = min(SendTimeUs)` (calibrate.go:215). Differences are origin-invariant. Replay is
the only consumer that uses `send_time_us` as an absolute tick.

**Second (latent) inconsistency:** the horizon is computed from a *different* column than the
injection. `cmd/replay.go computeHorizonFromMaxArrival(maxInjectedArrivalTimeUs(trace))` reads
`rec.ArrivalTimeUs` (relative, ~292s → horizon ~584s), while injection uses `SendTimeUs` (epoch).
So even the horizon and the injection disagree about which clock they're on. (`maxInjectedArrivalTimeUs`
reads the *arrival* column despite its name implying injection time.)

**CLI command to reproduce**

Minimal, self-contained. A 3-row trace with epoch `send_time_us` and relative
`arrival_time_us`/`deadline_us` (exactly what `blis observe` emits) reproduces it — no
KV-offload or special model needed.

`data.csv`:
```
request_id,client_id,tenant_id,slo_class,vllm_priority,session_id,round_index,prefix_group,prefix_length,streaming,input_tokens,output_tokens,text_tokens,image_tokens,audio_tokens,video_tokens,reason_ratio,model,deadline_us,server_input_tokens,arrival_time_us,send_time_us,first_chunk_time_us,last_chunk_time_us,num_chunks,status,error_message,finish_reason,x_request_id
50,c0,,standard,1,,0,,0,true,7400,600,0,0,0,0,0,,342838245,7455,42838245,1787274995712218,1787275030383086,1787275039779671,601,ok,,length,r0
51,c1,,standard,1,,0,,0,true,7400,600,0,0,0,0,0,,350814069,7481,50814069,1787275003688786,1787275030770092,1787275040083940,601,ok,,length,r1
52,c2,,standard,1,,0,,0,true,7400,600,0,0,0,0,0,,351339101,7351,51339101,1787275004213844,1787275036098173,1787275045443443,601,ok,,length,r2
```
`header.yaml`:
```yaml
trace_version: 3
time_unit: us
mode: real
warm_up_requests: 0
workload_seed: 42
server:
    type: vllm
    model: Qwen/Qwen3-14B
```
Run:
```bash
./blis replay \
  --trace-header header.yaml --trace-data data.csv \
  --defaults-filepath ./defaults.yaml --hardware-config ./hardware_config.json \
  --latency-model trained-physics --model Qwen/Qwen3-14B --tp 1 --hardware H100 \
  --max-num-seqs 128 --max-num-batched-tokens 2048 --block-size-in-tokens 16 \
  --gpu-memory-utilization 0.45 --max-model-len 12288 --total-kv-blocks 2621 \
  --model-config-folder ./model_configs/qwen3-14b \
  --results-path sim_result.json --log info
```

**Expected behavior**

Requests injected on the sim clock relative to the trace's earliest injection time; the sim
runs the workload and completes requests. Reference invariants:
- **Causality: arrival <= schedule <= completion** — violated: injection tick (epoch) is
  incomparable with the relative `deadline_us`, so causality against the deadline is broken.
- **Request conservation** — 3 injected, 0 completed / 0 queued / 3 timed_out: the sim does no
  work despite valid, servable requests.

**Actual behavior**

```
Loaded trace: 3 records (mode=real)
Built 3 requests for replay
Simulation horizon: 3574550008427688 ticks
[tick 1787275004229407] Simulation ended
[cluster] no requests completed — 3 of 3 requests timed out (client timeout exceeded, likely KV pressure)
SimResults written to sim_result.json (0 entries)     # <- empty []
```
Replay exits 0; `sim_result.json` is `[]`; downstream `blis calibrate` is fatal:
`No sim results found ... cannot calibrate with empty data`.

**Control (proves the origin is the sole cause)**

Take the identical trace, set `send_time_us := arrival_time_us` and shift
`first_chunk_time_us`/`last_chunk_time_us` by the same per-row offset (i.e. put `send`/chunks on
the relative origin). Replay then completes **3/3**, horizon = 102,678,202 ticks, sim ends at
tick 59,595,728. Only the injection origin changed. (Shown here only to isolate the cause — not
a proposed trace fix.)

- [x] Request conservation
- [x] Causality
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [ ] Work-conserving
- [ ] None / unknown

**Suggested fix (replay-side, preserves #1304 intent)**

Normalize the injection origin in `LoadTraceV2Requests` / `LoadTraceV2SessionBlueprints`:
compute `t0 = min(injectionTime(rec))` over all records and inject each request at
`injectionTime(rec) - t0`. This keeps #1304's semantics (send_time captures the concurrency-slot
wait, so intra-trace `send - arrival` gaps are preserved) while making injection origin-invariant.
Fix the horizon in the same change so it is computed from the same normalized injection values
(currently `maxInjectedArrivalTimeUs` reads `ArrivalTimeUs`, not the injection time). A regression
test should use a trace whose `send_time_us` is epoch-scale and assert `completed == injected`.

Alternatively, relativize `send_time_us` (and the chunk timestamps) at the observe writer, but
that is more invasive and must keep calibrate's `first_chunk - send` deltas intact.

**Environment**
- OS: macOS 15 (darwin/arm64)
- Go version: go1.24.4
- Commit hash: 63bd024f

**Additional context**

Discovered while running the BLIS Observe→Replay→Calibrate loop on a real Qwen3-14B multi-turn
observe trace (H100, multi-tier KV-offload experiment). Observe itself was clean (135 records,
0 timeouts, valid header); the failure is entirely in replay's trace ingestion. This affects
**any** real observe trace since #1304 (May 2026), because every such trace has epoch-scale
`send_time_us` from `time.Now().UnixMicro()`.

namasl:
## Summary

`blis replay` on a real `blis observe` trace injected **0 completions** — every request was instantly `timed_out`, `sim_result.json` was `[]`, and downstream `blis calibrate` was fatal. Root cause: `sim/workload/replay.go injectionTime()` returned the trace's `send_time_us` as the absolute DES injection tick, but `blis observe` writes `send_time_us` in **Unix-epoch µs** (`time.Now().UnixMicro()`) while `arrival_time_us`/`deadline_us` are on a **run-relative** origin (t≈0 at observe start). Injection landed at ~1.79e15 ticks carrying a relative deadline ~3.4e8, so `EnqueueRequest`'s past-due guard (`r.Deadline <= sim.Clock`, `sim/simulator.go`) fired at injection. Introduced by #1304, whose tests used toy values (`arrival=40000, send=180000`) that shared one origin.

## Fix

Re-base injection onto the `arrival_time_us`/`deadline_us` origin. Both loaders subtract a single per-trace constant `injectionOriginShift = min(injectionTime) − min(arrival_time_us)` at all three injection sites (`LoadTraceV2Requests` + `LoadTraceV2SessionBlueprints` round-0 and non-session):

- **Observe traces:** the shift ≈ the epoch offset, re-basing injection onto the relative clock → the deadline is far in the future, so requests are served (fixes the bug).
- **Generated `blis run` traces:** they write `send_time_us == arrival_time_us`, so `min(injectionTime) == min(arrival)` and the shift is exactly **0** ⇒ injection == arrival, **byte-identical** to pre-fix replay (INV-13/INV-6).

#1304's send-delta (concurrency-slot-wait) spacing is preserved — send deltas are origin-invariant, so subtracting one constant leaves spacing intact.

**Why not the issue's literal `injectionTime − min(injectionTime)` (earliest → tick 0)?** Throughput divides by the **absolute** `SimEndedTime` (`sim/metrics.go`: `vllmRuntime = SimEndedTime/1e6`), and generated traces start at `arrival ≈ 99999` (rate-dependent, **not** 0). That formula would shift generated-trace `SimEndedTime`/throughput while `blis run` keeps the absolute origin → **INV-13 break**. Anchoring at `min(arrival_time_us)` keeps generated-trace replay untouched **and** puts injection on the deadline clock.

**Second (latent) fix:** the closed-loop preliminary horizon now derives from the normalized injection (`workload.MaxNormalizedInjectionTimeUs`) instead of the raw `arrival_time_us` column, so the horizon and the injection are on one clock (the issue's noted secondary inconsistency; `maxInjectedArrivalTimeUs` removed).

**Hardening:** `LoadTraceV2` now rejects negative `arrival_time_us` (the injection-origin anchor; a negative value would yield a negative DES tick, INV-3), matching the existing loud checks on the other time/count fields (R1). `send_time_us` stays exempt — `injectionTime` deliberately falls back to `arrival_time_us` when `send_time_us <= 0`.

## Behavioral contracts

- **BC-1 / BC-4:** GIVEN a trace whose injection basis (`send_time_us`) is on a different origin than `arrival_time_us`, WHEN building requests, THEN `ArrivalTime = injectionTime(rec) − (min(injectionTime) − min(arrival_time_us))`, and injection spacing follows the **send** delta (not the arrival delta), earliest injection at `min(arrival_time_us)`.
- **BC-2 (fixes #1606):** GIVEN an epoch-scale `send_time_us` trace with relative `deadline_us > 0`, WHEN replayed, THEN `completed == injected`, `timed_out == 0` (INV-1, INV-5).
- **BC-3 (INV-13/INV-6):** GIVEN a `blis run` trace (`send == arrival`, `min(arrival) > 0`), WHEN replayed, THEN every injection == its `arrival_time_us` unchanged (shift == 0).
- **BC-6:** GIVEN a closed-loop epoch trace, WHEN sizing the preliminary blueprint horizon, THEN it derives from the normalized injection.
- **BC-7 (issue follow-up):** GIVEN an epoch trace with `deadline_us == 0`, WHEN replayed, THEN injection lands on the relative/arrival origin (sane TTFT/E2E clock), not the epoch clock.

## Testing

- **Issue's exact CLI reproducer:** before → `3 of 3 timed out`, `sim_result.json = []`; **after** → `completed_requests: 3`, `timed_out_requests: 0`, 3 entries, horizon 102,679,742 ticks (issue's control experiment: ~102,678,202).
- New unit tests (`sim/workload/replay_test.go`): epoch-origin normalization + send-delta spacing (BC-1/BC-4), generated-trace byte-identity (BC-3), `injectionOriginShift` all-fallback == 0 (BC-5), `MaxNormalizedInjectionTimeUs` (BC-6); retargeted the `#1304` table test to the raw `injectionTime` choice; updated `..._ConcurrencyModeInjection` to normalized values.
- New end-to-end cluster tests (`sim/cluster/replay_origin_e2e_test.go`): epoch-send + relative-deadline → `completed == injected` (BC-2); `deadline_us == 0` → relative-origin injection (BC-7).
- New validation tests (`sim/workload/tracev2_test.go`): negative `arrival_time_us` rejected; negative `send_time_us` tolerated.
- `go build ./...`, `go test ./...`, `golangci-lint run ./...` — all green.

Fixes #1606


namasl:

namasl:

## qa-review — PR #1611: ✅ PASS
_fix(replay): normalize injection origin so real observe traces replay (#1606) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer Claude Opus_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR implements all three things issue #1606 asks for: (1) origin normalization in BOTH LoadTraceV2Requests and LoadTraceV2SessionBlueprints via injectionOriginShift; (2) the horizon computed from the same normalized injection (MaxNormalizedInjectionTimeUs) rather than raw arrival; (3) a regression test asserting completed==injected on an epoch-scale-send trace. The one addition beyond the literal ask — rejecting arrival_time_us<0 in parseTraceRecord — is directly load-bearing for the fix's non-negativity guarantee (see G14), not gold-plating. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes, adequately. CLAUDE.md's observe/replay/calibrate section gains a paragraph describing #1606 injection-origin normalization. No other doc references send_time_us or injection-origin behavior, and docs/guide/observe-replay-calibrate.md does not describe the injection-timing internals, so there is no stale prose to update — the fix is transparent to existing traces. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale comments. The injectionTime doc comment was rewritten to 'raw DES injection basis' with an explicit #1606 caveat that callers MUST subtract injectionOriginShift; the '...only the initial injection point uses SendTimeUs' line became '...uses the injection basis'. The removed maxInjectedArrivalTimeUs has zero remaining references. Test comments referencing removed BC labels were updated. |
| G1 | correctness | ✅ CONFIDENT | `injectionOriginShift` computes `minInjection - minArrival` across all records in the trace. What happens if record A has the minimum `injectionTime` while record B has the minimum `arrivalTimeUs`? Is `injectionTime(rec) - originShift` mathematically guaranteed to be non-negative for every record `rec` in the trace? | Yes — non-negativity is guaranteed for every record, and it does NOT matter that the min-injection record and the min-arrival record may be different records. The mins are taken independently per column. For any record rec: injectionTime(rec) - shift = injectionTime(rec) - minInjection + minArrival. Since injectionTime(rec) >= minInjection by definition, this is >= minArrival, and minArrival >= 0 is enforced by parseTraceRecord's arrival<0 rejection. So the result is always >= minArrival >= 0. |
| G2 | correctness | ✅ CONFIDENT | `MaxNormalizedInjectionTimeUs` skips follow-up session rounds (`rec.SessionID != "" && rec.RoundIndex != 0`), but `injectionOriginShift` iterates over ALL records in `trace.Records`. If a follow-up session round record in a trace contains an unexpectedly small `SendTimeUs` or `ArrivalTimeUs`, could it distort `injectionOriginShift` and incorrectly shift the injection timing of initial round-0 requests? | For any well-formed trace, no. injectionOriginShift iterates over all records including follow-ups, but in a real observe trace follow-up rounds have LARGER send/arrival than round 0, so they never lower minInjection or minArrival and thus never affect the shift. Even if a follow-up carried the minimum (same clock), the shift is a single constant applied to every record, so it re-anchors the origin uniformly — round-0 spacing and non-negativity are preserved. Follow-ups are correctly EXCLUDED from the max in MaxNormalizedInjectionTimeUs but INCLUDED in the shift, consistent with the loader computing the shift the same way over all records. |
| G3 | invariants | ✅ CONFIDENT | Under INV-13 (run/replay parity), generated `blis run` traces have `send_time_us == arrival_time_us`, resulting in `originShift == 0`. However, if a trace has `send_time_us > arrival_time_us` for all records due to simulated queueing delays, does subtracting `originShift = min(send) - min(arrival)` shift the absolute DES injection ticks relative to simulation start tick 0 in a way that affects INV-5 (causality)? | No INV-5 impact. INV-5 governs the sim's internal event ordering on the sim clock; the injected ArrivalTime IS the arrival tick, and every downstream event is scheduled at >= that tick, so causality holds by construction regardless of the origin shift. When send>arrival for all records (concurrency wait), shift=min(send)-min(arrival) preserves the send-delta spacing while landing the earliest injection at min(arrival); it does not reorder anything. For a generated trace send==arrival so shift=0. |
| G4 | invariants | ✅ CONFIDENT | To satisfy INV-3 (clock monotonicity), DES injection ticks must remain non-negative. If a trace contains a record where `send_time_us` falls back to `arrivalTimeUs` (because `send_time_us <= 0`) while another record has a huge epoch-scale `send_time_us`, could `injectionTime(rec) - originShift` result in a negative `req.ArrivalTime`? | No — the result cannot be negative. General proof: injected = injectionTime - minInjection + minArrival >= minArrival >= 0. Concretely with rec X send<=0 (injectionTime=arrivalX) and rec Y epoch send: minInjection=arrivalX (relative << epoch); shift=arrivalX-minArrival>=0; injected(X)=minArrival>=0; injected(Y)=epoch-shift, huge-positive. INV-3 upheld. |
| G5 | rules | ✅ CONFIDENT | Rule R1 requires strict trace validation against silent fallback behavior. `parseTraceRecord` now loudly rejects `arrivalTimeUs < 0` with an explicit error, but tolerates `sendTimeUs < 0` by falling back to `arrivalTimeUs` in `injectionTime()`. Why is a negative `sendTimeUs` tolerated with silent fallback rather than rejected under R1? | Not an R1 violation. R1 targets error paths that silently DROP a request/metric/allocation. Nothing is dropped: the negative send_time_us is preserved verbatim, and injectionTime deterministically and documentedly falls back to arrival_time_us for send<=0. The asymmetry is intentional: arrival is the injection-origin ANCHOR (a negative anchor would yield a negative DES tick, violating INV-3), whereas send is never used as a tick when <=0, so a corrupt/absent send column must not fail loading of an otherwise-valid trace. Both behaviors are explicitly tested. |
| G6 | rules | ✅ CONFIDENT | Under rule R21 (canonical constructors and helpers), `MaxNormalizedInjectionTimeUs` is added as an exported function in `sim/workload/replay.go`. Does `MaxNormalizedInjectionTimeUs` handle `nil` or zero-record `TraceV2` pointers safely without panicking, and is that nil-safety contract explicitly tested? | Yes, nil-safe and tested. MaxNormalizedInjectionTimeUs returns 0 immediately for a nil or zero-record trace before dereferencing anything; injectionOriginShift likewise returns 0 for an empty slice. Both nil-safety contracts are explicitly asserted (TestMaxNormalizedInjectionTimeUs nil==0; TestInjectionOriginShift_AllFallback_IsZero nil==0). |
| G7 | parity | ✅ CONFIDENT | In `LoadTraceV2SessionBlueprints`, round-0 injection uses `injectionTime(r0) - originShift`, while think-times between rounds are calculated using `ArrivalTimeUs` deltas. If `r0` had a large concurrency slot wait in `send_time_us`, does mixing `send_time_us` for round 0 and `ArrivalTimeUs` for round 1 introduce a time-discontinuity in session step pacing? | No discontinuity. Round-1's arrival is not read as an absolute trace value — in closed-loop replay the SessionManager schedules the follow-up at round-0.CompletionTime + thinkTime, where thinkTime is a DURATION derived from ArrivalTimeUs deltas (origin-invariant). Round 0 using the send-based basis means it is injected at its true dispatch time (after any concurrency-slot wait); it completes later on the sim clock, and the think-time gap is added to THAT completion, so pacing stays continuous. The test pins think-time to the arrival gap (200000), not the send gap (180000). |
| G8 | parity | ✅ CONFIDENT | Real `blis observe` traces write `first_chunk_time_us` and `last_chunk_time_us` on the Unix epoch clock. When `blis replay` re-bases injection onto the relative `arrival_time_us` origin, how does `blis calibrate` ensure that real chunk timestamps from the trace remain on a compatible clock origin when comparing against simulated latencies? | calibrate is unaffected because it only takes DIFFERENCES on the trace's own epoch clock: realTTFT=FirstChunkTimeUs-SendTimeUs and realE2E=LastChunkTimeUs-SendTimeUs, all epoch, so origin-invariant. The replay-side origin shift touches only the sim's injection ticks; it never rewrites the trace's chunk/send columns. calibrate compares real DURATIONS against sim DURATIONS, both on compatible scales regardless of the injection re-basing. |
| G9 | tests | ✅ CONFIDENT | The new unit tests in `sim/workload/replay_test.go` cover all-epoch traces and all-generated traces. Is there a unit test covering a mixed trace where some records have `SendTimeUs > 0` and others have `SendTimeUs <= 0` (falling back to arrival time) to verify shift behavior under fallback? | No. Every new test uses a UNIFORM clock: all-epoch, all-generated (send==arrival), all-relative-send, or all-fallback (send<=0). None mixes send>0 (epoch) with send<=0 (fallback) records. |
| G10 | tests | ✅ CONFIDENT | Does the test suite include a law or invariant test asserting that for any valid `TraceV2` input, `req.ArrivalTime` is non-negative and preserves `send_time_us` intra-trace spacing across all generated `sim.Request` objects? | There is no single property/fuzz 'law' test over arbitrary valid TraceV2 inputs. Both properties ARE asserted via targeted example tests: non-negativity of req.ArrivalTime is checked in a loop over all reqs in TestReplay_EpochSendDeadlineZero_RelativeOrigin (also bounding max below epoch scale), and send-delta spacing preservation is checked in TestLoadTraceV2Requests_NormalizesEpochSendOrigin (delta==200000, the send delta, not the 50000 arrival delta). |
| G11 | rationale | ✅ CONFIDENT | Why was normalizing the injection origin on the replay loader side via `injectionOriginShift` chosen over changing `blis observe` to write run-relative `send_time_us` timestamps directly at trace capture time? | Three sound reasons, all consistent with the issue's own analysis. (1) Backward compatibility: the replay-side fix corrects EVERY real observe trace already written since #1304 with no re-capture; a writer change would only fix future captures. (2) Blast radius: the issue notes relativizing send_time_us at the writer is 'more invasive and must keep calibrate's first_chunk-send deltas intact'; the replay fix leaves calibrate untouched. (3) Locality: replay is the sole consumer that uses send as an absolute tick. |
| G12 | rationale | ✅ CONFIDENT | The PR summary explains that `min(injectionTime) - min(arrival_time_us)` was chosen instead of `injectionTime - min(injectionTime)`. How does anchoring the origin shift to `min(arrival_time_us)` specifically preserve `sim/metrics.go` throughput calculations (`vllmRuntime = SimEndedTime/1e6`) for generated traces? | By adding min(arrival) back, generated traces are left byte-identical, so SimEndedTime — and thus vllmRuntime and every derived throughput — is unchanged. For a generated trace send==arrival for every record, so minInjection==minArrival and shift=0, giving injected==arrival exactly. The issue's naive alternative (injectionTime - min(injectionTime)) would slide the earliest request to tick 0, shifting all ticks and changing SimEndedTime → breaking INV-13 byte-identity. Anchoring to min(arrival) preserves the arrival origin (and the runtime denominator). |
| G13 | scope | ✅ CONFIDENT | The PR removes `maxInjectedArrivalTimeUs` in `cmd/replay.go` and replaces it with `workload.MaxNormalizedInjectionTimeUs`. Is this preliminary horizon calculation change strictly required to fix issue #1606, or was it included as opportunistic cleanup? | Strictly required, not opportunistic. The issue's 'Second (latent) inconsistency' documents that the horizon was read from a different column (ArrivalTimeUs) than the injection, and the Suggested Fix explicitly says to compute it from the same normalized injection. Once injection is normalized, sizing the closed-loop preliminary horizon from raw arrival would put horizon and injection on different bases; MaxNormalizedInjectionTimeUs keeps both on the one normalized clock. The rename also fixes the old function's misleading name. |
| G14 | scope | ✅ CONFIDENT | Adding validation in `parseTraceRecord` to reject `arrivalTimeUs < 0` was not explicitly requested in issue #1606. What specific failure mode in replay or simulation justified adding this hardening check in this PR? | It is the precondition that makes the fix's non-negativity guarantee hold. After #1606, arrival_time_us is the injection-origin ANCHOR: injected = injectionTime - minInjection + minArrival, proven >=0 only because minArrival>=0. A negative arrival would yield a negative DES injection tick, violating INV-3. So rejecting arrival<0 up front (rather than silently producing a negative tick deep in the sim) is directly justified by the new role of the column, and mirrors the pre-existing negative-deadline_us rejection. |
| G15 | errors | ✅ CONFIDENT | When `parseTraceRecord` returns an error for negative `arrivalTimeUs`, how is this error propagated through `LoadTraceV2` to `cmd/replay.go`, and is it reported via `logrus.Fatalf` without leaking raw internal stack traces? | Clean propagation, no stack trace. parseTraceRecord returns fmt.Errorf('parsing arrival_time_us: negative value %d not allowed', v); LoadTraceV2 checks the error at the parse call and returns it up unchanged; cmd/replay.go reports via logrus.Fatalf('Failed to load trace: %v', err), which prints the formatted message and exits(1) without a Go stack trace. The user sees a single readable line. |
| G16 | errors | ✅ CONFIDENT | In `injectionOriginShift`, if `minInjection` is Unix-epoch scale (~1.79e15 µs) and `minArrival` is a negative or extreme value, is there any potential for `minInjection - minArrival` to overflow `int64` arithmetic, and how is integer arithmetic overflow guarded? | No overflow in the validated path. minInjection is either an epoch send (~1.79e15) or a relative arrival (>=0), and minArrival>=0 is enforced, so minInjection-minArrival lies in [0, ~1.79e15] — orders of magnitude below int64 max (~9.22e18). Downstream, computeHorizonFromMaxArrival has an explicit MaxInt64/2 guard for its 2x multiply. |

### Items to fix

### Important to consider
- **F2** — A one-line note in docs/guide/observe-replay-calibrate.md that real observe traces now replay correctly would help discoverability, but is not required.
  _diff CLAUDE.md (+line ~455); grep 'send_time_us\|injection\|origin' over docs/ returns only unrelated hits in observe-replay-calibrate.md:538,709_
- **F3** — computeHorizonFromMaxArrival's comment (cmd/replay.go:943-948) and its 'maxArrival' parameter name are now slightly imprecise — the caller passes a normalized injection value, not an arrival. Cosmetic only; the function is generic over an int64.
  _sim/workload/replay.go:28-57; grep 'maxInjectedArrivalTimeUs' returns nothing_
- **G1** — Non-negativity is preserved even in a mixed-clock trace, but correct re-basing is not — see G4/G9.
  _sim/workload/replay.go:83-98 (shift = minInjection-minArrival); replay.go:79-82; tracev2.go:600-602 (arrival>=0 enforced)_
- **G2** — The only way a follow-up could distort round-0 is if it carried send_time_us<=0 (fallback) while round-0 uses epoch send — the mixed-clock corner in G9. observe cannot produce that.
  _sim/workload/replay.go:113-118 (max skips follow-ups) vs replay.go:246 (loader shift over ALL records); replay.go:83-98_
- **G3** — A subtle fidelity edge: injecting at the send-based (post-wait) time against an arrival-based deadline could past-due a request whose concurrency wait exceeded its client timeout — but that request would also have timed out in observe, so it is defensible, not an invariant break.
  _sim/simulator.go:334,599; sim/workload/replay.go:77-81_
- **G4** — This exact scenario is where re-basing FAILS silently: rec Y (epoch) is injected at ~epoch scale (epoch - shift ≈ epoch), instant-timing-out — the #1606 bug — for the epoch-send records, while non-negativity still holds. See G9.
  _sim/workload/replay.go:83-98; sim/cluster/replay_origin_e2e_test.go:145-147 asserts r.ArrivalTime>=0 for all reqs_
- **G6** — The question cites R21 for 'canonical constructors and helpers', but R21 is actually 'No range over mutable slices'; canonical-constructor is R4. The substantive nil-safety answer stands regardless.
  _sim/workload/replay.go:110-112,84-86; sim/workload/replay_test.go_
- **G7** — The arrival-gap think time bundles prior-round service time (documented as lossy at replay.go:196-216); orthogonal to this PR.
  _sim/workload/replay.go:50-52,226-230; replay_test.go TestLoadTraceV2SessionBlueprints_ConcurrencyModeInjection_
- **G9** — The gap matters: a mixed trace mis-behaves. Records A{arrival=0,send=epoch}, B{arrival=50000,send=0}: shift=minInjection(50000)-minArrival(0)=50000, so A injects at epoch-50000 (epoch scale) → instant timeout — the #1606 bug, unfixed for A. Mitigating: I could not find a way for blis observe to actually emit such a trace (the only send=0 paths — JSON marshal / http.NewRequest error at cmd/observe.go:203-215 — are URL/body-level and fail identically for ALL records, yielding all-fallback which IS handled), so this is a hand-crafted/converter-trace corner, not a broken default path. Worth either a mixed-input test or an explicit 'uniform-clock' precondition guard.
  _sim/workload/replay_test.go and sim/cluster/replay_origin_e2e_test.go (all listed tests); no mixed case present_
- **G10** — Per CLAUDE.md 'test laws, not just values', a property-based test drawing random valid traces and asserting (a) all ArrivalTime>=0 and (b) injected deltas == send deltas would be a stronger companion invariant and would also surface the mixed-clock gap in G9.
  _sim/cluster/replay_origin_e2e_test.go:145-155; sim/workload/replay_test.go TestLoadTraceV2Requests_NormalizesEpochSendOrigin_
- **G12** — The PR's choice is strictly better than the issue's literal suggestion for INV-13 — worth calling out as a deliberate improvement.
  _sim/metrics.go:95,152-153; sim/workload/replay_test.go TestLoadTraceV2Requests_GeneratedTraceInjectionUnchanged ('shift must be 0')_
- **G16** — The subtraction itself has no explicit overflow guard; it is safe purely because the input domain is bounded by the arrival>=0 validation. A caller constructing a TraceV2 programmatically with a large-magnitude negative arrival (bypassing parseTraceRecord) could in principle overflow — a non-issue for the CLI, worth a one-line assumption comment.
  _sim/workload/tracev2.go:600-602; sim/workload/replay.go:83-98; cmd/replay.go:949-958_

namasl:
Thanks for the reviews. blis-pr-review returned **READY TO MERGE** and the cross-vendor qa-review **PASS** with no blocking findings. I pushed a follow-up (`eb8b1568`, test + comment only, no behavioral change) addressing the substantive non-blocking items:

- **G9 (mixed-clock corner):** Added two tests. `..._MixedSendSign_RelativeClock_ShiftZero` pins the *only* mixed-`send`-sign case a real producer emits — a generated trace whose first request arrives at `t=0` (so `send_time_us==0` → fallback) alongside later `send>0` records, all on one relative clock ⇒ `shift==0`, injection `==` arrival. That case is exactly why a "uniform send-sign" guard would be **wrong**: it would false-positive on legitimate `blis run` traces. `..._MixedEpochAndFallback_StaysNonNegative` documents the uniform-clock precondition via the INV-3 law — the pathological epoch+fallback mix is unreachable from observe/run/converters (all emit one uniform clock), and even there every injection stays `>= 0`.
- **F3 / G16:** Clarified `computeHorizonFromMaxArrival`'s doc (the arg is a normalized injection tick, not a raw arrival column) and noted the `int64` overflow bound on `injectionOriginShift` (both mins are bounded by the new `arrival_time_us >= 0` validation).

Non-code items acknowledged, no change needed: **G3** (a request whose concurrency wait exceeds its client timeout past-dues in replay — but it also timed out in observe, so it's faithful, not an invariant break); **G7** (arrival-gap think-time lossiness is pre-existing and orthogonal); **G12** (the `min(arrival)`-anchored shift is a deliberate improvement over the issue's literal formula, preserving INV-13 throughput byte-identity — called out in the PR body); **ARCHON `ARCHITECTURAL_CHANGE`** is the expected categorization for the one new exported helper (`MaxNormalizedInjectionTimeUs`) that replaced the deleted `cmd`-local `maxInjectedArrivalTimeUs`.

namasl:

susiejojo:
### BLIS PR Self-Review — PR #1611 (independent verification pass)

## Verdict: READY TO MERGE ✅

I independently reproduced the issue's experiment end-to-end, verified INV-13 byte-identity against the pre-fix binary, and traced the one substantive open concern (G9 mixed-clock corner). The fix is correct, minimal, and well-tested. This corroborates the prior `blis-pr-review` (READY TO MERGE), cross-vendor `qa-review` (PASS), and ARCHON (`ARCHITECTURAL_CHANGE`, expected) results.

---

### 1. Reproduced the issue's exact experiment

Ran the issue's literal CLI reproducer (3-row epoch-`send` / relative-`arrival` trace) on **both** the base commit `b5934c13` and this PR:

| | Base `b5934c13` (buggy) | This PR (fixed) | Issue's control |
|---|---|---|---|
| Simulation horizon | 3,574,550,008,427,688 ticks | **102,679,742 ticks** | ~102,678,202 |
| `completed_requests` | 0 | **3** | 3 |
| `timed_out_requests` | 3 | **0** | 0 |
| `sim_result.json` entries | 0 (`[]`) | **3** | 3 |

The bug is real (base binary reproduces `0 of 3 completed`, empty results) and this PR fixes it precisely — the horizon matches the issue's control experiment to within rounding.

### 2. INV-13 (run/replay parity) verified empirically, not just by unit test

Generated a `blis run --rate 10` trace, then replayed it with **both** the base binary and the PR binary. The two replay stdouts are **byte-identical** (`diff` clean). This confirms `send_time_us == arrival_time_us` for generated traces (verified: `arrival=99999 send=99999`) ⇒ `injectionOriginShift == 0` ⇒ no behavioral change on the `blis run` path. The unit test `TestLoadTraceV2Requests_GeneratedTraceInjectionUnchanged` is corroborated by this real cross-binary check.

### 3. G9 (mixed-clock corner) — confirmed genuinely unreachable

Traced the observe write path: `record.SendTimeUs = time.Now().UnixMicro()` is stamped at `cmd/observe.go:250`, **before** `httpClient.Do`. So a per-request HTTP timeout/error still carries an epoch send — it does **not** become a `send=0` fallback record. The only `send=0` paths are the pre-dispatch `json.Marshal` / `http.NewRequestWithContext` failures (`observe.go:203-215`), which are body/URL-level and fail identically for **every** record ⇒ an all-fallback trace (shift=0, handled). A trace mixing epoch-send with `send=0` fallback records cannot be emitted by observe/run/converters. The PR's choice to document this precondition (`TestLoadTraceV2Requests_MixedEpochAndFallback_StaysNonNegative`) rather than add a "uniform-clock" guard is correct — such a guard would false-positive on legitimate `blis run` traces whose first request arrives at `t=0` (`send=0` on a relative clock). INV-3 (`injection ≥ min(arrival) ≥ 0`) holds even in the pathological corner.

### 4. Cross-path parity (run / replay / observe)

- **`blis run`**: unaffected — never reads `send_time_us` from a file; INV-13 byte-identity confirmed (§2).
- **`blis replay`** (fixed + closed-loop): fixed at the loader (`LoadTraceV2Requests` + `LoadTraceV2SessionBlueprints` round-0 and non-session), all three injection sites use one `originShift`. Fixed-mode horizon (`computeReplayHorizon`) reads the already-normalized `req.ArrivalTime`; closed-loop uses `MaxNormalizedInjectionTimeUs` — both on the same (arrival) clock. Consistent.
- **`blis observe`** (incl. corpus-mode via `LoadTraceV2SessionBlueprints`): OTel/generated corpora are `send==arrival` ⇒ shift=0, no change. Write-side intentionally untouched (fix is in the reader).
- **`blis calibrate`**: unaffected — uses only epoch-clock **differences** (`FirstChunkTimeUs - SendTimeUs`), which are origin-invariant.

No parity gap. The fix normalizes data already being read; it adds no new behavioral dimension requiring propagation.

### 5. Timeout/deadline consistency

`DeadlineUs` is deliberately **not** shifted (it is already run-relative: `arrival + timeout`). After normalization `ArrivalTime = arrival_time_us`, so `Deadline > ArrivalTime` for any positive timeout ⇒ the `r.Deadline <= sim.Clock` guard (`sim/simulator.go`) is false at injection. BC-7 (`deadline_us == 0`) skips the guard entirely and still re-bases injection for clock-correct TTFT/E2E. Both covered by the e2e cluster tests.

### 6. Preemption safety

`ArrivalTime` is read-only after construction (injection scheduling + horizon only). No changes to per-request metrics, state transitions, or aggregate counters. No preemption concern.

### 7. Tests, build, standards

- `go build ./...`, `go test ./sim/workload/... ./sim/cluster/...` — green.
- All 15 new/retargeted tests pass individually (verified `-v`).
- `go vet ./sim/workload/... ./cmd/...` — clean. (golangci-lint not available in my environment; CI covers it.)
- Tests are behavioral (survive a rewrite): the strongest are the two e2e cluster tests replicating the exact symptom, and `NormalizesEpochSendOrigin` asserting send-delta spacing (200000) is preserved vs the arrival delta (50000).
- R1 (loud validation), R4 (single `originShift` per call). CLAUDE.md updated accurately.

---

### Non-blocking (already addressed by follow-up `eb8b1568`)

Nothing blocking. The minor doc-comment imprecision (F3/G16) and mixed-clock documentation (G9) were addressed in the follow-up commit. The double-assertion in the think-time test is intentional (documents the two-way law). No action required.

namasl:
Thanks for the thorough independent verification, @susiejojo — appreciate the cross-binary INV-13 byte-identity check and the G9 write-path trace confirming the mixed-clock corner is unreachable. Nothing outstanding on my end: all CI is green and every review pass (blis-pr-review, qa-review, ARCHON, and yours) is READY TO MERGE / PASS. The PR is `MERGEABLE` but still shows `REVIEW_REQUIRED`, so it's ready for a formal approval + merge whenever you are. I'll address anything further if it comes up.

susiejojo:
### Verified on real campaign traces (exp 4–6)

I re-ran each experiment's exact replay command against a real observe campaign (Qwen3-14B + Nemotron-30B, H100, multi-tier KV-offload), comparing the campaign's original (buggy) replay to this PR:

| Exp | Trace | Before (buggy) | After (this PR) |
|---|---|---|---|
| **5** | Nemotron-30B, 8091 reqs, multitier | **0 / 8091** completed → empty `sim_result.json` → `calibrate` **fatal** | **8046 / 8091** completed |
| **6** | Qwen3-14B agentic, 135 reqs, multitier | **0 / 135** completed → empty `sim_result.json` → `calibrate` **fatal** | **135 / 135** completed |
| **4** | Nemotron-30B, 150 reqs, `deadline=0` | 150/150 "completed" but injected at **epoch-scale** ticks (wrong clock) | 150/150, now on the **relative** origin (horizon 189M vs ~1.79e15 ticks) |

**What the fix did:** exp5/exp6 have epoch-scale `send_time_us` (~1.79e15) but run-relative `deadline_us` (~3.4e8), so every request injected *after* its deadline and instantly timed out — 0 completions, empty results, and a **fatal `blis calibrate`** (confirmed in the campaign's own logs: `cannot calibrate with empty data`). Re-basing injection onto the arrival/deadline origin fixes this: the full **observe → replay → calibrate** loop now runs end-to-end on all three.

Two notes (both *out of scope* for #1606, not fix problems):
- Exp5's 45 remaining timeouts are **real** saturation, not the clock bug — the fix correctly lets servable requests through while still timing out genuinely-overrunning ones.
- Calibration *accuracy* on these agentic/multitier workloads is poor (TTFT MAPE 50–76%). That's a model-fidelity question; this PR just restores the pipeline so it can be measured at all.

susiejojo (approved):
lgtm

namasl:
@claude /blis-pr-review final confirmation pass — no code changes since your last READY TO MERGE (the follow-up `eb8b1568` was test + comment only, and the PR is now formally approved). Please re-confirm so the commit status settles green.

PR #1612, opened 2026-08-24
fix(workload): non-lossy think_time_us encoding — distinguish recorded-0 from absent (#1608)

### LINKED ISSUE #1608: Non-lossy think_time_us encoding: distinguish a recorded-0 think time from an absent column
## Problem

`TraceRecord.ThinkTimeUs` (the `think_time_us` column, #1478) uses `0` as a **lossy sentinel** meaning "not recorded". A session whose every round records a genuinely-zero think time is therefore indistinguishable from one with no recorded think at all, and `LoadTraceV2SessionBlueprints` falls back to **arrival-gap derivation** for it (`sessionHasRecordedThinkTime` returns false when all rounds are 0). Arrival-gap think bundles service time into the gap — the very effect `think_time_us` exists to avoid.

This is real for the Weka converter (`blis convert weka`, #1604): overlapping main turns (round `i` arrives before round `i-1`'s response elapsed) legitimately recompute `ThinkTimeUs == 0` (clamped, INV-3). On the SemiAnalysis `051926` dataset ~2% of gaps clamp to 0; a session where *every* main turn overlaps would (incorrectly) fall back to arrival gaps.

## Why #1604 did not fix this

PR-F (#1604) was scoped to the Weka **reader** only; it emits the lossy `0` like the OTel converter and inherits this limitation from PR-A (#1478). In practice the fallback does not trigger for a real ≥20-main-turn Weka session (some round always has a positive gap), so the impact is bounded — but the encoding is still lossy, and the two comments below previously pointed to #1604 as the home of the fix.

## Proposed fix direction

Distinguish "recorded `0`" from "absent":
- a `-1` (or other reserved) sentinel for "not recorded", with the CSV column omitted only when *no* round recorded think; **or**
- a separate presence signal (bitmap / boolean column), so `sessionHasRecordedThinkTime` can test presence rather than non-zero.

`ExportTraceV2` / `LoadTraceV2` / `sessionHasRecordedThinkTime` and the converters would move to the presence test. Pinned regression: `TestLoadTraceV2SessionBlueprints_AllZeroRecordedThink_FallsBackToGap` documents current behavior and would be updated.

## Affected code

- `sim/workload/tracev2.go` (`ThinkTimeUs` doc + the `includeThinkTime` export gate)
- `sim/workload/replay.go` (`sessionHasRecordedThinkTime`)
- converters `sim/workload/weka_convert.go`, `sim/workload/otel_convert.go`

Discovered during #1605 (PR-F) review (qa-review F3): the two comments attributing this deferred work to "#1604" are now stale and are being retargeted to this issue.

namasl:
## Summary

`TraceRecord.ThinkTimeUs` (the `think_time_us` column, #1478) used `0` as a **lossy
sentinel** meaning "not recorded". A closed-loop session whose every round genuinely
recorded `think == 0` — real for the Weka converter's overlap-clamped turns (`blis
convert weka`, #1604/#1605) — was therefore indistinguishable from a session with no
recorded think at all, and `LoadTraceV2SessionBlueprints` fell back to **arrival-gap
derivation** for it. Arrival-gap think bundles service time into the gap — the very
effect `think_time_us` exists to avoid.

This PR makes the encoding **non-lossy** by turning `TraceRecord.ThinkTimeUs` and the
reader→encoder `NormalizedRound.ThinkUs` into `*int64` **presence signals**:

- `nil` = not recorded → CSV cell is empty; replay falls back to arrival gaps.
- non-nil (**including `&0`**) = recorded → CSV cell is the integer (`"0"` for a
  recorded zero); replay prefers it.

The CSV parser *already* distinguished an empty cell from `"0"` (it only parsed a
non-empty cell); the gaps were that the in-memory `int64` could not represent "absent"
and the writer always formatted the integer. `*int64` closes both, and — crucially —
makes the struct's **zero value (`nil`) mean the correct default** ("not recorded"), so
every existing `TraceRecord{…}` literal that omits think stays correct with no edit.

### Design choice (issue offered `-1` sentinel *or* a presence signal)

Chose the presence signal, realized as `*int64`, over a `-1` sentinel because:
- With `-1`, a fresh `TraceRecord{}` defaults to `0` = "recorded zero" — the wrong
  default and an R4 construction-site footgun. With `*int64`, `nil` = "not recorded" is
  the correct default.
- `-1` would need a carve-out in the parser, which currently rejects **all** negative
  think times (INV-3). `*int64` keeps that intact.
- Pointer-for-optional-where-zero-is-valid is the same principle rules.md R9 applies to
  YAML config fields.

**No trace-version bump** — this refines the *cell encoding* of an existing optional CSV
column, not the header schema. It degrades gracefully both directions: old code reading a
new trace reads an empty cell as its old `0` sentinel; new code reading an old trace reads
old `"0"` cells as recorded — which is the corrected behavior. (Consistent with precedent:
the trailing `adapter`/`think_time_us`/`x_request_id` columns were all added without a bump.)

> **Intentional behavior change:** replaying a pre-existing Weka trace whose main turns
> all overlapped (all `"0"`) now uses the recorded zeros (back-to-back rounds) instead of
> degrading to arrival gaps. This is the fix. `blis run` and `convert otel` outputs are
> unchanged (they never set think → nil → no column, INV-6/INV-13).

## Behavioral Contracts

- **BC-1** GIVEN a session whose every non-round-0 round records think == 0 (non-nil `&0`)
  WHEN loaded THEN closed-loop replay uses the recorded zeros, NOT arrival gaps.
- **BC-2 / BC-7** GIVEN a session whose non-round-0 rounds are all `nil` (OTel, generated)
  WHEN loaded THEN replay derives think from arrival gaps (byte-identical to pre-#1608).
- **BC-3** GIVEN records mixing `nil` / `&0` / `&N` WHEN exported then re-loaded THEN each
  round's presence and value round-trip exactly (`nil`↔empty cell, `&0`↔`"0"`, `&N`↔`"N"`).
- **BC-4** GIVEN every record's think is `nil` WHEN exported THEN no `think_time_us` column
  is written (`blis run` / `convert otel` byte-identical, INV-6).
- **BC-5** GIVEN at least one record with non-nil think (incl. `&0`) THEN the column is present.
- **BC-6** GIVEN a Weka session WHEN converted THEN round 0's think is `nil` and an
  overlapping turn is a recorded `&0`.
- **BC-8** A `think_time_us` cell with a negative value is still a hard load error (INV-3).

## Testing

`go build ./...` clean · `go test ./...` all pass (13/13 packages) · `golangci-lint run ./...` 0 issues.

New / updated tests (`sim/workload`, `cmd`):
- `TestLoadTraceV2SessionBlueprints_AllRecordedZeroThink_UsesRecordedZeros` (BC-1; replaces
  the old `…_AllZeroRecordedThink_FallsBackToGap`, which pinned the lossy-sentinel behavior).
- `TestLoadTraceV2SessionBlueprints_AbsentThink_FallsBackToGap` (BC-2, the surviving fallback).
- `TestTraceV2_ThinkTimeUs_RoundTrip` extended: `mixed presence: nil / &0 / &N round-trip
  distinctly` (BC-3) and `lone recorded &0 emits column` (BC-5).
- `TestLoadTraceV2SessionBlueprints_MixedThinkSessions` updated for per-session selection.
- `TestConvertWekaSession_NegativeGapThinkClampsToZero` now asserts the overlap is a recorded
  `&0` (non-nil), not absent (BC-6).

## Affected code

`sim/workload/tracev2.go` (`ThinkTimeUs *int64` + export gate/writer/parser),
`sim/workload/replay.go` (`sessionHasRecordedThinkTime` presence test + sampler deref),
`sim/workload/trace_encode.go` (`NormalizedRound.ThinkUs *int64`),
`sim/workload/weka_convert.go`, `sim/workload/otel_convert.go` (doc), plus `CLAUDE.md`
and `docs/guide/observe-replay-calibrate.md`.

Closes #1608

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,8 +234,10 @@ go build -o blis main.go
 # traffic (+312% on this dataset). Replayed input length / KV pressure / hit-rate is therefore
 # a substantial UPPER BOUND — do NOT read it as a faithful reproduction of the recorded ISL.
 # (Property of the PR-A/PR-B accumulate delta law, not this converter; the conversion is exact
-# per that law. Faithful compaction support is tracked in #1609; the separate think-time
-# lossy-0 sentinel is #1608.)
+# per that law. Faithful compaction support is tracked in #1609.) The recorded think time is
+# non-lossy (#1608): a genuinely-zero recorded think (an overlapping turn) is a &0 in the
+# think_time_us column, distinct from a not-recorded (empty) cell, so an all-overlap session
+# uses the recorded zeros rather than degrading to arrival-gap think at replay.
 ./blis convert weka --input traces.jsonl --trace-output corpus \
   --context-growth accumulate --max-think-time 0
 #
--- a/cmd/convert_weka_test.go
+++ b/cmd/convert_weka_test.go
@@ -51,9 +51,10 @@ func TestConvertWeka_EndToEndAndAccumulateReconstruction(t *testing.T) {
 			t.Errorf("record %d Model = %q, want empty (routing safety)", i, r.Model)
 		}
 	}
-	// think_time_us column round-trips: recorded 7s / 10s on rounds 1, 2.
-	if trace.Records[1].ThinkTimeUs != 7_000_000 || trace.Records[2].ThinkTimeUs != 10_000_000 {
-		t.Errorf("think_time_us = [_,%d,%d], want [_,7e6,10e6]", trace.Records[1].ThinkTimeUs, trace.Records[2].ThinkTimeUs)
+	// think_time_us column round-trips: recorded 7s / 10s on rounds 1, 2 (non-nil, #1608).
+	if trace.Records[1].ThinkTimeUs == nil || *trace.Records[1].ThinkTimeUs != 7_000_000 ||
+		trace.Records[2].ThinkTimeUs == nil || *trace.Records[2].ThinkTimeUs != 10_000_000 {
+		t.Errorf("think_time_us = [_,%v,%v], want [_,&7e6,&10e6]", trace.Records[1].ThinkTimeUs, trace.Records[2].ThinkTimeUs)
 	}
 
 	// Closed-loop reconstruction through the SessionManager (accumulate buffer).
--- a/cmd/observe_corpus_test.go
+++ b/cmd/observe_corpus_test.go
@@ -289,8 +289,8 @@ func TestObserveCorpusMode_MultiRoundAccumulate(t *testing.T) {
 	header := &workload.TraceHeader{Version: 3, TimeUnit: "microseconds", Mode: "generated", SessionContextGrowth: "accumulate"}
 	records := []workload.TraceRecord{
 		{RequestID: 0, SessionID: "s0", RoundIndex: 0, InputTokens: 100, OutputTokens: 10, ArrivalTimeUs: 0, Status: "ok"},
-		{RequestID: 1, SessionID: "s0", RoundIndex: 1, InputTokens: 40, OutputTokens: 10, ArrivalTimeUs: 1000, ThinkTimeUs: 1000, Status: "ok"},
-		{RequestID: 2, SessionID: "s0", RoundIndex: 2, InputTokens: 25, OutputTokens: 10, ArrivalTimeUs: 2000, ThinkTimeUs: 1000, Status: "ok"},
+		{RequestID: 1, SessionID: "s0", RoundIndex: 1, InputTokens: 40, OutputTokens: 10, ArrivalTimeUs: 1000, ThinkTimeUs: i64p(1000), Status: "ok"},
+		{RequestID: 2, SessionID: "s0", RoundIndex: 2, InputTokens: 25, OutputTokens: 10, ArrivalTimeUs: 2000, ThinkTimeUs: i64p(1000), Status: "ok"},
 	}
 	if err := workload.ExportTraceV2(header, records, headerPath, dataPath); err != nil {
 		t.Fatalf("export corpus: %v", err)
--- a/docs/guide/observe-replay-calibrate.md
+++ b/docs/guide/observe-replay-calibrate.md
@@ -395,8 +395,11 @@ The reader filters each session's `requests[]` to the **linear main-agent stream
 absorbed into the following main turn's think gap. Per-round **pure client think** is
 recomputed as `max(0, t_i − t_{i-1} − api_time_{i-1})` between consecutive main turns
 (carried in the `think_time_us` column, `--max-think-time` default `0` = uncapped,
-since Weka gaps are genuine away-from-keyboard times). The recorded `claude-*` model
-names are dropped during conversion (same routing-safety reason as OTel).
+since Weka gaps are genuine away-from-keyboard times). The column is **non-lossy**
+(#1608): a genuinely-zero recomputed think (an overlapping turn) is recorded as `&0`,
+distinct from a not-recorded (empty) cell, so an all-overlap session replays with the
+recorded zeros rather than degrading to arrival-gap think. The recorded `claude-*`
+model names are dropped during conversion (same routing-safety reason as OTel).
 
 !!! warning "Weka replay ISL is an upper bound (context compaction)"
     Weka input token counts are very large (p50 ≈ 110K, p90 ≈ 395K), so the
@@ -408,8 +411,8 @@ names are dropped during conversion (same routing-safety reason as OTel).
     total by ≈3–4×** (+312% on that dataset). Treat replayed input length, KV pressure,
     and hit-rate as a substantial **upper bound**, not a faithful reproduction of the
     recorded workload. (This is a property of the accumulate delta law, not the
-    converter; faithful compaction support is tracked in #1609, and the separate
-    think-time lossy-0 sentinel corner in #1608.)
+    converter; faithful compaction support is tracked in #1609. The separate
+    think-time lossy-0 sentinel was resolved in #1608 — see the non-lossy note above.)
 
 ---
 
--- a/sim/workload/otel_convert.go
+++ b/sim/workload/otel_convert.go
@@ -163,9 +163,9 @@ func ConvertOTelTrace(raw []byte, opts OTelConvertOptions) ([]TraceRecord, error
 	// (EncodeSessionToTraceRecords) computes per-round input deltas and leaves
 	// TraceRecord.Model empty (routing safety — a recorded cross-model name would
 	// drop every request at routing under a differing --model; see the encoder's
-	// doc comment and #1477). ThinkUs is left 0: OTel has no reliable
-	// response-complete time (end_time is often a placeholder), so per-round think
-	// is derived from arrival gaps at replay rather than recorded here.
+	// doc comment and #1477). ThinkUs is left nil (not recorded, #1608): OTel has no
+	// reliable response-complete time (end_time is often a placeholder), so per-round
+	// think is derived from arrival gaps at replay rather than recorded here.
 	t0 := spans[0].startUs
 	rounds := make([]NormalizedRound, 0, len(spans))
 	for i, sp := range spans {
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -249,8 +249,10 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 			for i := 1; i < len(rounds); i++ {
 				// derefInt64(nil) == 0: a recorded &0 and (defensively) a nil interior
 				// round both yield 0 think. sessionHasRecordedThinkTime guarantees at
-				// least one round here is non-nil; in practice a recorded session's
-				// rounds 1..N are all non-nil.
+				// least one round here is non-nil. No shipped converter produces a mixed
+				// interior (OTel leaves every round nil; Weka sets every round 1..N), so
+				// the nil-interior branch is unreachable today — the deref pins a defined
+				// behavior (0 think) if a future converter ever recorded think sparsely.
 				t := derefInt64(rounds[i].ThinkTimeUs)
 				if t < 0 {
 					t = 0 // defensive; LoadTraceV2 already rejects negatives (INV-3)
--- a/sim/workload/replay.go
+++ b/sim/workload/replay.go
@@ -119,17 +119,17 @@ func LoadTraceV2Requests(trace *TraceV2, seed int64) ([]*sim.Request, error) {
 // time is meaningless and ignored. When true, closed-loop replay prefers the recorded
 // per-round think time over arrival-gap derivation.
 //
-// LOSSY SENTINEL (F1, #1484 review): a value of 0 means "not recorded", so a session
-// whose every round has a genuinely-zero recorded think time (real for Weka's
-// overlap-clamped rounds) reads as false here and falls back to the arrival-gap path
-// — which bundles service time into the gap (see the gap-derivation note below), the
-// very effect think_time_us exists to avoid. The impact is bounded (a ~0 recorded
-// think replaced by the recorded arrival gap); a non-lossy encoding is deferred to
-// #1608 (the Weka converter #1604 intentionally kept this lossy 0-sentinel). Pinned by
-// TestLoadTraceV2SessionBlueprints_AllZeroRecordedThink_FallsBackToGap.
+// PRESENCE, not non-zero (#1608): ThinkTimeUs is now a *int64, so a recorded &0 is
+// distinguishable from an absent (nil) value. A session whose every round genuinely
+// recorded think == 0 (real for Weka's overlap-clamped rounds) now reads as recorded
+// here and uses those zeros (back-to-back rounds) — it no longer falls back to the
+// arrival-gap path, which bundles service time into the gap (see the gap-derivation
+// note below), the very effect think_time_us exists to avoid. Only a truly absent
+// (nil) value falls back. Pinned by
+// TestLoadTraceV2SessionBlueprints_AllRecordedZeroThink_UsesRecordedZeros.
 func sessionHasRecordedThinkTime(rounds []TraceRecord) bool {
 	for i := 1; i < len(rounds); i++ {
-		if rounds[i].ThinkTimeUs != 0 {
+		if rounds[i].ThinkTimeUs != nil {
 			return true
 		}
 	}
@@ -247,7 +247,11 @@ func LoadTraceV2SessionBlueprints(trace *TraceV2, seed int64, thinkTimeSampler L
 			// i-1's end). Round 0 carries no think; rounds 1..N supply the sequence.
 			thinkTimes := make([]int, len(rounds)-1)
 			for i := 1; i < len(rounds); i++ {
-				t := rounds[i].ThinkTimeUs
+				// derefInt64(nil) == 0: a recorded &0 and (defensively) a nil interior
+				// round both yield 0 think. sessionHasRecordedThinkTime guarantees at
+				// least one round here is non-nil; in practice a recorded session's
+				// rounds 1..N are all non-nil.
+				t := derefInt64(rounds[i].ThinkTimeUs)
 				if t < 0 {
 					t = 0 // defensive; LoadTraceV2 already rejects negatives (INV-3)
 				}
--- a/sim/workload/replay_test.go
+++ b/sim/workload/replay_test.go
@@ -322,9 +322,9 @@ func TestLoadTraceV2SessionBlueprints_PrefersRecordedThinkTime(t *testing.T) {
 	// gap derivation which bundles service time).
 	trace := &TraceV2{
 		Records: []TraceRecord{
-			{RequestID: 1, SessionID: "A", RoundIndex: 0, InputTokens: 100, OutputTokens: 50, ArrivalTimeUs: 0, ThinkTimeUs: 0},
-			{RequestID: 2, SessionID: "A", RoundIndex: 1, InputTokens: 200, OutputTokens: 80, ArrivalTimeUs: 5000, ThinkTimeUs: 300},
-			{RequestID: 3, SessionID: "A", RoundIndex: 2, InputTokens: 300, OutputTokens: 90, ArrivalTimeUs: 12000, ThinkTimeUs: 400},
+			{RequestID: 1, SessionID: "A", RoundIndex: 0, InputTokens: 100, OutputTokens: 50, ArrivalTimeUs: 0, ThinkTimeUs: nil},
+			{RequestID: 2, SessionID: "A", RoundIndex: 1, InputTokens: 200, OutputTokens: 80, ArrivalTimeUs: 5000, ThinkTimeUs: i64p(300)},
+			{RequestID: 3, SessionID: "A", RoundIndex: 2, InputTokens: 300, OutputTokens: 90, ArrivalTimeUs: 12000, ThinkTimeUs: i64p(400)},
 		},
 	}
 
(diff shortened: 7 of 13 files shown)


namasl:

namasl:

## qa-review — PR #1612: ✅ PASS
_fix(workload): non-lossy think_time_us encoding — distinguish recorded-0 from absent (#1608) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer Claude Opus_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR fully implements #1608: TraceRecord.ThinkTimeUs and NormalizedRound.ThinkUs become *int64 (nil=absent, &0=recorded zero), sessionHasRecordedThinkTime switches from `!= 0` to `!= nil` (presence test), the ExportTraceV2 gate and parseTraceRecord adopt the same presence semantics, both converters (Weka/OTel) are updated, the pinned regression test is renamed/updated as the issue directed, and the stale #1604-targeted comments are retargeted. It matches the issue's Affected-code list plus the necessary trace_encode.go change (the encoder maps ThinkUs->ThinkTimeUs). No unrelated additions. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes, for the affected behavior. The only docs that describe the TraceRecord think_time_us CSV-column semantics — CLAUDE.md (Weka convert example) and docs/guide/observe-replay-calibrate.md (Weka reader section) — are both updated to the non-lossy note. The other think_time_us references in docs (workloads.md, workload-spec.md, configuration.md, invariants.md INV-10) describe the workload-spec multi_turn.think_time_us config field / SessionBlueprint.ThinkTimeUs, which remain int64 and are correctly untouched. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale comments remain. The previously-stale comments that attributed the fix's home to #1604 (in CLAUDE.md, observe-replay-calibrate.md) and the 'deferred to #1608' notes (tracev2.go doc, replay.go doc, otel_convert.go, trace_encode.go) are all retargeted to describe #1608 as resolved. The surviving #1604 references (trace_encode.go:5, CLAUDE.md:223) correctly denote the Weka converter's own origin issue, not the deferred fix. |
| G1 | correctness | ✅ CONFIDENT | In `LoadTraceV2SessionBlueprints`, what happens when a multi-round session has a mix of `nil` and non-nil `ThinkTimeUs` records (e.g. round 1 is `nil`, round 2 is `&5000`)? Since `sessionHasRecordedThinkTime` returns `true` if any non-round-0 record is non-nil, `derefInt64(nil)` will evaluate round 1 to `0` — is treating a missing intermediate think time as `0` correct, or should it fall back to arrival gap for that specific round? | With a mixed session (round 1 nil, round 2 &5000), sessionHasRecordedThinkTime returns true (round 2 non-nil), so the recorded path runs and derefInt64(nil)==0 gives round 1 a think of 0 (back-to-back) rather than an arrival-gap fallback for that round. This is a documented, deterministic choice — and it cannot arise from either supported producer: convert weka sets ALL rounds 1..N non-nil (a computed gap or &0), convert otel leaves ALL nil. A mixed session is only reachable via a hand-edited/partially-filled CSV. Treating a missing interior think as 0 is defensible because think-time selection is per-session (whole session commits to recorded vs gap), not per-round. Correct within the model, not a flaw. |
| G2 | correctness | ✅ CONFIDENT | In `ExportTraceV2`, when `includeThinkTime` is true, a `nil` `ThinkTimeUs` writes an empty string `""` to the CSV row. If `think_time_us` is not the last column (or if trailing empty fields exist), how does `parseTraceRecord` handle cases where trailing commas exist vs omitted trailing fields, and is there any ambiguity in row slicing? | There is no ambiguity. think_time_us is ALWAYS the last column (appended after x_request_id and adapter), so a nil writes a trailing empty cell. Go's encoding/csv writes that as a trailing comma and reads it back as a preserved empty final field (len(row) unchanged) — verified empirically, including when the empty cell is on the very last row. parseTraceRecord looks it up by absolute header index (thinkTimeUsIdx) with a `< len(row)` guard, and LoadTraceV2 enforces `len(row) >= minCols` (which includes +1 for the think column), so a genuinely-short row errors rather than mis-slicing. Empty -> nil; "0" -> &0. |
| G3 | invariants | ✅ CONFIDENT | INV-3 states that recorded think time cannot be negative. `parseTraceRecord` rejects negative numbers parsed from CSV, but what prevents a Go struct literal from setting `ThinkTimeUs: i64p(-100)` directly in memory, and does `LoadTraceV2SessionBlueprints` validate pointers prior to sampler construction? | Nothing prevents an in-memory struct literal from holding i64p(-100) (true of any Go field), but it cannot violate the invariant. LoadTraceV2SessionBlueprints clamps `if t < 0 { t = 0 }` immediately after derefInt64 and BEFORE the value enters the SequenceSampler, so a negative in-memory pointer never reaches OnComplete. The CSV path additionally rejects negatives at parse time. The arrival-gap fallback path clamps too. So INV-3 (clock monotonicity — negative think would move the clock backward) is preserved for both CSV-loaded and in-memory-constructed traces. |
| G4 | invariants | ✅ CONFIDENT | Does changing `ThinkTimeUs` to `*int64` affect determinism (INV-6) or trace byte-identity across export and load cycles for traces generated via `blis run` or `convert otel` where `ThinkTimeUs` is `nil`? | Determinism/byte-identity is preserved for traces where ThinkTimeUs is nil. blis run and convert otel never set the field (or leave it nil), so all records are nil -> includeThinkTime is false -> the column is omitted, byte-identical to pre-#1608 where all-zero int64 tripped the `!= 0` gate to also omit it. No run/observe trace-export site constructs a non-nil TraceRecord.ThinkTimeUs. Only Weka all-overlap traces change bytes (column now emitted with recorded zeros) — the intended BC-5/BC-6 behavior change, not an INV-6 regression. |
| G5 | rules | ✅ CONFIDENT | Rule R9 calls for pointer types for optional fields where zero is valid. Here `ThinkTimeUs` uses `*int64` and test files define helper functions like `i64p(v int64) *int64`. Is `i64p` established as a canonical helper in a shared package or is it duplicated across test files? | i64p is NOT a shared canonical helper — it is duplicated per test package: sim/workload/trace_encode_test.go:8 (added by this PR) and cmd/kv_offload_test.go:26 (pre-existing, used by observe_corpus_test.go etc.). Two package-local copies. This is idiomatic Go (test helpers don't cross package boundaries without an exported testutil package) and is not a rule violation. |
| G6 | rules | ✅ CONFIDENT | Rule R17 prohibits silent fallback or context loss. In `LoadTraceV2SessionBlueprints`, `derefInt64(rounds[i].ThinkTimeUs)` silently converts `nil` to `0` inside a session where `sessionHasRecordedThinkTime` returned `true`. Does this violate R17 by silently masking a `nil` field in a session that was detected as having recorded think times? | No violation. First, R17 is not the silent-fallback rule — it is 'Document signal freshness for routing inputs' (routing scorers). The applicable rule is R1 (No silent data loss). The nil->0 conversion inside a recorded session is (a) explicitly documented at replay.go:250-253, so not silent, and (b) unreachable from shipped producers — convert weka sets every round 1..N non-nil, so when sessionHasRecordedThinkTime returns true, all interior rounds are non-nil. It masks nothing in practice; the defensive default is deterministic and documented. R1 is satisfied. |
| G7 | parity | ✅ CONFIDENT | Does changing `NormalizedRound.ThinkUs` to `*int64` maintain exact replay parity between traces loaded directly from CSV (`LoadTraceV2`) vs sessions converted in-memory via `EncodeSessionToTraceRecords` across `convert weka` and `convert otel`? | Yes. The CSV export/reload round-trip preserves the *int64 presence exactly — nil<->empty cell, &0<->"0", &N<->"N" — so a trace loaded from CSV (LoadTraceV2) carries the same ThinkTimeUs values that EncodeSessionToTraceRecords produced in memory for both convert weka and convert otel. Replay consumes those identical values, so parity holds. This is directly exercised by the mixed-presence round-trip test. |
| G8 | parity | ✅ CONFIDENT | When replaying a Weka trace where all turns overlapped (all recorded `&0`), does `blis run` produce identical timing outputs when replayed via closed-loop session blueprints vs open-loop request lists? | No, and they are not supposed to be. Open-loop replay (LoadTraceV2Requests) uses each record's recorded ArrivalTime directly and ignores think time entirely; closed-loop replay (LoadTraceV2SessionBlueprints) schedules round N+1 at round N's completion + recorded think (all-&0 -> back-to-back). These are two different replay semantics, and no invariant requires them to match — INV-13 governs run-vs-replay parity for the SAME configuration, not open-loop vs closed-loop. The whole purpose of the think_time_us column and this fix is to make closed-loop spacing differ from raw arrival timing. |
| G9 | tests | ✅ CONFIDENT | The PR adds unit tests for round-tripping mixed presence (`nil`, `&0`, `&N`), but is there a companion invariant test verifying that exporting a trace with `nil` think times and re-importing it produces byte-identical CSV content to pre-#1608 trace outputs? | There is a companion invariant test, though not a literal byte-diff-against-old-output golden. The subtest 'absent: no column when no record carries it (INV-6 byte-identity)' asserts the think_time_us column is OMITTED when all records are nil — which is the substance of byte-identity to pre-#1608 (both the old `!= 0` gate on all-zero and the new `!= nil` gate on all-nil omit the column). The complementary 'lone recorded &0 emits column (BC-5)' subtest pins the intended difference. Byte-identity only holds for nil-think traces; Weka traces intentionally differ, so a raw byte-vs-old golden would be inappropriate for them. |
| G10 | tests | ✅ CONFIDENT | Would the new test `TestLoadTraceV2SessionBlueprints_AllRecordedZeroThink_UsesRecordedZeros` survive a refactoring where think time presence is tracked via a boolean presence flag or bitmap rather than a pointer? | Yes. The test asserts observable behavior — the constructed sampler yields [0,0] (the recorded zeros) rather than the arrival gaps [5000,7000] — not the *int64 mechanism. A refactor that tracks presence via a boolean flag or bitmap while preserving 'all-recorded-zero uses recorded zeros' would still make the sampler emit [0,0] and pass. It survives the refactor-survival test. |
| G11 | rationale | ✅ CONFIDENT | Why was `*int64` selected over a boolean presence field (e.g., `HasThinkTime bool` + `ThinkTimeUs int64`) or a struct wrapper, considering pointer allocations in slice elements can increase GC pressure on large traces? | *int64 is the minimal, idiomatic Go presence signal (nil=absent, &0=recorded zero) and is consistent in spirit with R9 (pointer where zero is a valid value). GC pressure is bounded: the pointer is nil — zero allocations — for the entire common path (round 0 of every session, and every record of run/observe/otel traces). Allocations occur only for recorded rounds 1..N of converter traces (Weka), one per non-empty cell on parse. A bool+int64 wrapper avoids the allocation but widens every TraceRecord and still needs a distinct CSV encoding; the pointer keeps the single trailing column. A defensible tradeoff. |
| G12 | rationale | ✅ CONFIDENT | Why was no trace schema version bump required in `TraceHeader.Version` for encoding empty cell strings for optional `think_time_us`, and how was backward compatibility with older trace parsers verified? | No version bump is needed because think_time_us is an optional, name-detected trailing column: LoadTraceV2 sets reader.FieldsPerRecord=-1 and locates the column by header name (thinkTimeUsIdx), with no gate on TraceHeader.Version. Backward tolerance: a pre-#1608 parser that knew the column (#1478..#1607) treats an empty cell as 0 via its `if v != ""` guard — graceful lossy degradation, no crash; a pre-#1478 parser ignores the unknown trailing column. Forward: new parser reading an old trace works (old writer only emitted the column when non-zero). Compatibility rests on the additive/optional design, verified by the within-version round-trip tests. |
| G13 | scope | ✅ CONFIDENT | The PR touches `CLAUDE.md`, `observe-replay-calibrate.md`, and multiple converter doc comments. Do all documentation edits strictly pertain to non-lossy `think_time_us` encoding (#1608), or does the PR introduce documentation changes for unrelated features? | All documentation edits pertain strictly to non-lossy think_time_us (#1608). CLAUDE.md and observe-replay-calibrate.md edits are confined to the Weka-converter think-time note; the code doc-comment changes in otel_convert.go, trace_encode.go, tracev2.go, replay.go, weka_convert.go all describe the presence semantics. No unrelated feature documentation is introduced. |
| G14 | scope | ✅ CONFIDENT | Did the change to `NormalizedRound.ThinkUs` require any modifications to converters or encoders beyond Weka and OTel, or were any other trace importers affected? | No other importers were affected. NormalizedRound has exactly two producers — otel_convert.go:184 and weka_convert.go:139 — and EncodeSessionToTraceRecords has exactly two callers (the same two files), both updated. The encoder maps ThinkUs->ThinkTimeUs (both now *int64). Every other ThinkTimeUs in the tree is an unrelated config type (MultiTurnConfig/ClientConfig in generator.go/servegen.go/inference_perf.go/synthesis.go/stream.go, SessionBlueprint in session.go, spec fields in spec.go) that remains int64 and correctly needed no change. |
| G15 | errors | ✅ CONFIDENT | In `parseTraceRecord`, parsing an invalid string in `think_time_us` returns a wrapped error. Is this error guaranteed to abort trace loading with a hard failure in CLI entry points rather than degrading silently? | Yes. parseTraceRecord returns a wrapped error on an unparseable think_time_us; LoadTraceV2 propagates it. At CLI entry points this is a hard failure: replay.go logrus.Fatalf('Failed to load trace') and calibrate.go logrus.Fatalf('Failed to load TraceV2'); observe_corpus.go returns a wrapped error up its call chain. No silent degradation (R1 satisfied). |
| G16 | errors | ✅ CONFIDENT | If `sessionHasRecordedThinkTime` encounters a session with 0 rounds or missing records, does `LoadTraceV2SessionBlueprints` safely handle nil pointer dereferences under all possible slice bounds? | Yes, it is safe under all slice bounds. sessionHasRecordedThinkTime loops `for i:=1; i<len(rounds)` — a no-op for 0- or 1-round sessions — and reads only the pointer field (rounds[i].ThinkTimeUs != nil), never dereferences it, so a nil pointer is harmless. LoadTraceV2SessionBlueprints guards `if len(rounds)==0 { continue }` before use, and the recorded/gap think loops also start at i=1. No unchecked index or nil dereference is reachable. |

### Items to fix

### Important to consider
- **F1** — The issue proposed a -1 sentinel or a bitmap/bool presence column; the PR chose a third option (pointer presence), a clean fit for 'presence signal' — worth noting the deviation from the two listed options.
  _sim/workload/tracev2.go:195,278,384; sim/workload/replay.go:130-137,254; sim/workload/trace_encode.go:13,53; sim/workload/weka_convert.go:124-137; sim/workload/otel_convert.go:166_
- **F2** — There is no dedicated TraceV2 CSV-column reference doc enumerating think_time_us, so nothing else needed the nil/&0 distinction added.
  _CLAUDE.md:238-240; docs/guide/observe-replay-calibrate.md:398-415; grep of docs/ shows other refs are the config field (spec.go:121,219, session.go:35 — still int64)_
- **G1** — For a hand-crafted mixed trace, a per-round fallback (gap for nil interior rounds, recorded for non-nil) would be more faithful than 0; the PR consciously keeps per-session selection. Worth a one-line acknowledgement in the doc comment that the interior-nil path is unreachable from shipped converters.
  _sim/workload/replay.go:250-258 (documented derefInt64/clamp); sim/workload/weka_convert.go:124-137 (rounds 1..N always set); sim/workload/otel_convert.go:166 (all nil)_
- **G3** — INV-3 is formally 'clock monotonicity'; the questioner's paraphrase ('recorded think cannot be negative') is a consequence, not the invariant's statement — the code comments correctly connect the two.
  _sim/workload/replay.go:255-257 (defensive clamp before sampler); sim/workload/tracev2.go:668-670 (CSV rejects negative); replay.go:265-269 (gap-path clamp)_
- **G5** — R9 (as written) is specifically about YAML config structs using *float64 to disambiguate zero; it does not literally govern this CSV field or a test helper. The *int64 choice here is R9-in-spirit. The i64p duplication is minor; a shared internal test helper could consolidate it if more packages need it.
  _grep 'func i64p': sim/workload/trace_encode_test.go:8, cmd/kv_offload_test.go:26_
- **G6** — Questioner mis-cited R17; the concern maps to R1, which the documented+unreachable path does not violate.
  _docs/contributing/standards/rules.md:228 (R17 = routing freshness), :24-26 (R1 = no silent data loss); sim/workload/replay.go:250-258 (documented); weka_convert.go:124-137 (all rounds set)_
- **G9** — No test compares the emitted CSV byte-for-byte against a stored pre-#1608 fixture; the 'column omitted' invariant covers the substance transitively.
  _sim/workload/tracev2_test.go:1192-1214 (absent -> column omitted, INV-6), :1216+ (lone &0 -> column emitted)_
- **G10** — The test inputs still use i64p (pointer construction), so a producer-side refactor of how presence is stored would touch the test setup; the assertion itself is behavioral.
  _sim/workload/replay_test.go TestLoadTraceV2SessionBlueprints_AllRecordedZeroThink_UsesRecordedZeros: `if g1,g2 := s.Sample(nil),s.Sample(nil); g1!=0\|\|g2!=0` asserts outcome, not internal type_
- **G11** — For an extremely large Weka corpus, the per-recorded-round heap allocation is real but proportional to recorded rounds; if it ever mattered, a bool+int64 pair would remove it. Not a concern at current scale.
  _sim/workload/tracev2.go:195 (*int64 field), :662-671 (`&parsed` only on non-empty cell); nil for round 0 (weka_convert.go:124-137) and all non-converter traces_
- **G12** — There is no cross-version test against an actual older binary (typical for this repo); backward compat is argued from the parser structure, not exercised.
  _sim/workload/tracev2.go:418-446 (name-detected, FieldsPerRecord=-1, no version gate), :662-663 (empty->nil; old code's `v != ""` -> 0)_

namasl:

susiejojo:
### BLIS PR Self-Review (manual `/blis-pr-review`)

**PR:** fix(workload): non-lossy `think_time_us` encoding — distinguish recorded-0 from absent (#1608)
**Reviewed at head:** `6c29d17c` (one commit past the `232d03b5` that the earlier bot reviews saw)
**Verified locally:** `go build ./...` clean · `go test ./sim/workload/ ./cmd/` pass.

---

#### Verdict: ✅ READY TO MERGE — no blocking issues

The change is minimal, correct, and tightly scoped to #1608. It converts `TraceRecord.ThinkTimeUs` and `NormalizedRound.ThinkUs` from `int64` to `*int64` presence signals (`nil` = not recorded, `&0` = recorded zero), fixing the lossy-0-sentinel bug that made an all-overlap Weka session degrade to arrival-gap think at replay.

#### Correctness & invariants (independently verified)
- **INV-6 (byte-identity):** `RequestsToTraceRecords` never sets `ThinkTimeUs` → stays `nil` → export gate `if r.ThinkTimeUs != nil` suppresses the column for all `blis run` / `observe` / `convert otel` traces. Confirmed no non-pointer `TraceRecord{...ThinkTimeUs:...}` construction sites remain (the other `ThinkTimeUs` matches in the tree are distinct config types — `MultiTurnSpec`, `ClientConfig`, `SessionBlueprint` — correctly untouched, still `int64`).
- **INV-3 (no negative think / clock monotonicity):** parser still rejects negatives; `LoadTraceV2SessionBlueprints` re-clamps `t < 0 → 0` after `derefInt64`, before the value reaches the sampler. Preserved.
- **CSV round-trip:** `nil↔""`, `&0↔"0"`, `&N↔"N"`. Trailing empty cell is safe — `think_time_us` is always the last column, `FieldsPerRecord=-1`, and the `len(row) < minCols` guard (with `minCols++` when the column is present) rejects genuinely-short rows rather than mis-slicing.
- **No schema-version bump needed:** correct — this refines the *cell encoding* of an existing optional, name-detected trailing column; degrades gracefully both directions.

No parity gap. This is a converter→CSV→replay data-model change. `blis run` and `blis observe` don't produce recorded-think traces and are correctly unaffected (verified at the write path). `convert weka` sets rounds 1..N (incl. `&0`); `convert otel` leaves all `nil` — both encoders updated.

#### Preemption / timeout
Not applicable — no `Request` fields, per-request metrics, state transitions, aggregate counters, deadlines, or timeout mechanisms are touched.

#### Test coverage
Behavioral and refactor-survivable (assert sampler *output*, not the `*int64` mechanism). BC-1 (`AllRecordedZeroThink_UsesRecordedZeros`), BC-2 (`AbsentThink_FallsBackToGap`), BC-3 (mixed presence round-trip), BC-5 (lone `&0` emits column — regression guard), BC-6 (Weka overlap → recorded `&0`), BC-8 (negative → hard error) all present.

**Note on prior-review follow-up:** the earlier `/blis-pr-review` and `qa-review` (G1) both flagged, as an *optional* improvement, the absence of a test for the defensive nil-interior `derefInt64` path. Commit `6c29d17c` adds `TestLoadTraceV2SessionBlueprints_MixedInteriorThink_DerefDefensive` covering exactly `[nil, &0/&N]` interiors — that gap is now closed.

#### Architecture (per Archon `NO_CHANGE` + independent read)
Concur with Archon: internal-only, no package-boundary or module-contract change. `derefInt64` is reused from `generator.go` within the same `workload` package (no new arrow). The `*int64` choice is R9-in-spirit (pointer for optional-where-zero-is-valid) and makes the struct zero value the correct default (R4-clean). Sound.

#### Non-blocking observations (no action required)
- `i64p` test helper is duplicated (`sim/workload/trace_encode_test.go`, `cmd/kv_offload_test.go`) — idiomatic Go; a shared `testutil` helper is only worth it if more packages need it.
- The doc comment at `replay.go:250-258` correctly notes the nil-interior branch is unreachable from shipped converters; the new defensive test pins its behavior for a hypothetical future sparse-think converter. Good.

**No changes required.**

susiejojo (approved):
lgtm

Issue #1618, opened 2026-08-25
Design: peak-rate saturation detector needs redesign (all-time peak latches; rho=1 horizon-decided; collapses to a rate deficit)


sriumcp:
## Summary

Issue #1614 proposed two changes. **Change 1 (calibration knobs) is implemented** in
the PR closing #1614. **Change 2 (a `peak-rate` detector) is withdrawn**: design review
found two defects that cannot be fixed by tuning, and a third that collapses the
statistic into one BLIS already computes.

This issue records what was measured, so a successor design does not rediscover it.

## The proposed statistic

Track the highest backlog seen so far, divided by the number of requests completed
so far: `R = peak_backlog / completions`. Fire when `R` stays above a threshold.
Intuition: a server keeping up has a backlog that stops growing while completions
accumulate, so `R` decays; a server falling behind has both growing together, so `R`
settles at a positive constant.

The intuition is right, and the asymptotic limit is exact and elegant:

```
rho > 1:  peak(t) ~ (lambda - mu)t,  C(t) ~ mu*t   =>  R -> rho - 1
```

Measured against that prediction (M/M/1, 5 seeds): rho=1.1 -> 0.110 (predicted 0.100),
rho=1.5 -> 0.50 (predicted 0.500). So `R` is a consistent estimator of **excess load**,
and a threshold of 0.06 means "report overload above 6% excess load." That part is
genuinely nice, and worth keeping in any successor.

## Defect 1 (CRITICAL): the running maximum never forgets

`peak` is an all-time maximum, so a single early burst latches the verdict forever.
Decay is only O(1/C).

Measured: a 600-request flash crowd at t=0, then **6,000 requests at a healthy 0.5x
load**:

| completions | peak | R | verdict at 0.06 |
|---|---|---|---|
| 600 | 600 | 1.0000 | FIRES |
| 2,000 | 600 | 0.3000 | FIRES |
| 6,000 | 600 | 0.1000 | FIRES |
| 6,600 | 600 | 0.0909 | **still FIRES** |

Because the headline label comes from `ReduceAll`'s last-window plurality, this
corrupts the primary output: the final label reflects the worst instant of the run
rather than its steady state.

## Defect 2 (CRITICAL): at rho = 1 the verdict is decided by run length

At exactly critical load the queue is a null-recurrent walk, so the maximum grows
diffusively and `R ~ 1.77/sqrt(C)`.

Measured (25 seeds, rho = 1.000 exactly):

[...]

Issue #1617, opened 2026-08-25
Hardening: saturation.Event.InputTokens/OutputTokens are declared but never populated


sriumcp:
## Summary

`saturation.Event` declares `InputTokens` and `OutputTokens`, but nothing ever writes
them and nothing ever reads them. Any future token-aware detector that trusts these
fields will silently read zeros — the failure mode is a detector that reports STABLE
unconditionally while looking healthy.

## Where

Declared at `sim/saturation/detector.go`:

```go
type Event struct {
	Timestamp    int64
	Type         EventType
	RequestID    string
	LatencyMs    float64
	InputTokens  int      // never written
	OutputTokens int      // never written
}
```

`buildSortedEvents` (`sim/saturation/replay.go`) — the sole producer of the event
stream for all three detectors — populates only `Timestamp`, `Type`, `RequestID`, and
(on completions) `LatencyMs`.

`grep -rn "InputTokens\|OutputTokens" sim/saturation/` returns only the struct
definition.

## Why it matters

The data is available: `sim.RequestMetrics` carries `NumPrefillTokens` and
`NumDecodeTokens`, and `buildSortedEvents` already has the `RequestMetrics` row in
hand. So the fields are a two-line fix, not a design gap.

Leaving them declared-but-dead is the hazardous state: the struct advertises a
capability the pipeline does not provide. A work-conserving or token-weighted detector
would form a per-request work term of `kappa*InputTokens + OutputTokens`, get exactly
zero, and never leave its initial verdict — with no error and no warning.

This is worth fixing *before* the next detector rather than after, because the symptom
(a detector that always says STABLE) looks like a detector bug rather than a plumbing bug.

## Suggested fix

Populate both fields on both events in `buildSortedEvents`:

```go
InputTokens:  r.NumPrefillTokens,
OutputTokens: r.NumDecodeTokens,
```

The three current detectors ignore these fields, so this cannot change their verdicts
— but it should still carry a byte-identity golden (INV-6) as its companion test, plus
a regression guard asserting the fields are non-zero for a trace with non-zero token
counts (the guard is the point: it fails if the plumbing regresses).

## Discovery context

Found during the source audit for #1614 (saturation detector calibration knobs). Out
of scope there — that PR adds no token-aware detector — so filing separately per the
PR workflow's pre-existing-issue rule.

Issue #1616, opened 2026-08-25
Perf: CompositeDetector.Detect re-copies and re-sorts all completions every event (O(n^2), 6.5 GB at n=10k)


sriumcp:
## Summary

`CompositeDetector.Detect()` copies and re-sorts its entire completion history on
**every** event. Since the drive loop calls `Detect()` once per event (and the bank
calls it once per detector per event), this makes composite **O(n²) in both time and
allocated bytes** over a run.

The sort is also **provably redundant**: `buildSortedEvents` already emits events in
`(timestamp, type, request-id)` order, so completions arrive at `Observe` in
non-decreasing timestamp order.

## Where

`sim/saturation/composite.go`, in `Detect()`:

```go
sorted := make([]Event, len(c.completions))   // full copy, every event
copy(sorted, c.completions)
sort.Slice(sorted, func(i, j int) bool {      // full O(m log m) sort, every event
    return sorted[i].Timestamp < sorted[j].Timestamp
})
sortedLatencies := make([]float64, len(sorted)) // second full copy
```

Called from `replay.go` (`ReplayOneDetector`) and `bank.go` (`fanout`), both of which
do `Observe(e)` then `Detect()` for every event.

## Measured

| requests | composite time | composite allocated |
|---|---|---|
| 500 | 8.3 ms | 17 MB |
| 1,000 | 20 ms | 69 MB |
| 2,000 | 64 ms | 272 MB |
| 4,000 | 216 ms | 1.06 GB |
| 8,000 | 763 ms | 4.20 GB |
| 10,000 | 1.15 s | **6.54 GB** |

Doubling `n` quadruples both — quadratic in time *and* in allocated bytes. For
comparison at n=10,000: `backlog-drift` is 4.3 ms / 6 MB. `threshold` is O(n²) in
time but only O(n) in memory (`meanLatency` re-sums without copying).

Under `--detectors all`, composite accounts for ~94% of the total cost.

## Suggested fix

Two independent wins, smallest first:

1. **Drop the sort.** Completions already arrive in timestamp order, so
   `c.completions` is already sorted. This alone removes the `O(m log m)` factor and
   one of the two copies.
2. **Incrementalize.** The algorithm needs only `mean(first half)`, `mean(second half)`,
   and four quartile means. Those are computable from running sums plus an index,
   with no per-event copy at all — turning `Detect()` into O(1).

Either change must preserve byte-identical output (INV-6), so it wants a golden
[...]

Issue #1615, opened 2026-08-25
Bug: blis replay completes far fewer requests than the run it replays (INV-13 violation)


sriumcp:
## Summary

`blis replay` of a trace exported by `blis run` completes far fewer requests than the
original run, violating **INV-13** (run/replay parity). No saturation or detector
flags are involved — this reproduces with a plain `run --trace-output` /
`replay` pair.

The gap widens with load, which suggests replay is not reproducing the arrival
schedule or is terminating early rather than dropping a fixed set.

## Reproduction

Built from `upstream/main` at `b5934c13` (no local changes), run from the repo root:

```bash
for r in 20 60; do
  ./blis run    --model qwen/qwen3-14b --num-requests 200 --rate $r --seed 42 \
                --trace-output /tmp/p_$r
  ./blis replay --trace-header /tmp/p_$r.yaml --trace-data /tmp/p_$r.csv \
                --model qwen/qwen3-14b
done
```

| rate | `run` completed | `replay` completed |
|---|---|---|
| 20 | 200 | **130** |
| 60 | 200 | **29** |

At `--num-requests 400 --rate 60 --seed 42`:

```
run     completed_requests=400  still_queued=0   still_running=0
replay  completed_requests=82   still_queued=63  still_running=...
```

## Why it matters

INV-13 states: *"For any configuration supported by both `blis run` and `blis replay`,
a trace exported via `--trace-output` and replayed with identical flags MUST produce
identical per-request metrics."* Here the *count* of completed requests differs, so
every downstream aggregate (TTFT/E2E percentiles, throughput, goodput) is computed
over a different population.

Concretely, this makes the observe → replay → calibrate pipeline unsound at load:
`calibrate` compares simulated against real latencies, but the simulated side is
missing most of the requests.

It also silently degrades any post-hoc saturation analysis: with `--detectors`, the
trace is built from `CompletedRequestMetrics()`, so `run` and `replay` feed the
detectors different event streams and can report different final labels for what is
nominally the same workload (observed: `OVERLOADED` on run vs `BACKLOGGED` on replay,
800 vs 164 trace records).

## Discovery context

Found while verifying INV-13 for #1614 (saturation detector calibration knobs). The
parity check failed, so I re-ran it against a binary built from unmodified
`upstream/main` to isolate the cause — it fails identically there, with no detector
flags at all. Filing rather than fixing, per the PR workflow's pre-existing-issue rule.

## Suggested starting points

[...]

PR #1619, opened 2026-08-25
feat(saturation): give every post-hoc detector a calibration knob

### LINKED ISSUE #1614: Post-hoc saturation: give composite a sensitivity knob, and add the peak-rate detector
## Summary

Two independent changes to post-hoc saturation detection (`sim/saturation/`):

1. **Give `composite` one tunable knob.** It currently has none, which makes it impossible to compare fairly against the other detectors. Adding the knob turns it from "disqualified" into the best detector in the suite. **~15 lines.**
2. **Add a fourth detector, `peak-rate`**, built from a standard queueing-theory result. It matches `composite` and beats `threshold` on warning speed.

Both were built and measured in a controlled study; the measurements are quoted below. Change 1 is small, high-value, and independent — **land it first, on its own.**

Context: discussion #1596 stress-tested the three existing detectors with metamorphic tests and found `threshold` PASS, `composite` DISQUALIFIED, `backlog-drift` FAIL. That post ends by predicting `composite` "would likely become the best if it were given a tunable sensitivity knob." That prediction is correct, and Change 1 is the knob.

---

## Background: why a knob is not a nice-to-have

A saturation detector answers "is the server overloaded?" There is no ground truth to check it against, so we compare detectors by **holding their false-alarm rates equal** and asking which one warns earliest. This is the standard discipline (`papers/saturation`, `metamorphic_tests.md` §3.4).

That comparison requires each detector to have **at least one dial that trades sensitivity against false alarms.** Without a dial you cannot move a detector onto the common false-alarm rate, so it cannot be compared at all — only disqualified.

Two of the three current detectors have no such dial:

| Detector | Dial today | Comparable? |
|---|---|---|
| `threshold` | `threshold_ms` | yes |
| `composite` | **none** | **no** |
| `backlog-drift` | `slope_k` is **hardcoded `3.0`** | **no** |

`backlog_drift.go` says so itself:

> `// backlogDriftSlopeK is a tunable heuristic constant, NOT an empirically calibrated value.`

So `composite` was reported as failing when what actually happened is that it could not be tuned. **This is a measurement-apparatus bug that produced a wrong conclusion about a working detector.**

---

# Change 1 — a `sensitivity` knob for `composite` (and `slope_k` for `backlog-drift`)

### What `composite` does now

It computes a saturation score and compares it against a "noise floor" of `1/√arrivals` (`composite.go`, `computeComposite`):

```go
noiseFloor := 1.0
if arrivals > 0 {
    noiseFloor = 1.0 / math.Sqrt(float64(arrivals))
}
```

The floor is fixed. Nothing can move it.

### The change

Multiply the floor by one number:

```go
noiseFloor *= sensitivity   // default 1.0
```

A larger `sensitivity` raises the bar, so the detector fires less. That is the entire mechanism. Default `1.0` reproduces today's behaviour **exactly**, so this is byte-identical unless a user opts in (INV-6).

### Wiring

Follow the existing `ThresholdBlock` pattern precisely — this is a **policy-template extension**, the cheapest of the four extension types in the design guidelines, and needs no new interface:

- **`config.go`**: add `CompositeBlock{ Sensitivity *float64 \`yaml:"sensitivity"\` }` and a field on `SaturationConfig`. Pointer type so "absent" differs from "zero" (R9).
- **`config.go`**: register `composite` in the `blockOwners()` table so ownership checks work for both the single-detector and bank paths automatically. **Note:** `composite` currently *errors* if given a config block ("composite has no params"). That rejection must be replaced, not merely relaxed.
- **`composite.go`**: `NewCompositeDetectorWithSensitivity(float64)`; keep `NewCompositeDetector()` delegating at `1.0` so every existing call site is untouched (R4).
- **Validation**: reject non-positive, `NaN`, `Inf` with a message naming the field (R1 — never silently fall back).
- **`backlog_drift.go`**: same treatment for `slope_k`, replacing the `backlogDriftSlopeK` const with a config field that falls back to `3.0`.

### Usage

```yaml
composite:
  sensitivity: 2.0
```

```bash
blis run --model qwen/qwen3-14b --detectors composite \
  --saturation-config sat.yaml --saturation-report sat.json
```

### Measured effect

Same suite and calibration procedure as #1596, on a frozen apparatus (Llama-3.1-8B-Instruct, H100, TP=1, trained-physics; nominal capacity 20 rps measured by backlog divergence; 5 seeds; 6,000 requests/run):

| | before | after (`sensitivity: 2.0`) |
|---|---|---|
| false-alarm rate | 17–20 of 20 | **0 of 20** |
| verdict | DISQUALIFIED | **PASS all four tests** |

`composite` goes from worst to tied-best. **Nothing about its algorithm changed.**

### Acceptance criteria

- [ ] Absent `composite:` block ⇒ output byte-identical to today (INV-6). Assert against a committed golden.
- [ ] `sensitivity: 1.0` ⇒ byte-identical to absent.
- [ ] Larger `sensitivity` fires strictly no more often on a fixed trace (monotonicity — this is a **law**, not a value, per the BDD/TDD guidance).
- [ ] Invalid values error and name the field.
- [ ] A `composite:` block under `--detectors threshold` errors (block-ownership, R1).
- [ ] Same five for `backlog_drift.slope_k`.

---

# Change 2 — the `peak-rate` detector

### The idea, in plain terms

A server's backlog behaves like a **random walk that cannot go below zero**: it drifts up and down, but when the queue empties it sits at zero until more work arrives. Such walks have a sharp, well-known signature.

Track the **highest backlog seen so far**, and divide it by **elapsed time**:

```
R = peak_backlog_so_far / elapsed_seconds
```

- **Server keeping up:** the peak stops growing — there is a ceiling on how far behind it can fall — while time keeps ticking. So `R` **falls**.
- **Server drowning:** the peak grows roughly in step with time. So `R` **stays flat**.

The detector fires when `R` **holds** instead of falling.

### Why this is worth adding next to what exists

It needs **no latency threshold and no notion of how fast the server "should" be.** `threshold` needs an absolute millisecond target that must be re-tuned per model and per GPU. `peak-rate` reads a shape.

It is also the natural foil to `backlog-drift`, and the comparison explains why that detector fails:

> `backlog-drift` fits a **straight line** to backlog growth. But a server sitting *exactly at capacity* has backlog growing like **√t**, not `t`. A straight-line fit to a √t curve tends toward **zero slope** — so `backlog-drift` reports "fine" at precisely the load you most want flagged.

Dividing the *peak* by elapsed time does not have that failure mode. Two detectors, adjacent names, one mathematical distinction.

### Measured results

Full suite, 5 seeds, 11 load rungs, 6,000 requests/run, every detector calibrated to a matched false-alarm rate first:

| Detector | T1 rate | T2-IN prompt | T2-OUT output | T4 stability | false alarms | detection delay |
|---|---|---|---|---|---|---|
| **`peak-rate`** (new) | PASS | PASS | PASS | PASS | 0% | **0.0 s** |
| `composite` + knob | PASS | PASS | PASS | PASS | 0% | **0.0 s** |
| `threshold` | PASS | PASS | PASS | PASS | 0% | 6.6 s |
| `backlog-drift` | FAIL | PASS | FAIL | FAIL | 25% | — |

Firing pattern across the load ladder (`F` = silent, `T` = fired; capacity cliff at 1.0×):

```
0.3 F  0.4 F  0.5 F  0.6 F  0.7 F | 0.9 T  1.0 T  1.1 T  1.25 T  1.5 T  2.0 T
```

A clean step at the cliff: silent on every below-capacity rung, fires on every above-capacity rung, with its knob strictly interior to the search grid (so the 0% false-alarm rate is real, not a grid artifact).

**Stated plainly: `peak-rate` ties `composite`; it does not beat it.** Both are 0.0 s. Its value is not that it wins but that (a) it is a genuinely independent mechanism, so the two agreeing is real evidence rather than one detector's quirk, and (b) it needs no per-hardware latency target.

### One implementation caveat that must not be skipped

**The observation window is part of the algorithm.** The theory is asymptotic; how well the statistic separates healthy from overloaded depends on run length:

| requests per run | separation |
|---|---|
| 500 | 2.3× |
| 2,000 | 4.7× |
| 8,000 | **14.6×** |

The mechanism: at 0.5× capacity, peak backlog goes 66 → 71 over the run (it has levelled off, so `R` decays 13×); at 1.5× it goes 329 → **2743** (still climbing, so `R` barely moves). Short runs cannot tell these apart because *the healthy server's peak has not levelled off yet.*

This produced a **false negative in our own work** — early runs at 500 requests said the idea failed. It did not; we were watching too briefly. Two consequences for the implementation:

1. `min_observations` is **load-bearing**, not hygiene. Below it, return `STABLE` and do not evaluate.
2. `warmup_us` is **load-bearing**: `R` is largest early by construction (small `t`), so the initial transient must be discarded or the detector fires on every run's first moments.

Document both in the config comments. Anyone tuning this without knowing it will conclude the detector does not work.

### Shape

New file `sim/saturation/peak_rate.go`, implementing the existing four-method `Detector` interface — **no interface change**:

```go
Name() string
Observe(Event)
Detect() Result
Reset()
```

State is `O(1)`: running peak, in-flight count, first/last timestamps, consecutive-breach counter. No per-request retention.

Config block, all five fields optional with documented defaults:

```yaml
peak_rate:
  threshold: 0.5          # the FPR-calibration knob; larger fires less
  warmup_us: 2000000      # discard the initial transient (load-bearing)
  min_observations: 200   # below this, report STABLE (load-bearing)
  consecutive_k: 3        # consecutive breaches before firing (the T4 lever)
  backlog_k: 2.0          # multiplies threshold for the OVERLOADED band
```

Register in `blockOwners()` and append `"peak-rate"` to `rosterOrder` in `bank.go` **at the end** — the roster order is canonical and fixes report key order, so inserting mid-list would change existing output.

### Reuse — what is deliberately *not* built

This is the part that keeps the change small. Everything below already exists and must be used as-is:

| Need | Existing machinery | New code |
|---|---|---|
| Streaming over a trace | `ReplayOneDetector` (`replay.go`) | none |
| Multi-detector comparison | `Bank` (`bank.go`) | one roster entry |
| Event stream construction | `buildSortedEvents` | none |
| Final label | `ReduceAll` (`reduce.go`) — a pure function, so a new detector gets final labelling free | none |
| Report writing | `TraceSink` → `WriteCombinedReport` | none |
| CLI plumbing (3 commands) | `resolveSaturation` → `saturationTracer.run` (`cmd/saturation.go`) | none |
| Config ownership checks | the `blockOwners()` table | one row |

So `--detectors peak-rate`, `--detectors all`, `--saturation-report`, and the stdout `saturation` map all work with **no changes to `cmd/`** beyond the config block.

### Acceptance criteria

- [ ] Absent `peak_rate:` block ⇒ output byte-identical to today (INV-6).
- [ ] Fires on a monotonically growing backlog; silent on a bounded one (the **law**, not a golden value).
- [ ] Larger `threshold` fires no more often on a fixed trace (monotonicity).
- [ ] `--detectors all` output equals the explicit full comma-list byte-identically, and `peak-rate`'s records under a subset are byte-identical to its records under `all` (INV-6 — selection filters *which* detectors run, never *how* they see traffic).
- [ ] Determinism: same input ⇒ byte-identical report across runs (INV-6).
- [ ] `run` → `replay` of the same trace produces an identical report (INV-13).
- [ ] `observe` accepts `--detectors peak-rate` and produces the same report shape.
- [ ] Invalid values error and name the field; unknown key in the block errors (strict YAML, R10).
- [ ] A `peak_rate:` block without `peak-rate` selected errors (R1).

---

## On run/replay/calibrate parity

Worth stating precisely, because the answer is better than expected and one third of it is a non-goal:

- **`run` / `replay` / `observe`** all reach the detectors through the *same* `resolveSaturation` → `saturationTracer.run` path (R23), and both single-detector and bank paths share `buildSortedEvents`. Parity is **structural** — there is one pipeline with three input adapters, so a new detector inherits it. It still needs a test (INV-13), but no new code.

- **Detector config does *not* need to round-trip through the trace header.** Unlike `kv_offload` (whose config changes the *simulation*), a post-hoc detector only reads a finished trace. The same trace can be re-analysed with any detector or any knob setting; that is a feature, and it is why `--saturation-config` is a replay-time flag rather than a header field. **Do not add a header block.**

- **`calibrate` has no saturation surface at all** (verified: no detector references in `cmd/calibrate.go`) and should not gain one here. It compares simulated against real latencies; saturation labels are a separate axis. Out of scope.

---

## Suggested sequencing

**PR 1 — the knobs** (`composite.sensitivity`, `backlog_drift.slope_k`). ~15 lines plus tests. Independently valuable: it makes *every* existing detector comparison sound, and it resolves #1596's open question. **This one should not wait for PR 2.**

**PR 2 — `peak-rate`.** One new file, one roster entry, one config block.

Splitting matters because PR 1 fixes a defect in how we *measure* detectors, while PR 2 adds a detector. Bundling them would mean the measurement fix arrives with, and is validated by, the thing being measured.

---

## Provenance and honest limits

Measurements come from a controlled study on a frozen apparatus with the objective fixed **before** measuring (branching and thresholds content-hashed pre-registration, so nothing could be tuned toward a result already seen).

What we are **not** claiming:

- `peak-rate` does **not** beat `composite`. It ties. The bar "merely matching the level detectors is not enough" is **not met** by this configuration. It is worth landing for mechanism independence and for needing no per-hardware target, not for winning.
- Results are for **one** frozen apparatus (one model, one GPU, one TP). Generality across hardware is untested.
- Only **one** of six candidate statistics passed. The others (excursion counting, idle fraction, peak-ratio stability) are not proposed here.
- The absolute numbers hold at ~6,000 requests/run; shorter windows measure something weaker (see the caveat above).

sriumcp:
## What this does

Gives every post-hoc saturation detector a sensitivity dial, so detectors can be
compared at a matched false-alarm rate instead of one being disqualified for having
no dial at all.

This is **Change 1 of the two proposed in #1614**. Change 2 (a new `peak-rate`
detector) is withdrawn — design review found defects that cannot be fixed by tuning.
See *Scope* below.

## Why it matters

Detector scores are only comparable when every detector is first calibrated to the
same false-alarm rate. A detector with no knob cannot be moved onto that rate, so it
cannot be compared — only disqualified. Two of the three were in that state:

| Detector | Dial before | Dial now |
|---|---|---|
| `threshold` | `threshold_ms` | unchanged |
| `composite` | **none** | `sensitivity` |
| `backlog-drift` | `slope_k` **hardcoded to 3.0** | `slope_k` |

`backlog_drift.go` conceded the problem itself: *"a tunable heuristic constant, NOT an
empirically calibrated value."*

Discussion #1596 is the concrete cost: it stress-tested these three detectors with
metamorphic tests and had to disqualify `composite` (it fired on 17–20 of 20 healthy
runs) with the note that there was *"no knob to turn it down"*, predicting it *"would
likely become the best if it were given a tunable sensitivity knob."* That prediction
is now testable.

## The changes

**1. `composite.sensitivity`** multiplies composite's `1/√arrivals` noise floor.
Default `1.0`; IEEE-754 gives `1.0*x == x` exactly, so an absent block is
byte-identical. Applied *before* `Signals["noise_floor"]` is recorded, so the trace
reports the floor the banding actually used.

**2. `backlog_drift.slope_k`** replaces the hardcoded `3.0` band multiplier. Two
details that are easy to get wrong, both test-pinned:

- `NewBacklogDriftConfig` returns a fresh struct literal, so it **zero-fills** any new
  field — and a literal `SlopeK` of `0` would make `slope <= 0*noiseFloor` false for
  every positive slope, banding **every** rising trace OVERLOADED. `effectiveSlopeK()`
  defuses that for all construction paths, so the 9-parameter arity and its 14 call
  sites are untouched.
- The multiplier is read in **two** places (the band switch and the score denominator).
  It is hoisted into one local so they cannot diverge. `TestBacklogDrift_ScoreAndLevelAgreeAtCustomSlopeK`
  is mutation-verified: reverting either read site alone makes it fail.

**3. One ownership table.** `checkBlockOwnership` and `checkBlockOwnershipSet` were two
hand-written functions enumerating blocks in separate `if`s — and they had **already
drifted**: the bank path had no `composite` case at all, so once composite owned a
block, a `composite:` block supplied to a selection omitting composite would have been
**silently dropped** (R1). Both now derive from one `blockOwners()` table.

**4. Derived name lists.** The valid-detector-name list was hardcoded in three places;
the error message and `--detectors` help now derive from `AllDetectorNames()`.

## Behavioral contracts

- **BC-1** GIVEN `composite: {sensitivity: S}`, WHEN classifying a fixed stream, THEN
  for `S₁ < S₂` the `S₂` detector reports a saturated level on no event where `S₁`
  reported STABLE.
- **BC-2** GIVEN a stream that is BACKLOGGED at the default `slope_k`, WHEN
  re-classified with a smaller `slope_k`, THEN it escalates to OVERLOADED.
- **BC-3/BC-4** GIVEN no block, or an explicit default, THEN every `Result` field —
  including every `Signals` entry — equals the pre-change value (INV-6).
- **BC-5** GIVEN a `SlopeK` zero-filled by any construction path, THEN verdicts equal
  those of an explicit `3.0`.
- **BC-6** GIVEN any valid `slope_k`, THEN `Level` and `Score` derive from the same
  multiplier, so `Score == 1.0` coincides with the OVERLOADED edge.
- **BC-7** GIVEN a zero, negative, NaN or Inf knob, THEN construction returns an error
  naming the YAML field and no detector is built (R6).
- **BC-8/BC-9** GIVEN a block whose owner is not selected, THEN a hard error naming the
  block — on both paths, never a silent drop (R1).
- **BC-10** GIVEN a misspelled field in a new block, THEN strict parsing fails (R10).

## Two bugs the review found (and how)

Both were found by **rewriting structural tests as behavioral ones** — the change
that matters most in this PR's history.

**A subnormal knob decoupled Level from Score.** `slope_k = 5e-324` is positive and
finite, so it passed validation — then `slopeK * noiseFloor` underflowed to exactly
zero, the `denom > 0` guard skipped the score computation, and the detector reported
**OVERLOADED with Score 0** in 116 of 400 records, contradicting the invariant this
PR documents. `composite.sensitivity` had the same shape: a subnormal drove the floor
to zero, making *every* event non-STABLE.

The original test asserted that `effectiveSlopeK()` returned a finite positive value —
which it did. It checked the multiplier **in isolation rather than the product it
participates in**, so it passed while the bug was live. Rewritten to assert on the
verdict (`OVERLOADED ⟹ Score == 1.0`), it failed immediately. Fixed with a shared
`minCalibrationKnob = 1e-6` bound plus an explicit underflow branch.

**An unknown detector name reported the wrong error.** The ownership loop tested
`owner != name` for every row, so a typo'd name matched every row and blamed the first
config block. `--detectors bogus` with a valid `threshold:` block said *"threshold
block is not valid"* instead of *"unknown saturation detector"* — hiding the real
mistake and omitting the valid names. The old `switch` had no `default` and fell
through, so this was a regression. Fixed by validating the name first (the bank path
already did).

## Verification

```
go build ./...            exit 0
go test ./... -count=1     13/13 packages ok
golangci-lint run ./...    0 issues
```

Tests in `sim/saturation`: **89 → 155**, plus 5 end-to-end CLI tests in `cmd`.

Every knob test is behavioral — asserting over observable `Level`/`Score`/`Signals`
rather than reproducing formulas or touching unexported methods — and every
monotonicity assertion carries a positive control, so none can pass by both sides
being empty. Key contracts are mutation-verified: reverting either `slope_k` read site
alone, or swapping the order of the scale-and-record, makes the suite fail.

**INV-6 measured against a binary built from unmodified `upstream/main`** — stdout
byte-identical across 6 configurations (no detectors; each detector singly;
`--detectors all`; a comma-list; multiple seeds). Also 3/3 identical across repeated
runs *with* a tuned knob under `--detectors all`.

**End-to-end, the knob does what it claims:**

```
--detectors composite                             → BACKLOGGED
--detectors composite  (sensitivity: 50.0)         → STABLE
```

and for `slope_k`, per-event verdicts over the same trace:

```
slope_k=0.001  → BACKLOGGED=0    OVERLOADED=562
slope_k=1000.0 → BACKLOGGED=190  OVERLOADED=372
```

## Scope: why `peak-rate` is not here

Issue #1614 proposed a fourth detector tracking `peak_backlog / completions`. Design
review (six parallel perspectives) found two defects I reproduced independently:

1. **The all-time peak never forgets.** A 600-request burst followed by 6,000 requests
   at a healthy 0.5× load still fires after 6,600 healthy completions. Since the
   headline label is a last-window plurality, this corrupts the primary output.
2. **At ρ = 1 the verdict is decided by run length.** `R ≈ 1.77/√C` (measured: `R·√C`
   flat at 1.73–1.84 over a 32× range in `C`), so the exactly-critical server
   false-alarms on 72% of runs at the proposed gate.

Both have one root cause, so either fix needs a trailing window — and both windowed
forms fail: a windowed max saturates (0.998 at every ρ > 1), while windowed backlog
drift reduces *algebraically* to `dA/W − 1`, a rate deficit `composite` already
computes, because `backlog = arrivals − completions` by definition. The peak cancels.

Filed as **#1618** with the measurements and the requirements a successor must meet.
Issue #1614 itself recommends landing Change 1 first and alone.

- **#1615** — `blis replay` completes far fewer requests than the run it replays
  (INV-13). Reproduces on unmodified `upstream/main` with no detector flags: 200 → 130
  at rate 20, 200 → 29 at rate 60.
- **#1616** — `CompositeDetector.Detect()` re-copies and re-sorts all completions every
  event: O(n²) time *and* allocated bytes (6.5 GB at n=10,000). The sort is provably
  redundant, since `buildSortedEvents` already emits completions in timestamp order.
- **#1617** — `saturation.Event.InputTokens`/`OutputTokens` are declared but never
  populated, so a future token-aware detector would silently read zeros.
- **#1618** — the `peak-rate` redesign above.

## Notes for reviewers

Two accepted behaviours worth knowing, both now documented:

- **`slope_k <= 1` yields a two-level detector.** The BACKLOGGED band is
  `noiseFloor < slope <= slope_k×noiseFloor`, unsatisfiable at `slope_k <= 1`. That is
  a legitimate "maximally severe" setting, but a calibration sweep crossing 1 compares
  detectors of different shape.
- **Multiplicative knobs are rejected below `1e-6`** — see the subnormal bug above.

Read in this order: `config.go` (the ownership table), `composite.go` (where the
multiplier is applied relative to the recorded signal), `backlog_drift.go` (both read
sites), then the contract tests.

The `backlog_drift_config.go` struct-comment realignment is gofmt-mandated by adding a
field to that block — reverting it makes the file non-gofmt-clean.

`--saturation-report` now carries a `slope_k` diagnostic, matching the convention
`threshold` already follows (it reports its own `threshold`) and what the extension
guide asks for. Verdicts and stdout are unchanged.

Fixes #1614


sriumcp:
Superseded by #1620, which contains every commit from this branch plus the `peak-rate` detector, under a single consistent description.

Combining them because the two halves turned out to be one argument rather than two: the knobs exist so detectors can be compared at a matched false-alarm rate, and `peak-rate` is the detector that comparison was for. Splitting them made the second PR's motivation read as an assertion instead of a result.

Closing in favour of #1620. No work is lost — `git branch --contains` confirms this branch is an ancestor.

PR #1620, opened 2026-08-25
feat(saturation): give every post-hoc detector a calibration knob, and add peak-rate

### LINKED ISSUE #1614: Post-hoc saturation: give composite a sensitivity knob, and add the peak-rate detector

sriumcp:
Fixes #1614

## The problem

BLIS has three post-hoc saturation detectors. Two of them cannot be tuned at all.

That is not a cosmetic gap. Detector verdicts are only comparable when every detector
has first been calibrated to the same false-alarm rate — otherwise you are comparing a
strict detector against a lenient one and learning nothing. A detector with no dial
cannot be moved onto that common rate, so it cannot be compared. It can only be
disqualified.

Discussion #1596 is exactly that outcome: a metamorphic-test suite had to disqualify
`composite` because it fired on 17–20 of 20 healthy runs with *"no knob to turn it
down"*, and it closed by predicting that `composite` *"would likely become the best if
it were given a tunable sensitivity knob."*

This PR gives every detector a dial, and then adds a fourth detector whose signal is a
different kind of thing altogether.

## Part 1 — a dial for every detector

| Detector | Dial before | Dial now |
|---|---|---|
| `threshold` | `threshold_ms` | unchanged |
| `composite` | **none** | `sensitivity` |
| `backlog-drift` | `slope_k` **hardcoded to 3.0** | `slope_k` |

`composite` compared its score against a fixed `1/√arrivals` noise floor;
`sensitivity` multiplies that floor. `backlog-drift`'s band multiplier was a package
constant whose own comment conceded it was *"a tunable heuristic constant, NOT an
empirically calibrated value."* Both default to today's behaviour, so absent config the
output is byte-identical.

The prediction from #1596 holds: with the knob, `composite` calibrates to a 0%
false-alarm rate and passes.

Two details are load-bearing, and both are pinned by tests:

- `NewBacklogDriftConfig` returns a fresh struct literal, so it **zero-fills** any new
  field. A literal `slope_k` of 0 makes `slope ≤ 0·noiseFloor` false for *every*
  positive slope, banding every rising trace OVERLOADED. An accessor with a documented
  fallback defuses this for all construction paths, leaving the 9-parameter arity and
  its 14 call sites untouched.
- The multiplier is read in **two** places — the band switch and the score denominator.
  It is hoisted into one local so they cannot diverge, and the test is mutation-verified:
  reverting *either* site alone makes it fail.

Along the way, two hand-written functions deciding which config block belongs to which
detector are replaced by one table both derive from. They had already drifted: the bank
path had no `composite` case, so the moment `composite` owned a block, a `composite:`
block supplied to a selection omitting it would have been **silently dropped**.

## Part 2 — `peak-rate`, a detector that estimates nothing about the server

Backlog in a queue is a random walk reflected at zero (Lindley:
`W ← max(0, W + service − interarrival)`). The reflection is not incidental — it is what
makes the load regimes distinguishable, and it licenses a statistic that needs no
capacity estimate and no latency target:

```
R_t = Peak_t / t          (running max of in-flight requests over elapsed time)

  positive drift (ρ > 1)  →  R_t → a positive CONSTANT
  zero drift     (ρ = 1)  →  R_t → 0  as 1/√t     (Peak grows as √t)
  negative drift (ρ < 1)  →  R_t → 0  as 1/t      (Peak is a finite random variable)
```

So an overloaded server **holds** `R_t` while a healthy one lets it decay. The detector
fires when `R_t` holds above a threshold. That is the whole idea.

### Why this is worth having next to the other three

**`threshold` needs a number nobody knows.** It compares mean end-to-end latency against
an absolute millisecond target that must be re-derived for every model and every GPU.
`peak-rate` reads a shape instead.

**`backlog-drift` is weakest exactly where it matters most, and the reason is
structural.** It fits a *straight line* to backlog growth. But at ρ ≈ 1 backlog grows
like `√t` — so a linear fit tends toward **zero slope**, and the detector reports STABLE
at precisely the criticality you most want flagged. That is the worst possible direction
for a detector to fail. `R_t` has no degeneracy at the boundary. The two detectors now
sit adjacent in the roster, and the distinction is visible from their names.

### The knobs

```yaml
peak_rate:
  threshold: 0.5           # primary false-alarm dial; larger fires less
  min_observations: 20     # hold the verdict until enough EVENTS have been seen
  warmup_us: 0             # hold the verdict until this much TIME has elapsed
  consecutive_k: 3         # successive breaches before firing (anti-flapping)
  overload_multiple: 3.0   # OVERLOADED above this multiple of threshold (≥ 1)
```

**`min_observations` and `warmup_us` are not redundant, and the reason is
dimensional.** `R_t`'s numerator is counted in events; its denominator is measured in
seconds. So a dense burst satisfies an event gate while elapsed time is still
negligible — 300 arrivals inside 3 ms gives `R_t > 100000` and an OVERLOADED verdict
that *no* observation count can suppress, because the events really are there. Only a
time gate can. The converse also holds: a time gate does not help a slow trickle that
has run long enough to matter but produced too few samples. Default `warmup_us: 0`
leaves the gate off.

## The experiments

Every number below is measured, and each measurement answers a question that could have
gone the other way.

### 1. Where is the capacity cliff? (ground truth, established first)

You cannot score a saturation detector without knowing which loads *are* saturated. The
test is backlog divergence: at sub-capacity load, mean latency is independent of run
length; above capacity it grows without bound.

| rate | mean E2E at n=1500 | at n=6000 | growth | verdict |
|---|---|---|---|---|
| 10 | 11362 ms | 11297 ms | −1% | sub |
| 14 | 15640 ms | 16344 ms | +4% | sub |
| **16** | — | — | **+79%** | **super** |
| 18 | — | — | +132% | super |

The cliff is ~15 rps. This step mattered: an earlier ladder assumed 20 rps and made the
detector look like it was false-alarming at 0.9× when it was in fact correctly firing at
1.2×.

### 2. Does the detector separate the regimes? (n=6000, 5 seeds)

| load | 0.33× | 0.53× | 0.67× | 0.80× | 0.93× | 1.07× | 1.20× | 1.47× | 2.00× |
|---|---|---|---|---|---|---|---|---|---|
| verdict | STABLE | STABLE | STABLE | STABLE | **BACKLOGGED** | OVERLOADED | OVERLOADED | OVERLOADED | OVERLOADED |

A clean step at the cliff, **identical across all 5 seeds** at every rung. BACKLOGGED one
rung *before* the cliff is early warning, not error — it matches the lead time the
selection campaign recorded.

### 3. Is the observation window part of the algorithm? (yes — and this cost an experiment)

The result above is asymptotic; every real run is finite. So how well `R_t` separates the
regimes depends on how long you watch:

| requests per run | separation |
|---|---|
| 500 | 2.3× |
| 2,000 | 4.7× |
| 8,000 | **14.6×** |

The mechanism is exactly the one the theory predicts, visible in the peak trajectory: at
0.5× load the peak goes 66 → 71 over the run (it has levelled off, so `R_t` decays 13×),
while at 1.5× it goes 329 → **2743** (still climbing, so `R_t` barely moves). A short run
cannot tell them apart because the *healthy* server's peak has not levelled off yet.

This produced a false negative in our own work — early runs at n=500 said the idea did
not work. It did; we were watching too briefly. So `min_observations` is part of the
algorithm, not input validation, and the code says so.

### 4. Is the threshold a real ROC dial? (yes, and seed-stable)

| threshold | 0.80× load (healthy) | 1.47× load (overloaded) |
|---|---|---|
| 0.5 | BACKLOGGED, BACKLOGGED, BACKLOGGED | OVERLOADED ×3 |
| 1.0 | **STABLE ×3** | OVERLOADED ×3 |

Raising the dial silences the healthy runs on every seed while the overloaded runs still
fire. That is what makes it usable for false-alarm calibration.

### 5. Two properties we measured and then documented rather than hid

**`min_observations` does not move the stdout headline.** A gate of 7000 silences 7001 of
8000 per-event verdicts — and the headline does not change, because the headline is a
trailing-window plurality vote and that window is always past the gate. Use `threshold`
for the headline rate; use `min_observations` to keep the *trace* quiet through a known
ramp-up.

**Recovery after a transient takes `peak / threshold` completions.** The numerator is an
all-time high-water mark, so a burst that drives the peak to `P` keeps the detector firing
until roughly `P / threshold` further requests complete. Predicted 3483, observed 3484 —
a closed form, not a mystery. The honest consequence: `peak-rate` answers *"did this run
saturate?"*, not *"is it saturated right now?"* For a post-hoc verdict on a finished run
that is the right question, but it should be stated, not discovered.

### 6. Two silent-failure bugs, found by mutating the code against its own tests

Neither was found by reading. Both are the worst class of bug — a detector that reports
health.

**A verdict of STABLE, forever.** The first-timestamp field was set once and never
lowered, so if the first event carried the largest timestamp the elapsed span stayed 0,
the readiness gate was permanently false, and the detector reported STABLE on an
unboundedly growing backlog. Reproduced: **201 in-flight requests, zero completions,
verdict STABLE.** No live path hit it, because the event builder sorts — but `Observe` is
a public interface method. Now tracks minimum and maximum symmetrically.

**A verdict outliving its evidence.** An unrecognized event type advanced the elapsed
span *before* the early return, shrinking the statistic while leaving the breach streak
frozen. Reproduced: 1000 such events drove `R_t` from 10.18 to 0.092 — far below the 0.5
threshold — while the level stayed BACKLOGGED. Now returns before touching any state.

### 7. The test that was green while the detector was broken

This is the finding worth generalizing.

Replacing the running maximum with the **instantaneous backlog** — which destroys the
entire statistic and every asymptotic argument above — left the whole suite passing. Both
fixtures happened to keep the peak within 1 of the current backlog, so no assertion could
distinguish a high-water mark from a gauge.

A second instance, same shape: a test asserting that `effectiveSlopeK()` returned a
finite positive value passed while a subnormal `slope_k` silently produced OVERLOADED with
Score 0. It checked the multiplier **in isolation, rather than the product it participates
in.** Rewritten to assert on the verdict, it failed immediately.

Both were fixed by asking a different question of the test: not *"is this quantity
sane?"* but *"is the behaviour the user sees correct?"*

**Mutation score: 8 of 9.** The survivor is provably redundant — `statistic()`
independently returns 0 for a zero span, and 0 cannot exceed a threshold floored at
`1e-6`. It is kept as defence in depth, and the comment says so rather than leaving a
reader puzzled by an untested line.

## On the tests

Behavioral throughout: assertions over observable `Level` / `Score` / `Signals` on traffic
of a known regime — never over internal fields, and never reproducing the implementation's
own formula. Every monotonicity assertion carries a positive control, so none can pass
with both sides empty.

Contracts worth naming:

- **Boundary semantics of every knob.** `threshold` and `overload_multiple` are strict;
  `min_observations` and `consecutive_k` are inclusive. Ordering tests cannot see an
  off-by-one here, because both sides of the comparison shift together — so each boundary
  gets its own assertion.
- **Separation widens with the horizon**, asserted as a trend so the law is tested without
  pinning one apparatus's numbers.
- **Metamorphic**: rescaling every timestamp preserves the *ordering* of two streams'
  statistics — asserting the ordering rather than the value, which would be formula
  reproduction.
- **`Detect()` is a pure query.** Interleaving extra `Detect` calls cannot change the
  verdict sequence — the leg that actually fails if state advances inside `Detect`.
- **Peer isolation.** Tuning one detector leaves the other three detectors' records
  identical, with a vacuity guard proving the tuned one changed.
- **Exhaustive ownership truth table** over every (config block × selected detector) pair
  on both code paths, with a guard asserting the table covers the whole roster so it
  cannot go stale.

`sim/saturation`: **89 → 217** tests, plus 10 end-to-end CLI tests.

## Verification

```
go build ./...            exit 0
go test ./... -count=1     13/13 packages ok
golangci-lint run ./...    0 issues
```

**INV-6** — measured against a binary built from unmodified `upstream/main`: stdout is
byte-identical across 6 configurations (no detectors; each detector singly;
`--detectors all`; a comma-list; multiple seeds). Repeated runs *with* a tuned knob are
identical.

**INV-13** — run and replay saturation reports are byte-identical with a tuned knob,
asserted at the CLI boundary over the *whole roster* plus `all` (so a future detector
cannot escape it) and mutation-verified: a 1 µs shift in the replay-side extractor
fails the test.

**Performance** — `peak-rate` holds O(1) state (four scalars, no per-event retention) and
is the cheapest detector in the roster by 20×: 5 ms / 6.9 MB at 10k requests, against
`threshold` at 103 ms and `composite` at 1320 ms / 6.5 GB.

## Also fixed

`scripts/find-saturation.sh` runs `--detectors all` by default, so it was collecting a
`peak-rate` verdict and silently dropping it from both the CSV and the console line.
Roster-cardinality assertions in `cmd` and the bank tests now derive from
`AllDetectorNames()`, so a fifth detector cannot silently weaken them to subset checks.

- **#1615** — `blis replay` completes far fewer requests than the run it replays (INV-13).
  Reproduces on unmodified `upstream/main` with no detector flags: 200 → 130 at rate 20,
  200 → 29 at rate 60.
- **#1616** — `CompositeDetector.Detect()` re-copies and re-sorts all completions on every
  event: O(n²) in time *and* allocated bytes (6.5 GB at n=10,000). The sort is provably
  redundant, since the event builder already emits completions in timestamp order.
- **#1617** — `saturation.Event.InputTokens` / `OutputTokens` are declared but never
  populated, so a future token-aware detector would silently read zeros.
- **#1618** — tracking issue for this detector's design history, including the two
  limitations above (recovery time; the near-critical resolution floor) and the
  measurements behind them.


namasl:

## qa-review — PR #1620: ⛔ BLOCK
_feat(saturation): give every post-hoc detector a calibration knob, and add peak-rate · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ❌ FLAW_FOUND | Does this PR fully implement the issue it closes, without unnecessary additions? | No. The composite sensitivity and peak-rate detector are largely implemented, but the issue explicitly makes `warmup_us` load-bearing for peak-rate and requires five optional fields. The implementation has only four fields and computes elapsed time from the first accepted event, so it cannot discard a configured initial time transient. There are also unrelated gofmt-only edits in several `cmd/` files. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | No. The major user-facing documentation was updated for the implemented four-field peak-rate configuration and the new detector roster, but it does not document the issue-required `warmup_us` because that feature is absent. It also contains inaccurate statements about `min_observations` never affecting the headline and about the state size. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | Yes. There are stale or inaccurate code comments. In particular, `BacklogDriftConfig` still says the streaming detector reads only `WindowSize`, even though it now also reads `SlopeK`. Peak-rate's state comment says “Four scalars” despite tracking peak, in-flight count, observations, two timestamps, a boolean, and a breach counter. |
| G1 | correctness | ✅ CONFIDENT | In `peak_rate.go`, how are zero or subnormal elapsed timestamp spans handled when calculating $R_t = Peak_t / t$, especially when events arrive with identical timestamps or before the readiness gate `min_observations` is met? | Elapsed time is derived from an int64 microsecond timestamp difference, so the smallest positive span is 1 microsecond (1e-6 seconds), not a floating-point subnormal. If all accepted events have the same timestamp, `elapsedSec()` returns zero, `statistic()` returns zero without division, `ready()` remains false, the breach streak is reset, and `Detect()` returns STABLE with zero score/confidence. Before `min_observations`, the same STABLE gate applies even when elapsed time is positive. |
| G2 | correctness | ✅ CONFIDENT | In `BacklogDriftConfig.effectiveSlopeK()`, what happens if `SlopeK` is set to a subnormal positive float below `minCalibrationKnob`, and does `effectiveSlopeK()` fall back to the default `3.0` while YAML validation cleanly rejects the value? | Yes. An in-process `BacklogDriftConfig` whose `SlopeK` is positive but below `1e-6` is treated as unset and `effectiveSlopeK()` returns the default 3.0. The YAML resolution path checks the supplied pointer before constructing the detector and returns an error naming `backlog_drift.slope_k`, so user input is not silently defaulted. |
| G3 | invariants | ✅ CONFIDENT | How does the `peak-rate` detector maintain strict determinism (INV-6) when multiple events share identical microsecond timestamps, and does event processing order affect the running maximum calculation across runs? | The CLI pipeline deterministically sorts reconstructed events by timestamp, then event type, then request ID. At equal timestamps, arrivals sort before completions, so the order-dependent high-water mark has a fixed interpretation and cannot vary across identical runs. Peak-rate itself uses integer counters and min/max timestamps with no map iteration or randomness. Direct callers can choose a different event order and thereby change the peak, but identical ordered input remains deterministic. |
| G4 | invariants | ❌ FLAW_FOUND | Does running `blis replay` with `--detectors peak-rate` produce byte-identical JSON saturation traces and stdout final labels compared to `blis run` on the exact same workload (INV-13)? | The implementation routes run and replay through the same saturation resolver, completed-request adapter, tracer, event sorter, reducer, and JSON writer, so parity is structurally expected. However, the acceptance criterion explicitly requires a run-to-replay byte-identity test for the new detector, and no such peak-rate CLI test was added. Existing saturation E2E parity coverage uses `composite`, not `peak-rate`. |
| G5 | rules | ✅ CONFIDENT | How does `CompositeBlock.Sensitivity` in `sim/saturation/config.go` comply with Rule R9 (using pointer types for optional YAML scalar fields to distinguish zero from omitted) and Rule R1 (rejecting non-positive/NaN values without silent fallback)? | `Sensitivity` is a `*float64`, so omission leaves it nil and selects the historical default while an explicit zero remains distinguishable and is rejected. The builder rejects NaN, infinities, zero, negatives, and values below `1e-6` with an error naming `composite.sensitivity`; only direct in-process constructor misuse receives a defensive default. |
| G6 | rules | ✅ CONFIDENT | In `sim/saturation/config.go`, how does the unified `blockOwners()` table enforce strict block ownership rules (R1) for both single-detector selection and multi-detector bank selection without dropping unselected detector blocks? | `blockOwners()` maps every YAML block to exactly one detector and a presence predicate. The single path iterates it and rejects every present block whose owner differs from the selected detector. The bank builds a selected-name set and rejects every present block whose owner is absent. Only after that check does it construct detectors, so foreign blocks cannot be silently ignored. |
| G7 | parity | ✅ CONFIDENT | Are all four detector knobs (`composite.sensitivity`, `threshold.threshold_ms`, `backlog_drift.slope_k`, `peak_rate.*`) parsed, validated, and applied identically across `blis run`, `blis replay`, and `blis observe`? | Yes. All three commands register the same flags and call the same `resolveSaturation()`, which loads strict YAML once, performs ownership and value validation through `BuildDetector` or `NewBank`, and then uses the same `saturationTracer.run`. Thus all implemented knobs are parsed, validated, and applied through one path; only the source of `[]RequestMetrics` differs. |
| G8 | parity | ✅ CONFIDENT | When `peak_rate.min_observations` holds back early verdicts during warmup, how does this interact with the trailing-window plurality vote (`--saturation-final-window`) in `ReduceAll` to ensure consistent summary output between single-detector and bank runs? | Before readiness, peak-rate emits ordinary STABLE trace records. `ReduceAll` groups records by detector and applies the same timestamp-windowed plurality independently to each group. Since single and bank paths feed peak-rate the same sorted events and record the same per-event results, their peak-rate summary is consistent. Early gated STABLE records participate if they fall within the selected trailing window; therefore `min_observations` can affect the summary for short runs, large gates, or long final windows. |
| G9 | tests | ✅ CONFIDENT | How do the new saturation tests verify the mathematical monotonicity properties of `composite.sensitivity` and `peak_rate.threshold` (larger values fire strictly non-increasingly) without relying on fragile hardcoded golden trace values? | The tests use deterministic qualitative streams and compare whole per-event level sequences across increasing knob ladders. They assert severity never increases as the suppressing knob rises and include positive controls requiring the low setting to fire and the high setting to change or suppress verdicts, avoiding vacuous passes. They do not pin a serialized golden trace. |
| G10 | tests | ✅ CONFIDENT | How was mutation testing applied to verify that hoisting `slopeK` in `BacklogDriftDetector.Detect()` prevents silent divergence between the `Level` boundary switch and the `Score` calculation denominator? | `Detect()` resolves `slopeK` once, records it in Signals, and uses that same local value for both the BACKLOGGED/OVERLOADED boundary and the score denominator. This prevents one calculation from accidentally using the old constant while the other uses configuration. At exact equality, the existing `<=` contract permits BACKLOGGED with score 1.0; above the boundary, OVERLOADED has the capped score. |
| G11 | rationale | ✅ CONFIDENT | Why was `peak-rate` designed using an all-time high-water mark over total elapsed time ($R_t = Peak_t / t$) rather than a sliding window peak or instantaneous backlog gauge, and what are the trade-offs in recovery latency after transient bursts? | The all-time peak divided by total elapsed time directly represents the asymptotic reflected-random-walk distinction: bounded backlog leaves a fixed numerator over growing time, while positive drift grows the numerator with time. An instantaneous gauge would only say whether the queue is currently large, and a sliding peak would forget earlier saturation and change the detector into a local/current-state detector. The trade-off is intentionally slow recovery: after a burst the remembered peak remains fixed and the elapsed denominator must grow until `peak/elapsed <= threshold`, so recovery is roughly `peak/threshold` seconds from the time origin, subject to debounce. |
| G12 | rationale | ✅ CONFIDENT | Why was replacing separate hand-written block ownership checks with the single `blockOwners()` table in `sim/saturation/config.go` chosen, and how does it prevent past configuration drift bugs? | The table was chosen to make block-to-owner mapping a single source of truth. Previously the single and bank checks separately enumerated blocks, allowing one path to omit a newly introduced block and silently drop its configuration. Both checks now iterate the same rows, and exhaustive tests compare their decisions for every block and detector. |
| G13 | scope | ❌ FLAW_FOUND | Linked issue #1614 requested a knob for `composite` and the new `peak-rate` detector; why did this PR also extend `backlog-drift` to make `slope_k` configurable, and are there any backward-compatibility impacts for existing configs? | Making `slope_k` configurable is within the linked issue's Change 1: backlog-drift also lacked an exposed calibration dial. Its default effective value remains 3.0, so classification is backward-compatible. However, existing backlog-drift JSON traces are not byte-identical when the block is absent because the PR unconditionally adds a new `slope_k` entry to every result's Signals map, violating the stated absent-config byte-identity acceptance criterion. |
| G14 | scope | ✅ CONFIDENT | This PR updates `scripts/find-saturation.sh` to include `peak_rate_verdict` in the CSV output header; does adding this 13th column impact existing downstream benchmark parsing tools expecting 12 columns? | Yes, positional external consumers hard-coded for 12 columns will need updating because the script now always emits a 13th `peak_rate_verdict` column, including `n/a` when peak-rate is not selected. No in-repository parser of this CSV was found, and the accompanying README documents the new schema, so there is no demonstrated in-tree breakage. |
| G15 | errors | ✅ CONFIDENT | In `PeakRateDetector.Observe()`, how are unexpected or unhandled event types handled—does an unknown event type trigger an error or return early without modifying internal elapsed time or breach streak state? | An unknown event type returns immediately without an error and before changing timestamps, observation count, in-flight count, peak, or breach streak. This is deliberate defensive handling because only Arrival and Completion are meaningful to the detector. |
| G16 | errors | ✅ CONFIDENT | When invalid or misspelled keys are provided inside a `peak_rate:` YAML block in `--saturation-config`, does strict YAML parsing trigger a clear user error naming the key, or can it silently fall back to default struct values? | Misspelled or unknown keys are rejected during YAML decoding rather than ignored. `KnownFields(true)` applies recursively to `PeakRateBlock`, and the resulting parse error is wrapped with the config path and YAML's offending field information. Valid field names with invalid values are then rejected separately with errors naming the fully qualified field. |

### Items to fix
- **F1 · FLAW_FOUND** — No. The composite sensitivity and peak-rate detector are largely implemented, but the issue explicitly makes `warmup_us` load-bearing for peak-rate and requires five optional fields. The implementation has only four fields and computes elapsed time from the first accepted event, so it cannot discard a configured initial time transient. There are also unrelated gofmt-only edits in several `cmd/` files.
  _sim/saturation/config.go:41-62 defines only threshold, min_observations, consecutive_k, and overload_multiple; sim/saturation/peak_rate.go:163-178 anchors elapsed time to the minimum timestamp seen, with no warmup cutoff; sim/saturation/peak_rate.go:107-112 defines no warmup default._
- **G4 · FLAW_FOUND** — The implementation routes run and replay through the same saturation resolver, completed-request adapter, tracer, event sorter, reducer, and JSON writer, so parity is structurally expected. However, the acceptance criterion explicitly requires a run-to-replay byte-identity test for the new detector, and no such peak-rate CLI test was added. Existing saturation E2E parity coverage uses `composite`, not `peak-rate`.
  _cmd/root.go:2215-2218 and cmd/root.go:2308 invoke the shared tracer; cmd/replay.go:688-696 and cmd/replay.go:765 do likewise; cmd/saturation.go:105-142 is the shared execution/reduction/report path; sim/saturation/e2e_test.go:16-80 performs extractor byte-parity using `BuildDetector("composite", ...)`; no peak-rate run/replay parity test exists in cmd or sim/saturation._
- **G13 · FLAW_FOUND** — Making `slope_k` configurable is within the linked issue's Change 1: backlog-drift also lacked an exposed calibration dial. Its default effective value remains 3.0, so classification is backward-compatible. However, existing backlog-drift JSON traces are not byte-identical when the block is absent because the PR unconditionally adds a new `slope_k` entry to every result's Signals map, violating the stated absent-config byte-identity acceptance criterion.
  _sim/saturation/backlog_drift.go:170-171 unconditionally adds `signals["slope_k"]`; sim/saturation/backlog_drift_config.go:60-68 preserves the default value 3.0; sim/saturation/backlog_drift_slopek_test.go explicitly describes the added signal as changing the report while only asserting stdout is unchanged._

### Important to consider
- **F2** — These documentation inaccuracies are non-blocking by themselves; the missing warmup behavior is covered by F1.
  _CLAUDE.md documents peak_rate without warmup_us; docs/contributing/saturation-analyzer-extension.md:104-108 lists the same four fields; sim/saturation/config.go:51-55 claims the trailing window is always past the gate, although a short run or a min_observations value exceeding the run length leaves every record STABLE; sim/saturation/peak_rate.go:58-59 says four scalars while the struct at lines 68-78 has more state fields._
- **F3** — Stale comments are maintenance concerns, not runtime defects.
  _sim/saturation/backlog_drift_config.go:15-17 versus sim/saturation/backlog_drift.go:167-171; sim/saturation/peak_rate.go:56-78._
- **G7** — The issue-requested peak-rate `warmup_us` knob is not implemented, as noted in F1.
  _cmd/root.go:2215-2218 and 2308; cmd/replay.go:688-696 and 765; cmd/observe_cmd.go:609-615 and 747; cmd/saturation.go:29-42, 145-237; sim/saturation/config.go:100-118 and 189-220._
- **G8** — The comment claiming `min_observations` does not move the headline because the trailing window is always past the gate is too strong: sim/saturation/config.go:51-55.
  _sim/saturation/peak_rate.go:197-225; sim/saturation/reduce.go:28-72 performs the windowed vote; sim/saturation/reduce.go:75-94 groups independently; sim/saturation/replay.go:43-52 and sim/saturation/bank.go:127-151 use the same sorted event stream._
- **G11** — Documentation that describes `peak/threshold` as a number of completions is dimensionally imprecise because threshold is backlog per second; the implementation decays with elapsed time.
  _sim/saturation/peak_rate.go:9-23 states the statistical rationale; sim/saturation/peak_rate.go:193-199 maintains the all-time maximum; sim/saturation/peak_rate_test.go:623-670 tests bounded recovery and threshold dependence; sim/saturation/peak_rate_test.go:672-730 distinguishes the high-water mark from instantaneous backlog._
- **G14** — External positional CSV consumers should be treated as requiring a schema migration.
  _scripts/find-saturation.sh:51 defines the 13-column header; scripts/find-saturation.sh:104-114 extracts and appends the value; scripts/README.md:75-91 documents all columns and the `n/a` behavior._

sriumcp:
Thanks — three of these were real, and the two "important to consider" doc items were real too. All fixed in `b940a09a`, on a branch now rebased onto latest `main` (`cee7e179`).

**G13 — valid, and the sharpest of the three.** Confirmed by diffing against a binary built from unmodified `main`: an absent `backlog_drift:` block produced a report that was *not* byte-identical, because `signals["slope_k"]` was emitted unconditionally and `Signals` is serialized into `--saturation-report`. A default report differed from a pre-change one for a diagnostic that only restated the documented default.

The key is now emitted only for an explicit choice. That required a second fix the finding implies but does not name: `resolveBacklogDriftConfig` was seeding `SlopeK` with the *effective* default, so the resolved config could no longer tell "the operator chose 3.0" from "nobody chose anything". It now seeds zero and lets `effectiveSlopeK()` supply the default at read time. Verified: absent config is byte-identical again, and a tuned run still carries the diagnostic.

**G4 — valid.** Parity was structurally expected but never asserted for the new detector. Added a CLI-level `run` → `replay` byte-identity test over the **whole roster plus `all`**, with every knob configured, where the two legs resolve their requests exactly as the real commands do (simulator metrics vs a TraceV2 round trip). Mutation-verified: a 1 µs shift in the replay-side extractor fails it. The hermetic extractor-parity test in `e2e_test.go` is likewise generalized over `AllDetectorNames()`, so a future detector cannot escape either guarantee.

**F1 — valid, both parts.**

*`warmup_us` is now implemented*, and testing it settled the question of whether it was redundant with `min_observations`. It is not, and the reason is dimensional: `R_t`'s numerator is counted in events while its denominator is measured in seconds, so a dense burst satisfies an event gate while elapsed time is still negligible. Measured: **300 arrivals inside 3 ms yields `R_t > 100000` and an OVERLOADED verdict that no observation count can suppress** — the events really are there. Only a time gate can. Default `0` keeps the gate off, matching the campaign-validated level (`WARM=0`) and preserving absent-config byte-identity. Two behavioral tests pin the asymmetry and the fact that the gate only *delays* a verdict rather than blinding the detector to real saturation; removing the gate from `ready()` fails them.

*The gofmt churn was worse than reported.* Six files, not "several" — `git diff -w` was empty for all of them. The diff now contains no whitespace-only file at all.

**F2 / F3 / G8 / G11 — all four were real, so all four are fixed** rather than filed:

- "Four scalars" understated the state; now enumerated.
- `BacklogDriftConfig` still claimed the streaming detector reads only `WindowSize`.
- The `min_observations` claim was **too strong**, exactly as G8 says. It does not move the headline on a run long enough that the reducer's trailing window lies past the gate — the common case, but not a guarantee. Short runs, very large gates, and long `--saturation-final-window` values are now called out in both the code comment and the guide.
- "`peak/threshold` completions" was **dimensionally wrong**, as G11 notes: `threshold` is backlog per *second*, so recovery is ~`peak/threshold` seconds of elapsed time. Corrected in three places.

**G14 — acknowledged, no change.** The 13th CSV column is a schema change for any positional external consumer. No in-tree parser exists and `scripts/README.md` documents the new schema, so I've left it rather than pretending the column is optional.

Verification after the rebase: build clean, 13/13 packages, `golangci-lint` 0 issues. INV-6 re-checked against a baseline built from the *new* `main` — byte-identical at both stdout and report level across 5 configurations. The detector's step function at the capacity cliff is unchanged (STABLE through 0.80×, BACKLOGGED at 0.93×, OVERLOADED at 1.07×+).

Two notes on the review itself, offered as calibration rather than complaint: **G1 was a better answer than my own code comment** — it correctly observed that an `int64` microsecond span cannot be subnormal, only zero or ≥ 1 µs, which is more precise than what I had written. And **G11's dimensional catch is the kind of error that survives a lot of testing**, since every test that exercised recovery still passed; only the units were wrong. Both are useful signals for the two-agent experiment.

sriumcp:
@claude /blis-pr-review consider all the reviews from before and also the fixes, in addition to your usual checks.

sriumcp:
Thanks. One of the two documentation findings is valid and now fixed; the other does not reproduce. Details, since the distinction matters for calibrating this reviewer.

**✅ `--saturation-config` help omits `warmup_us` — valid, and worse than reported.**

Correct, and it is the one that matters most: the flag's own help string is what an operator reads first, before any guide. Fixed in `6959f64a`.

Auditing *every* site that enumerates the `peak_rate` knobs turned up **two more** the review did not flag:

| site | what it is |
|---|---|
| `cmd/saturation.go:38` | the flag help (reported) |
| `CLAUDE.md:320` | the run/replay/observe usage block |
| `sim/saturation/config.go:22` | the `SaturationConfig` doc comment |

All four enumeration sites now list all five knobs. Both YAML examples already showed `warmup_us`, which is why a spot check found only one gap — the omissions were in the *prose* enumerations, not the examples.

**❌ CLAUDE.md stale roster — does not reproduce.**

All three sub-items were fixed in an earlier round of this PR:

| claim | actual |
|---|---|
| single-name list omits `peak-rate` | `CLAUDE.md:561` — `` (`composite`, `threshold`, `backlog-drift`, `peak-rate`) `` |
| states "composite has no block" | `grep -n "composite has no block" CLAUDE.md` → no match |
| roster example omits `peak-rate` | `CLAUDE.md:568` — ``orders the roster canonically (`composite`, `threshold`, `backlog-drift`, `peak-rate`)`` |

The cited lines **541–548 are the Speckit section**, not the saturation section — so the review appears to have read a stale revision of the file. Worth noting for the tooling: the finding was well-formed and would have been correct against an earlier commit, which is exactly the failure mode that is hard to spot in a generated report.

**Verification after the fix:** build clean, 13/13 packages, `golangci-lint` 0 issues, no whitespace-only files in the diff. `blis run --help` now reads:

```
peak_rate: {threshold, min_observations, warmup_us, consecutive_k, overload_multiple}
```

I've kept the review's OBSERVATION about `peakRateWith` zero-filling `WarmupUs` as-is, since it reached the right conclusion: the helper deliberately varies only threshold/min-obs/consecutive-k, and the `warmup_us` tests construct their config directly for exactly that reason.

sriumcp:
@claude /blis-pr-review consider all the reviews from before and also the fixes, in addition to your usual checks.

sriumcp:
Valid finding — fixed in `054a1608`.

Confirmed by reading both bullets: three sentences describing `backlog_drift.slope_k` and `effectiveSlopeK()` were inside the `peak-rate` bullet, while the `backlog-drift` bullet — the detector that actually owns that knob — was a one-liner that never mentioned it. An operator reading the `peak-rate` description would hit references to a knob `peak-rate` does not have.

**Root cause, which is slightly different from "copy-paste" and worth recording.** The text started in the *right* place: `e8ec215c` correctly appended it to the `backlog-drift` bullet. The later `peak-rate` commit then inserted its new bullet by string-replacing on the `backlog-drift` **line** — and because that line had grown a long tail, the replacement swallowed it. Inserting a sibling by matching a neighbour's full line is fragile precisely when the neighbour is long, which is the general lesson.

**What moved and what stayed.** The `slope_k` paragraph is back on the `backlog-drift` bullet. The `peak_rate.overload_multiple` cross-reference stays in the `peak-rate` bullet, but restated on its own terms rather than as "for the same reason" as a sentence that is no longer adjacent:

> `peak_rate.overload_multiple` is rejected below 1, because below 1 the OVERLOADED boundary would sit under the firing threshold and BACKLOGGED would be unreachable.

The shared `1e-6` floor is now stated once, in the `backlog-drift` bullet, scoped explicitly to all three multiplicative knobs rather than implied for one.

**Audited the rest rather than fixing only what was reported.** Checked every detector bullet for references to another detector's knobs:

```
composite      foreign knobs mentioned: none
threshold      foreign knobs mentioned: none
backlog-drift  foreign knobs mentioned: none
peak-rate      foreign knobs mentioned: none
```

Verification: build clean, 13/13 packages, `golangci-lint` 0 issues.

Also confirming the note at the end of your review: the `M CLAUDE.md` you saw in the working tree was indeed not a real divergence — the committed `HEAD` was correct on the items from the previous round, which is why the earlier "stale roster" finding did not reproduce. This one did, and the difference is that it was checkable against `HEAD` rather than against a stale read.

namasl:

## qa-review adjudication — PR #1620: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The functional defect is fixed: `warmup_us` is now the fifth optional field, is resolved and validated, and gates readiness by elapsed time. However, the finding also covered unrelated formatting-only churn, and the current PR diff still removes only a trailing blank line from `cmd/default_config.go`. To clear the complete finding, remove that unrelated edit or provide a concrete, checkable reason it belongs in this PR. |
| G4 | FLAW_FOUND | ❌ STILL-OPEN | A byte-identity test now covers `peak-rate`, but it directly calls `resolveSaturation()` and `tracer.run()` for both legs rather than executing the actual `run` and `replay` command paths. This presents the specific acceptance-criterion ambiguity between requiring real CLI command-path coverage and accepting a shared-downstream byte-identity test. The author must pin that interpretation from the issue/maintainer, or add a test that invokes the actual run and replay paths. |
| G13 | FLAW_FOUND | ✅ RESOLVED | The resolved backlog-drift configuration now preserves whether `slope_k` was explicitly supplied, while the detector uses the effective default for classification and emits the signal only for an explicit value. Thus an absent block no longer adds the `slope_k` report entry. The relevant package tests also pass. |

### Still blocking
- **F1** — The functional defect is fixed: `warmup_us` is now the fifth optional field, is resolved and validated, and gates readiness by elapsed time. However, the finding also covered unrelated formatting-only churn, and the current PR diff still removes only a trailing blank line from `cmd/default_config.go`. To clear the complete finding, remove that unrelated edit or provide a concrete, checkable reason it belongs in this PR.
  _sim/saturation/config.go:41-70 defines all five optional peak-rate fields, including WarmupUs; sim/saturation/config.go:418-447 defaults, validates, and resolves warmup_us; sim/saturation/peak_rate.go:322-325 applies the elapsed-time gate. The base...head diff still contains a whitespace-only change to cmd/default_config.go after line 163, removing an otherwise unrelated trailing blank line._
- **G4** — A byte-identity test now covers `peak-rate`, but it directly calls `resolveSaturation()` and `tracer.run()` for both legs rather than executing the actual `run` and `replay` command paths. This presents the specific acceptance-criterion ambiguity between requiring real CLI command-path coverage and accepting a shared-downstream byte-identity test. The author must pin that interpretation from the issue/maintainer, or add a test that invokes the actual run and replay paths.
  _cmd/saturation_knobs_test.go:269-340 iterates all detector names plus `all`, but its `reportFor` helper calls resolveSaturation at line 315 and tracer.run at line 319 for both legs; it does not invoke the run or replay commands. sim/saturation/e2e_test.go:58-87 likewise compares requests through BuildDetector and ReplayOneDetector rather than the CLI command paths._

### Cleared
- **G13** (✅ RESOLVED) — The resolved backlog-drift configuration now preserves whether `slope_k` was explicitly supplied, while the detector uses the effective default for classification and emits the signal only for an explicit value. Thus an absent block no longer adds the `slope_k` report entry. The relevant package tests also pass.
  _sim/saturation/config.go:312-317 initializes slopeK to zero to represent unset, and sim/saturation/config.go:378-383,398-403 records it only when supplied. sim/saturation/backlog_drift.go:170-178 obtains the effective value but conditionally emits signals["slope_k"] only when config.SlopeK > 0. sim/saturation/backlog_drift_slopek_test.go:164-180 verifies the key is absent for an unconfigured detector. `go test ./sim/saturation` passed._

sriumcp:
Both still-open findings are addressed. `ef79076f` and `3ff50558`, branch current with `upstream/main` (`cee7e179`).

**F1 — valid, and my own audit was the reason it survived.**

`cmd/default_config.go` did still carry a trailing-blank-line deletion in a file this PR never functionally touches. Reverted.

Worth naming why two prior cleanup passes missed it: I was auditing with `git diff -w`, which reports empty for whitespace *changes* but **not** for a pure blank-line *deletion* — the line is genuinely gone, so there is nothing for `-w` to normalize. My "no whitespace-only files" claim was therefore true of the check I ran and false of the property I asserted. The audit now counts non-blank added/removed lines per file, which catches both shapes, and reports zero across the diff.

**G4 — I added the test rather than argue the interpretation.**

You're right that the acceptance criterion is ambiguous, and right that the in-process test does not settle it. Rather than ask a maintainer to pin the reading, `cmd/saturation_cli_parity_test.go` now executes the **actual commands**: `blis run --trace-output … --saturation-report …`, then `blis replay` over that exported trace, requiring byte-identical reports. Every detector plus `all`. Each leg re-execs the test binary so the real cobra tree runs, following the harness `TestSaturationStdout_FinalLabelShape` already established.

The distinction you drew is the right one, and sharper than I had appreciated. The in-process test proves the shared downstream pipeline is deterministic *given equivalent inputs*; it cannot prove the two commands **assemble** equivalent inputs, and everything upstream of the shared tracer differs between them — flag registration, trace export, trace parsing, each side's request construction.

**Mutation-verified, and the search corrected a wrong belief of mine.** I had claimed in an earlier reply that the in-process test was validated by perturbing `TraceRecordsToRequestMetrics`. Retrying that against the CLI test showed it changes nothing, and the reason is that **`blis replay` re-simulates and reads `CompletedRequestMetrics` — the same extractor as `run`.** `TraceRecordsToRequestMetrics` is the *observe* path. Perturbing the shared extractor also cannot fail a parity test, since both legs shift equally.

The mutation that *does* fail it is a 1 ms shift in `sim/workload/replay.go`'s replay-only arrival construction, and it fails **every** detector — so the test exercises the real run-vs-replay divergence surface. I've corrected the claim in the commit message rather than leave the earlier one standing.

Verification: build clean, 13/13 packages, `golangci-lint` 0 issues, no cosmetic-only files by either measure.

namasl:

## qa-review adjudication — PR #1620: ✅ PASS

|----|----|----|----|
| F1 | FLAW_FOUND | ✅ RESOLVED | The peak-rate configuration now exposes all five optional fields, including a validated and load-bearing `warmup_us` time gate. The default remains zero, and the prior unrelated whitespace-only command-file churn is absent from the current PR diff; the changed `cmd/` files now contain substantive saturation implementation or test changes. |
| G4 | FLAW_FOUND | ✅ RESOLVED | A real CLI-level run-to-replay byte-identity test now executes both Cobra command paths, exports and replays the trace, and includes `peak-rate` as well as the complete detector roster and `all`. |
| G13 | FLAW_FOUND | ✅ RESOLVED | The resolver now preserves whether `slope_k` was explicitly supplied, and the detector emits the `slope_k` signal only for an explicit nonzero configuration. Consequently, an absent block no longer adds the report key that caused the byte-level regression. |

### Still blocking

### Cleared
- **F1** (✅ RESOLVED) — The peak-rate configuration now exposes all five optional fields, including a validated and load-bearing `warmup_us` time gate. The default remains zero, and the prior unrelated whitespace-only command-file churn is absent from the current PR diff; the changed `cmd/` files now contain substantive saturation implementation or test changes.
  _sim/saturation/config.go:41-70 defines `WarmupUs *int64` alongside the other four optional fields; sim/saturation/config.go:442-447 validates and resolves it; sim/saturation/peak_rate.go:322-325 requires the elapsed span to meet `WarmupUs`; sim/saturation/peak_rate.go:328-334 computes that span from the minimum and maximum accepted timestamps. `go test ./sim/saturation` and `go test ./cmd` both pass. The current `pr_diff` changed-file list contains no unrelated gofmt-only files._
- **G4** (✅ RESOLVED) — A real CLI-level run-to-replay byte-identity test now executes both Cobra command paths, exports and replays the trace, and includes `peak-rate` as well as the complete detector roster and `all`.
  _cmd/saturation_cli_parity_test.go:12-27 documents that the actual `run` and `replay` command paths are exercised; lines 40-50 execute those paths; lines 60-65 configure `peak-rate`, including `warmup_us`; lines 83-115 iterate over `AllDetectorNames()` plus `all` and compare complete report bytes. The complementary extractor test also iterates every detector at sim/saturation/e2e_test.go:79-87. `go test ./cmd` passes._
- **G13** (✅ RESOLVED) — The resolver now preserves whether `slope_k` was explicitly supplied, and the detector emits the `slope_k` signal only for an explicit nonzero configuration. Consequently, an absent block no longer adds the report key that caused the byte-level regression.
  _sim/saturation/config.go:312-317 initializes `slopeK` to zero to represent unset; lines 378-383 assign it only when `backlog_drift.slope_k` is present; lines 398-403 preserve that value in the resolved configuration. sim/saturation/backlog_drift.go:167-178 obtains the effective default for calculations but conditionally emits `signals["slope_k"]` only when `b.config.SlopeK > 0`. sim/saturation/backlog_drift_slopek_test.go:164-180 explicitly verifies that an unconfigured detector omits the key. `go test ./sim/saturation` passes._

namasl (dismissed):

sim/saturation/config.go
```diff
@@ -266,8 +392,71 @@ func resolveBacklogDriftConfig(block *BacklogDriftBlock) (BacklogDriftConfig, er
...
+	}
+	if block == nil {
+		return out, nil
+	}
+	if block.Threshold != nil {
+		v := *block.Threshold
+		if math.IsNaN(v) || math.IsInf(v, 0) || v < minCalibrationKnob {
+			return peakRateConfig{}, fmt.Errorf("saturation config: peak_rate.threshold must be a finite value >= %v, got %v", minCalibrationKnob, v)
+		}
+		out.Threshold = v
```
jgchn, line 453: The comment for `minCalibrationKnob` (and the rationale copied here) claims that a subnormal value `"underflows the score denominator to zero, decoupling Level from Score."`
This rationale applies to multiplicative knobs (like composite.sensitivity) that multiply a noise floor, but it is mathematically incorrect for `peak_rate.threshold`. Threshold is an absolute value, and the score calculation in `Detect()`:

```go
case stat > overloadAt:
    score = 1.0
case overloadAt > 0:
    score = math.Max(0.0, stat) / overloadAt
```

is safely bounded to `[0, 1]`. Even if overloadAt underflows to exactly 0.0, the `stat > overloadAt` branch executes first (setting score to 1.0), preventing any division-by-zero or `+Inf` score. While rejecting subnormals/zero is a reasonable design choice for a false-alarm dial, the stated mathematical justification does not apply to this specific parameter.

sim/saturation/config.go
```diff
@@ -266,8 +392,71 @@ func resolveBacklogDriftConfig(block *BacklogDriftBlock) (BacklogDriftConfig, er
...
+		}
+		out.ConsecutiveK = *block.ConsecutiveK
+	}
+	if block.OverloadMultiple != nil {
+		v := *block.OverloadMultiple
+		if math.IsNaN(v) || math.IsInf(v, 0) || v < 1 {
+			return peakRateConfig{}, fmt.Errorf("saturation config: peak_rate.overload_multiple must be a finite value >= 1 (below 1 the BACKLOGGED band is unsatisfiable), got %v", v)
+		}
+		out.OverloadMultiple = v
+	}
```
jgchn, line 479: The PR documentation in `docs/contributing/saturation-analyzer-extension.md` explicitly defends `backlog_drift.slope_k <= 1` as a `"legitimate 'maximally severe' setting"` that intentionally collapses the detector into a two-level system (Stable vs. Overloaded) by making the BACKLOGGED band unsatisfiable.

However, this code explicitly rejects `overload_multiple < 1` for the exact same reason (`"below 1 the BACKLOGGED band is unsatisfiable"`). There is a logical contradiction here: if collapsing the detector to two levels is a legitimate "maximally severe" configuration for backlog-drift, it should be allowed for peak-rate as well (or the docs should clarify why one is allowed and the other is not).

sim/saturation/peak_rate.go
```diff
@@ -0,0 +1,352 @@
...
+	p.inFlight = 0
+	p.observations = 0
+	p.firstTsUs = 0
+	p.lastTsUs = 0
+	p.haveFirst = false
+	// Belt-and-braces: Observe already zeroes the streak on its first call after a
+	// reset (the peak is 0, so ready() is false and the else-branch fires), which
+	// makes this line unobservable from outside and therefore untestable
+	// behaviorally. It is kept so "Reset clears all accumulated state" holds as a
+	// property of Reset itself rather than depending on Observe's internals.
```
jgchn: The reasoning in this comment is factually incorrect. `ready()` does not depend on peak; it depends only on observations, elapsedUs(), and `elapsedSec()`.

If MinObservations is set to 1, `ready()` can be true on the very first event after a reset (provided the event spans a non-zero time). In that case, if `statistic() > Threshold`, `p.consecutive` will increment to 1 instead of being reset.

This means `p.consecutive = 0` is strictly necessary for correctness and is not just "unobservable behaviorally" or "belt-and-braces." The explicit reset is required to prevent a false streak from carrying over across replay legs.

sim/saturation/config.go
sriumcp: You're right, and I had copied the rationale across knobs without re-deriving it.

Verified your reasoning holds: with `overload_multiple=3.0` and a subnormal threshold, `overloadAt = 1.5e-323` — which doesn't even underflow — and even when it does reach exactly 0, the `stat > overloadAt` branch runs first, so `score = 1.0`. No division, no `+Inf`. Measured: `level=OVERLOADED score=1 finite=true in[0,1]=true`.

So the hazard is real for knobs that multiply a noise floor and then **divide** (`composite.sensitivity` underflows its floor to 0, making every event non-STABLE; `backlog_drift.slope_k` skips its guarded score computation, leaving Score 0 beside an OVERLOADED Level) and absent for `peak_rate`'s knobs, whose score switch compares before dividing.

Fixed in `f5740cd1`: the `minCalibrationKnob` comment now separates the two purposes explicitly — a mathematical hazard for the dividing knobs, a usability floor for `peak_rate`'s (a false-alarm dial below `1e-6` is indistinguishable from "fire on everything", so rejecting it names the mistake rather than silently accepting a value that means something else). Same correction applied to the extension guide and CLAUDE.md.

sim/saturation/config.go
sriumcp: This is the best finding in the review — a genuine contradiction I wrote both halves of. Resolved by **relaxing `peak-rate` to match `slope_k`** in `f5740cd1`, and the evidence points that way on three independent grounds.

**The campaign never validated the band split.** The selection campaign's searched factors were `walk_statistic`, `backlog_source`, `consecutive_k`, `horizon_ratio`, `warmup_ms`, `min_observations`. The band multiplier was **not** among them, and the scorer counted `fired = level in ("BACKLOGGED", "OVERLOADED")` — so every validated result (the clean step at the cliff, FPR 0.0, the lead time) is *invariant* to where the two saturated levels divide. It is the least evidence-backed parameter in the detector, which makes a hard rejection there severity without support.

**The theory is two-state.** `R_t` either decays or converges to a positive constant. BACKLOGGED-vs-OVERLOADED is a severity refinement the asymptotics say nothing about, so nothing privileges `>= 1`.

**Sub-1 is measurably well-behaved.** Bypassing the constructor clamp on a 200-round saturated stream:

```
om=0.5  S=19 B=0   O=581  Score/Level coherent
om=1.0  S=19 B=0   O=581  Score/Level coherent
om=2.0  S=19 B=90  O=491  Score/Level INCOHERENT   <-- bug at a LEGAL value
```

Exactly the collapse you describe, and coherent throughout. Your comment also led me to a real bug at a *permitted* value: at `om=2.0` the statistic landed exactly on `overload_multiple × threshold`, where the band is strict (BACKLOGGED, correctly) but `stat/overloadAt` is exactly `1.0` in float64 — so 90 events reported a capped Score beside a BACKLOGGED Level. Fixed by holding the ratio one representable value below the cap (`math.Nextafter`), mutation-verified.

And the decisive practical point is the one you implied: rejecting sub-1 for one detector while allowing it for the other would mean the two **cannot be swept over a common knob range**, which is precisely the comparability this configuration surface exists to provide. Docs updated to state the shared policy once rather than defending one and forbidding the other.

sim/saturation/peak_rate.go
sriumcp: You're right that the comment is factually wrong, and the correction matters more than the line: `ready()` does not read `peak`. I had asserted a mechanism I never verified, then used it to explain away a surviving mutation — which is the worse half of the mistake.

On the conclusion, though, the streak still cannot leak today, for a reason I had not identified. `elapsedSec()` requires `lastTsUs > firstTsUs`, and the **first event after a reset sets both to its own timestamp** — so a single event spans no time:

```
first event ever: first=5000000 last=5000000 elapsed=0 ready=false consecutive=0
```

That holds at *any* `MinObservations`, including 1: your scenario needs the first event to span non-zero time, and one event cannot. I verified the `MinObservations=1` case directly — `ready()` is false on event 1 and only becomes true on event 2, by which point the else-branch has already zeroed the streak.

So: right that my reasoning was wrong, right that the line should stay, and the line is defence-in-depth rather than strictly necessary. Rewritten in `f5740cd1` to name the real mechanism, say plainly that it is a fact about the timestamp bookkeeping rather than about `Reset`, and note that the clear removes the dependency should elapsed time ever become seedable from a prior leg. I've dropped the "unobservable / belt-and-braces" framing — it was doing rhetorical work that a wrong premise cannot support.

jgchn (approved):

PR #1613, opened 2026-08-24
feat(workload): represent context compaction in accumulate replay (#1609)

namasl:
## Summary

Accumulate context-growth replay (`session_context_growth: accumulate`) reconstructed a session's per-round input as a **strictly-growing** shared token buffer. Real agentic traffic (Claude Code) **compacts** context: when the model summarizes/trims, the recorded absolute `in_N < in_{N-1} + out_{N-1}`, the delta encoder clamps the negative delta to `0`, and the append-only buffer cannot shrink. On real Weka data this **over-counts** the reconstructed cumulative input by ≈3–4× (+312%, ~30% of rounds), and the surviving over-long shared prefix inflates prefix-cache hit-rate.

**Fix:** the shared agentic-trace encoder emits an optional per-round `input_tokens_reset` marker (the recorded absolute) on **exactly** the compaction rounds (pre-clamp delta `< 0` — the precise over-count trigger). Accumulate closed-loop replay **re-seeds** its growing buffer to that absolute at the boundary, which:
1. shrinks the reconstructed input to the recorded absolute — reconstruction becomes **exact at every round**, not merely bounded; and
2. **intentionally breaks strict prefix identity across the boundary** (a summary is not a literal prefix of the pre-compaction buffer — faithful; also fixes the hit-rate over-estimate direction).

The marker is a trailing conditional CSV column (`adapter` / `think_time_us` precedent), `*int64` presence (nil = absent, mirroring #1608/#1612), round-tripping losslessly. It is **absent** for monotone sessions and for spec-built `blis run` blueprints, so a trace with no compaction round replays **byte-identically** to before (INV-6).

### Design decision (the issue asked for one)

Chose issue candidates **1+2** (optional absolute-reset column consumed as a per-round buffer re-seed) over candidate 3 (an explicit compaction event type): no new `sim.Event`, no schema-version bump, mirrors the merged `think_time_us` `*int64` presence column exactly. **Re-seed (fresh tokens), not truncate-keeping-prefix** — the acceptance criteria require strict-prefix identity to be *intentionally (and only)* broken across a compaction boundary; truncating `buf[0:in_N]` keeps a literal prefix and still over-estimates reuse.

[...]

PR #1623, opened 2026-08-25
fix(replay): guard --trace-output against lossy closed-loop accumulate re-export

### LINKED ISSUE #1621: Accumulate closed-loop replay re-export (--trace-output) is lossy: drops follow-up rounds, emits absolute (not delta) input, omits session_context_growth header
## Summary

`blis replay --session-mode closed-loop --trace-output` of an **accumulate** corpus does not produce a trace that can be re-replayed to reproduce the original session. This is a pre-existing limitation of the merged closed-loop replay export path (#1478–#1499), independent of any single converter feature. Surfaced during review of #1613 (qa-review F1/G3/G10).

**Precise trigger:** *plain* closed-loop replay only — `--session-mode closed-loop` with `--concurrent-sessions == 0` (the default) and `--trace-output` set, on a header with `session_context_growth: accumulate`. The **pool path** (`--concurrent-sessions > 0`) is already guarded (see Evidence item 5); it is the plain closed-loop path that is the inconsistent outlier.

## Evidence (all in `cmd/replay.go` / `sim/workload/tracev2.go`, on current `main` @ `b5934c13`)

1. **Follow-up rounds are dropped.** Closed-loop replay collects generated follow-ups in `followUpRequests` (`cmd/replay.go:668–683`) but the re-export calls `workload.RequestsToTraceRecords(requests)` (`cmd/replay.go:722`) with only the **round-0** `requests` slice. The re-exported trace therefore contains one record per session (round 0), not the full round sequence.
2. **Delta encoding is lost.** `RequestsToTraceRecords` (`sim/workload/tracev2.go:707`) emits `inputTokens := int(req.InputLen()) - prefixLen` (`:735`) — the **absolute** suffix length, not the per-round delta the accumulate loader expects. Re-replaying the export in accumulate mode would misread absolutes as deltas.
3. **`session_context_growth` is omitted from the re-export header** (`cmd/replay.go:723+` header construction does not set `Header.SessionContextGrowth`), so the export would not even select accumulate mode on re-replay — the `:129` gate (`session_context_growth == "accumulate"` requires closed-loop) never fires because the field is empty.
4. **(post-#1613) `input_tokens_reset` compaction markers are not preserved** — `RequestsToTraceRecords` builds records from `sim.Request`, which carries no marker (the reset was applied to the token buffer during replay). This is the same class of loss as (1)–(3): a converter/loader-level signal that a `sim.Request → TraceRecord` mapping cannot reconstruct. **Note:** #1613 is an *open PR*, not merged; this item is forward-looking and becomes reachable only once #1613 lands.
5. **The sibling pool path already rejects this exact case — closed-loop does not.** `cmd/replay.go:149` already fails fast for the pool driver:
   ```go
   // Pool mode + output artifacts: guard against silent truncation (R1).
   // Trace re-export in pool mode would contain only the initial wave of
   // round-0 requests (follow-ups/clones are never collected for export), so
   // reject it outright — mirroring how replay fails fast on unsupported
   // features (INV-13). ...
   if replayConcurrentSessions > 0 && replayTraceOutput != "" {
       logrus.Fatalf("--trace-output is not supported with --concurrent-sessions (pool mode): ...")
   }
   ```
   That guard's reasoning (only the initial wave captured, follow-ups never collected, INV-13 fail-fast) is *identical* to items (1)–(3), but it only covers `--concurrent-sessions > 0`. Plain closed-loop (`--concurrent-sessions == 0`) reaches the lossy `:722` export unguarded. This both confirms the defect and points at the fix location and the established pattern. There is currently **no test** covering the `:149` guard.

## Impact

Re-exporting a replayed closed-loop corpus and replaying it again does not reproduce the original per-round ISL. Users must re-replay the **original** converter corpus, not a `--trace-output` of a prior closed-loop replay. Per INV-13's "never silent degradation" clause, this lossy re-export arguably warrants at least a loud warning/`Fatalf` (as node pools / autoscaler / the pool path already do), rather than silently writing an unfaithful trace.

- **`blis replay` — AFFECTED (plain closed-loop only).** Root cause of the asymmetry: replay re-exports the *static* round-0 input slice (`requests`) with no arrival hook, so follow-ups are never in the exported set. Pool mode is already guarded at `:149`; plain closed-loop is not.
- **`blis run` — NOT affected**, despite also calling `workload.RequestsToTraceRecords` (`cmd/root.go:2265`). Two reasons: (a) run sources records from `SetArrivalHook` (`cmd/root.go:2207–2213`), which captures *every* arrival at the cluster boundary including generated follow-ups; and (b) run never emits accumulate/delta corpora — its per-round inputs are absolute by design, so the absolute-suffix encoding is correct and `session_context_growth` is legitimately unset (header `Mode: "generated"`). This parity note is called out because a reader grepping for `RequestsToTraceRecords` will land on the run site and must not "fix" it.
- **`blis observe` — NOT affected.** The observe `Recorder` (`cmd/observe.go:570`) records rounds as they are dispatched against the real server, so all follow-up rounds are captured directly with real observed token counts; it does not round-trip a converter corpus through `sim.Request → TraceRecord`.

The bug is therefore confined to the `cmd/replay.go` re-export path; the fix must live there (adjacent to the `:149` pool guard), not in the shared `sim/workload` mapping.

## Suggested direction (design decision needed)

Either:

- **(a) Make closed-loop re-export faithful** — capture the full round sequence (adopt the same arrival-hook pattern `blis run` already uses, or export `append(requests, followUpRequests...)` in session order), re-derive per-round deltas + `input_tokens_reset` markers from the session sequence, and set `session_context_growth` on the re-export header; or
- **(b) Reject `--trace-output` for accumulate / closed-loop replay with a clear error** until (a) is implemented — the smaller, INV-13-honest stopgap, mirroring the existing `:149` pool-mode guard.

Option (b) is the recommended first step: it is a few lines next to `:149`, closes the silent-degradation hole immediately, and leaves (a) as a follow-up.

## Acceptance criteria

- **Option (b) (stopgap):** `blis replay --session-mode closed-loop --trace-output <prefix>` on a header carrying `session_context_growth: accumulate` (and, once #1613 lands, any header requiring reset markers) exits non-zero with a clear error naming `--trace-output` and closed-loop/accumulate. A test asserts the non-zero exit + message, and — for symmetry — the currently-untested `:149` pool guard gains a companion test.
- **Option (a) (faithful):** a round-trip test — convert → `blis replay --session-mode closed-loop --trace-output out` → `blis replay` of `out` — reproduces per-request per-round ISL/metrics identical to replaying the original converter corpus (INV-13 parity for this configuration). The re-export header carries `session_context_growth: accumulate`; per-round `input_tokens` are deltas; `input_tokens_reset` markers survive (post-#1613).

## Not a regression from #1613

#1613 (context compaction) does not touch `cmd/` at all; items (1)–(3) predate it and are reproducible on current `main` without #1613. #1613 documents this boundary explicitly and its CSV-layer round-trip (`TraceRecord ↔ CSV`) is lossless; only the `sim.Request → TraceRecord` re-export path is affected.

### LINKED ISSUE #1622: Docs: replay calibration example uses --trace-output in pool mode (rejected by the pool guard) and mismatches --sim-results
## Summary

The corpus-mode calibration example in `docs/guide/observe-replay-calibrate.md` (the "To calibrate real vs simulated over the same corpus" block, ~lines 339–345) cannot run as written on current `main`:

```bash
  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200 \
  --trace-output sim
blis calibrate --trace-header observed.yaml --trace-data observed.csv \
  --sim-results sim.results.json --report calibration.json
```

Two independent problems:

1. **`--trace-output` + `--concurrent-sessions` is rejected.** `cmd/replay.go` fails fast when pool mode is combined with `--trace-output`:
   ```
   --trace-output is not supported with --concurrent-sessions (pool mode): a re-export would capture only the initial N-session wave, not all pooled sessions. Re-run without --trace-output.
   ```
   So `blis replay ... --concurrent-sessions 8 ... --trace-output sim` exits non-zero immediately — the example never reaches `calibrate`.

2. **`--trace-output` does not produce `--sim-results`.** The `calibrate` step consumes `--sim-results sim.results.json`, but `--trace-output sim` writes TraceV2 files (`sim.yaml` / `sim.csv`), not a results JSON. The per-request results JSON `calibrate` needs comes from `--results-path`.

## Suggested fix

Replace `--trace-output sim` with `--results-path sim.results.json` in the replay step (pool mode + `--results-path` only warns about clone/follow-up coverage, it does not fail):

```bash
  --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200 \
  --results-path sim.results.json
```

## Discovery context

Found while implementing #1621 (guarding the sibling *plain* closed-loop `--trace-output` accumulate re-export). The two `--trace-output` pool/closed-loop guards in `cmd/replay.go` are the reason this documented command fails; the doc was not updated to match them. Pre-existing on `main` (independent of #1621's change, which only affects `--concurrent-sessions == 0`).

namasl:
## Summary

- Plain closed-loop replay (`--session-mode closed-loop`, `--concurrent-sessions 0`) of a `session_context_growth=accumulate` corpus **silently wrote an unfaithful trace** when `--trace-output` was set. The re-export sourced only the round-0 `requests` slice (dropping the `SessionManager`-generated follow-up rounds), `RequestsToTraceRecords` emitted **absolute** per-round `input_tokens` instead of the **deltas** an accumulate loader expects, and the re-export header omitted `session_context_growth`. Re-replaying such an export could not reproduce the original per-round input.
- **Option (b) stopgap** (the issue's recommended first step): fail fast instead of writing a lossy trace, mirroring the **existing sibling pool-mode guard** (`--concurrent-sessions > 0` + `--trace-output`). The new guard sits immediately after that pool guard and covers the remaining plain closed-loop case.
- Adds the **companion test the sibling pool-mode guard never had** (the issue flagged `cmd/replay.go`'s pool `--trace-output` guard as untested).
- **Docs:** documents both `--trace-output` re-export restrictions (pool mode + plain closed-loop accumulate) in the replay guide, and fixes a pre-existing broken calibration example that used `--trace-output` with `--concurrent-sessions` (rejected by the pool guard) and mismatched `--sim-results` — resolving #1622.

## Issue

Closes #1621
Fixes #1622

## Behavioral Contracts

- **BC-1** — GIVEN a header with `session_context_growth: accumulate`, WHEN `blis replay --session-mode closed-loop --trace-output <prefix>` runs with `--concurrent-sessions 0` (default), THEN the command exits non-zero (code 1) with a message naming `--trace-output`, `closed-loop`, and `accumulate`; no trace files are written.
- **BC-2** (companion / regression) — GIVEN any loadable trace, WHEN `blis replay --concurrent-sessions N>0 --trace-output <prefix>`, THEN it exits code 1 naming `--concurrent-sessions` / `pool mode` — locking in the previously-untested sibling guard.

## Cross-Path Parity

| `blis run` | no | N/A | Sources records from `SetArrivalHook` (captures every round incl. follow-ups) and never emits accumulate corpora — its absolute per-round encoding is correct. `RequestsToTraceRecords` (`cmd/root.go`) intentionally **not** touched. |
| `blis replay` | yes | yes | Fix site — the static round-0 re-export is the sole lossy path. |
| `blis observe` | no | N/A | The `Recorder` captures dispatched rounds directly with real token counts; no `sim.Request → TraceRecord` round-trip. |

## Invariants Affected

- **INV-13 (run/replay parity, 'never silent degradation')** — enforced: the lossy re-export now fails fast with a clear error, matching how node pools / autoscaler / the pool path already reject unsupported replay features, rather than silently writing an unfaithful trace.
- **INV-6 (determinism)** — unaffected: the guard fires only on the exact `accumulate` + plain-closed-loop + `--trace-output` combination; all other runs are byte-identical.

## Scope note (why Option (b), not (a))

The issue recommends Option (b) (fail fast) as the first step; faithful re-export (Option (a) — capture the full round sequence, re-derive per-round deltas, set the header) remains a follow-up. This PR closes the silent-degradation hole immediately with a few lines next to the existing guard.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (full suite green)
- [x] `golangci-lint run ./cmd/...` (0 issues)
- [x] `TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal` — new guard (red→green verified)
- [x] `TestReplayCmd_PoolTraceOutputFatal` — companion for the previously-untested pool guard

---

### DIFF
--- a/cmd/replay.go
+++ b/cmd/replay.go
@@ -148,6 +148,21 @@ Example:
 		if replayConcurrentSessions > 0 && replayTraceOutput != "" {
 			logrus.Fatalf("--trace-output is not supported with --concurrent-sessions (pool mode): a re-export would capture only the initial %d-session wave, not all pooled sessions. Re-run without --trace-output.", replayConcurrentSessions)
 		}
+		// Plain closed-loop + accumulate + output artifacts: the same silent-truncation
+		// hazard the pool guard above rejects (R1 / INV-13), for --concurrent-sessions == 0
+		// (#1621). A closed-loop re-export sources only the round-0 request slice (the
+		// export block below builds records from `requests`, never the generated
+		// follow-up rounds collected in followUpRequests); RequestsToTraceRecords then
+		// re-emits each per-round input_tokens as an absolute suffix length rather than
+		// the delta an accumulate loader expects; and the re-export header never carries
+		// session_context_growth. The exported trace therefore could NOT be re-replayed
+		// to reproduce the original per-round input. Reject it outright until faithful
+		// re-export (issue #1621, option (a)) lands. Accumulate implies closed-loop here
+		// (guaranteed by the gate above), and pool mode is already handled by the guard
+		// above, so this covers the remaining plain closed-loop case.
+		if traceData.Header.SessionContextGrowth == "accumulate" && replayConcurrentSessions == 0 && replayTraceOutput != "" {
+			logrus.Fatalf("--trace-output is not supported with --session-mode closed-loop on a session_context_growth=accumulate trace: the re-export would source only the round-0 requests (dropping the generated follow-up rounds), emit absolute per-round input_tokens instead of deltas, and omit session_context_growth from the header, so the result could not be re-replayed to reproduce the original per-round input. Re-run without --trace-output; to re-export, replay the original converter corpus (e.g. from `blis convert otel`) instead.")
+		}
 		if replayConcurrentSessions > 0 && resultsPath != "" {
 			logrus.Warnf("--results-path with --concurrent-sessions (pool mode): per-request results cover only the original corpus sessions' round-0; duplicated (clone) sessions and follow-up rounds are excluded (non-numeric request ids).")
 		}
--- a/cmd/replay_test.go
+++ b/cmd/replay_test.go
@@ -3199,3 +3199,210 @@ func TestReplayCmd_PoolRejectsNonSessionRecords(t *testing.T) {
 		t.Errorf("guard should pre-empt BuildSessionPool's internal 'count mismatch' error, but it surfaced:\n%s", out)
 	}
 }
+
+// TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal is a regression test for the
+// plain closed-loop accumulate re-export guard (issue #1621). Re-exporting a
+// closed-loop replay via --trace-output sources only the round-0 `requests` slice
+// (the export block never sees the generated follow-up rounds), and
+// RequestsToTraceRecords re-emits per-round input_tokens as absolute suffix
+// lengths (not the deltas an accumulate loader expects) and does not carry
+// session_context_growth onto the re-export header. The exported trace therefore
+// could not be re-replayed to reproduce the original per-round input, so — rather
+// than write a silently-unfaithful trace (INV-13, R1) — cmd/replay.go Fatalfs.
+//
+// This mirrors the sibling pool-mode guard (--concurrent-sessions + --trace-output,
+// TestReplayCmd_PoolTraceOutputFatal); the guard fires up front, before session
+// blueprints are built, so a minimal accumulate corpus reaches it. Fatalf os.Exit's,
+// so — following TestReplayCmd_AccumulateHeaderRequiresClosedLoop — the case runs in
+// a subprocess re-run (BLIS_TEST_SUBPROCESS=1) and is asserted via exit code + stderr.
+func TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal(t *testing.T) {
+	if os.Getenv("BLIS_TEST_SUBPROCESS") == "1" {
+		// Subprocess: an accumulate-header, multi-round session corpus replayed in
+		// plain closed-loop mode (--concurrent-sessions 0) WITH --trace-output set.
+		// Must Fatalf before completing.
+		dir := t.TempDir()
+		headerPath := filepath.Join(dir, "trace.yaml")
+		dataPath := filepath.Join(dir, "trace.csv")
+		header := &workload.TraceHeader{Version: 3, TimeUnit: "microseconds", Mode: "generated", SessionContextGrowth: "accumulate"}
+		records := []workload.TraceRecord{
+			{RequestID: 0, SessionID: "s1", RoundIndex: 0, InputTokens: 10, OutputTokens: 5, ArrivalTimeUs: 0, Status: "ok"},
+			{RequestID: 1, SessionID: "s1", RoundIndex: 1, InputTokens: 4, OutputTokens: 5, ArrivalTimeUs: 100_000, Status: "ok"},
+		}
+		if err := workload.ExportTraceV2(header, records, headerPath, dataPath); err != nil {
+			os.Exit(2)
+		}
+
+		mcFolder, hwPath, defaultsPath := setupTrainedPhysicsTestFixturesWithDefaults(t)
+		model = "test-model"
+		latencyModelBackend = "trained-physics"
+		totalKVBlocks = 1000
+		blockSizeTokens = 16
+		maxNumSeqs = 64
+		maxNumBatchedTokens = 2048
+		numInstances = 1
+		seed = 42
+		longPrefillTokenThreshold = 0
+		kvCPUBlocks = 0
+		kvOffloadThreshold = 0.9
+		kvTransferBandwidth = 100.0
+		kvTransferBaseLatency = 0
+		snapshotRefreshInterval = 0
+		admissionPolicy = "always-admit"
+		routingPolicy = "round-robin"
+		scheduler = "fcfs"
+		policyConfigPath = ""
+		maxModelLen = 0
+		traceLevel = "none"
+		counterfactualK = 0
+		traceHeaderPath = headerPath
+		traceDataPath = dataPath
+		modelConfigFolder = mcFolder
+		hwConfigPath = hwPath
+		gpu = "H100"
+		tensorParallelism = 1
+		defaultsFilePath = defaultsPath
+		replaySessionMode = "closed-loop"
+		replayConcurrentSessions = 0
+		replayTotalSessions = 0
+		replayThinkTimeMs = 0
+		replayThinkTimeDist = ""
+		resultsPath = ""
+		replayTraceOutput = filepath.Join(dir, "reexport")
+
+		testCmd := &cobra.Command{}
+		registerSimConfigFlags(testCmd)
+		testCmd.Flags().StringVar(&traceHeaderPath, "trace-header", "", "")
+		testCmd.Flags().StringVar(&traceDataPath, "trace-data", "", "")
+		if err := testCmd.ParseFlags([]string{
+			"--model", "test-model", "--latency-model", "trained-physics",
+			"--total-kv-blocks", "1000", "--hardware", "H100", "--tp", "1",
+			"--model-config-folder", mcFolder, "--hardware-config", hwPath,
+			"--trace-header", headerPath, "--trace-data", dataPath,
+			"--defaults-filepath", defaultsPath,
+		}); err != nil {
+			fmt.Fprintf(os.Stderr, "ParseFlags failed (test setup error): %v\n", err)
+			os.Exit(2) // distinct from logrus.Fatalf exit code (1)
+		}
+		replayCmd.Run(testCmd, nil) // must Fatalf before here
+		os.Exit(0)                  // reached only if no fatal = parent test failure
+	}
+
+	// Parent: re-run this test as subprocess and expect exit code 1 (logrus.Fatalf).
+	cmd := exec.Command(os.Args[0], "-test.run=TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal", "-test.v")
+	cmd.Env = append(os.Environ(), "BLIS_TEST_SUBPROCESS=1")
+	out, err := cmd.CombinedOutput()
+	if err == nil {
+		t.Fatal("expected non-zero exit when re-exporting a closed-loop accumulate replay (--trace-output), got exit 0")
+	}
+	var exitErr *exec.ExitError
+	if !errors.As(err, &exitErr) {
+		t.Fatalf("unexpected error type: %v", err)
+	}
+	if exitErr.ExitCode() != 1 {
+		t.Fatalf("expected exit code 1 (logrus.Fatalf), got %d; output:\n%s", exitErr.ExitCode(), out)
+	}
+	// Message must name the flag, the mode, and the header field so the operator
+	// knows exactly which combination is rejected and why.
+	for _, want := range []string{"--trace-output", "closed-loop", "accumulate"} {
+		if !strings.Contains(string(out), want) {
+			t.Errorf("fatal message should mention %q, got:\n%s", want, out)
+		}
+	}
+}
+
+// TestReplayCmd_PoolTraceOutputFatal is the companion regression test the sibling
+// pool-mode guard (--concurrent-sessions > 0 + --trace-output) never had (issue
+// #1621 notes it was untested). Pool-mode re-export would capture only the initial
+// wave of round-0 requests — follow-ups/clones are never collected — so the guard
+// in cmd/replay.go rejects it outright (INV-13, R1). Fatalf os.Exit's, so the case
+// runs in a subprocess re-run and is asserted via exit code + stderr content.
+func TestReplayCmd_PoolTraceOutputFatal(t *testing.T) {
+	if os.Getenv("BLIS_TEST_SUBPROCESS") == "1" {
+		// Subprocess: a pure-session corpus replayed with --concurrent-sessions 1
+		// (pool mode) AND --trace-output set. Must Fatalf before completing.
+		dir := t.TempDir()
+		headerPath := filepath.Join(dir, "trace.yaml")
+		dataPath := filepath.Join(dir, "trace.csv")
+		header := &workload.TraceHeader{Version: 3, TimeUnit: "microseconds", Mode: "generated"}
+		records := []workload.TraceRecord{
+			{RequestID: 0, SessionID: "s1", RoundIndex: 0, InputTokens: 10, OutputTokens: 5, ArrivalTimeUs: 0, Status: "ok"},
+		}
+		if err := workload.ExportTraceV2(header, records, headerPath, dataPath); err != nil {
+			os.Exit(2)
+		}
+
+		mcFolder, hwPath, defaultsPath := setupTrainedPhysicsTestFixturesWithDefaults(t)
+		model = "test-model"
+		latencyModelBackend = "trained-physics"
+		totalKVBlocks = 1000
+		blockSizeTokens = 16
+		maxNumSeqs = 64
+		maxNumBatchedTokens = 2048
+		numInstances = 1
+		seed = 42
+		longPrefillTokenThreshold = 0
+		kvCPUBlocks = 0
+		kvOffloadThreshold = 0.9
+		kvTransferBandwidth = 100.0
+		kvTransferBaseLatency = 0
+		snapshotRefreshInterval = 0
+		admissionPolicy = "always-admit"
+		routingPolicy = "round-robin"
+		scheduler = "fcfs"
+		policyConfigPath = ""
+		maxModelLen = 0
+		traceLevel = "none"
+		counterfactualK = 0
+		traceHeaderPath = headerPath
+		traceDataPath = dataPath
+		modelConfigFolder = mcFolder
+		hwConfigPath = hwPath
+		gpu = "H100"
+		tensorParallelism = 1
+		defaultsFilePath = defaultsPath
+		replaySessionMode = "closed-loop"
+		replayConcurrentSessions = 1
+		replayTotalSessions = 0
+		replayThinkTimeMs = 0
+		replayThinkTimeDist = ""
+		resultsPath = ""
+		replayTraceOutput = filepath.Join(dir, "reexport")
+
+		testCmd := &cobra.Command{}
+		registerSimConfigFlags(testCmd)
+		testCmd.Flags().StringVar(&traceHeaderPath, "trace-header", "", "")
+		testCmd.Flags().StringVar(&traceDataPath, "trace-data", "", "")
+		if err := testCmd.ParseFlags([]string{
+			"--model", "test-model", "--latency-model", "trained-physics",
+			"--total-kv-blocks", "1000", "--hardware", "H100", "--tp", "1",
+			"--model-config-folder", mcFolder, "--hardware-config", hwPath,
+			"--trace-header", headerPath, "--trace-data", dataPath,
+			"--defaults-filepath", defaultsPath,
+		}); err != nil {
+			fmt.Fprintf(os.Stderr, "ParseFlags failed (test setup error): %v\n", err)
+			os.Exit(2) // distinct from logrus.Fatalf exit code (1)
+		}
+		replayCmd.Run(testCmd, nil) // must Fatalf before here
+		os.Exit(0)                  // reached only if no fatal = parent test failure
+	}
+
+	// Parent: re-run this test as subprocess and expect exit code 1 (logrus.Fatalf).
+	cmd := exec.Command(os.Args[0], "-test.run=TestReplayCmd_PoolTraceOutputFatal", "-test.v")
+	cmd.Env = append(os.Environ(), "BLIS_TEST_SUBPROCESS=1")
+	out, err := cmd.CombinedOutput()
+	if err == nil {
+		t.Fatal("expected non-zero exit when re-exporting a pool-mode replay (--trace-output), got exit 0")
+	}
+	var exitErr *exec.ExitError
+	if !errors.As(err, &exitErr) {
+		t.Fatalf("unexpected error type: %v", err)
+	}
+	if exitErr.ExitCode() != 1 {
+		t.Fatalf("expected exit code 1 (logrus.Fatalf), got %d; output:\n%s", exitErr.ExitCode(), out)
+	}
+	for _, want := range []string{"--trace-output", "--concurrent-sessions", "pool mode"} {
+		if !strings.Contains(string(out), want) {
+			t.Errorf("fatal message should mention %q, got:\n%s", want, out)
+		}
+	}
+}
--- a/docs/guide/observe-replay-calibrate.md
+++ b/docs/guide/observe-replay-calibrate.md
@@ -232,7 +232,25 @@ Replay also accepts all shared simulation config flags (`--latency-model`, `--to
 | **Horizon** | From `--horizon` flag or spec | Auto-computed as 2x max arrival time (override with `--horizon`) |
 | **Output format** | Full `MetricsOutput` JSON | `SimResult` JSON array (request_id, ttft_us, e2e_us, input_tokens, output_tokens) |
 | **Session support** | Session manager creates follow-ups | Session structure encoded in trace (no manager needed) |
-| **Trace export** | `--trace-output` (header `mode: "generated"`) | `--trace-output` (header `mode: "replayed"`) |
+| **Trace export** | `--trace-output` (header `mode: "generated"`) | `--trace-output` (header `mode: "replayed"`); see restrictions below |
+
+!!! warning "`--trace-output` cannot re-export closed-loop follow-up rounds"
+    Replay **rejects `--trace-output` (fails fast, non-zero exit)** in two cases,
+    because a re-export sources only the round-0 request slice — the follow-up rounds
+    generated during a closed-loop run are never in the exported set:
+
+    - **Pool mode** (`--concurrent-sessions > 0`): the export would capture only the
+      initial round-0 wave, not the pooled/cloned sessions.
+    - **Plain closed-loop replay of an accumulate corpus** (`--session-mode
+      closed-loop` on a header with `session_context_growth: accumulate` — e.g. a
+      `blis convert otel` / `blis convert weka` corpus): besides dropping the
+      follow-ups, the export would emit absolute per-round `input_tokens` instead of
+      the deltas an accumulate loader expects and omit `session_context_growth`, so it
+      could not be re-replayed faithfully.
+
+    To re-export, replay the **original converter corpus** rather than a prior
+    closed-loop replay's output. (Faithful closed-loop re-export is a planned
+    follow-up.)
 
 !!! warning "Latency model matters"
     The replay command simulates token generation using the configured latency model. For accurate calibration, choose the latency model that best matches the server's behavior. See [Latency Models](latency-models.md) for guidance on selecting between roofline and trained-physics modes.
@@ -339,7 +357,7 @@ To calibrate real vs simulated over the same corpus:
 ```bash
 blis replay --trace-header corpus.yaml --trace-data corpus.csv \
   --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200 \
-  --trace-output sim
+  --results-path sim.results.json
 blis calibrate --trace-header observed.yaml --trace-data observed.csv \
   --sim-results sim.results.json --report calibration.json
 ```


namasl:

namasl:

## qa-review — PR #1623: ⛔ BLOCK
_fix(replay): guard --trace-output against lossy closed-loop accumulate re-export · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR implements acceptance option (b): it rejects plain closed-loop re-export for valid accumulate traces, gives an actionable error, tests the non-zero exit and message, and adds the specifically requested companion test for the existing pool guard. It does not alter the shared exporter or unrelated run/observe behavior. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. User-facing replay documentation still presents `--trace-output` as generally supported and does not document that it is rejected for pool mode or for plain closed-loop accumulate traces. The run/replay comparison table likewise lists replay trace export without qualification. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale code comment was found. The new comments accurately describe the current export source, missing follow-ups, absolute token encoding, omitted growth header, and the temporary fail-fast behavior. |
| G1 | correctness | ✅ CONFIDENT | The check `traceData.Header.SessionContextGrowth == "accumulate"` performs an exact string comparison. Is `SessionContextGrowth` guaranteed to be normalized to lowercase at parse time, or could a trace header containing `"Accumulate"` or surrounding whitespace bypass this guard and write an unfaithful trace? | `SessionContextGrowth` is not normalized during YAML parsing: capitalization and surrounding whitespace are retained. However, in closed-loop mode such values do not reach trace export; blueprint loading explicitly accepts only the empty string or exact `"accumulate"` and fails before simulation/export. Shipped converters emit the canonical lowercase value. A malformed value used in fixed mode can bypass this validation, but that is a hand-authored invalid-header case rather than a producer-generated accumulate trace. |
| G2 | correctness | ✅ CONFIDENT | The guard logic checks `replayConcurrentSessions == 0`. If `replayConcurrentSessions` is set to a negative integer via CLI or config, would it bypass both the pool mode guard (`> 0`) and this closed-loop guard (`== 0`), reaching the lossy export path? | No. Negative `--concurrent-sessions` values are rejected before either trace-output guard, so they cannot bypass both guards and reach export. |
| G3 | invariants | ✅ CONFIDENT | How does rejecting `--trace-output` in closed-loop accumulate replay uphold INV-13 (run/replay parity) when `blis run` supports trace output via arrival hooks for multi-round workloads? | The rejection upholds INV-13 by changing an unsupported replay configuration from silent, incorrect output into an explicit startup failure. INV-13 requires parity for configurations supported by both paths and explicitly requires unsupported replay features to fail rather than degrade. It does not require replay to support every feature that run supports. |
| G4 | invariants | ✅ CONFIDENT | To ensure determinism under INV-6 across environments, does executing `exec.Command(os.Args[0], ...)` in the new subprocess tests introduce test non-determinism when running `go test` with cross-compilation or custom test binary paths? | No test nondeterminism is introduced. `os.Args[0]` is the currently running test executable, so custom build/output paths remain valid. Ordinary cross-compilation only builds the test binary and cannot execute an incompatible target binary without an appropriate executor; that is an environment capability issue, not seed- or ordering-dependent behavior in this test. |
| G5 | rules | ✅ CONFIDENT | In accordance with rule R1 (never silently degrade or truncate), does `cmd/replay.go` perform any file creation or path initialization for `replayTraceOutput` prior to executing this guard check that could leave partial artifacts on disk upon `logrus.Fatalf`? | No replay-output artifact is created before the guard. The command first reads the input YAML and CSV, validates flags, and then executes the guard. Output files are only opened by `ExportTraceV2` after simulation, a path the guard prevents from being reached. |
| G6 | rules | ✅ CONFIDENT | The check evaluates `replayTraceOutput != ""`. Is there any scenario where `--trace-output` is provided as a whitespace-only string or whitespace-padded path that evades this guard while still attempting output generation later? | Whitespace-only and whitespace-padded values do not evade the guard because both are non-empty strings. If the configuration is rejected, either form triggers the guard; otherwise the unchanged string is later used as the output prefix. |
| G7 | parity | ✅ CONFIDENT | The cross-path parity table states `blis observe` is unaffected. How does `blis observe` ensure session context growth fields and per-round delta encodings remain faithful when recording live server dispatches? | `blis observe` does not recreate an accumulate/delta corpus. It records every request when that round is actually dispatched, including generated follow-ups, and writes the dispatched request's absolute suffix count (`InputTokens - PrefixLength`) with its session ID and round index. Its header intentionally leaves `session_context_growth` empty, so downstream replay interprets those values as independent absolute per-round inputs rather than accumulate deltas. This preserves observed per-round token counts without requiring reconstruction from `sim.Request` after the fact. |
| G8 | parity | ✅ CONFIDENT | Why does `blis replay` reject accumulate `--trace-output` outright instead of reusing the arrival-hook pattern that `blis run` uses in `cmd/root.go` to capture follow-up rounds as they arrive? | Reusing the run arrival hook would solve only collection of follow-up requests. A faithful accumulate export must additionally regroup requests by session/round, convert each accumulated absolute context back into the required per-round deltas, preserve ordering and think-time semantics, set the growth header, and eventually preserve reset-marker information that is not present on `sim.Request`. The PR therefore chose the issue's smaller accepted stopgap rather than partially fixing only the truncation. |
| G9 | tests | ✅ CONFIDENT | In `TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal`, the test mutates global package variables (`model`, `replaySessionMode`, `replayConcurrentSessions`, etc.). Does this global state mutation prevent running `cmd` package tests with `t.Parallel()`, or risk leaking configuration to concurrent tests? | The new mutations occur only in a separately executed test process, which terminates via `Fatalf`; they cannot leak into the parent test process. The tests themselves do not call `t.Parallel()`, and the cmd test suite already treats package-level CLI variables as incompatible with parallel in-process tests. |
| G10 | tests | ✅ CONFIDENT | The subprocess test re-executes the test runner using `exec.Command(os.Args[0], "-test.run=TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal")`. If `-test.run` matches additional test names or sub-tests in future refactorings, could the subprocess execute unintended test cases before hitting the exit condition? | The current names select only the intended top-level test, so no unintended case runs today. Go's `-test.run` argument is a regular expression and these expressions are not anchored, so a future top-level test whose name contains the current name could also match. In practice the intended child test reaches `Fatalf` before another matched test could run. |
| G11 | rationale | ✅ CONFIDENT | Option (b) (fail fast) was implemented instead of Option (a) (faithful export of `followUpRequests`). What concrete technical limitation or complexity prevented implementing full session sequence capture in this PR? | The concrete complexity is that full capture is more than appending `followUpRequests`. Follow-ups are generated dynamically and hold accumulated absolute token buffers, while the source format requires per-round deltas and a growth header; session ordering and metadata must also be reconstructed. Reset markers discussed in the issue cannot be inferred from the current `sim.Request` representation. The stopgap avoids claiming fidelity until all of those transformations exist. |
| G12 | rationale | ✅ CONFIDENT | Why was the guard placed directly inside the CLI layer in `cmd/replay.go` rather than inside `workload.RequestsToTraceRecords` or `ExportTraceV2` where trace export transformation logic is defined? | The CLI layer is the correct location because unsupportedness depends on command context: the input header, replay session mode, pool setting, and requested output flag. `RequestsToTraceRecords` sees only requests and cannot know whether their token counts originated as accumulate deltas, while `ExportTraceV2` is a generic serializer also used by valid run and observe paths. Moving the rejection into either shared function would reject legitimate exports or require leaking replay-specific policy into the workload package. |
| G13 | scope | ⚠️ CANNOT_ANSWER | Issue #1621 notes that context compaction reset markers (`input_tokens_reset` from open PR #1613) are also lost during `sim.Request -> TraceRecord` export. Does this guard inherently protect those reset cases as well, or will a separate guard be required once #1613 lands? | The current guard inherently protects any reset-marker trace only if the future format guarantees that every such trace also has exact `session_context_growth: accumulate`. It checks no reset-specific field. Because #1613's reset field and validation rules are not present in this worktree, it is not possible to verify whether that implication will be guaranteed; if reset markers can appear under any other header value, a separate or broadened guard will be required. |
| G14 | scope | ✅ CONFIDENT | The PR adds `TestReplayCmd_PoolTraceOutputFatal` as a companion regression test for the existing pool-mode guard. Are there any behavioral differences or edge cases in the pool guard itself that were uncovered while adding this test? | No behavioral difference or new pool-guard edge case was uncovered. The added test exercises the existing behavior: any positive pool size plus non-empty trace output fails before pool construction, with the expected flag and mode information. The production pool guard itself is unchanged. |
| G15 | errors | ✅ CONFIDENT | The guard calls `logrus.Fatalf`, terminating the process immediately with `os.Exit(1)`. Is `Fatalf` standard for CLI validation in `cmd/replay.go`, or should error handling return a structured `error` up through Cobra's `RunE` to allow clean teardown and in-process unit testing? | `logrus.Fatalf` is the established validation style in `cmd/replay.go` and is permitted in `cmd/`; the project rule prohibits it only in library code under `sim/`. The replay command is defined with Cobra `Run`, not `RunE`, and numerous neighboring validation and execution failures already use `Fatalf`, so the subprocess test follows existing conventions. |
| G16 | errors | ✅ CONFIDENT | If a user specifies both `--trace-output` and `--results-path` in pool mode or closed-loop accumulate mode, which error/warning path takes precedence, and is that evaluation order deterministic? | Evaluation order is deterministic. In pool mode, the trace-output fatal executes first and terminates the process, so the later `--results-path` pool warning is never emitted. In plain closed-loop accumulate mode, the accumulate trace-output fatal likewise executes before results handling; there is no special plain-mode results warning. If trace output is absent, the pool results warning executes normally. |

### Items to fix
- **F2 · FLAW_FOUND** — No. User-facing replay documentation still presents `--trace-output` as generally supported and does not document that it is rejected for pool mode or for plain closed-loop accumulate traces. The run/replay comparison table likewise lists replay trace export without qualification.
  _docs/guide/observe-replay-calibrate.md:215-223 describes replay `--trace-output` without the new restriction; docs/guide/observe-replay-calibrate.md:225-235 lists replay trace export without qualification; the actual restrictions are cmd/replay.go:142-165_
- **G13 · CANNOT_ANSWER** — The current guard inherently protects any reset-marker trace only if the future format guarantees that every such trace also has exact `session_context_growth: accumulate`. It checks no reset-specific field. Because #1613's reset field and validation rules are not present in this worktree, it is not possible to verify whether that implication will be guaranteed; if reset markers can appear under any other header value, a separate or broadened guard will be required.
  _sim/workload/tracev2.go:17-67 contains the current header schema with SessionContextGrowth but no reset capability; cmd/replay.go:163 guards solely on exact SessionContextGrowth; current CSV serialization at sim/workload/tracev2.go:330-389 has no reset-marker column_

### Important to consider
- **G1** — Central validation in LoadTraceV2 would make malformed-header handling consistent between fixed and closed-loop replay.
  _sim/workload/tracev2.go:397-409 decodes YAML without normalization; sim/workload/replay.go:298-305 rejects unknown values such as `"Accumulate"`; cmd/replay.go:197-218 performs blueprint loading before export at cmd/replay.go:735-750; cmd/convert_otel.go:171 and cmd/convert_weka.go:118 populate the canonical growth value_
- **G3** — Faithful replay export remains desirable as follow-up option (a), but fail-fast satisfies the stated stopgap acceptance criterion.
  _docs/contributing/standards/invariants.md:178 defines parity for configurations supported by both commands and requires fail-fast for unsupported replay features; cmd/replay.go:151-165 applies that policy to accumulate re-export; run's faithful arrival-hook implementation is cmd/root.go:2183-2213 and cmd/root.go:2254-2277_
- **G6** — The CLI does not trim output paths, so an allowed whitespace-only value would create unusually named files, but that is pre-existing path handling and does not bypass this guard.
  _cmd/replay.go:148 and cmd/replay.go:163 test only `replayTraceOutput != ""`; cmd/replay.go:750 appends `.yaml` and `.csv` to that same value_
- **G7** — This preserves per-round sizes, not accumulate-mode token-prefix identity; observe uses the independent absolute representation.
  _cmd/observe_cmd.go:1089-1118 records each dispatched round and sends completions through the session handler; cmd/observe.go:527-566 writes suffix input count, session ID, and round index; cmd/observe_cmd.go:693-708 constructs a real-trace header without SessionContextGrowth; cmd/observe_cmd.go:721-726 exports all recorder records_
- **G9** — A future conversion of these tests to in-process execution or `t.Parallel()` would require state capture/restoration.
  _cmd/replay_test.go:3218-3287 confines setup mutations to the `BLIS_TEST_SUBPROCESS` branch; cmd/replay_test.go:3290-3293 starts a separate process; the pool test follows the same pattern at cmd/replay_test.go:3319-3390; existing warnings against parallel global mutation appear at cmd/replay_test.go:705 and cmd/parity_run_replay_test.go:359_
- **G10** — Using `^TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal$` and the equivalent anchored pool expression would make future selection unambiguous.
  _cmd/replay_test.go:3291 uses `-test.run=TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal`; cmd/replay_test.go:3392 uses the analogous unanchored pool name; the child invokes replay and exits at cmd/replay_test.go:3286-3287_
- **G15** — A future command-wide migration to `RunE` could improve in-process testability, but doing it only for this guard would be inconsistent and unnecessary.
  _cmd/replay.go:92-106, cmd/replay.go:129-176, and cmd/replay.go:215-241 use `logrus.Fatalf`; docs/contributing/standards/rules.md:90-98 permits termination in cmd while forbidding it in sim; the new guard is cmd/replay.go:163-165_

namasl:
Thanks — addressing the qa-review findings (pushed in 1e96c86b):

**F2 (FLAW_FOUND, docs) — fixed.** Added an admonition to `docs/guide/observe-replay-calibrate.md` documenting both `--trace-output` re-export restrictions (pool mode, and plain closed-loop replay of an `accumulate` corpus) with the workaround (re-export from the original converter corpus), and annotated the run/replay comparison-table row. While there I also fixed the pre-existing broken corpus-mode calibration example (it used `--trace-output` with `--concurrent-sessions`, which the pool guard rejects, and mismatched `--sim-results`, which comes from `--results-path`) — this resolves the separately-filed #1622, now folded in via `Fixes #1622`.

**G13 (CANNOT_ANSWER, reset markers / #1613) — intentionally out of scope.** #1613 (`input_tokens_reset`) is an open PR, not merged; its field/validation are not in this tree, so this guard deliberately does not key on a reset marker. Issue #1621 explicitly scopes reset-marker preservation as forward-looking ("becomes reachable only once #1613 lands"). When #1613 merges, whether this accumulate-keyed guard already covers reset traces (only if every reset trace also carries `session_context_growth: accumulate`) or needs broadening will be handled there — no change belongs in this stopgap.

**Non-blocking observations — acknowledged, no change.** The answerer itself marked these CONFIDENT:
- **G1** (exact-string `"accumulate"` match): capitalized/whitespace values cannot reach export in closed-loop — `LoadTraceV2SessionBlueprints` rejects anything other than empty or exact `"accumulate"` before simulation/export, and shipped converters emit the canonical value. Central normalization in `LoadTraceV2` is a reasonable consistency improvement but is orthogonal to this guard.
- **G2/G5/G6/G15**: negative `--concurrent-sessions` is rejected upstream; no output artifact is created before the guard; whitespace paths are non-empty so they still trip the guard; `Fatalf` is the established `cmd/` validation convention (prohibited only in `sim/`).
- **G10** (unanchored `-test.run`): only the intended top-level test matches today, and the child reaches `Fatalf` before any other matched test could run. Harmless; can anchor in a future cleanup.

Re-running `/archon-pr-review` + `/blis-pr-review` on the new commit.

namasl:

namasl:

## qa-review adjudication — PR #1623: ✅ PASS

|----|----|----|----|
| F2 | FLAW_FOUND | ✅ RESOLVED | The current documentation now explicitly states that replay rejects `--trace-output` both in pool mode and for plain closed-loop replay of an accumulate corpus. The run/replay comparison table also points readers to those restrictions. |
| G13 | CANNOT_ANSWER | 🟡 WAIVED (justification) | The reset-marker concern is concretely forward-looking and concerns unmerged code. The current TraceV2 schema contains `SessionContextGrowth` but no reset-marker field, and a repository-wide search finds no reset-marker implementation. GitHub issue/PR #1613 remains open and describes introducing `input_tokens_reset` specifically for accumulate replay. Therefore no reset-specific guard can be required or verified in this current tree; compatibility must be assessed when that schema change lands. |

### Still blocking

### Cleared
- **F2** (✅ RESOLVED) — The current documentation now explicitly states that replay rejects `--trace-output` both in pool mode and for plain closed-loop replay of an accumulate corpus. The run/replay comparison table also points readers to those restrictions.
  _docs/guide/observe-replay-calibrate.md:235 qualifies replay trace export with “see restrictions below”; docs/guide/observe-replay-calibrate.md:237-253 documents the pool-mode and closed-loop accumulate restrictions and the original-corpus alternative. These correspond to the enforced guards at cmd/replay.go:148-165._
- **G13** (🟡 WAIVED (justification)) — The reset-marker concern is concretely forward-looking and concerns unmerged code. The current TraceV2 schema contains `SessionContextGrowth` but no reset-marker field, and a repository-wide search finds no reset-marker implementation. GitHub issue/PR #1613 remains open and describes introducing `input_tokens_reset` specifically for accumulate replay. Therefore no reset-specific guard can be required or verified in this current tree; compatibility must be assessed when that schema change lands.
  _sim/workload/tracev2.go:17-67 defines the current TraceHeader, including session_context_growth at lines 31-37, without any reset capability; repository grep found no input_tokens_reset/InputTokensReset/reset-marker implementation. GitHub #1613 is OPEN and proposes the not-yet-merged reset field for accumulate context-growth replay. GitHub #1621 is OPEN and explicitly identifies reset-marker preservation as post-#1613, forward-looking behavior._

susiejojo:
### BLIS PR Self-Review — PR #1623 (independent re-review)

**Verdict: ✅ READY TO MERGE** — a clean, minimal, correctly-scoped stopgap. All defect claims independently verified against base code; both new tests pass; build/vet/CI green. No blocking issues.

I re-verified the whole chain from scratch (checked out the branch, re-ran tests, re-read the guarded export path) rather than relying on the prior claude/qa-review/archon rounds.

---

#### Correctness — all three issue claims verified against actual code

The guard (`cmd/replay.go:163`) sits immediately after the sibling pool guard:
```go
if traceData.Header.SessionContextGrowth == "accumulate" && replayConcurrentSessions == 0 && replayTraceOutput != "" {
```
The lossy export path it protects is real and reproduces exactly as the issue describes:
1. **Follow-ups dropped** — `cmd/replay.go:722` calls `RequestsToTraceRecords(requests)` on the round-0 slice; `followUpRequests` (collected at `:668–685`) is never in the export set. ✓
2. **Absolute (not delta) encoding** — `sim/workload/tracev2.go:819` emits `int(req.InputLen()) - prefixLen`, an absolute suffix length, which an accumulate loader would misread as a delta. ✓
3. **`session_context_growth` omitted** — the re-export header (`:723`) sets Version/TimeUnit/Mode/GoodputSLOTargets/KVOffload/ObservedKVMetrics but never `SessionContextGrowth`. ✓

**Guard logic is exact, not redundant.** The third conjunct `replaySessionMode == "closed-loop"` is correctly *omitted* because the upstream gate at `:129` already guarantees `accumulate ⇒ closed-loop`. The comment at `:159–162` documents this dependency explicitly. Negative `--concurrent-sessions` is rejected at `:105` before either guard, so no bypass. No output file is opened before the guard, so no partial artifact on a rejected run.

#### Invariants / standards
- **INV-13** (never silent degradation): converts a silently-unfaithful trace into a loud `Fatalf` — matches the autoscaler / node-pool / pool-mode precedent. ✓
- **INV-6** (determinism): guard is inert for every non-(accumulate + plain-closed-loop + trace-output) run; byte-identical elsewhere. ✓
- **R1** (no silent continue): fail-fast with an actionable message naming the flag, mode, and header field. ✓

Correctly confined to `blis replay`. `blis run` is unaffected because it sources records via `SetArrivalHook` (captures every arrival incl. follow-ups) and never emits accumulate corpora; `blis observe` records dispatched rounds directly with no `sim.Request → TraceRecord` round-trip. Parity table in the description is accurate.

#### Tests
Both `TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal` and `TestReplayCmd_PoolTraceOutputFatal` **pass on my checkout** and are genuinely behavioral: subprocess re-run (`BLIS_TEST_SUBPROCESS=1`, following the `AccumulateHeaderRequiresClosedLoop` precedent) asserting exit code 1 + stderr content. The pool test is the companion the sibling guard never had (issue-flagged). They assert observable behavior (exit + message), not internal structure — refactor-survivable.

#### Docs
The admonition (`observe-replay-calibrate.md:237–253`) documents both rejection cases with the correct workaround (re-export from the original converter corpus). The `--trace-output sim` → `--results-path sim.results.json` fix to the calibration example correctly resolves the separately-filed #1622 (the old example combined `--trace-output` with `--concurrent-sessions`, which the pool guard rejects, *and* mismatched `--sim-results`).

---

#### Non-blocking observations (no change required)

1. **Residual (documented, in-scope) gap:** the guard keys on `accumulate`, but the *follow-up-dropping* defect (issue item 1) is general to **any** multi-round closed-loop trace. A **non-`accumulate`** multi-round session corpus re-exported via `--trace-output` would still silently drop follow-ups (without the extra delta-misencoding). This is consistent with the issue's declared option-(b) scope, but it means the silent-truncation hole is not *fully* closed. The faithful-re-export follow-up (option (a)) is the right place to close it; worth a one-line note or tracking issue so it isn't forgotten.
2. **G1 (already adjudicated):** `SessionContextGrowth == "accumulate"` is an exact string match; a capitalized/whitespace value isn't normalized at parse time. As qa-review established, such values can't reach export in closed-loop (`LoadTraceV2SessionBlueprints` rejects anything but empty/exact-`accumulate` upstream), so this is not exploitable via producer-generated traces. Central normalization in `LoadTraceV2` would be a nice orthogonal hardening.
3. **Reset markers (#1613):** correctly deferred — that field doesn't exist in this tree; the accumulate-keyed guard will need revisiting when #1613 lands (only covers reset traces if every reset trace also carries `accumulate`). Already tracked in #1621.

**Prior gates:** archon `NO_CHANGE` (fast-track eligible), qa-review adjudication PASS, CI all-green (build + all test shards + lint). Consistent with my findings.

namasl:
Thanks for the independent re-review (checking out the branch and re-running the chain from scratch, rather than trusting the prior rounds, is exactly the right rigor) — and for the sharp non-blocking observations.

**Observation 1 (residual non-`accumulate` gap) — tracked in #1630.** You're right that the follow-up-dropping defect is general to *any* multi-round closed-loop trace, whereas this guard keys on `accumulate`, so a non-`accumulate` multi-round corpus re-exported in plain closed-loop still silently drops follow-ups. This is within the issue's declared option-(b) scope, but it shouldn't be forgotten — so I've filed #1630 (faithful closed-loop re-export, option (a)) which captures this residual gap explicitly and notes *why* broadening the fail-fast isn't the right fix (it would over-guard legitimately-faithful single-round / no-session re-exports, and detecting "has follow-ups" needs post-blueprint knowledge). #1630 also owns removing both fail-fast guards once faithful re-export lands.

**Observation 2 (G1, exact-string match):** agreed and already adjudicated — capitalized/whitespace growth values can't reach export in closed-loop (`LoadTraceV2SessionBlueprints` rejects anything but empty/exact-`accumulate` upstream), and shipped converters emit the canonical value. Central normalization in `LoadTraceV2` is a reasonable orthogonal hardening; noted on #1630 as adjacent cleanup rather than blocking this stopgap.

**Observation 3 (reset markers / #1613):** correctly deferred — that field doesn't exist in this tree; #1630 carries the "revisit the accumulate-keyed guard when #1613 lands" note.

No code change for this PR (option-(b) scope stands); the residual work is now tracked in #1630. Appreciate the thorough pass.

susiejojo:
### Follow-up: interaction with the `convert weka → observe → replay → calibrate` flow

For the common pipeline where **replay only ever consumes observe-produced traces**, this guard is **unreachable by construction** — worth stating explicitly so it isn't mistaken for a new restriction on that flow:

- Observe has a single output-header build site (`cmd/observe_cmd.go:694`) that hardcodes `Mode: "real"` and **never sets `SessionContextGrowth`** — even in corpus mode (`observe_corpus.go` only *reads* the `--corpus-*` input; it never writes growth onto the output header). The field is `omitempty`, so it's absent.
- The guard's first conjunct is `Header.SessionContextGrowth == "accumulate"`; `"" == "accumulate"` is false, so replay of an observe trace proceeds normally regardless of `--trace-output` / `--concurrent-sessions`.
- Observe also records **absolute** per-round token counts (`observe.go:532`, `suffix = total − prefix`), not accumulate deltas — it doesn't round-trip a converter corpus through `sim.Request → TraceRecord`, so the loss class this guard protects against doesn't arise for observe output at all.

The accumulate encoding lives only in the `convert weka → observe` hop (as observe's **input** corpus); observe "launders" it into plain `Mode:"real"` absolute observed timings. By the time replay runs on that trace, nothing accumulate-shaped remains.

Net: this guard is a safety net for **replaying a converter corpus directly** with `--trace-output` — not for the observe-trace replay path — so it's inert for that pipeline. (Reminder unchanged from the review: replay's calibrate leg should use `--results-path`, which `blis calibrate --sim-results` consumes; `--trace-output` writes a TraceV2 trace, not the SimResult JSON.)

namasl:
Confirmed — your analysis is exactly right, and thanks for tracing it end-to-end.

The guard is inert for the `convert weka → observe → replay → calibrate` pipeline: observe's single output-header site hardcodes `Mode: "real"` and never sets `SessionContextGrowth` (corpus mode only *reads* the `--corpus-*` input), and observe records absolute per-round suffix counts — so `Header.SessionContextGrowth == "accumulate"` is `false` on any observe trace and replay proceeds normally regardless of `--trace-output` / `--concurrent-sessions`. The accumulate encoding exists only as observe's *input* corpus; observe launders it into plain `Mode:"real"` absolute timings, so nothing accumulate-shaped reaches replay. This matches the PR's cross-path parity table (observe unaffected) and qa-review G7.

On the "worth stating explicitly" point: the admonition is deliberately scoped to name only the two *rejected* cases (accumulate-corpus plain closed-loop, and pool mode), so it doesn't implicate the observe path by its own terms — I held off adding a "observe traces are unaffected" line to avoid documenting-what-isn't-restricted on an otherwise ready PR. Happy to add one explicit sentence to the admonition if you'd prefer it be called out for the weka→observe→replay audience — just say the word and I'll push it.

(And yes — the calibrate leg uses `--results-path` for `--sim-results`; that's the example fix folded in here as #1622.)

susiejojo (approved):
lgtm

PR #1629, opened 2026-08-26
feat(latency): inter-node network cost for cross-node TP/EP traffic (#1530)

### LINKED ISSUE #1530: Inter-node network cost not modeled (cross-node TP/EP traffic is free)

namasl:
## Summary

- BLIS never charged for cross-node collective traffic: `tpAllReduceBasis` (TP all-reduce) and `moeDispatchBasis` (DEP expert all-to-all) both divide byte volume by `bwHbmUs` (on-package HBM/NVLink), so any multi-node TP or Wide-EP (`multi_node_dep`) deployment carried **zero interconnect penalty** — optimistic in a way that's invisible to the user.
- This PR adds a minimal inter-node network cost to the **trained-physics** latency model. A new `sim.NetworkConfig` (`--gpus-per-node`, `--inter-node-bandwidth` GB/s, `--inter-node-latency` ms) charges a collective whose participant group exceeds `--gpus-per-node`: its cross-node fraction `(n-1)/n` of the ring/all-to-all byte volume is charged at the inter-node bandwidth (an effective rate, like `--pd-transfer-bandwidth`; **no learned β** — β₄ absorbs the NVLink/HBM ratio, which must not re-apply to the fabric) plus a fixed per-collective base latency.
- Delivered boundary: **both** comm legs (they share one mechanism) — multi-node TP all-reduce (group = `tp`) **and** DEP all-to-all (group = `moeGroup` = `TP·DP`).

## Issue

Closes #1530 · Part of #1532 (GLM-5.2 support).

## Design notes (source-document audit)

The issue body says "implement after #1529" (multi-node placement, **not yet merged**), while the parent tracking issue #1532 says "#1530 is the foundation … buildable and unit-testable first … suggested order: #1530 → then #1529/#1531." Resolved toward the parent's guidance:

- **Topology is a config input, not a placement-derived signal.** `--gpus-per-node` + the existing `--tp`/`--dp` degrees determine whether a collective crosses a node boundary — no dependency on #1529's multi-node placement. This makes the mechanism unit- and integration-testable today, and it is the exact seam #1529 will later drive from real GPU spans.
- **No `NewModelHardwareConfig` churn:** injected through a `latency.WithNetworkConfig` functional option (mirroring `WithAdapterCost`), carried on `SimConfig` as an 8th zero-value-inert sub-config (mirroring `LoRAConfig`).
- Two comm bases (`tpAllReduceBasis`, `moeDispatchBasis`) were refactored to expose their raw byte volume (`*Bytes`) so the same volume feeds both the intra-node basis (÷ `bwHbmUs`) and the new cross-node term (÷ inter-node bandwidth); the operation order is preserved exactly, so the existing terms are byte-identical (guarded by the trained-physics golden + INV-BC-DP1 suites, which pass unchanged).

## Cross-Path Parity (run / replay / observe)

| `blis run` | yes | yes | `--gpus-per-node` / `--inter-node-bandwidth` / `--inter-node-latency` |
| `blis replay` | yes | yes | Same flags via the shared `registerSimConfigFlags`; INV-13 holds by construction (identical flags → identical config → identical `StepTime`). No TraceV2 round-trip needed (signal is not `node_pools`-derived). |
| `blis observe` | no | N/A | Takes timing from the real server; unaffected. |

## Invariants

- **INV-6 / INV-BC-DP1:** inert by default (`--gpus-per-node 0`) and when a group fits in one node → the term is exactly `0.0` → `StepTime` byte-identical to a pre-feature build; dense `DP=1` unchanged across the TP matrix.
- **INV-13:** run/replay reconstruct the same `StepTime` from identical CLI flags; a run→replay parity test covers a cross-node config.
- **R1:** `--gpus-per-node` requires `--latency-model trained-physics`; an active config without a positive `--inter-node-bandwidth`, or fabric knobs set without `--gpus-per-node`, is a hard error (never a silent no-op).

## Test Plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (all packages green, incl. trained-physics golden + INV-BC-DP1)
- [x] `golangci-lint run ./sim/... ./cmd/...` (0 issues)
- [x] **AC1** `TestInterNodeCost_BC1_CrossNodeCharged` — cross-node > single-node for both legs
- [x] **AC2** `TestInterNodeCost_BC2_MonotonicFabricQuality` — monotonic in bandwidth and latency
- [x] **AC3** `TestInterNodeCost_BC3_NoOpByteIdentical` — inert + fits-in-node byte-identical (incl. dense DP=1)
- [x] **AC4** `TestInterNodeCost_BC4_Deterministic`
- [x] **INV-13** `TestParity_InterNodeNetwork_RunReplay_INV13`
- [x] **R1 guards** `TestInterNodeCost_BC6_BackendAndConfigGuards`, `TestResolveNetworkConfig_HappyPaths`, `NetworkConfig.Validate` table
- [x] End-to-end smoke: `blis run … --tp 8 --gpus-per-node 4 --inter-node-bandwidth 50` raises mean E2E vs the same run without the flags; all three misconfig guards fatal with clear messages.

## Not in scope (documented)

Hierarchical-overlap refinement of the (currently serial-additive, per-layer) cost model; deriving `--gpus-per-node` from real multi-node placement (#1529). `blis observe` does not model the fabric.

---

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,18 @@ go build -o blis main.go
 # Run with default model
 ./blis run --model qwen/qwen3-14b
 
+# Run with inter-node network cost (#1530, trained-physics only). --gpus-per-node sets the
+# node size; a TP collective (group=tp) or DEP expert all-to-all (group=TP·DP) larger than it
+# is charged for its cross-node hops at --inter-node-bandwidth (GB/s) + --inter-node-latency
+# (ms), instead of on-package NVLink. Omitting --gpus-per-node (default 0) is inert and
+# byte-identical to a pre-feature build. Same flags on `blis replay` reproduce the StepTime
+# (INV-13). Multi-node TP all-reduce example (tp=8 spanning 2 nodes of 4 over ~50 GB/s IB):
+./blis run --model qwen/qwen3-14b --hardware H100 --tp 8 \
+  --gpus-per-node 4 --inter-node-bandwidth 50 --inter-node-latency 0.005
+# Wide-EP (multi_node_dep) example: TP within node, DP across nodes (moeGroup=TP·DP spans):
+./blis run --model deepseek-ai/deepseek-v2-lite --hardware H100 --tp 4 --dp 2 \
+  --enable-expert-parallel --gpus-per-node 4 --inter-node-bandwidth 50
+
 # Run with goodput SLO targets (#1413). --slo-ttft / --slo-itl / --slo-e2e accept
 # class=duration[,class=duration...] using Go duration syntax. Precedence:
 # CLI > trace header > workload spec. Distinct from --slo-targets (dispatch ordering).
@@ -614,6 +626,8 @@ Two latency model modes (trained-physics, roofline), selected via `--latency-mod
 **Roofline model**: Pure analytical model available via explicit `--latency-model roofline` flag. Useful when you want FLOPs/bandwidth-based estimation without learned corrections.
 
 See [`docs/guide/latency-models.md`](docs/guide/latency-models.md) for details.
+
+**Inter-node network cost (#1530, trained-physics only)**: By default the trained-physics comm terms (`tpAllReduceBasis` TP all-reduce, `moeDispatchBasis` DEP expert all-to-all) divide byte volume by `bwHbmUs` (on-package HBM/NVLink), so cross-node collective traffic is free. `sim.NetworkConfig` (`--gpus-per-node`, `--inter-node-bandwidth` GB/s, `--inter-node-latency` ms) charges it: a collective whose participant group exceeds `--gpus-per-node` (TP group = `tp`; MoE all-to-all group = `moeGroup` = `TP·DP`) is scored as spanning `ceil(group/gpus_per_node)` nodes, and its cross-node fraction `(n-1)/n` of the ring/all-to-all volume is charged at the inter-node bandwidth (an effective rate, like `--pd-transfer-bandwidth`; **no learned β** — β₄ absorbs the NVLink/HBM ratio, which must not re-apply to the fabric) plus a fixed per-collective latency. Injected via the `latency.WithNetworkConfig` option (no `NewModelHardwareConfig` churn). **Inert by default** (`--gpus-per-node 0`): the term is exactly `0.0`, so every config expressible today is byte-identical to a pre-feature build (INV-6/INV-BC-DP1) — the guarantee is conditional on the network model being off or the group fitting in one node. Topology is a **config input, not a placement-derived signal** (mirroring `--tp`/`--dp`): both `blis run` and `blis replay` build it from the same flags, so run/replay parity (INV-13) holds by construction with no TraceV2 round-trip. `--gpus-per-node` requires `--latency-model trained-physics`, and an active config without a positive `--inter-node-bandwidth` (or fabric knobs set without `--gpus-per-node`) is a hard CLI error (R1). This is the foundation the multi-node placement work (#1529) will drive from real GPU spans; until then the node span is declared. `blis observe` takes timing from the real server, so it is unaffected.
 
 **Quantized model support**: Three-tier auto-detection of weight precision: (1) `quantization_config` in HF `config.json` — GPTQ/AWQ (`bits`), FP8 (implicit), compressed-tensors (`config_groups.*.weights.num_bits`); (2) model name conventions (`w4a16` → 0.5, `FP8` → 1.0 via `InferWeightBytesFromModelName`); (3) fallback to `BytesPerParam` from `torch_dtype`. Uses quantized weight precision for weight bandwidth and KV capacity calculations while keeping compute dtype for KV cache and activations. `ModelConfig.WeightBytesPerParam` (0=fallback to `BytesPerParam`) with `EffectiveWeightBytesPerParam()` accessor decouples weight storage precision from compute/KV dtype.
 
--- a/cmd/network_cost_parity_test.go
+++ b/cmd/network_cost_parity_test.go
@@ -1,0 +1,197 @@
+package cmd
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+
+	"github.com/inference-sim/inference-sim/sim/workload"
+	"github.com/spf13/cobra"
+)
+
+// replayTraceWithNetwork replays a TraceV2 through replayCmd at the given TP and
+// inter-node network config (#1530). When gpusPerNodeVal == 0 the network flags are
+// omitted (inert). Modeled on replaySpecTrace; adds --tp / --gpus-per-node /
+// --inter-node-bandwidth so INV-13 parity can be exercised for a cross-node config.
+//
+// NOTE: mutates package-level CLI vars; not for t.Parallel().
+func replayTraceWithNetwork(t *testing.T, traceHeaderFile, traceDataFile string, tpVal, gpusPerNodeVal int, bwGBps float64) []workload.SimResult {
+	t.Helper()
+	tmpDir := t.TempDir()
+	resultsFile := filepath.Join(tmpDir, "results.json")
+	mcFolder, hwPath, defaultsPath := setupTrainedPhysicsTestFixturesWithDefaults(t)
+
+	orig := captureCmdLevelVars()
+	defer orig.restore()
+
+	origTraceHeader := traceHeaderPath
+	origTraceData := traceDataPath
+	origSessionMode := replaySessionMode
+	origGpusPerNode := gpusPerNode
+	origInterNodeBw := interNodeBandwidth
+	origInterNodeLat := interNodeLatency
+	defer func() {
+		traceHeaderPath = origTraceHeader
+		traceDataPath = origTraceData
+		replaySessionMode = origSessionMode
+		gpusPerNode = origGpusPerNode
+		interNodeBandwidth = origInterNodeBw
+		interNodeLatency = origInterNodeLat
+	}()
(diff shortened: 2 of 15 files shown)


namasl:

namasl:
Closing without merge. This standalone/config-driven foundation is being set aside: per the #1532 dependency ordering, inter-node network cost should derive its cross-node topology from real multi-node placement (#1529 / PR #1537), which has not yet merged. The work will be restarted on top of that placement once #1537 lands. Not deleting via branch-delete-first so the close reason is on record.

Issue #1633, opened 2026-08-27
specs/**/*.plan.json is silently gitignored — archon plan cannot be committed


mtoslalibu:
## Summary

`.gitignore:44` ignores `*.json` repo-wide. None of its negations cover the compiled archon plan, so `specs/<feature>/*.plan.json` is silently untrackable. The archon-based planning flow introduced in #1625 depends on that file being committed to the feature branch, so PR0 of any planned feature appears to succeed while omitting the one artifact every later PR needs.

## Reproduction

On `issue-1624-sdlc-modernize` (or any branch), after `archon-go plan compile` writes a plan:

```
$ git check-ignore -v specs/008-hello-greeting/feature.plan.json
.gitignore:44:*.json	specs/008-hello-greeting/feature.plan.json
```

`git add specs/008-hello-greeting/` then stages `design.md` and `feature.archon` but **silently drops `feature.plan.json`** — no warning, and `git status` shows nothing missing.

## Why it matters

The compiled plan is the machine-readable half of the methodology:

- `docs/contributing/archon/pr-review-context.md` has CI read the plan off the base branch to run `archon-go pr-review --plan`.
- Every sub-issue generated by `docs/contributing/templates/rfc-to-plan.md` carries an `archon-plan: specs/<feature>/<name>.plan.json` line pointing at that exact path.
- `dist` tracking (baseline → 0) is the merge gate for a multi-PR feature.

Failure mode: PR0 lands and looks correct. The gap surfaces one or more PRs later, when plan-aware review either errors on a missing file or degrades to a plan-less delta review. The cause — a `.gitignore` line unrelated to archon — is far from the symptom.

Existing negations cover `testdata/`, `model_configs/*/config.json`, `hardware_config.json`, `specs/**/testdata/*.json`, and `cmd/testdata/**/*.json`, which is why this was not noticed: `specs/` is partially negated already, just not for plan files.

## Suggested fix

Add a negation next to the others in `.gitignore`:

```gitignore
# compiled archon plans (read from the repo by /archon-pr-review)
!specs/**/*.plan.json
```

`git add -f` also works per-file but leaves the trap armed for the next feature, so the negation is preferable.

Optionally, one line in `docs/contributing/archon/encode-plan.md` noting the compiled plan must be tracked, since the sub-issue templates assume it is present on the branch.

## How this was found

[...]

PR #1625, opened 2026-08-25
docs: modernize SDLC — archon-based planning, remove stale methodology

mtoslalibu:
## What this PR achieves

Establishes a lightweight, Claude-assisted design/planning methodology using archon, replacing unused infrastructure. The flow after this PR:

    Bug fix / small feature:
      Issue -> pr-workflow.md (implement) -> blis-pr-review + archon-pr-review (FAST-TRACK)

    Large feature (new package boundaries):
      Tracking issue with RFC (rfc.md) -> team discusses -> encode .archon + sub-issues (rfc-to-plan.md)
      -> PR0 persists plan -> PR1+ per hole (pr-workflow.md) -> archon-pr-review --plan -> blis-pr-review 
      -> merge when verdict is ok and delta_dist <= 0 + tests pass

User experience (what people type):
- Bug fix: `Resolve issue #N as per @docs/contributing/pr-workflow.md`
- Large feature RFC: `Follow @docs/contributing/rfc.md to design feature X`
- Large feature planning: `Follow @docs/templates/rfc-to-plan.md. Agreed design is in #N`
- PR review: `@claude /blis-pr-review` + `/archon-pr-review` (CI runs both)

## Removed files

| File | What it was | Why removed |
|------|-------------|-------------|
| .claude/commands/speckit.*.md (9) | Alternative SDD slash commands | Never used |
| .specify/ (12 files) | Speckit templates/scripts | Never adopted |
| .claude/skills/implement-issue/ | Auto-implement issues | Not used recently |
| .claude/skills/convergence-review/ | 10-perspective convergence loop | Never triggered; perspectives moved to perspectives.md |
| .claude/checkpoint.md | Session scratch | Stale |
| .bob/ + docs/bob-commands/ | Legacy review harness | Replaced by blis-pr-review |
| docs/contributing/convergence.md | Convergence protocol doc | Content inlined; never linked from active workflows |
| docs/contributing/design-process.md | 7-step design process | Replaced by rfc.md |
| docs/contributing/macro-planning.md | Macro PR decomposition | Replaced by rfc-to-plan.md |
| docs/contributing/templates/macro-plan-prompt.md | Claude prompt for macro plans | Replaced by rfc-to-plan.md |
| docs/contributing/templates/micro-plan-prompt.md | Claude prompt for micro plans | Informal prompts used; template kept |
| docs/guide/experimentation.md | Hypothesis guide page | Covered by hypothesis.md |

## Added files

| File | Purpose | Key content |
|------|---------|-------------|
[...]

PR #1642, opened 2026-08-28
feat(ci): auto-detect archon-plan for dist tracking in PR reviews

### LINKED ISSUE #1631: feat(ci): auto-detect archon-plan for dist tracking in PR reviews
## Summary

`.github/workflows/archon.yml` runs `archon-go pr-review` delta-only. Teach it to detect an `archon-plan: <path>` declaration and add `--plan` to that same invocation, so multi-PR features get dist tracking and a plan verdict.

One command, one comment, in every case.

Follow-up to #1624. Docs are already in place; this is the CI half.

**Depends on #1625** — that PR lands two prerequisites:

- `.archon-version` (currently `v0.5.0`), read by `scripts/archon-build.sh`, replacing the hardcoded `v0.2.0` fallback in the workflow. `--plan` needs v0.5.0: earlier tags report `EXCEEDS` for nearly every PR.
- `.gitignore` negation `!specs/**/*.json`, without which the compiled plan is silently untrackable (#1633) and no plan ever reaches a branch.

Do not test this before #1625 is on `main`, since `issue_comment` runs the default branch's workflow.

## Current behavior

```bash
archon-go pr-review . "$BASE" "$HEAD" --out .archon
```

## Target behavior

One invocation, with `--plan` added when a plan is found:

```bash
# plan found
archon-go pr-review . "$BASE" "$HEAD" --plan /tmp/archon-plan.json --out .archon

# no plan declared (unchanged)
archon-go pr-review . "$BASE" "$HEAD" --out .archon
```

`--plan` adds `### G5 — Plan distance ratchet` to `review.md`; everything else is identical. Both forms produce all three views (component, witness delta, interface-contract) with no extra flags.

## Cases

| Case | Target | Plan file | CI does |
|---|---|---|---|
| PR0 — create branch + persist plan | no PR (direct push) | — | nothing; there is no PR to review |
| Hole PR (PR1..N) | feature branch | on the base (put there by PR0) | 3 views + plan checks |
| Final PR | `main` | on the head only — never on base | 3 views + plan checks |
| Normal PR (bug fix, small feature) | `main` | none anywhere | 3 views only, silently |

## Logic

1. Always run `archon-go pr-review`. That alone gives the three views.
2. Look for an `archon-plan: <path>` line in the PR body. If absent, look in the body of the issue the PR closes.
3. **No line found** → run without `--plan`. Three views, no warning. This is the normal-PR case and must stay quiet.
4. **Line found** → fetch the file with `git show`: try the base commit first, then the head commit.
   - Got it → add `--plan /tmp/archon-plan.json`. Three views plus dist and verdict. Report which commit it came from.
   - Could not get it → run without `--plan` **and post a warning line** saying the path was declared but unreadable. A PR that asks to be checked against a plan and silently isn't is indistinguishable from a PR with no plan, which quietly removes dist as the merge gate.

**Why base-then-head:** for a hole PR the base *is* the feature branch, so the plan comes from base and a PR cannot edit the plan to make itself pass. For the final PR, `main` has no plan and only the head does, so head is the only option — acceptable because that plan was already reviewed when PR0 landed. Emitting the plan's source commit in the comment keeps that visible.

## Two things to get right

**1. Read the plan from git, not the filesystem.**

`issue_comment` workflows always check out the **default branch**, so the working tree is `main`. The plan lives on the feature branch, so a repo-relative path fails and `test -f` always reports missing:

```
archon-go: read plan specs/008-x/feature.plan.json: no such file or directory
```

`fetch-depth: 0` plus the existing `git fetch origin pull/N/head` means both commits' objects are local, so extract the blob and pass the temp path:

```bash
git show "$BASE:$PLAN_PATH" > /tmp/archon-plan.json
```

Verified locally: byte-identical to the feature-branch copy, and `--plan` accepts a path outside the repo. This also avoids checking out PR head code, which matters for fork PRs.

**2. Treat the declared path as untrusted input.**

The path comes from a PR or issue body, which anyone can write, and this workflow runs on a self-hosted runner with `pull-requests: write`.

- Read the body inside `actions/github-script` and hand it out via `$GITHUB_OUTPUT` → `env:`. Never interpolate `${{ github.event.issue.body }}` into a `run:` block.
- Validate the path the way the existing archon-ref check does: `^[A-Za-z0-9._/-]+$`, reject `..`, reject a leading `-` (flag injection into `--plan`).
- Read it only with `git show`, which cannot escape the tree. A filesystem fallback would become an arbitrary file read exfiltrated into a public PR comment.

## Detection details

- Closing issue: use the PR's `closingIssuesReferences` (GraphQL), not a `Closes #N` regex over the body.
- Multiple `archon-plan:` lines: first match wins.
- `$BASE` is the **merge-base** the workflow already computes, not `pr.base.sha`. `$HEAD` is the fetched PR head SHA.

## Verification

Confirmed by hand on a two-hole test feature (PR against a feature branch, plan persisted by PR0):

```
dist(P,G): 4 → 2 — OK
Plan verdict: EXCEEDS
```

### LINKED ISSUE #1643: Bug: archon.yml posts from a fixed /tmp path, so concurrent reviews can swap comment bodies
## Location

`.github/workflows/archon.yml` — the `archon-review` job.

## Impact

The comment body is written to a fixed path, `/tmp/archon-output.md`, and posted from there by a later step:

```yaml
      - name: Run archon pr-review
        run: scripts/archon-review.sh      # OUTPUT_FILE: /tmp/archon-output.md
      - name: Post PR comment
        run: gh pr comment "$PR_NUMBER" --body-file /tmp/archon-output.md
```

The runner is self-hosted, so that is the same file on the same machine for every run. Two reviews in flight at once can interleave:

```
review for PR #10  →  writes /tmp/archon-output.md
review for PR #20  →  overwrites /tmp/archon-output.md
review for PR #10  →  posts PR #20's review body
```

The result is a review comment on the wrong PR, with nothing in the logs to suggest anything went wrong.

## Suggested fix

Derive the path from `$RUNNER_TEMP`, which GitHub allocates per job, instead of `/tmp`:

```yaml
        env:
          OUTPUT_FILE: ${{ runner.temp }}/archon-output.md
```

and read the same path in the `Post PR comment` step. `archon-review.sh` already takes the path as `OUTPUT_FILE`, so no script change is needed.

## Discovery context

Found while reviewing the plan for #1631. Out of scope there, so filed rather than fixed.

#1642 covers the adjacent half of this: it truncates the output file up front, so a *stale* body left by an earlier run can no longer be posted when a later run aborts before composing its own. Two runs racing each other is what remains.

mtoslalibu:
## Summary

`/archon-pr-review` reviewed every PR as an isolated diff. Now, when a PR declares which archon plan it belongs to, `--plan` is added to that same invocation — so a planned multi-PR feature gets a dist ratchet and a plan verdict.

One command, one comment, in every case.

```
plan = first `archon-plan: <path>` line in the PR body, else in a closing issue body
if <path> is a committed .json file at the base branch tip, else at the PR head:
  archon-go pr-review . $BASE $HEAD --plan <extracted> --out .archon   # 3 views + dist
else:
  archon-go pr-review . $BASE $HEAD --out .archon                      # 3 views
```

| Case | Target | Plan file | CI does |
|---|---|---|---|
| PR0 — create branch + persist plan | no PR (direct push) | — | nothing; there is no PR to review |
| Hole PR (PR1..N) | feature branch | on the base | 3 views + plan checks |
| Final PR | `main` | on the head only | 3 views + plan checks |
| Normal PR (bug fix) | `main` | none | 3 views only, silently |

Builds on #1625 (merged), which pinned archon at v0.5.0 and added the `!specs/**/*.json` negation without which a compiled plan cannot be committed (#1633).

## Behavioral contracts

**Positive** — BC-1 quiet by default (no note, no warning, unchanged output). BC-2/BC-3 plan resolved from the base branch tip, else the PR head. BC-4 the base copy wins. BC-5 declaration found in a closing issue. BC-6 the earliest declaration wins. BC-7 one invocation, one comment. BC-8 a head-sourced plan is labelled as not independently verified.

**Negative** — BC-9 unsafe paths refused (absolute, `..`, leading `-`, over 256 chars, outside `[A-Za-z0-9._/-]`, not `.json`). BC-10 a declared path is never dereferenced through the filesystem.

**Error handling** — BC-11 declared but unusable is loud. BC-12 an unusable base copy is not silently replaced by the head copy. BC-13 a failing plan-aware review still yields the three views. BC-14 an unreachable commit is an error, not an absent plan. BC-15 pre-existing failure reporting survives. BC-16 a resolver that cannot run does not swallow the review.

## Why the review step moved into a script

Over half those contracts describe *what gets posted*, which no test can reach inside a YAML `run:` block. `scripts/archon-review.sh` is the same logic in a file a stub `archon-go` can drive, so the pre-existing merge-base and missing-bundle paths are now tested too — they never were before.

## Three things worth a reviewer's attention

**The declared path is untrusted.** It comes from a PR or issue body, and this workflow runs on a self-hosted runner with `pull-requests: write`. The text crosses as file *contents*, never a shell argument; the path is validated before any use; it is dereferenced only with git plumbing against an explicit commit, with object type, mode, and size gated before a byte is written; and anything echoed back has out-of-allowlist characters replaced.

**Base is preferred over head, and there are three ways that could have silently collapsed.** A hole PR racing the base (BC-4), a base copy that exists but is broken (BC-12), and a base commit git has never fetched (BC-14). `git ls-tree` prints nothing for the last two, exactly as it does for an absent file, so each needed an explicit check. Each has a test that fails if the check is removed.

**The workflow never checks out the PR head**, so a PR cannot get its own copy of these scripts executed. That property is load-bearing and now documented in `archon-review.sh`: adding a `ref:` to the checkout step would turn it into arbitrary code execution on the runner.

## Testing

`scripts/` gains a test-only Go package driving both scripts against throwaway git repositories and a stub archon. `ci.yml` lists packages explicitly rather than `./...`, so a `./scripts` matrix entry was needed or none of it would run.

```
go build ./...          exit 0
go test ./... -count=1  exit 0   (all packages ok; ./scripts 40 cases)
```

Six mutations were applied to confirm the tests discriminate rather than merely pass:

| Mutation | Test that caught it |
|---|---|
| prefer head over base | `TestResolve_PlanOnBoth_PrefersBase` |
| read the path from the filesystem instead of git | `TestResolve_PlanOnlyInWorkingTree_ReportsError` |
| drop the commit-reachability check | `TestResolve_UnknownBaseCommit_ReportsError` |
| drop the retry without `--plan` | `TestReview_PlanAwareRunFails_RetriesWithoutPlan` |
| stop clearing `.archon` between invocations | `TestReview_RetryProducesNoBundle_DoesNotPostTheFailedRun` |
| drop the plan warning from failure bodies | `TestReview_UnusablePlan_WarnsAndReviewsWithoutPlan` |

The first pass of the `.archon`-clearing test did *not* catch its mutation — the stub rewrote the bundle on every call, masking it. That is why a second test exists for it.

`golangci-lint` was not run locally: the installed binary is v1.53.3 (built with Go 1.20) and cannot read Go 1.24 export data. CI runs it.

## Docs corrected

`pr-review-context.md` documented a design that was never built — two invocations with both outputs posted, base-branch-only lookup, and a silent fall back to delta-only. All three are wrong and are fixed here. The `.json` and must-be-committed requirements were implicit at all four sites that define the convention; `archon-issue-examples.md` and `rfc-to-plan.md` now also say the bracketed placeholders must be substituted, since a literal `specs/[NNN-feature]/...` is rejected.

## Also fixed here

`archon.yml` posted the comment from a fixed `/tmp/archon-output.md`. Adding `if: always()` to the comment step made that reachable in a new way: when a step *before* the review fails, the review script never runs to truncate the file, so a body left by an earlier job on that host would be posted — possibly another PR's review. The path now lives under `${{ runner.temp }}`, referenced from both steps (env does not carry between steps). Closes #1643.

Fixes #1631

### DIFF
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -96,6 +96,7 @@ jobs:
             });
             core.setOutput('base_sha', pr.base.sha);
             core.setOutput('head_sha', pr.head.sha);
+            core.setOutput('body', pr.body || '');
 
       - name: Fetch PR head ref
         run: git fetch origin "pull/${{ github.event.issue.number }}/head"
@@ -103,19 +104,18 @@ jobs:
       - name: Collect plan declarations
         id: decl
         uses: actions/github-script@v8
+        env:
+          PR_BODY: ${{ steps.refs.outputs.body }}
         with:
           script: |
             const fs = require('fs');
             const os = require('os');
             const path = require('path');
-            const { data: pr } = await github.rest.pulls.get({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              pull_number: context.issue.number
-            });
             // PR body first: the earliest declaration wins, so a PR's own declaration
-            // outranks one in a third-party issue it happens to close.
-            const parts = [pr.body || ''];
+            // outranks one in a third-party issue it happens to close. In practice the PR
+            // body is usually the ONLY source: GitHub links closing issues only for a PR
+            // targeting the default branch, so a hole PR against a feature branch links none.
+            const parts = [process.env.PR_BODY || ''];
             const res = await github.graphql(
               `query($owner: String!, $repo: String!, $num: Int!) {
                  repository(owner: $owner, name: $repo) {
@@ -126,8 +126,13 @@ jobs:
                }`,
               { owner: context.repo.owner, repo: context.repo.repo, num: context.issue.number }
             );
-            for (const node of res.repository.pullRequest.closingIssuesReferences.nodes) {
-              parts.push(node.body || '');
+            // Tolerate a partial response rather than failing the step: losing the fallback
+            // costs a warning on the PR, whereas failing here costs the whole review.
+            const nodes = res?.repository?.pullRequest?.closingIssuesReferences?.nodes;
+            if (nodes) {
+              for (const node of nodes) parts.push(node.body || '');
+            } else {
+              core.warning('Could not read closing issues; using the PR body alone.');
             }
             // Written as file CONTENTS, never handed to a shell as an argument.
             const dir = fs.mkdtempSync(path.join(process.env.RUNNER_TEMP || os.tmpdir(), 'archon-'));
@@ -141,22 +146,27 @@ jobs:
           BASE_SHA: ${{ steps.refs.outputs.base_sha }}
           HEAD_SHA: ${{ steps.refs.outputs.head_sha }}
           DECL_FILE: ${{ steps.decl.outputs.file }}
-          OUTPUT_FILE: /tmp/archon-output.md
+          OUTPUT_FILE: ${{ runner.temp }}/archon-output.md
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: scripts/archon-review.sh
 
-      # always(): the review script exits 0 for every reviewed outcome, but an environment
-      # error exits non-zero, and a trigger that silently posts nothing is the failure mode
-      # this feature exists to avoid. The script truncates the output file up front, so an
-      # aborted run leaves it empty and the guard below skips rather than posting a stale body.
+      # always(): the review script reports every reviewed outcome through its output file and
+      # exits 0, but an environment error exits non-zero, and a trigger that silently posts
+      # nothing is the failure mode this feature exists to avoid.
+      #
+      # The body lives under runner.temp, which the runner empties at the start of every job.
+      # A fixed /tmp path would not be safe here: when a step BEFORE the review fails, the
+      # review script never runs to truncate it, and this step would post whatever the
+      # previous job on this self-hosted runner left behind — possibly another PR's review.
       - name: Post PR comment
         if: always()
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.issue.number }}
+          OUTPUT_FILE: ${{ runner.temp }}/archon-output.md
         run: |
-          if [[ ! -s /tmp/archon-output.md ]]; then
+          if [[ ! -s "$OUTPUT_FILE" ]]; then
             echo "::warning::Archon produced no output. Skipping PR comment."
             exit 0
           fi
-          gh pr comment "$PR_NUMBER" --body-file /tmp/archon-output.md
+          gh pr comment "$PR_NUMBER" --body-file "$OUTPUT_FILE"
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -116,23 +116,33 @@ jobs:
             // body is usually the ONLY source: GitHub links closing issues only for a PR
             // targeting the default branch, so a hole PR against a feature branch links none.
             const parts = [process.env.PR_BODY || ''];
-            const res = await github.graphql(
-              `query($owner: String!, $repo: String!, $num: Int!) {
-                 repository(owner: $owner, name: $repo) {
-                   pullRequest(number: $num) {
-                     closingIssuesReferences(first: 10) { nodes { body } }
+            // Losing this fallback costs at most a warning on the PR, whereas failing the
+            // step costs the whole review — so neither a thrown error (transient API failure,
+            // missing scope) nor a malformed response is allowed to propagate.
+            let nodes;
+            let failure;
+            try {
+              const res = await github.graphql(
+                `query($owner: String!, $repo: String!, $num: Int!) {
+                   repository(owner: $owner, name: $repo) {
+                     pullRequest(number: $num) {
+                       closingIssuesReferences(first: 10) { nodes { body } }
+                     }
                    }
-                 }
-               }`,
-              { owner: context.repo.owner, repo: context.repo.repo, num: context.issue.number }
-            );
-            // Tolerate a partial response rather than failing the step: losing the fallback
-            // costs a warning on the PR, whereas failing here costs the whole review.
-            const nodes = res?.repository?.pullRequest?.closingIssuesReferences?.nodes;
+                 }`,
+                { owner: context.repo.owner, repo: context.repo.repo, num: context.issue.number }
+              );
+              nodes = res?.repository?.pullRequest?.closingIssuesReferences?.nodes;
+              if (!nodes) { failure = 'the response contained no closingIssuesReferences'; }
+            } catch (error) {
+              // A GraphQL error can still carry partial data alongside it.
+              nodes = error?.data?.repository?.pullRequest?.closingIssuesReferences?.nodes;
+              if (!nodes) { failure = error.message; }
+            }
             if (nodes) {
               for (const node of nodes) parts.push(node.body || '');
             } else {
-              core.warning('Could not read closing issues; using the PR body alone.');
+              core.warning(`Could not read closing issues (${failure}); using the PR body alone.`);
             }
             // Written as file CONTENTS, never handed to a shell as an argument.
             const dir = fs.mkdtempSync(path.join(process.env.RUNNER_TEMP || os.tmpdir(), 'archon-'));
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -2,6 +2,9 @@ name: Archon PR Review
 
 permissions:
   contents: read
+  # closingIssuesReferences returns issue objects, which the pull-request scope does not
+  # cover. Without this the declaration-collection step fails and nothing is posted.
+  issues: read
   pull-requests: write
 
 on:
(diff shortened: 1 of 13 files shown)


mtoslalibu:
## Review: ready to merge

Reviewed against the behaviour we established while validating the SDLC end-to-end on a throwaway two-hole feature, and re-verified empirically — the two shell scripts were run directly against real branches, not just through the test suite.

### Original issues, all closed

| Issue | Status |
|---|---|
| One invocation, not two | ✅ `--plan` output is a strict superset of the delta-only one — verified by diff, zero lines unique to the delta run |
| Plan read from git, not the worktree | ✅ `ls-tree` + `cat-file` by object sha; the `issue_comment` checkout is the default branch, so a filesystem read would find the wrong file |
| base → head fallback | ✅ verified on real branches: hole PR resolves from base, final PR from head |
| Silent for no plan, loud for a broken one | ✅ verified both — a declared-but-unreadable path posts a warning, an absent line posts nothing |
| `--archon-ref` override preserved | ✅ including branch names; empty ref skips validation and falls through |
| Version from `.archon-version` | ✅ builds v0.5.0, no literal anywhere |
| Comment truncation | ✅ 60k cap with a link to the run; full report still in the job summary |
| `ci.yml` runs on `feature/**` | ✅ plus a `./scripts` matrix entry, so the new tests actually run |
| Resolver silent fall-throughs | ✅ an unreadable declaration file and an `ls-tree` failure are now both errors, not "absent" |
| `archon-plan:` reaching the PR body | ✅ see below |
| Fixed `/tmp` output path (#1643) | ✅ `${{ runner.temp }}` in both the review and post steps |

### The `archon-plan:` fix is the one that mattered most

GitHub links closing issues only for a PR targeting the default branch. Every hole PR targets a feature branch, so nothing is linked and the sub-issue body — the only place the line was required — is never read. The final PR does close the tracking issue, but that issue carries no such line. Verified across three real PRs:

```
PR #11  base=feature/hello-greeting     closingIssues=0
PR #13  base=feature/hello-greeting     closingIssues=0
PR #14  base=issue-1624-sdlc-modernize  closingIssues=0
```

The dist ratchet would have been skipped **silently** on the one PR where `dist = 0` is the merge gate. Now fixed in four places — `pr-workflow.md` Step 5 (the doc an agent follows when creating a PR), the final-PR template, the sub-issue template, and `rfc-to-plan.md` — each carrying the reason, so it will not get simplified away later.

### Re-verified empirically

Seven resolver cases against real commits, all correct:

| Case | Result |
|---|---|
| Hole PR, plan on base | `status=resolved plan_source=base` |
| Final PR, plan only on head | `status=resolved plan_source=head`, comment labels it not independently verified |
| No declaration | `status=none`, no note at all |
| Declared but not committed | `status=error` + visible warning, delta review still posted |
| `../../etc/passwd.json` | rejected as an unsafe path |
| Unsubstituted `specs/[NNN-feature]/…` | rejected, warning posted |
| Unreadable declaration file | `status=error`, not silently "none" |

End-to-end on the final-PR shape produced `dist(P,G): 4 → 0 — OK` / `Plan verdict: REALIZES`. 35 tests pass.

### Details worth calling out

- `git ls-tree` reachability is checked separately from path absence, because both otherwise print nothing — conflating them would let an unfetched base tip degrade silently to the PR's own copy.
- A plan that is present but unusable on base is **terminal**, not a fall-through, so a broken base copy cannot hand grading to the PR's own.
- `neuter()` maps `:` to `?`, so a rejected path cannot smuggle a `::workflow::` command into a posted comment.
- The comment explaining why adding `ref:` to the checkout would be arbitrary code execution on the runner.

### Two nits, neither blocking

1. **The GraphQL call is not in a try/catch.** The comment says a partial response should not cost the whole review, but only a malformed response is handled (`res?.…`) — a thrown exception would still fail the step and post nothing. Two lines to fully deliver the stated intent, and this is a file you are already touching.

2. **The stale-branch baseline case is undocumented.** When a PR is behind its feature branch, the before/after distances are measured from the fork point, so they can differ from the reduction the sub-issue predicted. One sentence in `pr-review-context.md` would stop that reading as a bug.

**Verdict: ready to merge.** The plan-detection gap was the real find, and it is fixed at the source rather than patched at the edge.

susiejojo:
Reviewed this end-to-end (both scripts line-by-line, both test suites, workflow + docs). Overall it looks solid and I'd be comfortable with it merging — no correctness blockers. The security model is careful (untrusted path crosses as file contents, validated before use, extracted via `git cat-file` by object SHA, `neuter()` scrubbing anything echoed back), and the three silent-collapse cases around `git ls-tree` (base racing head, broken base copy, unfetched base) are each separated with a dedicated test. The `#1643` stale-comment fix (`/tmp` → `runner.temp` under `if: always()`) is a real catch.

A few low-severity notes, none blocking:

1. **`abspath` (and the `set -uo pipefail` rationale block) is duplicated verbatim across both scripts.** Acceptable given they're standalone executables with no shared-lib mechanism — flagging only as the one real redundancy.

2. **The resolver's new commit-reachability check assumes `BASE_SHA` is in the local object store.** Not a regression — the old inline `git merge-base` had the same requirement, and `actions/checkout@v5` with `fetch-depth: 0` is what satisfies it. But for a hole PR the base is a feature-branch tip and the workflow only explicitly fetches the PR *head*, so the whole feature's viability there rests on `fetch-depth: 0` landing that base. Worth one confirmation on a real hole PR in CI, since the Go tests (which build their own local repos) structurally can't cover it.

3. **Nit:** the anchor `^[^A-Za-z0-9]*archon-plan:` can match line-initial prose (e.g. `- archon-plan: is what we call it`), which would try to resolve `is` and warn. `TestResolve_ProseMention` only covers the mid-line case. Fails loudly (visible warning, never silent wrong grading), so purely a nit — a one-line test would close the gap.

Nice work on the mutation table and the refactor-survival-style tests (asserting observable status/warning/which-invocation rather than internals).

mtoslalibu:
Thanks @susiejojo — checked all three. Two are valid as flagged, one is valid but narrower than described. Only the third needed a change.

**1. Duplicated `abspath` — agreed, leaving it.** Confirmed byte-identical in both scripts. It's one line; extracting it means a third file both scripts `source`, adding a failure mode (missing/unreadable lib) to dedupe one line.

**2. Base commit reachability — valid, but already bounded by pre-existing code.** `git merge-base "$BASE_SHA" "$HEAD_SHA"` runs at `archon-review.sh:71`, *before* the resolver at line 82, and exits 128 on an unreachable SHA (verified). So an unfetched base produces the pre-existing "Could not compute merge-base" comment and the resolver never runs. Two things follow:

- The risk isn't new and isn't plan-specific — it would have broken the old delta-only review identically.
- Which means my reachability check (BC-14) is effectively **unreachable via the workflow**, since `merge-base` always gates it first. It still matters when running the resolver by hand, but I overstated it as a live CI guard.

Your recommendation to confirm on a real hole PR still stands and is on the post-merge checklist.

**3. Line-initial prose — reproduced and pinned in `1794d2a1`.** `- archon-plan: is what we call it` does try to resolve `is` and warn. I added the test but deliberately did **not** tighten the anchor: requiring the token to look like a path would turn a genuinely malformed declaration into `status=none` — a PR that asked for a plan check and silently didn't get one, which is the failure this feature exists to prevent. Warning on prose is the safe direction, so I pinned the behaviour instead of changing it.

Now 52 test cases. Remaining gap is unchanged and worth flagging for the merger: everything is stub-tested, nothing has run against the real `archon-go v0.5.0`. First live `/archon-pr-review` is the first exercise of the real binary, `issues: read`, and `runner.temp` together — and is also the check for note 2.

susiejojo (approved):
lgtm

PR #1645, opened 2026-08-28
feat(replay): faithful closed-loop --trace-output re-export (option a of #1621)

### LINKED ISSUE #1630: Faithful closed-loop --trace-output re-export (option a of #1621); closes residual non-accumulate multi-round follow-up drop
## Context

#1621 closed the *silent-degradation* hole for plain closed-loop replay of an **accumulate** corpus by failing fast on `--trace-output` (option (b), the recommended stopgap — PR #1623). This issue tracks the **faithful re-export** (option (a)) and a residual gap that option (b) does not cover.

## Residual gap (surfaced in PR #1623 independent review)

The #1623 guard keys on `session_context_growth == "accumulate"`. But the *follow-up-dropping* defect (issue #1621, item 1) is general to **any** multi-round closed-loop trace: a **non-`accumulate`** multi-round session corpus re-exported via `--trace-output` in plain closed-loop (`--concurrent-sessions 0`) **still silently drops the generated follow-up rounds** (without the delta-misencoding, since its per-round inputs are absolute). So the silent-truncation hole is not *fully* closed for non-accumulate multi-round corpora.

Why not just broaden the fail-fast guard? Broadening it to *all* closed-loop + `--trace-output` would over-guard legitimately-faithful cases (single-round sessions; no-session traces where the round-0 re-export is complete), and correctly detecting "this corpus has generated follow-ups" requires post-blueprint knowledge (the guard fires up front, before `LoadTraceV2SessionBlueprints`). The honest fix is faithful re-export, not a broader `Fatalf`.

## Option (a): faithful closed-loop re-export

- Capture the full round sequence — adopt the arrival-hook pattern `blis run` already uses (`cmd/root.go` `SetArrivalHook`), or export `append(requests, followUpRequests...)` in session order.
- Re-derive per-round **deltas** from the accumulated absolute buffers, and set `session_context_growth` on the re-export header.
- Once #1613 lands, preserve `input_tokens_reset` compaction markers (a `sim.Request → TraceRecord` mapping cannot currently reconstruct them).
- Remove the #1623 fail-fast guards (plain closed-loop **and** the pool-mode guard) once re-export is faithful for both.

## Acceptance

Round-trip test (INV-13 for this configuration): convert → `blis replay --session-mode closed-loop --trace-output out` → `blis replay` of `out` reproduces per-request per-round ISL/metrics identical to replaying the original converter corpus. Re-export header carries `session_context_growth: accumulate`; per-round `input_tokens` are deltas; `input_tokens_reset` markers survive (post-#1613). Covers both accumulate and non-accumulate multi-round corpora.

namasl:
## Summary

`blis replay --session-mode closed-loop --trace-output <prefix>` (and pool-mode `--concurrent-sessions N --trace-output`) previously wrote an **unfaithful** trace: the export sourced only the round-0 `requests` slice, so the `SessionManager`/pool-driver-generated **follow-up rounds were silently dropped**; for an `accumulate` corpus it also re-emitted **absolute** per-round `input_tokens` instead of the **deltas** an accumulate loader expects and omitted `session_context_growth`. PR #1623 closed that silent-degradation hole by **failing fast** (option (b)).

This PR implements **option (a): faithful re-export**, and **removes both #1623 guards**. A closed-loop / pool replay now captures **every** round (`append(requests, followUpRequests...)`, as the issue suggests), re-derives per-round **deltas + `input_tokens_reset` compaction markers** from the replayed absolute inputs via the shared `EncodeSessionToTraceRecords` law (#1613), captures per-round think time, and sets `session_context_growth` on the header. Fixed-mode export is untouched (byte-identical, INV-6). Covers **accumulate and non-accumulate** multi-round corpora, plain closed-loop and pool.

### The one subtlety that makes the round-trip exact

The accumulate buffer grows by the **actual** accumulated output (`ProgressIndex − InputLen`), which can be below the oracle `MaxOutputLen` (e.g. the final decode token isn't always counted, or a length-capped round). The new `ReExportClosedLoopRecords` therefore feeds the encoder `accumulatedOutputLen` as `prevOut` — making the re-derived deltas the **exact inverse** of the accumulate growth — while emitting the oracle `MaxOutputLen` in the `output_tokens` column so re-replay reproduces the same completion → same actual output → the absolute input sequence reconstructs exactly. (Without this, every follow-up delta was off by one.)

## Issue

Closes #1630

## Behavioral contracts

- **BC-1 (accumulate encoding):** an accumulate corpus replayed closed-loop `--trace-output` writes `session_context_growth: accumulate`, per-round `input_tokens` **deltas** (round 0 = full absolute), `input_tokens_reset` on exactly the non-monotone rounds, and every follow-up round (not just round 0).
- **BC-2 (accumulate reproduction, INV-13):** replaying that export closed-loop reproduces the original run's metrics **byte-for-byte** (plain closed-loop ⇒ per-request identity). Scope: `session_context_growth: accumulate` is set only by `blis convert otel`/`weka`, so every reachable accumulate corpus is converter-produced.
- **BC-3 (non-accumulate round-trip):** every follow-up round is captured with absolute per-round `input_tokens`; follow-ups inherit their session's round-0 `prefix_group`/`prefix_length` (no prefix double-count) and carry the captured `think_time_us`.
- **BC-4 (pool completeness + aggregate):** pool re-export captures **all** sessions (originals + clones), every round, as a complete closed-loop corpus (replay with `--session-mode closed-loop`, not re-cloned). Aggregate conservation metrics reproduce; per-request bit-identity is not guaranteed for pool (data-dependent admission timing).
- **BC-5 (fixed-mode byte-identity):** fixed mode (and closed-loop with no session records) keeps the byte-identical `RequestsToTraceRecords(requests)` path (INV-6).
- **BC-6:** both #1623 `logrus.Fatalf` guards removed.
- **BC-7:** non-session single-shot records mixed into a closed-loop replay are re-exported as absolute single-round records regardless of the accumulate header.

## Cross-path parity

| Path | Applies | Notes |
|------|---------|-------|
| `blis run` | no | Uses `SetArrivalHook` + `RequestsToTraceRecords` (already all rounds, absolute, replayed fixed). Untouched. |
| `blis replay` fixed | no-op | `RequestsToTraceRecords(requests)` unchanged (BC-5). |
| `blis replay` closed-loop / pool | **fix site** | New faithful path via `workload.ReExportClosedLoopRecords`. |
| `blis observe` | no | Recorder captures dispatched rounds directly; no `sim.Request → TraceRecord` round-trip. |

## Invariants

- **INV-13 (run/replay parity):** enforced positively — the lossy re-export is replaced by a faithful one; the round-trip reproduces per-round metrics (accumulate byte-for-byte).
- **INV-6 (determinism):** fixed-mode export byte-identical; the new path reads only deterministic collections + a lookup-only think map; RequestIDs assigned over a deterministic ordering.

## Known boundaries (documented)

- **Length-capped round:** reproduces its input length but may diverge in prefix-cache token *content* across the cap boundary (same class as the existing `blis run` export convention; the huge-ISL agentic corpora use a large `--max-model-len` to avoid capping).
- **Pool:** re-export is a plain closed-loop corpus; per-request bit-identity not guaranteed (admission timing is data-dependent). Aggregate conservation reproduces.

## Test plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (full suite green)
- [x] `golangci-lint run ./...` (0 issues)
- [x] `TestReExportClosedLoopRecords_*` (unit): accumulate deltas+resets, actual-vs-oracle output, non-accumulate absolute+think+prefix, non-consecutive error, determinism
- [x] `TestReplayCmd_ClosedLoopAccumulate_FaithfulReExport` (round-trip): re-export replay reproduces original metrics byte-for-byte — **replaces** the removed `TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal` guard test
- [x] `TestReplayCmd_Pool_FaithfulReExport` (round-trip): all sessions captured, aggregate conservation reproduced — **replaces** the removed `TestReplayCmd_PoolTraceOutputFatal` guard test


namasl:

namasl:
Thanks for the thorough review — all three blocking findings addressed in c8eb14cd, plus the minors and the missing integration test:

- **Issue A (R23 metadata drop):** the accumulate branch now copies the surviving per-request metadata (`ClientID, TenantID, SLOClass, SLOTargetUs, DeadlineUs, Model, Adapter, Streaming, Text/Image/Audio/Video tokens, ReasonRatio`) from the replayed request via a new `copyReExportMetadata`, giving field-coverage parity with the non-accumulate `RequestsToTraceRecords` path. `PrefixGroup`/`PrefixLength` are deliberately excluded (accumulate folds the prefix into round-0's absolute; a prefix on a delta record would double-count). `Model` is copied per your note — it came from the original trace and already drove routing in this replay.
- **Issue B (pool claim too strong):** narrowed the log message **and** the docs to aggregate **conservation** metrics (which is what `TestReplayCmd_Pool_FaithfulReExport` actually asserts), calling out that per-request and cache/latency aggregates are not guaranteed — including the non-accumulate clone + `prefix_group` cache-busting loss you identified. Added `TestReplayCmd_Pool_ClonesCapturedInReExport` (`--total-sessions 4` > corpus 2) asserting the two `_dup` clone sessions and all 8 rounds are captured.
- **Issue C (ordering):** moved the `--trace-output` re-export block to run **last**, after `EmitOutput`/`--results-path`, so a reconstruction or write failure of the secondary artifact never discards the primary simulation metrics (also fixes the pre-existing `ExportTraceV2` Fatalf ordering).
- **Finding 2:** `captureFollowUps` now gates the think capture on `fu.RoundIndex > 0` (writer-side), so pool round-0 admissions never write a spurious think entry.
- **Findings 1 & 3:** added `logrus.Warnf` diagnostics for the (currently unreachable) prefix-shorter-than-prefix-length and think-map-miss fallbacks.
- **BC-3 evidence gap:** added `TestReplayCmd_ClosedLoopNonAccumulate_FaithfulReExport` — the non-accumulate integration round-trip (byte-identical metrics), mirroring the accumulate one.

`go build ./...`, full `go test ./...`, and `golangci-lint run ./...` all green. Re-triggering the review below.

namasl:

susiejojo:
## BLIS PR Review — #1645 `feat(replay): faithful closed-loop --trace-output re-export (option a of #1621)`

Independent verification pass (per `/blis-pr-review`), focused on the Archon architectural signal, whether #1630 is correctly resolved, and code correctness/quality. Built, tested, and traced the round-trip law myself against `main`.

### Archon signal — no genuine architectural change

Archon flagged `ARCHITECTURAL_CHANGE` (1 surface, 2 invariant). Verified this is a **boundary *strengthened*** signal, not a regression:
- `cmd → sim/workload` and `sim/workload → sim` edges are **STRENGTHENED, not ADDED** — both already existed (`sim/workload/tracev2.go:13`, `session.go:13` already import `sim`; `RequestsToTraceRecords` already lives in `tracev2.go`). Archon's own witness delta confirms `edgesAdded: 0`, no cycles.
- The new public surface `workload.ReExportClosedLoopRecords` is a sibling of `RequestsToTraceRecords` in the same package/file — correct altitude. Dependency direction `cmd/ → sim/workload/ → sim/` is respected.
- Invariant delta = the two removed `*Fatal` guard tests + the new positive round-trip tests. Expected and correct for an option-(b)→option-(a) migration.

**No architectural concern.** REALIZES in practice.

### Issue #1630 correctly resolved

The PR implements option (a) and removes both #1623 fail-fast guards. Verified against the issue's acceptance criteria:
- ✅ Captures **every** round via `append(requests, followUpRequests...)` — the follow-up-drop (item 1) is closed for **both** accumulate and non-accumulate multi-round corpora (the residual gap #1630 specifically names).
- ✅ Accumulate: per-round **deltas** + `input_tokens_reset` markers re-derived via the shared `EncodeSessionToTraceRecords` law (#1613); header carries `session_context_growth: accumulate`.
- ✅ Both #1623 guards removed (`TestReplayCmd_ClosedLoopAccumulateTraceOutputFatal` / `TestReplayCmd_PoolTraceOutputFatal` confirmed gone).
- ✅ INV-13 round-trip test present and green.

### Correctness — the delta-inverse law holds

The one subtlety that makes the round-trip exact is verified sound. `accumulatedOutputLen(req) = ProgressIndex − InputLen` (`reexport.go`) is fed to the encoder as `prevOut`; `SessionManager.OnComplete` (`session.go:206-211`) grows the accumulate buffer by exactly this `actualOutputLen`. Since the encoder's law is `delta_k = abs_k − abs_{k-1} − prevOut`, feeding `actualOutputLen` (not the oracle `MaxOutputLen`) makes the re-derived deltas the **exact inverse** of buffer growth, while the emitted `output_tokens` column keeps the oracle so re-replay reproduces the same completion. `TestReExportClosedLoopRecords_Accumulate_ActualOutputVsOracle` pins the off-by-one this prevents. The compaction (`input_tokens_reset`) re-seed path is covered by the integration round-trip (byte-for-byte metrics equality).

### Prior blocking findings — all fixed (verified in HEAD c8eb14cd)

The two automated review passes raised 3 blocking issues; all are correctly resolved in the current tree:
- **A (R23 metadata drop):** `copyReExportMetadata` now copies ClientID/TenantID/SLOClass/SLOTargetUs/DeadlineUs/Model/Adapter/Streaming/modality-tokens/ReasonRatio onto accumulate-branch records. `PrefixGroup`/`PrefixLength` correctly excluded (accumulate folds prefix into round-0 absolute); rationale documented.
- **B (pool clone divergence):** claim narrowed to *conservation* metrics in both the log message and docs; `TestReplayCmd_Pool_ClonesCapturedInReExport` (`--total-sessions 4 > corpus 2`) added, asserting clone capture completeness.
- **C (Fatalf ordering):** re-export block moved to run **after** `EmitOutput`/`--results-path`/`--metrics-path` (`replay.go:912-961`), so a secondary-artifact failure never discards primary metrics.

### Cross-path parity — accurate

- `blis run` — untouched (uses `SetArrivalHook` + `RequestsToTraceRecords`, already all-rounds/absolute). ✓
- `blis replay` fixed — byte-identical `RequestsToTraceRecords(requests)` path preserved (INV-6, BC-5). ✓
- `blis replay` closed-loop/pool — the fix site. ✓
- `blis observe` — independently confirmed exempt: `Recorder.RecordRequest` captures each dispatched round (incl. `roundIndex`) directly; no `sim.Request → TraceRecord` round-trip, so the round-0-drop bug never applied. ✓

### Local verification
- `go build ./...` ✓
- `go vet ./cmd/ ./sim/workload/` ✓
- `go test ./sim/workload/ -count=1` ✓ (incl. all `TestReExportClosedLoopRecords_*`)
- `go test ./cmd/ -run TestReplay -count=1` ✓ (all 4 new round-trip integration tests pass)
- CI: build/lint/all test shards **green**.

### Minor (non-blocking) observations
- The three "unreachable today" defensive paths (think-map miss, `input < prefix_length`) correctly `logrus.Warnf` rather than silently degrade (R1 spirit). Good.
- No integration test exercises an intermediate **timed-out/incomplete** round in the re-export (`reexportStatus` maps it, but only the "ok" path is round-trip tested). Low value — status is informational and not read by the loader — but a future edge worth noting.
- `accumulatedOutputLen`'s `ProgressIndex < InputLen` clamp-at-0 is untested; genuinely unreachable under current sim guarantees (mirrors the `session.go` guard). Not blocking.

### Verdict

**✅ Ready to merge.** The PR fully and correctly resolves #1630 for both accumulate and non-accumulate multi-round corpora, the delta-inverse law is verifiably exact, all 7 behavioral contracts (BC-1…BC-7) are positively evidenced by tests, the three prior blocking findings are addressed, cross-path parity is accurate, and CI is green. The Archon `ARCHITECTURAL_CHANGE` verdict reflects an edge-strengthening (not a new boundary or cycle) and carries no genuine architectural risk.

*Independent review by Claude Code (Opus 4.8). Read-only — no code modified.*

namasl:
Thanks for the independent verification! Implemented both actionable non-blocking observations in e1063f45 (test-only, no production change):

- **Timed-out/incomplete round coverage:** added `TestReExportClosedLoopRecords_StatusMapping` — pins `reexportStatus` for all three branches (completed → `ok`, terminal timed-out → `timeout`, still-running → `incomplete`), closing the "only the ok path is round-trip tested" gap.
- **`accumulatedOutputLen` clamp:** added `TestAccumulatedOutputLen_ClampsNegative` — pins the `ProgressIndex < InputLen` clamp-at-0 (plus the boundary and normal cases).

(The third observation — the defensive paths using `logrus.Warnf` rather than silently degrading — was a confirmation, so nothing to change there.)

`go build ./...`, full `go test ./...`, and `golangci-lint run ./...` all green.

namasl:

susiejojo (approved):
lgtm

Issue #1545, opened 2026-08-10
High-TP KV-block sizing correctness at TP=8/16


namasl:
> Surfaced during the PR #1537 (multi-node TP, #1529) review. A **pre-existing** condition that #1537 does not introduce, but which its whole-node multi-node-TP path makes reachable (high TP degrees). Part of #1532.

**Rescoped 2026-08-12:** #1545 now tracks only **Item 1 (high-TP KV sizing)** — the directly GLM-relevant leg. The other two items were split out per the one-PR-per-issue convention:
- span-aware `MarkNodeTerminated` → #1550
- tautological `VerifyConservation` (INV-A) → #1551

## KV capacity may be ~TP× too large at high TP

`CalculateKVBlocks` in [`sim/latency/kv_capacity.go`](https://github.com/inference-sim/inference-sim/blob/18c2c9260cdf5af7007ffd39103c5304ecd805ac/sim/latency/kv_capacity.go) pairs an **aggregate** memory numerator with a **per-GPU** block-size denominator:

- **Numerator** — `totalAvailableGiB = hc.MemoryGiB · gpuMemoryUtilization · tp` — aggregate across all TP GPUs.
- **Denominator** — `perBlockBytes` derives from `KVBytesPerToken`, which **already divides by TP** (`perTokenKVBytesPerGPUF = perTokenKVBytesF / tp`) → per-GPU.
- `totalBlocks = allocatableBytes / perBlockBytes` therefore divides an aggregate (~×tp) numerator by a per-GPU (~÷tp) denominator ⇒ block count ≈ **`tp × correct`**. Exactly 1 at TP=1 (a genuine no-op), inflating by ~TP× as TP grows.

Multi-node TP (#1529) makes TP=8/16 a reachable, natural config, so this latent error is now exercisable at scale.

**Directly GLM-5.2-relevant:** GLM-5.2's FP8 checkpoint must run **multi-node on H100 at TP=8/16**, so a TP-scaling KV error directly corrupts its capacity-planning numbers.

## Acceptance criteria

- [ ] The KV-block budget formula is validated at TP=8 and TP=16 against a first-principles per-GPU calculation. The cross-check must be **self-consistent** — per-GPU available ÷ per-GPU block bytes, **or** aggregate ÷ aggregate, never mixed. Hypothesis to confirm/refute: block count is ≈ TP× too large.
[...]

PR #698, opened 2026-03-16
feat(latency): decouple quantized weight precision from compute dtype

susiejojo:
## Summary

Closes #443 — Roofline latency model and KV capacity auto-calculation now correctly handle quantized models (GPTQ, AWQ, FP8).

- **Problem**: `BytesPerParam` was derived from `torch_dtype` (the compute dtype), so quantized models like W4A16 reported 2.0 bytes/param instead of 0.5. This caused roofline to overestimate weight bandwidth by 4x and KV capacity to undercount available blocks.
- **Fix**: Added `WeightBytesPerParam` field to `ModelConfig` (zero-sentinel = fallback to `BytesPerParam`), parsed from HF `quantization_config`. Roofline `model_weights` and KV capacity `computeModelWeightBytes()` now use `EffectiveWeightBytesPerParam()`, while KV cache and activations continue using `BytesPerParam`.
- **CLI**: New `--weight-bytes-per-param` flag for manual override when auto-detection is insufficient.

## Changes

| File | Change |
|------|--------|
| `sim/model_hardware_config.go` | `WeightBytesPerParam` field + `EffectiveWeightBytesPerParam()` method |
| `sim/latency/config.go` | Parse `quantization_config` + validation |
| `sim/latency/roofline.go` | Use `EffectiveWeightBytesPerParam()` for `model_weights` |
| `sim/latency/kv_capacity.go` | Use `EffectiveWeightBytesPerParam()` for weight memory |
| `cmd/root.go` | `--weight-bytes-per-param` flag + quantization logging |
| `CLAUDE.md` | Documentation updates |

## Test plan

- [x] 3 unit tests for `EffectiveWeightBytesPerParam()` (sentinel, set, both-zero)
- [x] 6 tests for `quantization_config` parsing (GPTQ, FP8, AWQ, no-quant, bits=0, regression)
- [x] 3 tests for `ValidateRooflineConfig` (negative, NaN, valid)
- [x] 3 tests for roofline memory split (W4A16 weights 1/4, KV unchanged, conservation)
- [x] 3 tests for KV capacity (W4A16 more blocks, monotonicity FP16<FP8<W4, regression)
- [x] All existing tests pass (zero regressions)
- [x] `go build ./...` passes
- [x] `go vet ./...` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1646, opened 2026-08-28
feat(latency): model --kv-cache-dtype (independent KV storage precision, vLLM parity)

### LINKED ISSUE #1565: feat(latency): model --kv-cache-dtype fp8 — independent KV storage precision (vLLM parity)
> **Also applies to Kimi K3** ([#1638](https://github.com/inference-sim/inference-sim/issues/1638)) — K3's config declares a `kv_cache_scheme`, so this shared KV-precision capability covers it too.

**What external behavior are we replicating?**

vLLM's `--kv-cache-dtype` engine argument (`CacheConfig.cache_dtype`), which lets the KV cache be stored at a **different precision from the compute/activation dtype** — most importantly `fp8` (1 byte/element) while compute stays `bfloat16` (2 bytes). vLLM recipes for large MoE/MLA models (DeepSeek-V3.2, GLM-5.2-FP8) commonly run `--kv-cache-dtype fp8` to roughly **halve KV memory and double KV-block capacity**.

BLIS has no equivalent knob. [`KVBytesPerToken`](https://github.com/inference-sim/inference-sim/blob/70e9ba855ecdf00e08c40984a17c1934b5aa8d77/sim/latency/kv_capacity.go#L85-L154) sizes the KV cache using the **compute** dtype `mc.BytesPerParam` on **both** the standard MHA/GQA path (L147) and the MLA compressed-latent path (L114). So a run of `zai-org/GLM-5.2-FP8` with `--kv-cache-dtype fp8` in vLLM is modeled by BLIS at 2 bytes/element → **KV capacity under-counted by ~2×** (pessimistic: fewer KV blocks → smaller batches → lower throughput, more preemption than reality).

Note: BLIS already splits **weight** precision (`WeightBytesPerParam`, #698) from compute; this issue asks for the analogous split on the **KV** axis. Weight-quant ≠ KV-quant — they are independent flags in vLLM.

**Reference implementation (MANDATORY)**

Target: **vLLM v0.11.0** (commit `b8b302cde434df8c9289a2b465406b47ebab1c2d`).

The `CacheConfig.cache_dtype` field and its accepted values (`auto`, `fp8`, `fp8_e4m3`, `fp8_e5m2`, …):
https://github.com/vllm-project/vllm/blob/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/config/cache.py#L25-L59

The `--kv-cache-dtype` CLI argument wiring into `CacheConfig.cache_dtype`:
https://github.com/vllm-project/vllm/blob/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/engine/arg_utils.py#L769-L770

The dtype→torch-dtype map showing fp8 variants map to `torch.uint8` (**1 byte/element**, vs 2 for bf16):
https://github.com/vllm-project/vllm/blob/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/utils/__init__.py#L123-L133

`get_kv_cache_torch_dtype` — the resolution logic (`auto` ⇒ follow model/compute dtype; explicit ⇒ use the mapped KV dtype), which is exactly the auto-vs-override precedence BLIS should mirror:
https://github.com/vllm-project/vllm/blob/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/utils/__init__.py#L1006-L1026

**Key behaviors to preserve**
- [ ] KV storage precision is **independent** of compute/activation precision (fp8 KV under bf16 compute) — ref: cache.py#L25-L59
- [ ] `fp8` / `fp8_e4m3` / `fp8_e5m2` KV ⇒ **1 byte/element**; `bfloat16` ⇒ 2 — ref: utils/__init__.py#L123-L133
- [ ] `auto` (default) resolves to the model/compute dtype — i.e. **default is byte-identical to today** (critical for INV-6) — ref: get_kv_cache_torch_dtype#L1006-L1026
- [ ] Applies on **both** the standard MHA/GQA and the MLA compressed-latent KV paths (both currently read `mc.BytesPerParam`) — ref BLIS: kv_capacity.go#L114 and #L147

**Known intentional deviations**
- Deviation: BLIS need not model fp8 KV **scale tensors** or accuracy effects — only the **byte-width → capacity** consequence. vLLM's `calculate_kv_scales` and per-block scale storage are out of scope; BLIS is a capacity/latency simulator, not a numerics simulator.
- Deviation: mamba/ssm cache dtypes (`--mamba-cache-dtype`) are out of scope here — that belongs with Nemotron-H work (#1552/#1558), not this KV-dtype knob.

**Target parity version**
vLLM **v0.11.0**, commit `b8b302cde434df8c9289a2b465406b47ebab1c2d`.

- [x] Core simulator (`sim/`) — add a KV-precision field on `ModelConfig` (e.g. `KVBytesPerParam`, 0 = fall back to `BytesPerParam`, mirroring the `WeightBytesPerParam` pattern) + an `EffectiveKVBytesPerParam()` accessor
- [ ] Cluster simulation (`sim/cluster/`) — validation call sites of `KVBytesPerToken` already pass `ModelConfig`, so likely no signature change
- [x] KV cache (`sim/kv/`) / `sim/latency/kv_capacity.go` — use the KV precision in `KVBytesPerToken` on both paths (L114, L147)
- [x] CLI (`cmd/`) — a `--kv-cache-dtype` flag (semantic: `auto|fp8|fp8_e4m3|fp8_e5m2|bfloat16`) mapping to bytes, mirroring #629's proposed `--quantization` UX for weights

**Extension friction check**
- Estimated files: `sim/model_hardware_config.go` (field + accessor), `sim/latency/kv_capacity.go` (2 usage sites), `sim/latency/config.go` (optional auto-detect from HF `quantization_config`/dtype), `cmd/root.go` (flag). ~4 files.
- No new interface — extends the existing `ModelConfig` precision-split pattern already established for weights.
- Invariants: **INV-6** — default (`auto`) MUST be byte-identical to current output (KV precision == compute dtype when unset); **INV-4** KV conservation unaffected (only per-block byte width changes); **INV-13** run/replay parity preserved since the KV formula is shared and `blis run`/`replay` read the same `ModelConfig`.

- **Not currently tracked.** Verified against the GLM-5.2 epic **#1532** and its full gap inventory (A–E): gap A / #1527 covers **FP8 weight** precision (`weight_block_size` → `WeightBytesPerParam`), explicitly **not** KV-storage dtype. **#1545** covers a **~TP× KV-block *count*** error (aggregate÷per-GPU mismatch) — orthogonal to per-token **byte width**. **#629** proposes a friendlier `--quantization` flag for **weights** only and considers the weight/KV split "done." None models an independent KV-cache dtype.
- **Do-together candidate:** **#1545** (same file `kv_capacity.go`, both about KV capacity fidelity at high TP — GLM-5.2 runs multi-node TP=8/16 with fp8 KV, so the two errors compound).
- **Pattern to mirror:** the `WeightBytesPerParam` split from **#698** and the semantic-flag UX proposed in **#629**.
- Natural home: a new sub-item ("Gap F: KV-cache dtype") under the **#1532** GLM-5.2 tracking epic.

namasl:
## Summary

Closes #1565. Models vLLM's `--kv-cache-dtype` engine arg (`CacheConfig.cache_dtype`, vLLM v0.11.0) — the KV-cache **storage** precision, which vLLM lets differ from the compute/activation dtype. Most importantly `fp8` (1 byte/element) KV under `bfloat16` (2 bytes) compute, the recipe large MoE/MLA deployments (GLM-5.2-FP8, DeepSeek-V3.2) use to roughly **halve KV memory / double KV-block capacity**.

Before this PR, `KVBytesPerToken` sized the KV cache with the **compute** dtype `BytesPerParam` on both the standard MHA/GQA path and the MLA compressed-latent path, so a run with `--kv-cache-dtype fp8` in vLLM was modeled at 2 bytes/element → KV capacity under-counted ~2× (pessimistic: fewer blocks → smaller batches → lower throughput, more preemption than reality).

This mirrors the existing **weight**-precision split (`WeightBytesPerParam` / `EffectiveWeightBytesPerParam()`, #698) on the **KV axis**. KV-quant, weight-quant, and compute dtype are three independent vLLM engine args.

## What changed (~4 production files + tests)

- **`sim/model_hardware_config.go`** — `ModelConfig.KVBytesPerParam float64` (`json:"kv_bytes_per_param,omitempty"`; 0 = `auto` → fall back to `BytesPerParam`) + `EffectiveKVBytesPerParam()` accessor.
- **`sim/latency/kv_capacity.go`** — `KVBytesPerToken` uses `EffectiveKVBytesPerParam()` for the per-token byte width on **both** branches (MLA latent **and** standard MHA/GQA — both previously multiplied `BytesPerParam`). Added a positive/finite guard for an explicitly-set value (mirrors the `WeightBytesPerParam` guard). PD KV-transfer sizing and the offload `per_block_bytes` derivation inherit this automatically (single source).
- **`sim/latency/config.go`** — `KVCacheDtypeToBytes(string) (float64, bool)`: `auto`/`""`→0; `fp8`/`fp8_e4m3`/`fp8_e5m2`→1 (vLLM maps every fp8 variant to `torch.uint8`); `bf16`/`bfloat16`/`fp16`/`float16`/`half`→2; `fp32`/`float32`/`float`→4; unrecognized → `(0,false)`.
- **`cmd/`** — `--kv-cache-dtype` flag (default `auto`) registered in the shared `registerSimConfigFlags` (so it's on **run** and **replay**, not **observe** — a black-box dispatcher that derives no KV capacity, same boundary as `--kv-offload-config`). `applyKVCacheDtype` sets the field alongside `applyWeightPrecisionFallback` at every `ModelConfig`-resolution site; `resolveLatencyConfig` fails fast on an unrecognized value.

## Scope boundary

KV-capacity byte-width only. The step-time (trained-physics/roofline) models deliberately keep `BytesPerParam` for the KV-read bandwidth term, so step-time goldens and **INV-BC-DP1** are byte-identical. Independent of the KV-block *count* work (#1545) and of weight quant (#1527). fp8 KV **scale tensors** and numerical accuracy are out of scope (BLIS is a capacity/latency simulator) — matching the issue's stated deviations.

## Behavioral contracts

- **GIVEN** `--kv-cache-dtype fp8` on a bf16 model, **WHEN** `KVBytesPerToken` computes per-token bytes, **THEN** the result is exactly half the `auto` value on both the MHA/GQA and MLA branches, and the auto-computed KV block count is ~2×.
- **GIVEN** `--kv-cache-dtype auto` (default / flag absent), **THEN** `KVBytesPerParam` stays 0, `EffectiveKVBytesPerParam()` returns `BytesPerParam`, and per-token bytes + block count are byte-identical to a build without the flag (**INV-6**).
- **GIVEN** an unrecognized value, **THEN** the CLI fails loudly listing valid names (**R1**); a hand-built `ModelConfig` with a negative/NaN/Inf `KVBytesPerParam` errors from `KVBytesPerToken`.
- **GIVEN** a trace exported by `blis run` with `--kv-cache-dtype fp8`, **WHEN** replayed with identical flags, **THEN** per-request metrics are identical (**INV-13**).

## INV-13 round-trip decision (the main design choice)

**Chosen: re-supply the flag on the replay CLI — NOT round-trip through the TraceV2 header.**

Both `run` and `replay` register the flag via the shared `registerSimConfigFlags` and recompute KV capacity through the **same** `resolveLatencyConfig` → `applyKVCacheDtype` → `CalculateKVBlocks` path. Replaying with identical flags is therefore byte-identical.

This is the treatment BLIS already gives every capacity-affecting knob: `--gpu-memory-utilization`, TP/DP, weight precision, and auto-computed `--total-kv-blocks` are all re-derived from the CLI on replay. I verified the trace header's `gpu_memory_utilization` field is **write-only** (documentary) — replay's capacity comes from the CLI flag (`sim/cluster/kv_autocalc.go`, populated at `cmd/root.go`), never read back from the header. By contrast `kv_offload` **is** header-authoritative because replay reads it back to drive a runtime mechanism; KV dtype drives no runtime mechanism (it only scales a byte width), so it needs no round-trip. Keeping it out of the header also matches the issue's stated ~4-file scope.

## Verification

```
go build ./...                → exit 0
go test ./... -count=1        → ok (all packages; cmd 29.9s, sim/cluster 74.6s, sim, sim/latency, …)
golangci-lint run ./... (v2.9.0) → 0 issues
```

New tests: accessor law (`sim`); fp8 halves both branches + auto byte-identity + invalid-value guard + ~2× `CalculateKVBlocks` capacity + auto block-count identity (`sim/latency`); `KVCacheDtypeToBytes` mapping table (`sim/latency`); flag-on-run-and-replay + flag→`ModelConfig.KVBytesPerParam` wiring through the real `resolveLatencyConfig` (`cmd`). All pre-existing goldens unchanged (INV-6 evidence).


namasl:

namasl:
@claude /blis-pr-review a follow-up commit addresses your recommended-before-merge findings A/B/C (the three stale 'KV uses compute dtype' text/log/doc instances now that KV precision is independent) plus E (added test coverage for the accepted dtype aliases half/float32/float/fp8e4m3/fp8e5m2). Deferred as cosmetic per your classification: D (the harmless unreachable defensive Fatalf guard) and F (captureCmdLevelVars — the new test already self-manages kvCacheDtype save/restore). Comment/doc/test-only, no behavior change. Please confirm; also consider the /archon-pr-review output above once available.

susiejojo:
## `/blis-pr-review` — PR #1646 `feat(latency): model --kv-cache-dtype`

**Verdict: ✅ Ready to merge** (no blocking findings). A well-scoped, faithful mirror of the `WeightBytesPerParam` split on the KV axis. Below are 1 recommended fidelity fix and a handful of doc/test gaps — all non-blocking.

Reviewed via archon context + 10 code-review perspectives + a 3-agent fan-out (vLLM-fidelity, test/parity/docs, 12-question Q&A) + direct empirical runs on the reporter's exact model. Archon verdict was `ARCHITECTURAL_CHANGE` (2 surface, 1 schema, 3 invariant-test additions; only `cmd→sim/latency` and `sim/latency→sim` call edges *strengthened* — no boundary moved, no cycle, no violation). The "architectural change" flag is purely the new `ModelConfig.KVBytesPerParam` field + `EffectiveKVBytesPerParam`/`KVCacheDtypeToBytes` surface; the dependency graph is unchanged. No concern.

### Empirical validation (built the PR, ran GLM-5.2-FP8 + Llama-3.1-8B)

Reproducing the reporter's `vllm serve zai-org/GLM-5.2-FP8 --tensor-parallel-size 16 ... --kv-cache-dtype fp8` in BLIS:

| Model / branch | `--kv-cache-dtype auto` | `--kv-cache-dtype fp8` | ratio |
|---|---|---|---|
| GLM-5.2-FP8 (MLA, TP=16) | 326,966 blocks | 653,932 blocks | **2.000×** |
| Llama-3.1-8B (GQA, TP=1) | 26,312 blocks | 52,625 blocks | **2.000×** |
| Llama-4-Scout (quantized weights + fp8 KV) | 1,279,198 | 2,558,397 | **2.000×** — weight-precision log byte-identical, i.e. weight-quant and KV-quant are independent ✓ |

- **Capacity binds correctly under pressure**: Llama TP=1 at rate 60, 4–8K prompts — fp8's 2× blocks cut preemptions **415 → 338**. The knob is direction-correct (more blocks → less preemption). `fp8_e5m2` and `fp8e5m2` (no-underscore alias) also work end-to-end.
- **INV-6 (auto == flag-absent)**: `--kv-cache-dtype auto` and no flag produce **byte-identical stdout** (md5 match), on both the analytical and MLA paths.
- **INV-13 (run→replay)**: The global KV auto-calc lives in the **shared `resolveLatencyConfig`** (`cmd/root.go:636-660`), so **both** run and replay auto-calc identically (both logged `total-kv-blocks=52625` for the fp8 trace). Re-supplying `--kv-cache-dtype fp8` on replay reproduces run metrics. ✓ (One agent hypothesized replay skips the global auto-calc — I verified in code + logs that it does **not**; that claim is refuted.)
- **INV-BC-DP1**: step-time (roofline/trained-physics) KV-read term still uses `BytesPerParam`, not the KV dtype (`roofline.go:250,256`). Byte-identical step time. ✓
- Full `go build ./...`, `go vet`, and `go test ./cmd/... ./sim/ ./sim/latency/...` all clean.

### Correctness (verified independently)
- Both `KVBytesPerToken` branches (MLA latent `kv_capacity.go:133`, MHA/GQA `:171`) use `EffectiveKVBytesPerParam()`. No stray `BytesPerParam` KV-width read remains. MLA fp8 halving is exactly TP-invariant (latent not divided by TP). ✓
- Guard ordering is correct: `BytesPerParam` positivity first, then `KVBytesPerParam` (NaN/Inf/negative rejected before the accessor is reached). Mirrors the `WeightBytesPerParam` guard. ✓
- `applyKVCacheDtype` is wired at all 3 resolution sites (analytical `root.go:598`, run-PD `:1523`, replay-PD `replay.go:435`); node-pool per-instance recalc and PD KV-transfer sizing inherit fp8 via the single resolved `ModelConfig`. ✓
- The `json:"kv_bytes_per_param"` tag is dormant — HF configs are parsed field-by-field via `GetModelConfigFromHF`, never `json.Unmarshal` into `ModelConfig`, so there is no config.json backdoor; the flag is the sole driver. ✓
- Correctly **absent from `observe`** (which never calls `registerSimConfigFlags` and derives no KV capacity) — matches the `--kv-offload-config` boundary. ✓

### Findings

**F1 — Recommended (minor fidelity gap): `fp8_inc` is a valid vLLM v0.11.0 value but BLIS hard-errors on it.** `sim/latency/config.go` `KVCacheDtypeToBytes`. vLLM v0.11.0 `CacheDType = Literal["auto","bfloat16","fp8","fp8_e4m3","fp8_e5m2","fp8_inc"]`. `fp8_inc` (Intel Gaudi fp8) → 1 byte/element, so it should map to `1.0` alongside the other fp8 variants. A user copying a valid vLLM `--kv-cache-dtype fp8_inc` command hits `logrus.Fatalf`. One-line addition to the `case`.

**F2 — Wording (non-blocking): "vLLM parity" overstates the accepted set.** BLIS accepts `bf16/fp16/float16/half/fp32/float32/float`; vLLM v0.11.0's `--kv-cache-dtype` accepts *only* `bfloat16` (not `bf16`/`fp16`/`fp32`) plus the fp8 variants. The byte *values* are physically correct and the superset is a reasonable usability extension — but the flag help / docstring should say "superset of vLLM's accepted values" rather than pure parity.

**F3 — Docs (non-blocking): stale KV-precision claim outside the touched files.**
- `docs/concepts/roofline.md:10` still says KV cache "continue[s] to use the compute dtype (`BytesPerParam`)" — now wrong; the follow-up commit fixed this exact sentence in CLAUDE.md + docstrings but missed this file.
- `docs/guide/latency-models.md:221` couples KV capacity to weight precision and doesn't mention `--kv-cache-dtype`.
- `docs/reference/configuration.md` flag table lists every other KV flag but **omits `--kv-cache-dtype` entirely** (`grep -rn kv-cache-dtype docs/` → 0 hits). A user-facing flag should appear in the config reference.

**F4 — Test gaps (non-blocking):**
- No CLI-boundary test that `--kv-cache-dtype garbage` fatals (R1). The pure `KVCacheDtypeToBytes(...) == (0,false)` mapping is tested, but the loud-failure contract at the `cmd` layer is not (unlike sibling guards that use the `BLIS_TEST_SUBPROCESS` pattern).
- No e2e run→replay byte-identity (INV-13) test at the CLI/integration level — the headline claim is proven only at the `resolveLatencyConfig` unit level. (I verified it manually; a committed golden would lock it.)

**F5 — Concern (documented, non-blocking): silent capacity divergence if the flag isn't re-supplied on replay.** The KV dtype is intentionally *not* round-tripped through the TraceV2 header (CLI re-supply model, matching `--gpu-memory-utilization` and auto `--total-kv-blocks` — I confirmed the header records none of these). This is consistent and documented in the flag help + CLAUDE.md. It is, however, the *weaker* of two existing precedents — `kv_offload` **is** header-authoritative with a hard error on mismatch. Replaying an fp8 trace without `--kv-cache-dtype fp8` silently halves capacity with only an (unrelated) info-log. Acceptable given the gpu-mem-util precedent; worth a one-line note in the replay docs.

### Cosmetic (already classified by the prior review, no action)
- The `logrus.Fatalf` inside `applyKVCacheDtype` (`cmd/hfconfig.go`) is unreachable: `resolveLatencyConfig` validates the value up-front (`root.go:453`) before every `applyKVCacheDtype` call. Harmless defense-in-depth.
- No `ValidateModelConfig`-level guard for `KVBytesPerParam` (unlike `WeightBytesPerParam`); the runtime `KVBytesPerToken` guard covers it.

### Bottom line
Correct, invariant-preserving, well-tested for the math, and empirically validated on the exact fp8-KV MLA deployment the issue targets. The only change I'd *recommend* before merge is **F1 (`fp8_inc`)**; F2–F5 are polish and can land as a follow-up. This realizes issue #1565's checklist (both KV paths, fp8→1 byte, auto byte-identical, independent of weight quant) without scope creep.

namasl:
@claude /blis-pr-review addressed the review findings: F1 (accept fp8_inc — the valid vLLM v0.11.0 value that was hard-erroring), F2 (flag help now says 'superset of vLLM values' not 'parity'), and the factually-wrong F3 instance (docs/concepts/roofline.md's stale 'KV uses compute dtype' line). Remaining F3 discoverability docs, F4 tests (CLI garbage-fatal + e2e INV-13 golden), and the F5 replay-docs note are deferred to #1648 (non-blocking, per the review). Please confirm; also consider the /archon-pr-review output above once available.

namasl:
@claude /blis-pr-review commit 95d02b51 folds in the previously-deferred minor findings, so ALL of your F1-F5 are now addressed in this PR (nothing deferred): F3 discoverability docs (configuration.md flag table + latency-models.md), F4a (CLI garbage-fatal subprocess test), F4b (e2e run→replay INV-13 golden with fp8 load-bearing — built at the cluster level since the existing spec harness pins --total-kv-blocks and would make the flag inert), and F5 (replay-docs re-supply note). Docs+tests only, no production change; build/test/lint clean. #1648 closed as absorbed. Please confirm; also consider the /archon-pr-review output above once available.

susiejojo:
## `/blis-pr-review` follow-up — verdict: ✅ **Ready to merge**

Re-reviewed at `95d02b51`. Both follow-up commits (`939ca793`, `95d02b51`) resolve every finding from my earlier review; verified by building and running the latest code.

| Finding | Status |
|---|---|
| **F1** — `fp8_inc` (valid vLLM v0.11.0 value) hard-errored | ✅ Fixed — added to `KVCacheDtypeToBytes` + Fatalf lists + help. Verified: `fp8_inc` → 653,932 blocks = 2× `auto` (326,966) on GLM-5.2-FP8, identical to `fp8`. |
| **F2** — "vLLM parity" overstated accepted set | ✅ Fixed — help/docstring now say "superset of vLLM's values". |
| **F3** — stale doc + missing config-ref entry | ✅ Fixed — `roofline.md`, `latency-models.md`, `configuration.md` (new row), replay-guide note. |
| **F4** — no CLI garbage-fatal test, no e2e INV-13 golden | ✅ Fixed — `TestResolveLatencyConfig_KVCacheDtype_GarbageRejected` (subprocess) + `TestINV13_RunReplayParity_FP8KV` (run→replay per-request TTFT/E2E identity, asserts ~2× via real `CalculateKVBlocks`). Both PASS. |
| **F5** — silent capacity divergence if flag not re-supplied on replay | ✅ Documented — new replay-guide admonition, contrasted with header-authoritative `--kv-offload-config`. |

Build, `go vet`, and changed-package tests (`cmd`, `sim/latency`) all clean; the two new tests run and pass. Correctness/invariants (INV-6, INV-13, INV-BC-DP1) were confirmed empirically in the first pass and remain intact. No blocking or open concerns.

susiejojo (approved):
lgtm

Issue #1638, opened 2026-08-28
Tracking: Kimi K3 support — hybrid attention, MoE-shape, and multi-node deployment


namasl:
## Purpose

Tracking issue for everything BLIS needs to **faithfully model Kimi K3**. K3 is architecturally a close cousin of GLM-5.2 (both are DeepSeek-V3-class MLA MoE models), so most of the **multi-node deployment** machinery is shared with the GLM-5.2 tracking issue (#1532) and is **inherited** here — landing it for GLM lands it for K3. What is genuinely K3-specific is a small set of **model-shape** gaps.

As with #1532, none of these are crashes — BLIS runs and emits numbers. The concern is fidelity: without the shape fixes below, K3's KV-capacity and MoE-compute numbers are wrong; without the (inherited) deployment work, K3's real Wide-EP multi-node topology can't be placed.

## Background: what Kimi K3 is

From the HF config ([`moonshotai/Kimi-K3/config.json`](https://huggingface.co/moonshotai/Kimi-K3/blob/main/config.json), `model_type: kimi_k3`, `KimiK3ForConditionalGeneration`):

- **Multimodal wrapper**, language model nested in `text_config` (BLIS already pivots `text_config` into the flat map — [`config.go#L156-L161`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/config.go#L156-L161) — so the nesting itself is handled).
- **Hybrid attention:** **24** full **MLA** layers (`full_attn_layers`) + **69** **KDA** linear-attention layers (`kda_layers`) = 93 total. Only the MLA layers bear a KV cache.
- **MLA:** `kv_lora_rank: 512`, `q_lora_rank: 1536`, `qk_rope_head_dim: 64`, `v_head_dim: 128`.
- **MoE:** `num_experts: 896`, `num_experts_per_token: 16`, `num_shared_experts: 2`, `first_k_dense_replace: 1`.
- **Activation:** `situ` (SiTU-GLU) — already whitelisted, [`kv_capacity.go#L52-L58`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/kv_capacity.go#L52-L58) (#1526).
[...]

Issue #1635, opened 2026-08-28
feat(latency): Kimi-K3 hybrid attention — only MLA layers bear KV cache (KDA layers do not)


namasl:
> Part of #1638 (Kimi K3 support tracker).

## Problem

Kimi K3 is a **hybrid-attention** model. Per its config ([`moonshotai/Kimi-K3/config.json`](https://huggingface.co/moonshotai/Kimi-K3/blob/main/config.json), `text_config.linear_attn_config`), of its **93** layers:

- **24 are full Multi-head Latent Attention (MLA)** — `full_attn_layers: [4, 8, 12, …, 92, 93]` (24 entries)
- **69 are Kimi Delta Attention (KDA)** — `kda_layers: [1, 2, 3, 5, 6, 7, …]` (69 entries), a gated linear/delta-net attention (`short_conv_kernel_size: 4`, `use_full_rank_gate: true`, `gate_lower_bound: -5.0`)

(24 + 69 = 93.) KDA is **linear attention**: it maintains a **fixed-size recurrent + short-conv state per layer** (independent of sequence length), *not* a per-token KV cache that grows with context. Only the **24 MLA layers** store the compressed KV latent (`kv_lora_rank + qk_rope_head_dim = 512 + 64`).

BLIS sizes KV over **all** `NumLayers` on the MLA path:
- [`sim/latency/kv_capacity.go#L112-L114`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/kv_capacity.go#L112-L114) — `perTokenKVBytesF := float64(mc.NumLayers) * float64(latentWidth) * mc.BytesPerParam`

For K3 this counts **93** KV-bearing layers instead of **24** → the per-token KV footprint, and therefore the KV block count / max concurrent tokens / batch sizes, is **overcounted by ~3.9× (93/24)**. This is a first-order capacity **and** throughput error (too-large KV footprint → fewer blocks → smaller batches → lower predicted throughput).

## What to do

Introduce a **"KV-bearing layer count"** distinct from `NumLayers` for hybrid models, and use it in the MLA branch of `KVBytesPerToken` ([`kv_capacity.go#L114`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/kv_capacity.go#L114)) in place of `mc.NumLayers`:

[...]

PR #1650, opened 2026-08-31
feat(latency): Kimi-K3 KDA linear-attention step-time (69 of 93 layers)

### LINKED ISSUE #1636: feat(latency): Kimi-K3 KDA linear-attention step-time (69 of 93 layers are not full attention)
> Part of #1638 (Kimi K3 support tracker).

## Problem

After #1635 corrects KV **capacity**, the **step-time** model still treats all 93 K3 layers as full attention. The default trained-physics attention-compute term scales with `numLayers`:

- [`sim/latency/trained_physics_model.go#L203`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/trained_physics_model.go#L203) — `L := float64(m.numLayers)` (drives the per-layer attention terms below it).

The roofline backend scales its attention FLOPs by `num_layers` in the same way ([`sim/latency/roofline.go`](https://github.com/inference-sim/inference-sim/blob/597df51b49c13ee15e0e17a3037a82624cc16371/sim/latency/roofline.go)).

K3's **69 KDA layers** are linear attention — in decode their per-token cost is ~O(state), not O(sequence) — so charging all 93 layers full-attention compute **overcounts decode step time / TTFT** for K3 (74% of its layers). This is the same *class* of documented approximation as "MLA step-time is pessimistic" (#1527), but much larger in magnitude for K3.

## What to do

Split the attention step-time contribution by layer type (full-attention MLA vs KDA linear attention), charging KDA layers the appropriate linear-attention cost, in the **default trained-physics** model (and mirror in roofline for backend parity). **Reuse the layer-type split parsed in #1635** (do not re-parse `linear_attn_config`).

## Cross-path parity

Step time lives in the DES latency model, shared by **`blis run` and `blis replay`** — both are fixed together (**INV-13**). **`observe` is unaffected** (it measures a real server; it does not compute step time).

## Acceptance

- Decode step time for K3 reflects **24 full-attention + 69 linear-attention** layers (lower than the all-93-full-attention estimate).
- Non-hybrid models **byte-identical** (INV-6); dense-DP=1 step time preserved (**INV-BC-DP1**).
- Metamorphic test: growing sequence length raises the full-attention term but **not** the KDA term.

## Dependencies / priority

**Blocked by #1635** (needs the layer-type split). **Latency-fidelity refinement (Tier 1)** — promote toward the bare-minimum set if latency/TTFT accuracy (not just capacity) is the priority.

namasl:
## Summary

After #1635 corrected KV **capacity** for hybrid-attention models, the **step-time** model still charged all 93 Kimi-K3 layers as full attention — overcounting decode step time / TTFT for the 69 (of 93) linear-attention layers. This PR splits the sequence-length-dependent per-layer attention cost by layer type in **both** latency backends (default **trained-physics** `sim/latency/trained_physics_model.go` + **roofline** `sim/latency/roofline.go`, for backend parity).

- **Full-attention (KV-bearing) layers** keep the O(context) / O(N²) attention-score compute and the growing-KV read/write bandwidth.
- **KDA (linear-attention) layers** charge a linear cost — **O(N) in prefill, O(state) per token in decode** (context-independent).

The full-attention layer count is reused from #1635 (`ModelConfig.EffectiveKVBearingLayers()`); the KDA count is derived as `numLayers − kvBearing` (NOT re-parsed from `linear_attn_config`). KDA layer **weights** stay charged as full attention (out of scope, #1638).

## KDA cost model (and its physics basis)

The delta-rule state update/readout in Kimi Delta Attention are `d_head × d_head` GEMMs, so the "context" each token attends to is the **fixed state dimension `d_head`**, not the growing sequence. The model **substitutes the growing context dimension with `d_head`** in the *existing* FlashAttention FLOP convention:

| term | full-attention (per layer) | KDA (per layer) |
|---|---|---|
| prefill attn-score FLOPs | `4·h·tᵢ·(sᵢ + tᵢ/2)·d_head`  (O(N²)) | `4·h·tᵢ·d_head·d_head`  (O(N)) |
| decode attn-score FLOPs / token | `4·h·context·d_head`  (O(context)) | `4·h·d_head·d_head`  (O(state)) |
| growing-KV read/write bandwidth | charged | **not charged** (no growing KV cache — #1635) |

This reuses the **existing** `β₁ₐ`/`β₂ₐ` compute coefficients and the existing FLOP-counting constant — **no new empirical coefficient is introduced.**

### Calibration honesty

The KDA term is a **principled physics/roofline basis, not empirically calibrated against K3 measurements.** Two documented approximations (in code comments + `CLAUDE.md`), flagged as calibration follow-ups:
1. KDA layers reuse the **full-attention head count / head_dim** (since `linear_attn_config` is not re-parsed here — the KDA count comes from `numLayers − kvBearing` per #1635).
2. The small **context-independent KDA recurrent-state bandwidth** is left folded into the per-layer overhead `β₅·L` (charged over all layers) rather than given a separate fabricated state-bandwidth coefficient.

No validated-accuracy claim is made beyond "correct order (O(N)/O(state)), and - for context at or above dHead (~85-128 tokens for K3) - materially below the pessimistic all-93-full-attention estimate." Note: at very short context the fixed-state KDA cost can slightly exceed full attention's tiny O(N^2) cost, so the reduction is not monotone at all lengths (caveat documented in code, per review FLAW 3).

## Behavioral contracts

- **GIVEN** a non-hybrid model (`EffectiveKVBearingLayers() == numLayers`, i.e. no `linear_attn_config`) **WHEN** `StepTime` is computed **THEN** it is byte-identical to pre-#1636 (`lFull == L`, every KDA branch guarded by `if lKDA > 0` is skipped) — **INV-6 / INV-BC-DP1**.
- **GIVEN** a K3-shaped hybrid model (24 full of 93) **WHEN** a decode-heavy batch is stepped **THEN** step time is strictly **lower** than charging all 93 as full attention.
- **GIVEN** a hybrid model **WHEN** sequence length grows **THEN** the full-attention term rises but the KDA term does **not** — the per-context step-time increase equals the KV-bearing layer fraction (24/93) exactly (β₂ₐ=0 default ⇒ growth is entirely the KV-read term, which scales with `lFull`).
- **GIVEN** a hybrid model **WHEN** prompt length grows **THEN** prefill rises **less** than all-full-attention (KDA prefill is O(N), not O(N²)).

## Cross-path parity

Step time lives in the shared DES latency model → **`blis run` and `blis replay` change together (INV-13)**. **`observe` is unaffected** (it measures a real server; it computes no step time).

## Tests (`sim/latency/kda_step_time_test.go`)

- `TestKDAStepTime_NonHybridByteIdentical` — INV-6/INV-BC-DP1 no-op guard (KVBearingLayers=0 vs =NumLayers, TP matrix).
- `TestKDAStepTime_HybridDecodeLowerThanAllFull` — acceptance: hybrid decode < all-93-full.
- `TestKDAStepTime_ContextGrowthScalesWithFullAttentionLayers` — metamorphic: context-growth ratio == 24/93.
- `TestKDAStepTime_HybridPrefillGrowsSubQuadratically` — KDA prefill O(N) not O(N²).
- `TestRooflineKDA_*` — roofline backend byte-identity + reduction + metamorphic (parity).

The existing `TestINVBCDP1_DenseStepTimeByteIdentical` golden (unchanged) pins the non-hybrid absolute values. No existing hybrid step-time golden exists (#1635 was capacity-only), so no goldens changed.

## Which step-time goldens changed

**None.** The split is a strict no-op for non-hybrid models, and there was no pre-existing hybrid/K3 step-time golden.

## Verification

```
go build ./...            → exit 0
go test ./... -count=1     → ALL PASS (cmd, sim, sim/cluster, sim/latency, sim/kv, sim/workload, ...)
golangci-lint run ./...    → 0 issues  (v2.9.0, matches CI)
```

Closes #1636


namasl:

namasl:
@claude /blis-pr-review commit 2fa78ffa addresses all your findings: FLAW 1 (KVBearingLayers field comment now matches the accessor's 'two consumers' framing — capacity + step-time, not weights), FLAW 2 (ratio-test tolerance tightened 0.02 → 1e-3 to catch off-by-one), FLAW 3 (short-context 'pessimistic for context < dHead, hybrid <= all-full only holds for context >~ dHead' caveat added at all three substitution sites, and the PR-body claim softened accordingly), FLAW 4 (new TestKDAStepTime_DecodeComputeBranchWiredWhenBeta2aPositive exercises the β2a>0 decode-compute path), and the ropeOps pessimism note. INV-6/INV-BC-DP1 intact (only production change is a doc comment; goldens unchanged). build/test/lint clean. Please confirm; also consider the /archon-pr-review output above once available.

namasl:

## qa-review — PR #1650: ⛔ BLOCK
_feat(latency): Kimi-K3 KDA linear-attention step-time (69 of 93 layers) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ❌ FLAW_FOUND | Does this PR fully implement the issue it closes, without unnecessary additions? | The two latency backends implement the requested layer split and tests pass, but the change is not fully complete because every valid hybrid analytical run/replay still emits a user-visible warning claiming that step time charges all layers as full attention. That statement is now false and directly contradicts this PR's behavior. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The principal architecture documentation was updated: CLAUDE.md and ModelConfig/latency-model comments describe the new trained-physics and roofline behavior. One user-facing explanatory message was not updated and still documents the old step-time approximation. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | Yes. The comment immediately above the hybrid warning still says step time charges all layers as full attention, even though this PR changes that behavior. |
| G1 | correctness | ✅ CONFIDENT | In roofline.go, nKDA is calculated as nLayers - nFull where nFull is float64(config.EffectiveKVBearingLayers()). If a model configuration has NumLayers equal to zero or negative due to improper initialization, how does roofline.go prevent nKDA from becoming zero or negative and producing invalid memory access and GEMM FLOP counts? | roofline.go itself relies on its documented precondition rather than locally repairing invalid layer counts. The shipped constructor validates NumLayers > 0 before rooflineStepTime can run, so zero or negative NumLayers cannot reach this calculation through the actual latency-model path. |
| G2 | correctness | ✅ CONFIDENT | In trained_physics_model.go under StepTime for decode, KDA decode FLOPs are calculated as lKDA * 4 * hPerGPU * dH * dH * totalDecodeTokens / dpf. If a decode request batch contains requests with different numbers of new decode tokens per request (e.g., NumNewDecodeTokens > 1 in spec decode), does totalDecodeTokens (which counts request items) undercount token FLOPs compared to totalDecodeTokens in prefill? | Yes, a hand-constructed decode request with NumNewTokens > 1 would be counted as one token by trained-physics, because totalDecodeTokens is incremented once per request. However, this is not a regression in a realistic shipped path: BLIS batch formation currently schedules exactly one decode token per request, and speculative/multi-token decode is out of scope. The pre-existing decode projection, FFN, KV, and context accounting already use the same one-token-per-request assumption. |
| G3 | invariants | ✅ CONFIDENT | INV-6 requires byte-identical step time for non-hybrid models. In trained_physics_model.go, L (float64(m.numLayers)) was replaced by lFull (float64(m.numKVBearingLayers)) for attention and KV bandwidth calculations. When m.numKVBearingLayers == m.numLayers, does using lFull instead of L introduce any compiler instruction reordering or floating-point precision differences in 64-bit float math? | No precision difference is introduced when the counts are equal. L and lFull are each independently converted from equal int values to float64, yielding identical values, and the non-hybrid lKDA == 0 branch is skipped. Tests compare exact int64 StepTime results across prefill/decode/mixed batches and TP values. |
| G4 | invariants | ✅ CONFIDENT | INV-13 requires strict execution parity between blis run and blis replay. Does blis replay deserialize and pass ModelConfig.KVBearingLayers correctly when reconstructing ModelHardwareConfig from replay execution logs, or could replayed runs fall back to KVBearingLayers = 0 and simulate hybrid models as non-hybrid? | Replay does not depend on execution logs to serialize this architecture field. It runs the same resolveLatencyConfig function as run, reparses the configured HuggingFace model JSON, and passes the resulting complete ModelConfig into NewModelHardwareConfig. Therefore KVBearingLayers is preserved for replay. |
| G5 | rules | ✅ CONFIDENT | Rule R1 mandates that fallback logic or silent corrections must emit visible warnings. When EffectiveKVBearingLayers() falls back to NumLayers because KVBearingLayers is 0 or absent, is there a warning logged during model construction in TrainedPhysicsModel or roofline.go to inform users whether hybrid attention scaling was activated? | A zero KVBearingLayers value is the normal representation for every non-hybrid model, so it intentionally falls back without warning. If linear_attn_config is present but its full_attn_layers list is unusable, parsing emits a visible warning before returning zero. A valid hybrid also triggers the shared hybrid-detection warning, although that warning's step-time wording is stale. |
| G6 | rules | ✅ CONFIDENT | Rule R2 requires canonical constructors for latency models. TrainedPhysicsModel derives numKVBearingLayers directly inside NewTrainedPhysicsModel without exposing KDA parameters in LatencyCoeffs. Does hardcoding the KDA linear FLOP substitution inside StepTime without dedicated structural options violate clean model parameterization? | No. The cited R2 is the deterministic-map-iteration rule, not a constructor/parameterization requirement. The layer split is architecture structure already carried by ModelConfig, while LatencyCoeffs remains reserved for empirical coefficients. Both backends are still selected through the canonical NewLatencyModel path, and trained-physics uses its existing canonical constructor. |
| G7 | parity | ✅ CONFIDENT | In roofline.go, KDA GEMM FLOPs use nHeads * newT * dHead * dHead (total heads across the model), whereas in trained_physics_model.go prefill KDA FLOPs use hPerGPU * ti * dH * dH (heads per GPU). How does roofline.go account for Tensor Parallelism (TP) splitting in the KDA GemmOps and SramOps calculations? | Roofline computes total-model FLOPs using all heads, then divides the complete FLOP result—including KDA GemmOps and SramOps—by tpFactor when converting it to per-rank compute time. The same division is used for prefill and decode. |
| G8 | parity | ✅ CONFIDENT | The PR updates both default trained-physics (trained_physics_model.go) and roofline (roofline.go) latency backends. Are there any subtle differences in how prefill sequence lengths (si + ti/2) vs (newT) are treated for KDA layers between trained-physics and roofline when processing chunked prefill steps? | There is a backend difference for full-attention chunked-prefill context: trained-physics uses InputLen plus ti/2, while roofline uses ProgressIndex plus (newT-1)/2. That difference is pre-existing. For the new KDA term, both backends replace the context axis entirely with d_head, so KDA cost depends only on the number of newly processed tokens and is independent of the chunk's starting sequence position. |
| G9 | tests | ✅ CONFIDENT | In TestKDAStepTime_ContextGrowthScalesWithFullAttentionLayers, the metamorphic context growth ratio test asserts an exact delta of 1e-3 against 24.0/93.0. Because default test coefficients have BetaCoeffs[1] (β₂ₐ) set to 0, growth comes entirely from KV-reads. Would this test still pass if β₂ₐ is positive in future empirical coefficient calibrations? | Yes, under the model as implemented it should still pass. A positive β₂ₐ adds another context-dependent delta that is also scoped solely to lFull, while the KDA decode compute term remains context-independent and cancels between the small- and large-context measurements. Thus both compute and KV-read growth retain the 24/93 layer ratio, aside from final int64 truncation. |
| G10 | tests | ❌ FLAW_FOUND | TestKDAStepTime_DecodeComputeBranchWiredWhenBeta2aPositive tests trained-physics decode compute by overriding BetaCoeffs. Is there an equivalent test for roofline.go verifying that decode compute FLOPs for KDA layers remain context-independent when roofline rooflineStepTime is called? | No equivalent roofline test isolates the new KDA decode-compute formula. The roofline tests only compare final roofline step times; because roofline takes max(compute, memory), the long-context decode cases can remain memory-bound and pass even if the KDA GEMM formula becomes context-dependent. |
| G11 | rationale | ✅ CONFIDENT | The implementation substitutes the context dimension with d_head in the FlashAttention FLOP formula (4·h·d_head·d_head) to estimate KDA linear attention cost. Why was this substitution chosen over adding a explicit empirical coefficient for linear attention state updates? | The substitution models the KDA delta-rule state update/readout as d_head-by-d_head work: replacing the growing context axis with the fixed state dimension makes prefill O(N) and decode O(state) per token. Reusing the existing FlashAttention FLOP convention and β₁ₐ/β₂ₐ avoids claiming a newly fitted empirical parameter without calibration data. |
| G12 | rationale | ✅ CONFIDENT | Why does the KDA step-time model reuse the full-attention head_dim and head count rather than parsing linear_attn_config from the model architecture JSON during step-time calculation? | The issue explicitly requires reuse of the layer-type split parsed by the prerequisite work rather than reparsing linear_attn_config. The latency models therefore consume EffectiveKVBearingLayers and derive KDA count as total minus full-attention layers. Head count and head_dim are reused as an acknowledged approximation so no second architecture parser or divergent source of truth is introduced. |
| G13 | scope | ✅ CONFIDENT | The PR PR body notes that for sequences shorter than ~d_head (~85–128 tokens), the fixed KDA state update cost can exceed the full-attention context cost. Does this PR include any guard or fallback for short sequence lengths, or is the pessimistic cost strictly intentional? | There is no short-sequence guard or fallback. The fixed d_head state-cost substitution is intentionally retained even when it exceeds the corresponding full-attention context term; both implementations explicitly document this pessimism. |
| G14 | scope | ✅ CONFIDENT | Issue #1636 addresses Kimi-K3's 69 of 93 layers being linear attention. Does the implementation support arbitrary hybrid attention layer counts (e.g., 50/100 or 1/32) across any model config, or are there hidden assumptions tailored specifically to K3's 24/93 ratio? | The implementation supports arbitrary valid counts. It uses EffectiveKVBearingLayers as the full-attention count and computes KDA layers by subtraction; no production formula contains the K3-specific 24/93 constants. Those values occur only in examples and acceptance tests. |
| G15 | errors | ✅ CONFIDENT | If NewTrainedPhysicsModel receives a ModelHardwareConfig where NumLayers is zero or negative, how does the constructor handle construction errors before numKVBearingLayers is assigned via hw.ModelConfig.EffectiveKVBearingLayers()? | The constructor returns an error before constructing the model or assigning numKVBearingLayers. NumLayers must be strictly positive. |
| G16 | errors | ✅ CONFIDENT | In calculateTransformerFlops (roofline.go), if config.EffectiveKVBearingLayers() returns a value greater than nLayers, nKDA becomes negative. What safeguards prevent negative nKDA from silently reducing GEMM and SRAM ops in roofline execution? | For actual roofline execution, two safeguards apply: NumLayers must be positive, and EffectiveKVBearingLayers clamps any positive KVBearingLayers greater than NumLayers down to NumLayers. Consequently nKDA cannot be negative on validated inputs. |

### Items to fix
- **F1 · FLAW_FOUND** — The two latency backends implement the requested layer split and tests pass, but the change is not fully complete because every valid hybrid analytical run/replay still emits a user-visible warning claiming that step time charges all layers as full attention. That statement is now false and directly contradicts this PR's behavior.
  _sim/latency/trained_physics_model.go:304-313 and sim/latency/trained_physics_model.go:346-381 implement the trained-physics split; sim/latency/roofline.go:110-134 and sim/latency/roofline.go:272-287 implement roofline parity. However, cmd/root.go:622-631 warns that "step-time estimation still charge[s] all ... layers as full attention." Both run and replay call resolveLatencyConfig (cmd/root.go:1502; cmd/replay.go:287), so K3 realistically receives this incorrect warning on both commands. `go test ./sim/latency -count=1` passes._
- **G10 · FLAW_FOUND** — No equivalent roofline test isolates the new KDA decode-compute formula. The roofline tests only compare final roofline step times; because roofline takes max(compute, memory), the long-context decode cases can remain memory-bound and pass even if the KDA GEMM formula becomes context-dependent.
  _The trained-physics isolation test is sim/latency/kda_step_time_test.go:135-169. Roofline coverage at sim/latency/kda_step_time_test.go:181-225 checks non-hybrid equality, lower hybrid time, and final step-time context growth, but never calls calculateTransformerFlops or otherwise forces/asserts compute-bound behavior. The new roofline compute branch is at sim/latency/roofline.go:124-130, while final latency masks one resource using math.Max at sim/latency/roofline.go:375-380._

### Important to consider
- **F2** — Update the hybrid-model warning to say step time is split while KDA weights remain pessimistically modeled as full-attention weights.
  _CLAUDE.md:616 updates the hybrid-attention description; sim/model_hardware_config.go:27 and sim/model_hardware_config.go:63-72 document both consumers. The stale user-facing description is at cmd/root.go:622-631._
- **F3** — The adjacent warning text at cmd/root.go:627-630 is stale as well and is the user-visible part of the same concern reported in F1.
  _cmd/root.go:622-625 contains the stale comment; the new split is implemented at sim/latency/trained_physics_model.go:304-313, sim/latency/trained_physics_model.go:346-381, and sim/latency/roofline.go:110-134._
- **G1** — The unexported calculation helpers assume validated input; direct same-package tests can bypass that precondition.
  _sim/latency/roofline.go:63-65 and sim/latency/roofline.go:319-320 state the validation precondition. ValidateRooflineConfig rejects NumLayers <= 0 at sim/latency/config.go:529-531, and NewLatencyModel invokes it at sim/latency/latency.go:155-162._
- **G2** — If speculative decode is later implemented, both backends and all decode basis terms—not only KDA—will need token-count auditing.
  _sim/latency/trained_physics_model.go:293-297 increments totalDecodeTokens once; the KDA term consumes it at sim/latency/trained_physics_model.go:346-365. Actual decode scheduling fixes decodeTokens to 1 and sets NumNewTokens=1 at sim/batch_formation.go:139-159 and sim/batch_formation.go:215-227. Roofline likewise explicitly passes one token for decode at sim/latency/roofline.go:355-361._
- **G5** — The malformed-config warning mentions KV sizing only even though the same fallback now also affects step time; expanding that wording would be clearer.
  _sim/model_hardware_config.go:73-85 defines the fallback. sim/latency/config.go:152-167 distinguishes absent linear_attn_config from malformed hybrid configuration and warns for the latter. cmd/root.go:626-631 emits the valid-hybrid warning; both run and replay use resolveLatencyConfig._
- **G8** — The differing full-attention chunk-context conventions are pre-existing and outside this KDA change.
  _Trained-physics full attention uses si := InputLen and si+ti/2 at sim/latency/trained_physics_model.go:271-277, while its KDA expression is ti*dH*dH at sim/latency/trained_physics_model.go:278-291. Roofline full attention computes effectiveCtx from sequenceLength/ProgressIndex at sim/latency/roofline.go:95-104, while KDA uses newT*dHead*dHead at sim/latency/roofline.go:124-130. Roofline receives ProgressIndex at sim/latency/roofline.go:345-349._
- **G9** — A future formula change adding context-dependent terms not proportional to lFull could require relaxing or restructuring the exact-ratio assertion.
  _The full-attention decode compute uses lFull*sumCtx while KDA uses totalDecodeTokens without sumCtx at sim/latency/trained_physics_model.go:346-365. Growing KV also uses lFull at sim/latency/trained_physics_model.go:381-388. The test's current default β₂ₐ=0 is documented at sim/latency/kda_step_time_test.go:80-112, and the positive-β₂ₐ branch is exercised separately at sim/latency/kda_step_time_test.go:135-169._
- **G11** — A dedicated coefficient may be appropriate after KDA measurements are available, but its absence is intentional calibration honesty rather than a correctness defect.
  _The rationale and calibration constraints are documented in sim/latency/trained_physics_model.go:151-173 and at the prefill implementation in sim/latency/trained_physics_model.go:279-291. The roofline equivalent is documented at sim/latency/roofline.go:110-128._
- **G14** — The zero value intentionally means non-hybrid/all-full, so a hypothetical valid hybrid with exactly zero full-attention layers cannot currently be represented.
  _sim/latency/trained_physics_model.go:239-248 computes lKDA=L-lFull, and sim/latency/roofline.go:124-125 computes nKDA=nLayers-nFull. sim/model_hardware_config.go:73-85 clamps an explicit positive full-layer count to NumLayers. K3-specific values appear in tests such as sim/latency/kda_step_time_test.go:65-76._

namasl:
@claude /blis-pr-review commit d38d9638 addresses two findings surfaced by the experimental cross-vendor qa-review (RFC #1603) that this round caught beyond the prior official review: (F1/F3) the hybrid-detected runtime warning in cmd/root.go falsely claimed step-time still charges all layers as full attention — this PR made step-time KDA-aware, so the warning + comment (and a now-closed models.md follow-up bullet) are corrected to say only weight-memory remains pessimistic (#1638); (G10) added TestRooflineKDA_DecodeComputeGemmContextIndependentForKDA which isolates the roofline KDA decode-compute GEMM (asserts context-independence directly via calculateTransformerFlops, ratio 24/93 at 1e-9), closing the max(compute,memory)-masking gap symmetric to the trained-physics isolation test. cmd/doc/test only; no sim/ logic change, INV-6/INV-BC-DP1 intact; build/test/lint clean. Please confirm; also consider the /archon-pr-review output above once available.

namasl:

## qa-review adjudication — PR #1650: ✅ PASS

|----|----|----|----|
| F1 | FLAW_FOUND | ✅ RESOLVED | The stale user-visible warning has been corrected. It now says step-time estimation splits attention cost by layer type and identifies only weight-memory as pessimistic for KDA layers. |
| G10 | FLAW_FOUND | ✅ RESOLVED | A roofline-specific test now bypasses the compute-versus-memory maximum and directly tests the KDA decode-compute FLOP formula. It verifies that context-dependent GEMM growth is exactly proportional to the 24/93 full-attention-layer fraction, so KDA layers contribute no context-dependent decode GEMM term. |

### Still blocking

### Cleared
- **F1** (✅ RESOLVED) — The stale user-visible warning has been corrected. It now says step-time estimation splits attention cost by layer type and identifies only weight-memory as pessimistic for KDA layers.
  _cmd/root.go:622-627 documents the corrected behavior, and cmd/root.go:628-634 emits the corrected warning for hybrid models._
- **G10** (✅ RESOLVED) — A roofline-specific test now bypasses the compute-versus-memory maximum and directly tests the KDA decode-compute FLOP formula. It verifies that context-dependent GEMM growth is exactly proportional to the 24/93 full-attention-layer fraction, so KDA layers contribute no context-dependent decode GEMM term.
  _sim/latency/kda_step_time_test.go:238-260 directly calls calculateTransformerFlops with attention enabled and MLP disabled, computes context-growth deltas, and asserts the 24/93 ratio. The production split it covers is at sim/latency/roofline.go:124-133. `go test ./sim/latency` passes at the current PR head._

susiejojo (approved):
lgtm

PR #1651, opened 2026-08-31
feat(calibrate): add real-vs-sim throughput comparison to calibration report (#1647)

### LINKED ISSUE #1647: calibrate: add throughput comparison (output tokens/sec + requests/sec) to calibration report

`blis calibrate` today reports only **per-request latency** fidelity — the `metrics` map holds exactly `ttft`, `itl`, `e2e` (`sim/workload/calibrate.go:536-550`), each with workload-level percentiles and request-level MAPE/Pearson. There is **no throughput comparison** anywhere in `CalibrationReport` (`sim/workload/calibrate.go:56`). A full scan of a produced `calibration_report.json` confirms this: top-level keys are `trace_info`, `metrics` (`ttft`/`itl`/`e2e` only), `config_match`, `known_limitations`, `ttft_tolerance` — no throughput/tokens-per-sec/requests-per-sec field.

So a user calibrating BLIS against a real server can answer *"does the sim predict each request's latency?"* but **not** *"does the sim reproduce the server's aggregate output-token throughput and request-completion rate?"* — which is a primary capacity-planning question and the headline metric of most serving benchmarks. Latency fidelity and throughput fidelity are distinct: a sim can track per-request latency ordering well (high Pearson) yet mispredict aggregate throughput if it mismodels batching/contention. We hit exactly this calibrating weka-real ORC runs (Llama-3.1-70B TP4): under concurrency the latency MAPE was "poor" and we had no throughput number to corroborate whether the divergence was a scaling issue or a batching-contention issue.

**Proposed solution**

Add an optional `throughput` block to `CalibrationReport`, populated over the same `request_id`-matched set already used for latency, comparing real vs. simulated:
- **output-token throughput** = `total_output_tokens / runtime_seconds`
- **request throughput** = `matched_requests / runtime_seconds`
- **output-token throughput per GPU** = `output_token_throughput / num_gpus` — the standard normalized metric in serving benchmarks (lets a 70B/TP4 run be compared against a 7B/TP1 run fairly). Emitted **only** when the GPU count is supplied via a new `--num-gpus` flag (see below); omitted otherwise.

Both real and sim, with `error` / `percent_error` and (mirroring `ttft_tolerance` / `hit_rate`) an optional within-tolerance verdict. Emitted only when derivable, omitted otherwise so the legacy report shape is unchanged (same convention as the existing `Goodput`/`HitRate` optional sub-reports).

**On the per-GPU metric and GPU count sourcing.** The total GPU count for a deployment is `TP × PP × DP × num_instances`, but the trace header's `TraceServerConfig` records **only `TensorParallel`** (`sim/workload/tracev2.go:141`) — no DP/PP degree and no instance count — so the true GPU count is **not recoverable from existing calibrate inputs** for DP>1 or multi-instance runs. Deriving it from TP alone would silently undercount (a fidelity trap). Therefore the GPU count is supplied explicitly by the operator via a new `--num-gpus N` flag (same "operator declares it" convention as `--vllm-commit`); the per-GPU fields appear only when it is set. Note that since real and sim model the **same** deployment, one GPU count divides both sides, so `percent_error` on tokens/sec/GPU is identical to `percent_error` on raw tokens/sec — the per-GPU metric adds reporting/comparability value (benchmark parity, cross-config comparison), not a new tolerance dimension. The `--throughput-tolerance-pct` verdict therefore applies to the raw throughput; the per-GPU figures are reported alongside for reference.

Most of the primitives already exist:
- The **real-side runtime** is already computed in `cmd/calibrate.go:208-224` (`runtimeSec` = span from earliest `SendTimeUs` to latest `LastChunkTimeUs` over the matched set) — currently used only for the goodput block.
- The **output-throughput formula already exists** in the adapter-reference path: `loadBLISAggregate` (`cmd/calibrate.go:429`) computes `OutputThroughput = TotalOutputTokens / VllmDurationSec`. It just isn't wired into the main trace-calibration report.
- Real output tokens come from the matched `trace.Records`; sim output tokens + duration come from `sim_result.json` (per-request `output_tokens` + `e2e_us`) or the `--sim-metrics` `MetricsOutput` (`TotalOutputTokens`, `VllmDurationSec`).

The raw throughput comparison needs no new CLI flag (computed from existing inputs); two optional flags: `--throughput-tolerance-pct` (mirroring `--hit-rate-tolerance-pp` / `--ttft-mape-threshold`) for the within-tolerance verdict, and `--num-gpus` to enable the per-GPU normalization.

```bash
# no new flag needed — raw throughput block appears in the report automatically:
./blis calibrate --trace-header t.yaml --trace-data d.csv \
  --sim-results results.json --report calibration.json
# optional tolerance verdict:
./blis calibrate ... --throughput-tolerance-pct 15
# optional per-GPU normalization (also emits output_tokens_per_sec_per_gpu, real + sim):
./blis calibrate ... --throughput-tolerance-pct 15 --num-gpus 4
```

- [ ] Core simulator (`sim/`)
- [x] Workload generation (`sim/workload/`) — `CalibrationReport` struct + comparison logic in `sim/workload/calibrate.go`
- [ ] KV cache (`sim/kv/`)
- [x] CLI (`cmd/`) — wire real runtime + sim tokens into the new comparison in `cmd/calibrate.go` (optional new `--throughput-tolerance-pct` and `--num-gpus` flags)
- [ ] New package needed

**Extension friction check**
- **Files to change: ~2** (`sim/workload/calibrate.go` for the struct + compute function; `cmd/calibrate.go` to populate it and register the `--throughput-tolerance-pct` / `--num-gpus` flags — the runtime and throughput-formula primitives already live there). Plus tests.
- **New interface?** No — extends the existing `CalibrationReport` with an optional pointer field, exactly like `Goodput`/`HitRate`/`TTFTTolerance` already do.
- **Invariants?** None affected. This is a post-hoc analysis over completed run outputs — no change to sim execution, so conservation/causality/determinism (INV-1/-5/-6) are untouched. Omitting the field when not derivable preserves byte-identical legacy report shape for existing consumers.

**Alternatives considered**
- *Derive throughput externally from `sim_result.json` + observe `data.csv`.* Works (we did it ad-hoc) but every consumer re-implements the real-side runtime window and token-sum, and it's not reproducible/versioned alongside the latency verdicts — the calibration report should be the single source of fidelity truth.
- *Use the goodput block instead.* Goodput measures fraction-of-requests-within-SLO, not aggregate throughput, and requires SLO targets; it answers a different question.
- *Reuse the adapter-reference throughput path.* That path (`--adapter-reference`) is DT-specific and doesn't consume the observe trace / `request_id` join, so it can't compare against a real observe run.

- Extends the observe→replay→calibrate pipeline (#659/#689/#701).
- Directly parallels the KV hit-rate comparison (#1583) and goodput comparison (#1413), which added optional `HitRate`/`Goodput` sub-reports to the same struct — this issue applies the identical pattern for throughput.
- Reuses the runtime computation added for goodput (`cmd/calibrate.go:208-224`) and the throughput formula from the adapter-fidelity path (#1470, `cmd/calibrate.go:429`).

susiejojo:
## Summary

`blis calibrate` today reports only per-request **latency** fidelity (`ttft`/`itl`/`e2e`). This adds an optional `throughput` block comparing real vs simulated **aggregate output-token throughput** and **request-completion rate** — the headline metric of most serving benchmarks and a primary capacity-planning question. Latency and throughput fidelity are distinct: a sim can track per-request latency well yet mispredict aggregate throughput if it mismodels batching/contention.

The block appears automatically from the already-required inputs (trace + sim-results); two optional flags add a tolerance verdict and per-GPU normalization. It mirrors the existing optional `Goodput`/`HitRate`/`TTFTTolerance` sub-reports.

Closes #1647.

## Behavioral Contracts

- **BC-1 — Raw block from required inputs.** GIVEN a trace + matching sim-results with a positive makespan, WHEN `blis calibrate` runs with no throughput flag, THEN the report gains a `throughput` block with real & sim `output_tokens_per_sec` and `requests_per_sec`, each with signed `error` and `percent_error`.
- **BC-2 — Emitted only when derivable; never Inf/NaN.** GIVEN a non-positive real *or* sim makespan or zero matched output tokens, THEN the `throughput` key is omitted (a non-finite value would fail `json.MarshalIndent` and discard the whole report). Present-when-derivable, the `TTFTTolerance` precedent.
- **BC-3 — Client-frame makespan isolates the batching signal.** `real_runtime_sec` = latest last-chunk − earliest send; `sim_runtime_sec` = latest (send + client-frame sim E2E) − earliest send, where sim E2E is normalized (`+ RTT + upload + download`) identically to the latency leg. Shared `min(send)` origin ⇒ the block isolates real-vs-sim batching/contention, consistent with the `e2e` block.
- **BC-4 — Per-GPU normalization.** `--num-gpus N` (>0) adds `output_tokens_per_sec_per_gpu` (real+sim); one GPU count divides both sides, so per-GPU `percent_error` equals the raw value.
- **BC-5 — Tolerance verdict.** `--throughput-tolerance-pct P` (>0) adds `tolerance_pct` + `within` on raw output-token throughput.
- **BC-6 — Input validation.** `--num-gpus` ≥ 0 and `--throughput-tolerance-pct` finite ≥ 0, else fatal naming the flag.

## Validity boundary

The reconstructed sim makespan (`send + simE2E`) is a physical sim timeline only for **fixed-mode replay** (the standard observe→replay→calibrate path). Under closed-loop / concurrent-session replay the sim regenerates the arrival schedule, so treat the number as open-loop only. Documented in the guide and the struct comment.

## Testing

`go build ./...`, `go test ./...` (full suite green), `golangci-lint run` (CI config, clean). New tests: perfect match, metamorphic E2E×3 scaling, client-frame network shift, per-GPU, tolerance within/exceeds + inclusive boundary, corrupt-record skip, degenerate→nil (unit); block present, per-GPU + verdict (cmd e2e).

## Files
- `sim/workload/calibrate.go` — `ThroughputComparison` struct, `CalibrationReport.Throughput` field, `ComputeThroughputComparison` (returns nil, no library `Fatalf`).
- `cmd/calibrate.go` — wiring + `--num-gpus`/`--throughput-tolerance-pct` flags + validation + stderr summary; hoisted `simByID`/`matched` to share with the goodput block.
- `docs/guide/observe-replay-calibrate.md`, `CLAUDE.md` — docs.


susiejojo:
@claude /blis-pr-review

susiejojo:
Thanks for the thorough review. Addressed in f035c4b8:

**Important**
1. **Silent omission (R1)** — `cmd/calibrate.go` now warns when `Throughput == nil` *and* the operator passed `--throughput-tolerance-pct`/`--num-gpus`, so "not derivable" is distinguishable from "feature absent" (matches the hit-rate/ITL skip warnings).
2. **Coverage masking** — added a coverage guard: warns when `matched / (matched + unmatched_real) < 0.9`, so a sim that dropped requests can't spuriously PASS (real tok/s measured over the survivor subset would deflate toward sim). Documented the `matched_pairs`/`num_requests` cross-check in the guide.
3. **Closed-loop validity** — `cmd` now reads `trace.Header.SessionContextGrowth` and warns when set; the boundary is no longer documentation-only on a machine-readable verdict.

**Minor**
5. **Non-`ok` records** — `ComputeThroughputComparison` now skips `Status != "ok"` (matches the goodput numerator); failed/timeout records with partial tokens no longer bias the metric.
6. **Unit doc** — added a note that `*_percent_error` is a fraction (0–1, BLIS-wide convention) while `tolerance_pct` is a percentage (0–100), with the ×100 relationship.
7. **Boundary test** — reworked to an *exact* 10% error (100 vs 90 tok/s, 1s makespan) so `boundaryEps` is genuinely load-bearing; asserts the error is exactly 0.10 before checking WITHIN.
8. **BC-6 fatal guards** — added a table test (negative `--num-gpus`, negative/NaN/Inf `--throughput-tolerance-pct`) via `logrus.ExitFunc` interception.

**Deferred with rationale**
4. **Two makespan windows** — left as-is intentionally. Extracting a shared helper would change `goodput`'s denominator on traces containing failed records (goodput counts `ok/error/timeout` in its window; throughput now counts only `ok`), altering already-shipped `goodput.*_rps` output. The guards differ *because* the two metrics have different eligibility (goodput scores attainment including failures in the denominator; throughput measures completed-work rate). Added a comment noting the intentional difference rather than unifying and perturbing goodput.

**CLAUDE.md** — the working tree already has the `#1647` block committed (both HEAD and working tree contain it); the earlier "reverted" state the bot saw was a transient snapshot from a since-corrected worktree mix-up.

Build / `go test ./...` / `golangci-lint` (CI v2.9.0) all clean.

susiejojo:
@claude /blis-pr-review

namasl:

## qa-review — PR #1651: ⛔ BLOCK
_feat(calibrate): add real-vs-sim throughput comparison to calibration report (#1647) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ❌ FLAW_FOUND | Does this PR fully implement the issue it closes, without unnecessary additions? | The requested report fields, tolerance, GPU normalization, CLI wiring, and tests are implemented, and the extra warnings are relevant safeguards rather than unrelated scope. However, the implementation is not fully correct for all realistic calibrate inputs: it reconstructs simulated completion timestamps from real send times even when replay used a closed-loop schedule, and it removes real failures together with their successful simulated counterparts, which can hide a completion-rate fidelity gap. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The user-facing guide, CLI help, and project usage documentation cover the new report block, formulas, units, optional flags, omission behavior, per-GPU rationale, and validity boundary. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No materially stale comment was found. The new comments generally match the implementation. |
| G1 | correctness | ✅ CONFIDENT | In `ComputeThroughputComparison`, `clientSimE2E` is computed as a `float64` and then cast to `int64` via `int64(clientSimE2E)` when calculating `simEndUs := rec.SendTimeUs + int64(clientSimE2E)`. Why is sub-microsecond precision truncated before finding `maxSimEnd`, and could this precision loss distort makespan calculations for high-throughput or short-duration traces? | The cast truncates only the fractional part of a microsecond, so the reconstructed tail can move earlier by less than 1 µs. Simulator E2E values originate in microsecond ticks and are normally integral; fractional values mainly arise from bandwidth normalization. For realistic serving makespans this is negligible and does not violate the stated microsecond trace resolution. |
| G2 | correctness | ✅ CONFIDENT | In `ComputeThroughputComparison`, `minSend` is initialized to `-1` and updated when `rec.SendTimeUs < minSend`. If all matched records in a trace have `SendTimeUs == 0` (e.g. synthetic or batch-submitted traces) and `LastChunkTimeUs > 0`, `minSend` becomes `0` and passes the `minSend < 0` check. However, if all valid records have `SendTimeUs == 0` and `LastChunkTimeUs == 0`, how does the method prevent division by zero before reaching the `realRuntimeSec <= 0` guard? | There is no division before the guard. A record with LastChunkTimeUs equal to SendTimeUs is rejected in the loop; if every record is like that, matched remains zero and the function returns nil. Otherwise runtimes are calculated without division, checked for <= 0, and only then used as divisors. |
| G3 | invariants | ❌ FLAW_FOUND | INV-13 requires run/replay parity and INV-5 requires temporal causality. `ComputeThroughputComparison` computes `simRuntimeSec` by adding simulated `clientSimE2E` to the real trace's `rec.SendTimeUs`. When replaying traces under closed-loop mode or when arrival schedules shift, does relying on real trace `SendTimeUs` rather than actual simulated arrival timestamps break run/replay makespan parity? | Yes, for closed-loop replay the reconstructed sim makespan can be wrong because follow-up arrivals are generated from simulated completions, not the original real SendTimeUs values. This does not mutate simulator causality or directly violate INV-5, and INV-13 is specifically defined for fixed replay, but it does produce an invalid calibration metric for a supported realistic mode. The warning is incomplete because it checks SessionContextGrowth rather than the replay mode, which is not carried by SimResult. |
| G4 | invariants | ❌ FLAW_FOUND | INV-1 enforces request conservation across the simulation pipeline. In `ComputeThroughputComparison`, filtering out records where `rec.Status != "ok"` drops real failed requests, but what happens if a request failed in the real observe trace (`rec.Status != "ok"`) yet completed in `simByID`? Does filtering it on the real side create an asymmetry with how total simulated output tokens are aggregated? | The function excludes the whole pair, so that request contributes neither real nor simulated tokens, requests, nor makespan. That is symmetric arithmetically, but it masks a realistic fidelity failure: a real request that failed while the simulator completed it disappears instead of reducing real request throughput or otherwise surfacing the completion mismatch. The existing coverage warning also does not catch this because its denominator is based on request-ID matching, not successful real completion. |
| G5 | rules | ✅ CONFIDENT | Rule R1 forbids silent continues or drops without log visibility or explicit handling. In `ComputeThroughputComparison`, records failing `rec.Status != "ok"` or `rec.SendTimeUs < 0 \|\| rec.LastChunkTimeUs <= rec.SendTimeUs \|\| sr.E2E < 0` are skipped silently via `continue`. Why are corrupt or non-OK records skipped without returning a count or warning of skipped records back to the report caller? | The timestamp checks are defensive handling of corrupt or non-derivable rows, and returning nil when no usable aggregate remains is explicit handling rather than emitting invalid JSON. Non-OK status is an intentional domain filter documented in the guide. A separate skipped-row count would improve observability but is not required for the valid core calculation. |
| G6 | rules | ✅ CONFIDENT | Rule R20 requires strict CLI flag validation. In `cmd/calibrate.go`, `--num-gpus` is checked for `< 0`. If an operator explicitly passes `--num-gpus 0`, no per-GPU metrics are generated and no warning is printed about `--num-gpus` being ignored. Why is `numGPUs == 0` allowed silently when the user explicitly provides the flag on the command line? | Zero is the documented disabled/unset sentinel for this optional flag, so explicitly passing zero has the same intentional meaning as omitting it. Negative values are invalid and rejected; positive values enable normalization. |
| G7 | parity | ✅ CONFIDENT | Flags `--throughput-tolerance-pct` and `--num-gpus` were added to `blis calibrate`. Are corresponding throughput metrics and per-GPU normalization options supported in `blis run` or `blis replay` outputs (e.g. `--sim-metrics`), or does this introduce a reporting parity gap between simulator runs and calibration reports? | Run/replay already expose aggregate raw throughput inputs and rates, but they do not expose this calibration-specific real-vs-sim comparison, tolerance verdict, or per-GPU normalization. That is not a run/replay parity gap: the new flags operate on a CalibrationReport comparing two data sources, while INV-13 concerns equivalent simulator results between run and fixed replay. |
| G8 | parity | ✅ CONFIDENT | In `ComputeThroughputComparison`, network upload and download delays are added to `sr.E2E` using `computeUploadDelay` and `computeDownloadDelay`. How does this network adjustment compare with the exact network normalization implemented in `PrepareCalibrationPairs` for `e2e` latency metrics, and are both guaranteed to produce identical client-frame timestamps? | The network-adjustment expression is the same: E2E plus NetworkRTTUs plus upload delay based on simulated input tokens plus download delay based on simulated output tokens, using the same helper functions. Thus the client-frame latency value is identical before throughput converts it to an integer completion timestamp. |
| G9 | tests | ✅ CONFIDENT | The new tests in `sim/workload/calibrate_throughput_test.go` cover unit fixtures and basic flag behavior. Is there a companion invariant/law test verifying metamorphic properties—for example, that doubling all request output tokens in both real and sim traces exactly doubles both `RealOutputTokensPerSec` and `SimOutputTokensPerSec` while leaving `OutputTokensPerSecPercentError` unchanged? | There is no exact token-doubling metamorphic test. There is, however, a companion metamorphic law test that scales all simulated E2E values and verifies inverse throughput scaling and unchanged real throughput, plus direct tests for perfect matching, per-GPU scaling, tolerance boundaries, networking, and degenerate inputs. The absent token-doubling variant is additional coverage rather than an untested core path. |
| G10 | tests | ✅ CONFIDENT | In `cmd/calibrate_throughput_test.go`, tests like `TestCalibrateCmd_Throughput_Present` depend on mock trace CSV and JSON fixtures written to temporary directories. If the underlying matched-pair construction in `cmd/calibrate.go` changes how warm-up requests or missing `RequestID`s are mapped, would these tests detect regressions or break silently? | The command integration fixture would detect regressions that changed ordinary matching or caused the three expected IDs to disappear: it asserts matched_requests == 3 and exact rates. It does not itself exercise warm-up or missing-ID behavior, but those behaviors have dedicated PrepareCalibrationPairs tests, so such changes would not generally break silently. |
| G11 | rationale | ✅ CONFIDENT | Why was simulated makespan reconstructed post-hoc in `ComputeThroughputComparison` using `rec.SendTimeUs + int64(clientSimE2E)` rather than logging actual aggregate makespans directly from the simulator's execution engine during simulation run time? | Post-hoc reconstruction is needed to calculate both sides over exactly the request-ID-matched, warm-up-filtered survivor set and in the same client/network frame. The aggregate simulator metrics cover the whole simulation and do not carry matched-subset timing, while SimResult contains per-request E2E but no arrival or completion timestamp. Using aggregate engine duration would therefore mix unmatched requests and use a different frame. |
| G12 | rationale | ✅ CONFIDENT | The PR introduces `--num-gpus` because `TraceServerConfig` records only `TensorParallel` (TP) and not DP/PP or instance counts. Why was `--num-gpus` added as an imperative CLI flag rather than expanding `TraceServerConfig` or the trace header YAML schema to explicitly specify full cluster GPU topologies? | The CLI flag is appropriate because GPU count is calibration/deployment metadata supplied by the operator, not information recoverable from the current trace. Expanding TraceServerConfig would change the strict trace schema and still require the observer/operator to supply DP, PP, and instance count. Keeping it on calibrate avoids claiming that legacy traces contain a complete topology and limits the change to the report feature. |
| G13 | scope | ✅ CONFIDENT | Issue #1647 requested real-vs-sim throughput comparison fields in `CalibrationReport`. Does adding the coverage warning for `coverage < 0.9` and the `SessionContextGrowth` closed-loop validity check go beyond the requested issue scope, and why were those specific warning criteria included in this PR? | The warnings are ancillary safeguards directly tied to whether the new comparison is meaningful. The 90% coverage threshold flags survivor bias without suppressing output, while the session warning communicates the documented fixed/open-loop validity boundary. They do not alter simulation or report calculations. |
| G14 | scope | ✅ CONFIDENT | The PR adds optional per-GPU normalization fields for output tokens (`real_output_tokens_per_sec_per_gpu` and `sim_output_tokens_per_sec_per_gpu`). Was per-GPU request throughput (`requests_per_sec_per_gpu`) considered, and what was the reason for omitting per-GPU request rates? | Per-GPU request throughput was omitted consistently with the acceptance criteria, which specifically requests output-token throughput per GPU as the standard normalized capacity metric. Request rate depends strongly on request-size distribution, so dividing it by GPU count adds less cross-configuration comparability than output tokens/sec/GPU. |
| G15 | errors | ✅ CONFIDENT | In `cmd/calibrate.go`, when `report.Throughput == nil` while `--throughput-tolerance-pct > 0` or `--num-gpus > 0`, the CLI logs a warning (`logrus.Warnf`) and continues execution. Why does this log a warning instead of failing fast with `logrus.Fatalf`, given that requested tolerance checks cannot be performed? | The flags request optional report adornments, not that calibration become impossible when throughput is non-derivable. The established contract is to omit the optional block rather than fabricate Inf/NaN, while warning makes the ineffective requested flag visible. Failing would also prevent the still-valid latency calibration report from being produced. |
| G16 | errors | ✅ CONFIDENT | In `ComputeThroughputComparison`, when inputs are invalid or non-derivable (e.g. `realRuntimeSec <= 0` or `realOutTok == 0`), the function returns `nil`. Should `ComputeThroughputComparison` return a rich Go `error` detailing the reason for omission (e.g., zero makespan vs zero output tokens) rather than a bare `nil` pointer? | A rich error would improve diagnostics, but bare nil is consistent with the API's meaning that this optional sub-report is not derivable. The caller distinguishes silent omission from an explicitly requested option by warning, and no invalid value is serialized. |

### Items to fix
- **F1 · FLAW_FOUND** — The requested report fields, tolerance, GPU normalization, CLI wiring, and tests are implemented, and the extra warnings are relevant safeguards rather than unrelated scope. However, the implementation is not fully correct for all realistic calibrate inputs: it reconstructs simulated completion timestamps from real send times even when replay used a closed-loop schedule, and it removes real failures together with their successful simulated counterparts, which can hide a completion-rate fidelity gap.
  _sim/workload/calibrate.go:213-250 reconstructs the sim makespan and excludes every pair whose real status is not "ok"; cmd/calibrate.go:277-307 still emits the comparison and only warns when SessionContextGrowth is non-empty, while cmd/replay.go:181-205 supports closed-loop replay independently of that header field._
- **G3 · FLAW_FOUND** — Yes, for closed-loop replay the reconstructed sim makespan can be wrong because follow-up arrivals are generated from simulated completions, not the original real SendTimeUs values. This does not mutate simulator causality or directly violate INV-5, and INV-13 is specifically defined for fixed replay, but it does produce an invalid calibration metric for a supported realistic mode. The warning is incomplete because it checks SessionContextGrowth rather than the replay mode, which is not carried by SimResult.
  _docs/contributing/standards/invariants.md:176-180 limits INV-13 parity to --session-mode fixed; cmd/replay.go:181-205 shows closed-loop replay injects round zero and lets SessionManager drive follow-ups; sim/workload/calibrate.go:143-148 acknowledges real send times are invalid in that mode; cmd/calibrate.go:305-307 warns only when SessionContextGrowth is non-empty, although cmd/replay.go:99-119 permits --session-mode closed-loop on traces without that marker._
- **G4 · FLAW_FOUND** — The function excludes the whole pair, so that request contributes neither real nor simulated tokens, requests, nor makespan. That is symmetric arithmetically, but it masks a realistic fidelity failure: a real request that failed while the simulator completed it disappears instead of reducing real request throughput or otherwise surfacing the completion mismatch. The existing coverage warning also does not catch this because its denominator is based on request-ID matching, not successful real completion.
  _sim/workload/calibrate.go:213-223 performs the real-status filter before adding either side; cmd/calibrate.go:288-299 computes coverage from pairs.MatchedCount and pairs.UnmatchedReal; sim/workload/calibrate.go:410-443 shows latency matching increments MatchedCount without filtering status. Scenario: one status="error" trace record with a matching completed SimResult is counted as matched for coverage but omitted entirely from throughput._

### Important to consider
- **F2** — The guide equates SessionContextGrowth with detecting closed-loop replay, but ordinary session traces can be replayed closed-loop with an empty value; this is the same limitation described under G3.
  _docs/guide/observe-replay-calibrate.md:468-475 documents both flags; docs/guide/observe-replay-calibrate.md:560-597 documents the throughput JSON and semantics; cmd/calibrate.go:615-619 registers descriptive help; CLAUDE.md:158-172 adds an example._
- **F3** — The phrase "same request_id-matched set used for latency" is imprecise because throughput additionally excludes non-OK and non-positive-duration records.
  _sim/workload/calibrate.go:134-172 describes the client-frame formula and closed-loop limitation; sim/workload/calibrate.go:213-231 documents the status and timestamp filters; cmd/calibrate.go:270-307 describes invocation and warnings._
- **G1** — Keeping maxSimEnd as float64 would avoid the sub-microsecond truncation and could matter for synthetic sub-microsecond fixtures.
  _sim/workload/calibrate.go:234-250 computes clientSimE2E, casts it to int64, and converts the resulting microsecond span to seconds; sim/workload/calibrate.go:343-352 defines SimResult latency units as microseconds; sim/workload/network.go:27-45 shows bandwidth delays may be fractional._
- **G5** — Reporting excluded-status and invalid-timestamp counts would make diagnosis easier; the substantive failed-real/successful-sim masking problem is captured under G4.
  _sim/workload/calibrate.go:213-231 documents and applies the filters; sim/workload/calibrate.go:252-264 explicitly returns nil for non-derivable aggregates; cmd/calibrate.go:281-286 warns when a requested optional result cannot be produced; docs/guide/observe-replay-calibrate.md:590 states that only status=="ok" contributes._
- **G6** — Cobra's Flags().Changed could distinguish explicit zero if maintainers want a diagnostic, but zero is not an ambiguous operational value here.
  _cmd/calibrate.go:146-148 rejects negative values; cmd/calibrate.go:617 describes per-GPU fields as emitted only when greater than zero; sim/workload/calibrate.go:283-291 only populates the fields for numGPUs > 0; docs/guide/observe-replay-calibrate.md:471 documents the default as 0._
- **G7** — Run/replay do not currently emit tokens/sec/GPU, so consumers wanting normalized standalone simulator metrics must divide externally.
  _sim/metrics_utils.go:58-70 includes TotalOutputTokens, VllmDurationSec, ResponsesPerSec, and TokensPerSec in MetricsOutput; sim/metrics.go:94-105 populates aggregate duration and token totals; docs/contributing/standards/invariants.md:176-180 defines run/replay parity; cmd/calibrate.go:615-618 registers the new flags only on calibrate._
- **G8** — The final throughput timestamp truncates fractional microseconds via int64, whereas the E2E metric retains float64 precision.
  _sim/workload/calibrate.go:446-452 implements PrepareCalibrationPairs normalization; sim/workload/calibrate.go:234-237 implements the throughput normalization with the same fields and helpers; sim/workload/network.go:27-45 defines those helpers._
- **G9** — A token-scaling law should use zero bandwidth or hold network delay constant, because with bandwidth normalization changing token counts also changes the makespan.
  _sim/workload/calibrate_throughput_test.go:93-126 contains TestComputeThroughput_MetamorphicScaleE2E; lines 52-91 test perfect matching; lines 153-228 test per-GPU and tolerance behavior; lines 255-275 test degenerate inputs. No token-doubling test appears in that file._
- **G10** — A throughput command-level test combining warm-up and an unmatched ID would provide stronger end-to-end coverage of the newly duplicated matched-map construction.
  _cmd/calibrate_throughput_test.go:57-111 runs the real command path and asserts a non-nil block, three matches, and expected rates; sim/workload/calibrate_test.go:123-165 contains warm-up and unmatched-request tests; cmd/calibrate.go:201-217 constructs the shared matched map._
- **G11** — Adding actual simulated arrival/completion timestamps to the per-request artifact would make closed-loop comparison accurate, but would broaden the artifact schema.
  _sim/workload/calibrate.go:343-355 shows SimResult has per-request latency and tokens but no arrival/completion timestamp; sim/metrics_utils.go:58-70 exposes only whole-run aggregate duration and rates; cmd/calibrate.go:201-217 builds the calibration-specific matched set; sim/workload/calibrate.go:234-250 reconstructs its client-frame tail._
- **G12** — A future resolved deployment-topology header could be useful for broader replay provenance, but it is not necessary for this calibration-only normalization.
  _sim/workload/tracev2.go:139-148 shows TraceServerConfig contains TensorParallel but no DP, PP, instance count, or total GPUs; cmd/calibrate.go:617 explicitly describes num-gpus as TP×PP×DP×instances and operator-supplied; sim/workload/tracev2.go:43-45 notes that trace schema additions affect strict readers/versioning._
- **G13** — The 90% threshold is a policy choice, and the SessionContextGrowth check does not detect every closed-loop replay; see G3.
  _cmd/calibrate.go:287-300 computes and logs the coverage warning; cmd/calibrate.go:301-307 logs the session validity warning; sim/workload/calibrate.go:143-148 documents why reconstructed timing is only physically valid for fixed replay._
- **G15** — A separate strict mode could be useful for CI users who require a throughput verdict.
  _sim/workload/calibrate.go:252-264 defines nil omission for degenerate inputs; cmd/calibrate.go:281-286 warns specifically when tolerance or GPU normalization was requested; sim/workload/calibrate.go:88 uses omitempty; docs/guide/observe-replay-calibrate.md:560-562 documents omission when non-derivable._
- **G16** — Returning a result plus an omission-reason enum would permit more actionable warnings without treating normal non-derivability as a fatal error.
  _sim/workload/calibrate.go:252-264 enumerates the nil conditions; cmd/calibrate.go:281-286 handles nil with a warning when optional flags were supplied; sim/workload/calibrate.go:81-88 documents optional omission from CalibrationReport._

susiejojo:
@claude /blis-pr-review check to see if all above concerns have been addressed

namasl:

## qa-review adjudication — PR #1651: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The follow-up does not fully fix either correctness concern. Plain traces replayed in closed-loop mode still produce a silently invalid simulated makespan, and the new status-mismatch warning is only executed when a throughput block is derivable. If all matched real requests failed while the simulator completed them, throughput is nil and the mismatch remains hidden. An implementation fix is still required, or a real open follow-up issue must cover any deliberately deferred limitation. |
| G3 | FLAW_FOUND | ❌ STILL-OPEN | Documentation of the blind spot does not correct or waive the defect. Closed-loop replay remains supported for plain traces, but calibrate cannot identify that provenance and still emits a makespan reconstructed from real SendTimeUs. The author cited no open follow-up issue and provided no independently verifiable basis that this supported mode is outside scope. A trustworthy replay-mode signal, refusal to emit the metric when validity cannot be established, or a genuine covering follow-up issue is still required. |
| G4 | FLAW_FOUND | ❌ STILL-OPEN | The warning fixes only mixed datasets containing enough successful real records to derive throughput. In the finding's single-error scenario, the failed real record is filtered out, ComputeThroughputComparison returns nil, and the status-mismatch guard is never reached. With no throughput flags, even the generic nil warning is suppressed. The mismatch must be surfaced independently of whether the throughput block is derivable. |

### Still blocking
- **F1** — The follow-up does not fully fix either correctness concern. Plain traces replayed in closed-loop mode still produce a silently invalid simulated makespan, and the new status-mismatch warning is only executed when a throughput block is derivable. If all matched real requests failed while the simulator completed them, throughput is nil and the mismatch remains hidden. An implementation fix is still required, or a real open follow-up issue must cover any deliberately deferred limitation.
  _sim/workload/calibrate.go:216-223 excludes every non-ok real record and lines 253-254 return nil when none remain. cmd/calibrate.go:285-291 places all mismatch handling in the non-nil throughput branch, while lines 316-328 contain the warning inside that branch. cmd/calibrate.go:335-343 explicitly acknowledges that plain closed-loop replay remains undetectable and silently invalid._
- **G3** — Documentation of the blind spot does not correct or waive the defect. Closed-loop replay remains supported for plain traces, but calibrate cannot identify that provenance and still emits a makespan reconstructed from real SendTimeUs. The author cited no open follow-up issue and provided no independently verifiable basis that this supported mode is outside scope. A trustworthy replay-mode signal, refusal to emit the metric when validity cannot be established, or a genuine covering follow-up issue is still required.
  _cmd/replay.go:99-119 accepts or promotes plain traces to closed-loop mode, and lines 181-205 generate closed-loop arrivals through SessionManager. sim/workload/calibrate.go:143-147 states that real SendTimeUs is not the simulated arrival schedule in this mode. cmd/calibrate.go:335-343 expressly says the plain-trace case remains silently open-loop-invalid and only warns when SessionContextGrowth is present. docs/guide/observe-replay-calibrate.md:593-594 merely documents that residual invalid-output behavior._
- **G4** — The warning fixes only mixed datasets containing enough successful real records to derive throughput. In the finding's single-error scenario, the failed real record is filtered out, ComputeThroughputComparison returns nil, and the status-mismatch guard is never reached. With no throughput flags, even the generic nil warning is suppressed. The mismatch must be surfaced independently of whether the throughput block is derivable.
  _sim/workload/calibrate.go:216-223 filters the failed real record and lines 253-254 return nil when no records remain. cmd/calibrate.go:285-290 only emits the generic skipped warning when a throughput option was requested; lines 291-328 put the real-failed/sim-completed warning exclusively in the non-nil branch. cmd/calibrate_throughput_test.go:171-183 avoids the failing all-error case by adding two successful records, so it does not verify the original one-error scenario._

### Cleared
- _None._

namasl:

## qa-review adjudication — PR #1651: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | Both correctness concerns remain. The closed-loop limitation is only documented, while calibrate still emits a reconstructed simulated makespan that can be invalid for plain traces replayed in closed-loop mode. The status-mismatch warning also remains conditional on a non-nil throughput comparison, so an all-failed-real/matched-sim dataset silently hides the completion mismatch. An implementation fix is required for both cases, or each deliberately deferred limitation must be covered by a genuine open follow-up issue. |
| G3 | FLAW_FOUND | ❌ STILL-OPEN | Documenting the known blind spot does not fix or independently justify waiving it. Plain traces can be replayed in supported closed-loop mode, but calibrate receives no trustworthy replay-mode provenance and still emits a makespan based on the real send schedule. The author must provide an implementation that carries or validates replay-mode provenance, decline to emit the metric when validity cannot be established, provide a concrete independently verifiable scope justification, or cite a genuine open issue covering the defect. |
| G4 | FLAW_FOUND | ❌ STILL-OPEN | The added warning does not cover the finding's all-error scenario. If the only matched real request failed while the simulator completed it, no record survives the ok-only filter, throughput is nil, and execution never reaches the status-mismatch warning. With no throughput flags, the nil branch emits no warning either. The mismatch detection must run independently of throughput derivability, with coverage for the original single-error scenario. |

### Still blocking
- **F1** — Both correctness concerns remain. The closed-loop limitation is only documented, while calibrate still emits a reconstructed simulated makespan that can be invalid for plain traces replayed in closed-loop mode. The status-mismatch warning also remains conditional on a non-nil throughput comparison, so an all-failed-real/matched-sim dataset silently hides the completion mismatch. An implementation fix is required for both cases, or each deliberately deferred limitation must be covered by a genuine open follow-up issue.
  _sim/workload/calibrate.go:216-223 filters all non-ok real records, and sim/workload/calibrate.go:253-254 returns nil if none survive. cmd/calibrate.go:285-291 branches on nil throughput, while the mismatch scan and warning are only inside the non-nil branch at cmd/calibrate.go:310-328. For closed-loop replay, cmd/replay.go:99-119 supports and auto-promotes closed-loop mode, and cmd/replay.go:181-205 regenerates arrivals through the session machinery. sim/workload/calibrate.go:143-147 acknowledges that real SendTimeUs is not the simulated schedule in this mode, while cmd/calibrate.go:335-343 admits plain closed-loop provenance is undetectable and still emits the metric without warning. No open follow-up issue was cited._
- **G3** — Documenting the known blind spot does not fix or independently justify waiving it. Plain traces can be replayed in supported closed-loop mode, but calibrate receives no trustworthy replay-mode provenance and still emits a makespan based on the real send schedule. The author must provide an implementation that carries or validates replay-mode provenance, decline to emit the metric when validity cannot be established, provide a concrete independently verifiable scope justification, or cite a genuine open issue covering the defect.
  _cmd/replay.go:99-119 permits closed-loop mode and may force it through --concurrent-sessions; cmd/replay.go:181-205 uses SessionManager-driven follow-up arrivals. sim/workload/calibrate.go:143-147 states that the reconstructed makespan is not a physical simulated timeline under closed-loop replay. cmd/calibrate.go:335-343 explicitly says plain closed-loop replay is not detected and remains silently invalid; docs/guide/observe-replay-calibrate.md:593-594 merely transfers responsibility to the operator. No follow-up issue was cited._
- **G4** — The added warning does not cover the finding's all-error scenario. If the only matched real request failed while the simulator completed it, no record survives the ok-only filter, throughput is nil, and execution never reaches the status-mismatch warning. With no throughput flags, the nil branch emits no warning either. The mismatch detection must run independently of throughput derivability, with coverage for the original single-error scenario.
  _sim/workload/calibrate.go:216-223 excludes the failed real record and sim/workload/calibrate.go:253-254 returns nil when no records survive. cmd/calibrate.go:285-290 warns for nil throughput only when a throughput option was supplied, while cmd/calibrate.go:291-328 places status-mismatch detection exclusively in the non-nil branch. cmd/calibrate_throughput_test.go:171-183 includes two successful records alongside the failed record, and cmd/calibrate_throughput_test.go:217-221 explicitly requires throughput to remain derivable, so it does not test the original one-failed-record case._

### Cleared
- _None._

susiejojo:
Addressed the adjudication in b01cf24c — the fix now **enforces** the validity boundary instead of documenting it.

**G3 + F1 (silently-invalid closed-loop makespan) → fail-loud refusal.** Rather than emit a throughput verdict calibrate cannot verify, it now emits the block **only** when provenance confirms the makespan is a physical timeline:

```
deltaCorpus            := trace.Header.SessionContextGrowth != ""
throughputProvenanceOK := calibrateReplayMode == "fixed" && !deltaCorpus
```

- New `--replay-mode {fixed|closed-loop}` flag (default `fixed`) lets the operator affirm the `blis replay --session-mode` choice calibrate cannot itself observe — the "trustworthy replay-mode signal" resolution the adjudicator listed as acceptable.
- Under `--replay-mode closed-loop` **or** a delta-corpus header, the block is **omitted** and a loud `REFUSED` warning names the reason (R1). This closes the plain-trace-closed-loop hole: it no longer depends on a header marker.
- The durable auto-detection (stamp the replay mode into `SimResult` so the operator flag isn't needed) remains scoped to the open follow-up **#1652** — a genuinely narrower cross-command serialization change, not a blocker for this feature.

**G4 + F1 (single-error case swallowed the mismatch) → unconditional guard.** The status-mismatch scan now runs **before** the nil check, with a three-way suffix (emitted / refused-on-provenance / not-derivable), so a matched real-failed/sim-completed request surfaces even when *every* matched request failed and the block collapses to nil. Added `TestCalibrateCmd_Throughput_AllRealFailed_StillWarns` covering exactly the one-error scenario the adjudication said the prior test avoided.

Also removed the now-dead delta-corpus warning in the non-nil branch (the gate makes `report.Throughput` non-nil only for a fixed-mode plain trace) and suppressed the duplicate "not derivable" message after a refusal.

New tests: closed-loop refusal, delta-corpus refusal, all-real-failed-still-warns, invalid-`--replay-mode` rejected. Docs updated (`--replay-mode` row + rewritten validity-boundary section). `go test ./cmd/ ./sim/workload/` green; no new lint findings.

namasl:

## qa-review adjudication — PR #1651: ✅ PASS

|----|----|----|----|
| F1 | FLAW_FOUND | 🟡 WAIVED (deferred) | The real-failed/sim-completed masking is fixed: mismatch detection now runs independently of throughput derivability and warns even when every matched real request failed. The remaining closed-loop provenance limitation is genuinely covered by open follow-up #1652. The new operator flag mitigates it but does not independently establish provenance because it defaults to "fixed"; issue #1652 tracks stamping replay mode into the simulation artifact so calibrate can detect it reliably. |
| G3 | FLAW_FOUND | 🟡 WAIVED (deferred) | The current code adds an operator-supplied --replay-mode gate, but its default of "fixed" still cannot prove how the SimResults were produced. The author cited open issue #1652 for the durable provenance fix, and that issue directly covers stamping replay mode into the simulation artifact and having calibrate refuse or warn for plain-trace closed-loop replay. |
| G4 | FLAW_FOUND | ✅ RESOLVED | The mismatch guard now executes before and independently of the nil-throughput branch. Consequently, an all-error matched dataset still emits a real-failed/sim-completed warning even though no throughput block is derivable. The exact single-error scenario is covered by a passing test. |

### Still blocking

### Cleared
- **F1** (🟡 WAIVED (deferred)) — The real-failed/sim-completed masking is fixed: mismatch detection now runs independently of throughput derivability and warns even when every matched real request failed. The remaining closed-loop provenance limitation is genuinely covered by open follow-up #1652. The new operator flag mitigates it but does not independently establish provenance because it defaults to "fixed"; issue #1652 tracks stamping replay mode into the simulation artifact so calibrate can detect it reliably.
  _cmd/calibrate.go:324-358 performs the status-mismatch scan unconditionally and emits a warning even when report.Throughput is nil; cmd/calibrate_throughput_test.go:340-400 covers the single failed-real/matched-sim case. GitHub issue #1652 is OPEN and explicitly covers adding a trustworthy replay-mode signal to prevent silently invalid throughput for plain traces replayed closed-loop._
- **G3** (🟡 WAIVED (deferred)) — The current code adds an operator-supplied --replay-mode gate, but its default of "fixed" still cannot prove how the SimResults were produced. The author cited open issue #1652 for the durable provenance fix, and that issue directly covers stamping replay mode into the simulation artifact and having calibrate refuse or warn for plain-trace closed-loop replay.
  _cmd/calibrate.go:46-53 documents that calibrate cannot observe replay mode; cmd/calibrate.go:307-320 omits throughput when --replay-mode=closed-loop or a delta-corpus marker is present. GitHub issue #1652 is OPEN and its problem statement and acceptance criteria specifically cover the undetectable plain-trace closed-loop case and a trustworthy serialized replay-mode signal._
- **G4** (✅ RESOLVED) — The mismatch guard now executes before and independently of the nil-throughput branch. Consequently, an all-error matched dataset still emits a real-failed/sim-completed warning even though no throughput block is derivable. The exact single-error scenario is covered by a passing test.
  _cmd/calibrate.go:324-358 scans all matched non-ok real records and warns without requiring report.Throughput to be non-nil; cmd/calibrate.go:360-368 handles nil throughput only afterward. cmd/calibrate_throughput_test.go:340-400 verifies the one-failed-real-request scenario. `go test ./cmd/` and `go test ./sim/workload/` both pass._

namasl (approved):

Issue #1658, opened 2026-09-02
Bug: closed-loop accumulate sessions under-count each round's output by 1 token (missing prefill-token compensation in SessionManager.OnComplete)


namasl:
**Describe the bug**

`SessionManager.OnComplete` derives a closed-loop round's actual output length as

```go
// sim/workload/session.go:206
rawOutputLen := int(req.ProgressIndex) - int(req.InputLen())
```

but BLIS charges output token #1 to prefill completion and only the remaining tokens to decode steps, so a request that generated its **full** assigned output finishes at `ProgressIndex == InputLen + len(OutputTokens) - 1`. `recordRequestCompletion` compensates for exactly this when counting output tokens:

```go
// sim/simulator.go:733-736
decodeTokens := int(req.ProgressIndex) - int(req.InputLen())
if decodeTokens < len(req.OutputTokens) {
	decodeTokens++ // prefill-generated first token (vLLM parity)
}
```

`session.go` applies no such `+1`. Two consequences for **every** normally-completed round of a `context_growth: accumulate` session:

1. The accumulate buffer grows by `L-1` output tokens instead of `L` — a systematic 1-token-per-round shortfall in the reconstructed context (so round N's input is N tokens short).
2. The round takes the `actualOutputLen < len(outTokens)` branch and logs `... round N length-capped — accumulating L-1/L output tokens` at Debugf, even though nothing was length-capped. The label is misleading for anyone debugging a corpus.

`sim/workload/reexport.go`'s `accumulatedOutputLen` deliberately mirrors `OnComplete` (that mirroring is correct and load-bearing for the #1630 round-trip), so it inherits the same shortfall.

**Which invariant is violated?**

- [x] None / unknown — it is a fidelity/accounting shortfall, not an invariant break. Conservation (INV-1) is unaffected: `Metrics.TotalOutputTokens` uses the compensated count.

**Expected behavior**

A round's accumulated output should be the same `L` tokens the metrics count, i.e. `OnComplete` should apply the same prefill-token compensation as `recordRequestCompletion` (or, better, both should read one shared helper so they cannot drift). The genuine length-capped case (`actualOutputLen` short by more than the prefill token) must still truncate and log.

**Additional context**

[...]

Issue #1627, opened 2026-08-26
feat(sim): reserve spec-decode KV/token-budget by lookahead footprint (K+1), not accepted count — occupancy contention under saturation (follow-up to #1538)


namasl:
> Follow-up from PR #1538 (speculative decoding / MTP throughput, closes #1528). Non-blocking fidelity-boundary raised by @susiejojo in [PR review](https://github.com/inference-sim/inference-sim/pull/1538). Part of #1532 (GLM-5.2 support).

**What external behavior are we replicating?**

vLLM's **speculative KV / token-budget occupancy reservation**. PR #1538 models spec-decode *throughput* (fewer, wider decode steps) and per-step *latency* (verify width `w=K+1`), but it accounts **occupancy** by the *accepted* count `g=1+α·K`: it reserves `g` KV token-slots and debits `g` from the batched-token budget per decode step. Real vLLM reserves **`num_lookahead_tokens = K`** extra slots per running request per step **unconditionally** (before knowing acceptance), and a spec step consumes ~`K+1` of `max_num_batched_tokens`. So under KV saturation / high occupancy, real vLLM fits **fewer** concurrent requests and **preempts more**; BLIS sees the verify-width latency cost but not the batch-occupancy contention. Net: BLIS **over-predicts max batch size and under-predicts preemption** in memory-bound regimes — it shows the MTP speedup without the contention penalty that erodes it.

**Reference implementation (MANDATORY)**

Permalinks pinned to vLLM `903a02192fca19e4c89705af7017c0f24971ea4f` (main HEAD at filing) and verified to resolve.

`num_lookahead_tokens` set on the scheduler from config:
https://github.com/vllm-project/vllm/blob/903a02192fca19e4c89705af7017c0f24971ea4f/vllm/v1/core/sched/scheduler.py#L257-L258

Passed **unconditionally** into `allocate_slots` for every running-request KV allocation (independent of how many drafts are later accepted):
https://github.com/vllm-project/vllm/blob/903a02192fca19e4c89705af7017c0f24971ea4f/vllm/v1/core/sched/scheduler.py#L653-L658

(Definition: `num_lookahead_tokens` derives from `num_speculative_tokens` in `vllm/config/__init__.py` / `SpeculativeConfig`; the scheduler usage above is the load-bearing reservation site.)

**Key behaviors to preserve**
[...]

Issue #1656, opened 2026-09-02
EP-aware weight-footprint sizing: KV auto-calc replicates routed experts across TP, blocking large MoE+EP deployments


susiejojo:
> Discovered while attempting to replay/calibrate a real InferenceX Kimi-K3 H200 agentic run (id 438866: TP16 / DP2 / EP32, MXFP4). See "Relationship to existing work" — this is the KV-capacity half of #1548, which currently blocks even *sizing* a large EP-sharded MoE, independent of the step-time/comm work #1548 tracks.

**What problem does this solve?**

`CalculateKVBlocks` / `computeModelWeightBytes` (`sim/latency/kv_capacity.go`) always charge **every routed expert's weights to every GPU in the TP group** — i.e. weights are modeled as tensor-sharded (÷TP) but never **expert-sharded** across an expert-parallel (EP) group. For a large MoE served with EP, this makes the auto weight footprint far too large, so KV auto-calc fails outright and the model cannot be sized on its real topology.

Concrete case — Kimi-K3 (MXFP4) on H200, `--tp 16`:

```
--latency-model: KV capacity auto-calculation failed: CalculateKVBlocks:
model overhead (2569.58 GiB = 2551.98 weights + 8.00 activation + 9.60 non-torch)
exceeds available GPU memory (2030.40 GiB = 141.0 GiB x 90% util x 16 GPUs).
Minimum GPUs required per instance: 21
```

BLIS's footprint (2552 GiB) is itself **correct** — I verified against the HF safetensors metadata (`moonshotai/Kimi-K3`): the real checkpoint is 2.78T *packed* params (`U8: 2.72T` MXFP4 + `BF16: 57B`), which is 5.50T *elements*; BLIS computes 5.48T elements x 0.5 byte/param = 2552 GiB ≈ the real 2642 GiB disk footprint. **The param count and precision are right.** The problem is *placement*: the real run fits because **EP32 shards the 2536 GiB of routed experts across all 32 GPUs** (~86 GiB/GPU incl. experts, ~52 GiB/GPU left for KV). BLIS at `--tp 16` replicates all experts within the TP group → 2642 GiB / 16 → doesn't fit → fatal. So a faithful, real-world MoE+EP deployment that runs on the hardware **cannot be sized in BLIS at all**.

**Proposed solution**

[...]

PR #1659, opened 2026-09-03
fix(sim): clamp spec-decode progress at the completion boundary (Fixes #1657)

### LINKED ISSUE #1657: Bug: MTP spec-decode overshoots MaxOutputLen, cancelling closed-loop sessions (actualOutputLen > OutputTokens)
**Describe the bug**

Speculative decoding (MTP, #1528) can advance a request's `ProgressIndex` **past** its target output length, because decode steps advance by up to `1 + K` tokens (the verified draft block) rather than exactly 1. In a **closed-loop accumulate session**, `SessionManager.OnComplete` derives `actualOutputLen = ProgressIndex − InputLen`, and its over-cap guard treats `actualOutputLen > len(OutputTokens)` as accounting corruption and **cancels the entire session** — so an entire multi-round conversation is dropped after round 0.

With `--num-speculative-tokens 2`, round 0 of a session whose target output is 2180 tokens completes with `actualOutputLen = 2181` (a +1 overshoot), tripping the guard:

```
SessionManager.OnComplete: session <id> round 0 actualOutputLen=2181 > len(OutputTokens)=2180
  — cancelling session to contain corruption (ProgressIndex accounting drift)
```

The overshoot is inherent to speculative decoding (you verify a K-token draft block, so the final step can land 1..K tokens past the target). The completion path does not clamp `ProgressIndex` to the request's `MaxOutputLen` when a spec-decode step overshoots, and the session guard assumes exact landing.

**CLI command to reproduce**

```bash
# any closed-loop accumulate corpus + MTP triggers it; here a Kimi-K3 corpus:
./blis replay \
  --trace-header <accumulate-corpus>.yaml --trace-data <accumulate-corpus>.csv \
  --model kimi-k3 --hardware H200 --tp 32 \
  --session-mode closed-loop --max-model-len 1000000 \
  --num-speculative-tokens 2 --speculative-acceptance-rate 0.755
```

Observed on commit `9af90b71` (the #1528 merge). Behavior by config on the same corpus:
- `--num-speculative-tokens 2` → round 0 overshoots (2181 > 2180) → **session cancelled, 1 request completes**.
- `--num-speculative-tokens 1` → no overshoot observed → 10 requests complete.
- no spec flags → 7 requests complete (baseline).

**Expected behavior**

A spec-decode step must not advance a request beyond its `MaxOutputLen`: the final step should stop at the target (clamp `ProgressIndex` to `InputLen + MaxOutputLen`), so `actualOutputLen == len(OutputTokens)` at completion. Under-cap (length-capped) is already handled; over-cap by ≤K is a legitimate spec-decode outcome that should be clamped at generation, not flagged as corruption.

Relevant invariant: **Request lifecycle** — a request should complete at exactly its output budget; a completion overshoot then wrongly triggers **Session completeness** (INV-11) mishandling by cancelling a session that did nothing wrong.

**Actual behavior**

`ProgressIndex` overshoots the target by up to K tokens; `SessionManager.OnComplete` (`sim/workload/session.go`, over-cap branch ~L290) sees `actualOutputLen (2181) > len(OutputTokens) (2180)`, sets `sess.state = sessionCancelled`, and returns — dropping every remaining round of the session. The message is logged as "corruption / ProgressIndex accounting drift", but the drift here is a normal spec-decode overshoot, not corruption.

Root cause chain:
- `sim/workload/session.go` ~L206: `rawOutputLen := int(req.ProgressIndex) - int(req.InputLen())`.
- `sim/workload/session.go` ~L291: `case actualOutputLen > len(outTokens): ... sess.state = sessionCancelled`.
- The generation/step path (spec-decode, #1528) lets `ProgressIndex` exceed `InputLen + MaxOutputLen` on the overshooting final verified block.

- [x] Request lifecycle
- [x] Session completeness (INV-11) — a valid session is cancelled
- [ ] Determinism
- [ ] None / unknown

**Environment**
- OS: macOS 15 (darwin 24.6.0)
- Go version: as repo (1.22+)
- Commit hash: `9af90b71` (feat: model speculative decoding / MTP throughput #1528/#1538)

**Additional context**

Discovered while attempting a closed-loop replay of a real InferenceX Kimi-K3 H200 agentic run (id 438866), which used MTP (`num_speculative_tokens=2`, `synthetic_acceptance_length=2.51`). MTP (#1528) and closed-loop accumulate sessions (#1445) appear not to have been exercised together. This blocks any faithful replay of a spec-decode deployment against a multi-round session corpus. (`--tp 32` above is only to make the large MoE fit for the repro; the topology is orthogonal to the bug — the overshoot is in the decode/completion path, not placement.)

namasl:
## Summary

A speculative-decoding / MTP decode step advances `ProgressIndex` by `g = 1 + α·K` tokens, which could carry it **past** the request's completion boundary. `SessionManager.OnComplete` derives `actualOutputLen = ProgressIndex − InputLen` for closed-loop `accumulate` context growth, saw it exceed `len(OutputTokens)`, and **cancelled the entire multi-round session** as "ProgressIndex accounting drift" — dropping every remaining round after round 0.

`FormBatch` now caps each decode grant by the request's distance to its completion boundary, so the final verified block lands exactly **on** the target. This mirrors vLLM's `_update_from_output`, which appends the accepted tokens one at a time and trims the tail once `check_stop` fires: a request never records more output tokens than it was budgeted, while the forward pass still cost the full verify width.

**Evidence** (`examples/multiturn-chat-demo.yaml`, 100 accumulate sessions × 5 rounds, `--num-speculative-tokens 2 --speculative-acceptance-rate 0.755`, 2 instances):

| | completed rounds | `cancelling session to contain corruption` log lines |
|---|---|---|
| `9af90b71` (before) | 701 / 1000 | **121** |
| this PR | **1000 / 1000** | **0** |
| either build, `K=0` | 1000 / 1000 | 0 (stdout byte-identical between builds) |

## Behavioral contracts

- **BC-1 (no overshoot).** GIVEN spec-decode is enabled (`K > 0`), WHEN a decode step would advance a request past its completion boundary, THEN the step is granted only the tokens remaining to that boundary, so the final `ProgressIndex == InputLen + max(len(OutputTokens),1) − 1`.
- **BC-2 (`K=0` byte-identity, INV-6).** GIVEN spec-decode is off, WHEN a batch is formed, THEN grants are byte-identical to a pre-fix build — both caps sit strictly inside the `ctx.DecodeTokensPerStep != nil` gate. Verified by diffing `blis run` stdout against `9af90b71`.
- **BC-3 (accounting parity).** GIVEN the same workload run with and without spec-decode, THEN each request's final `ProgressIndex` and each accumulate round's `InputLen` sequence are **identical**. Spec-decode buys fewer, wider decode steps — never different token accounting.
- **BC-4 (session completeness, INV-11).** GIVEN a closed-loop accumulate session and any `K`/`α`, WHEN every round completes normally, THEN the session completes all its rounds — never `cancelled`.
- **BC-5 (PD liveness, INV-8/INV-11).** GIVEN a PD decode sub-request whose remaining budget is 0 (a 1-output-token sub-request, admitted already at its boundary), THEN the grant is floored at 1 — a 0-token grant would `break` the admission loop every step forever, since the sub-request is not yet in `RunningBatch` and nothing would force-complete it. This is exactly what the pre-feature (`K=0`) PD path did.
- **BC-6 (conservation, INV-1/INV-4).** Output-token and KV conservation hold for every `K`/`α`, including the previously-overshooting configs. (Bonus: with the clamp, `processCompletions`' final-token `AllocateKVBlocks` now fires under spec-decode exactly as at `K=0` — an overshoot used to skip it silently.)

## Implementation

- `sim/request.go` — new (package-private) `Request.completionProgressIndex()`: the single source of truth for the boundary `InputLen + max(len(OutputTokens),1) − 1` (R23). BLIS charges output token #1 to prefill completion and the rest to decode steps, hence the `−1`.
- `sim/simulator.go` — `processCompletions` now tests against that accessor (same expression, byte-identical), so the completion test and the new cap **cannot** disagree.
- `sim/batch_formation.go` — `decodeTokensToCompletion(req)` applied in Phase 1's decode branch (plain `min`) and Phase 2's PD decode-sub-request branch (floored at 1, per BC-5).
- Reading `OutputTokens` here is per-step execution planning, which **INV-9's Scope paragraph permits for `FormBatch`** (the decode-phase gate one line above already reads `len(req.OutputTokens) > 0`) — not a servability decision. The Phase-2 call site sits in the *admission* loop, so rather than lean on that sentence this PR adds an **explicit bounded carve-out** to INV-9: the cap is monotone-shrinking and floored at 1, so it can never reject a request, reorder admissions, or change *whether* a request is admitted — it only makes admission under KV pressure marginally **more** likely. A future oracle read there that could gate admission is a violation.

**Timing footprint.** The GPU step cost is unchanged: both latency backends size a decode step by verify width `w = K+1`, not by the granted count, so a clamped final step still pays the full verify pass (as in vLLM). The one effect is per-token detokenization (`NumNewTokens × OutputTokenProcessingTime`), now charged for the tokens actually emitted — total OTPT is exactly `(L−1)×OTPT`, identical to `K=0`, where before it was over-charged by up to `K×OTPT`. So spec-decode E2E/ITL shift slightly downward; `K=0` output is untouched.

## Testing

- `sim/latency/trained_physics_model_test.go` — `TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens` pins the #1528 cost/progress split that makes this clamp safe: a decode step costs the **verify width** `K+1`, never the granted count, so a boundary-clamped final step is **not** cheaper. Nothing guarded that before. Verified by mutation on both backends (sizing roofline's decode width by `req.NumNewTokens` fails it; collapsing trained-physics' `verifyWidth()` to 1 fails its non-vacuity leg).
- `sim/simulator_decode_test.go` — `TestCompletionProgressIndex_BoundaryLaw` pins the `−1` / `max(…,1)` semantics that the completion test and the clamp now share. `TestSpecDecode_PDDecodeSubRequest_AtBoundary_GrantsOneToken` asserts the floor-at-1 grant directly: verified by mutation that replacing the floor with a plain `min()` previously surfaced only as a package-level **timeout** (the sub-request is granted 0, `break`s, and is never force-completed) and now fails in 6 ms with a one-line diagnostic.
- `sim/cluster/spec_decode_session_test.go` (new) — differential closed-loop accumulate tests: 13 output lengths × 4 spec configs, each compared against its own `K=0` baseline for round count, per-round terminal `ProgressIndex`, per-round `InputLen`, and completed-request count; plus a **PD** twin (2P+2D) that exercises the structurally different Phase-2 branch, with `outLen=1` pinning the floor-at-1 liveness case. Both fail on `9af90b71` with the reported symptom (`actualOutputLen=5 > len(OutputTokens)=2 — cancelling session`; `PD rounds = 1, want 3`). Runs in ~0.1s.
- `sim/simulator_decode_test.go` — `TestSpecDecode_OutputTokenConservation` previously *permitted* overshoot (its inner assertion was unreachable); it now asserts the exact boundary and adds the two reported repro rows. New `TestSpecDecode_FinalProgressIndexMatchesBaseline` states BC-3 directly as a differential law over 9 output lengths × 5 spec configs, with a "fewer steps than baseline" leg so it cannot pass by spec-decode silently doing nothing.
- No golden values: every expectation is either re-derived from the boundary law or taken from a `K=0` run computed in the same test.

Verification gate: `go build ./...` clean, `go test ./... -count=1` all pass, `go vet ./...` clean, `gofmt` clean on all touched files (`golangci-lint` is not installed in this container). INV-6 re-verified by diffing `blis run` stdout against `9af90b71` (no-spec: byte-identical; determinism: 3 identical runs with spec-decode on). INV-13 re-verified end-to-end: `blis run --trace-output` → `blis replay` under `K=3, α=0.7` reproduces the same completions and output tokens.

## Docs

- `CLAUDE.md` and `docs/guide/latency-models.md` — the `#1528` sections now state the clamp, the vLLM parity, the accounting-parity consequence, and the OTPT timing footprint.
- `docs/contributing/standards/invariants.md` — INV-11 gains this bug as evidence; INV-9's grep-based verification now also names `CompletionProgressIndex()` as an oracle-derived accessor that must stay out of the servability files (a plain `OutputTokens` grep would not catch it).
- `docs/reference/models.md` — (unrelated to the fix, called out here deliberately) dropped the stale "MTP throughput is not modeled … tracked in #1528" bullet (#1528 is merged and this PR refines it); it now points at the guide and at the remaining #1627 occupancy gap.

## Review rounds

- **archon (package altitude):** 0 violations, no package edge changed. Its only flagged item was the newly *exported* accessor, so it is now package-private — both callers are inside `sim`.
- **Substance / DES / vLLM-fidelity and standards / tests / docs review:** no CRITICAL findings. Verified independently that reverting only `sim/batch_formation.go` makes both new cluster tests fail with the exact issue signature; a ~37k-simulation sweep (KV / token-budget / chunked-prefill / MaxModelLen configs × K × α × input × output × concurrency) with `decodeTokensToCompletion(req)==0` instrumented as a panic never fired in Phase 1, no request was ever stranded, and every non-length-capped completion landed exactly on the K=0 boundary; the fractional carry stays bounded and drift-free (`granted ≤ peek` ⇒ `carry ≥ 0`, and a capped step is the completing step, after which the carry is reset). The review also surfaced a **silent** pre-fix corruption mode this closes: PD `outLen=4, K=5` did *not* cancel the session but its accumulate buffer had already drifted (`InputLen 44` vs `43`). `length_capped_requests` is unchanged, and the length-capped TPOT denominator is untouched (when the output cap binds, `processCompletions` takes the *normal* completion branch).

## Discovered issues

- **#1665** (filed, not fixed here) — a **zero-output-token request through PD** panics in `AllocateKVBlocks` (`index out of range [0] with length 0`). Reproduces on `9af90b71` with spec-decode entirely off, and this PR's floor-at-1 leaves the faulty allocation range exactly as the pre-feature path had it.
- **#1658** (filed, not fixed here) — `SessionManager.OnComplete` omits the prefill-generated-first-token compensation that `recordRequestCompletion` applies, so an accumulate round grows the context by `L−1` instead of `L` and logs every normal round as "length-capped" at Debugf. This holds at `K=0` too, so this PR's "identical to a `K=0` run" bar is met either way; fixing it changes accumulate context length for every existing closed-loop run and must move in lockstep with `sim/workload/reexport.go`, so it needs its own PR.

Fixes #1657


namasl:
@claude /blis-pr-review

Archon PR review for this head (`e6b87c8e`) returned **`NO_CHANGE`** — "No architectural change. Internal-only — fast-track eligible": 0 violations, 0 package edges added/removed, 0 public-surface changes (the boundary accessor `completionProgressIndex` is package-private), and 2 guarded-promise (invariant/test) additions within the existing boundary — `TestSpecDecode_FinalProgressIndexMatchesBaseline` + the modified `TestSpecDecode_OutputTokenConservation` in `sim`, and the two new `TestSpecDecode_*AccumulateSession_AccountingMatchesBaseline` tests in `sim/cluster`.

Please focus on what archon cannot see:

1. **The clamp target.** `decodeTokensToCompletion` caps a spec-decode grant by the distance to `Request.completionProgressIndex()` = `InputLen + max(len(OutputTokens),1) − 1`, now the single source of truth shared with `processCompletions` (R23). Is that the right boundary, and is the `−1` / `max(…,1)` reasoning sound in both directions (zero-output, 1-output, PD decode sub-requests)?
2. **Liveness.** Phase 1 uses a plain `min` (a running decode request is always ≥1 token short of its boundary, since `processCompletions` finishes any batch member that reaches it in the same step); Phase 2 floors at 1 because a PD decode sub-request can be admitted already AT its boundary and is not yet in `RunningBatch`, so a 0-token grant would `break` the admission loop forever (INV-8/INV-11). Is the asymmetry justified and the argument airtight?
3. **INV-9.** The Phase-2 read of `len(OutputTokens)` sits in the *admission* loop, so this PR adds an explicit bounded carve-out to INV-9 rather than leaning on the Scope paragraph: the cap is monotone-shrinking and floored at 1, so it can never reject or reorder an admission. Is the carve-out worded tightly enough?
4. **Cross-path parity.** INV-6 (K=0 stdout byte-identical to `9af90b71`, verified), INV-13 (run→replay under identical spec flags, verified), and the deliberate spec-decode metric shift: step time and step count are unchanged, but per-token detokenization (`NumNewTokens × OTPT`) now bills only the tokens actually emitted, so spec-decode ITL/E2E move slightly downward.
5. **Timeout / preemption consistency** and whether the length-capped path (MaxModelLen boundary, `LengthCappedRequests`, the spec-decode TPOT denominator) is genuinely untouched when the output cap binds.

Two pre-existing issues were filed rather than fixed here: #1665 (PD + zero-output request panics in `AllocateKVBlocks`; reproduces on `9af90b71` with the feature off) and #1658 (accumulate rounds under-count each round's output by one token; holds at K=0 too, so this PR's "identical to a K=0 run" bar is met either way).

namasl:
@claude /blis-pr-review

Archon on this head (`8fc1f52d`) again returned **`NO_CHANGE`** — "No architectural change. Internal-only — fast-track eligible": 0 violations, 0 package edges, 0 public-surface changes, guarded-promise (test) additions only.

The only delta since your READY-TO-MERGE review of `e6b87c8e` is **one new test file change**, no production code: `sim/latency/trained_physics_model_test.go` gains `TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens`. It pins the #1528 cost/progress split that your review confirmed at point 4 — that step timing is unaffected because both backends scale decode token-population terms by `w = K+1`, independent of the granted count. Nothing was actually *guarding* that, so a future refactor sizing decode cost by `req.NumNewTokens` would make a boundary-clamped final step silently cheaper and bill speculative decoding for progress it never paid for. The test asserts `StepTime` is identical for a granted count of 1 (the clamp) and `K+1` (fully accepted) on **both** backends, plus a non-vacuity leg (`K>0` must still cost more than `K=0` at the same granted count). Verified by mutation: sizing roofline's decode width by `req.NumNewTokens` fails the equality leg, and collapsing trained-physics' `verifyWidth()` to 1 fails the non-vacuity leg.

I also checked your suggestion to source the cap from `MaxOutputLen` (the client budget) instead of `len(OutputTokens)`, since that would sidestep the INV-9 question entirely. It is **not viable**, and the failure is loud: with a `MaxOutputLen`-derived boundary, any request whose `MaxOutputLen` is 0 — hand-built requests, and any client that declares no budget while `--max-model-len` is unset, so the `EnqueueRequest` auto-fill never fires — gets boundary `InputLen + max(0,1) − 1 = InputLen` and completes at prefill end, generating **1** output token instead of 9 (`TotalOutputTokens=1, want 9`) **even at K=0**. That is an INV-1 conservation break in the feature-off path. The completion boundary `processCompletions` tests is defined by `len(OutputTokens)`, so the cap must read the same quantity or the two definitions diverge — which is exactly why this PR made `completionProgressIndex()` the single source of truth (R23) rather than adding a second, non-equal boundary.

Nothing else changed: the three non-blocking style observations from your review (the verbose `if outRoom >= 1` floor form, the DEFENSE-IN-DEPTH relabel, the `output > 3` gating note) were left as-is per your own assessment, and #1665 / #1658 are both recorded under "Discovered issues" in the PR body.

namasl:

## qa-review — PR #1659: ✅ PASS
_fix(sim): clamp spec-decode progress at the completion boundary (Fixes #1657) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR fixes the reported overshoot at the execution-planning point for both ordinary and PD decode paths, preserves the session corruption guard, and adds focused unit and closed-loop regression coverage. The documentation changes are directly related to the corrected accounting semantics; the unrelated-looking whitespace-only test edits are harmless. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. The user-facing latency guide, model reference, canonical invariants, and project summary all describe the completion clamp, its timing/accounting effects, run/replay applicability, and the remaining speculative-occupancy limitation. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No materially misleading comment remains. One comment is slightly imprecise: `processCompletions` says zero output is 'set ... to 1 manually', while that normalization now occurs inside `completionProgressIndex()`. The described behavior remains correct. |
| G1 | correctness | ✅ CONFIDENT | In `FormBatch` Phase-1 decode step planning (`sim/batch_formation.go`), `decodeTokens = min(decodeTokens, decodeTokensToCompletion(req))` caps the grant without flooring at 1. If `ProgressIndex` reaches `completionProgressIndex()`, `decodeTokensToCompletion` returns 0, making `decodeTokens = 0`. What happens if a running batch request gets a 0-token grant—does `executeBatchStep` execute a 0-token step, and is `processCompletions` guaranteed to remove the request in that same step without executing an idle forward pass? | A running request with a zero grant remains in `RunningBatch`, but it is excluded from the `scheduled` slice, so it does not participate in a model forward pass. `StepTime` is still invoked, possibly on an empty slice, and the simulator advances by at least one tick. Immediately afterward, in the same `Step` call, `processCompletions` sees `ProgressIndex >= completionProgressIndex()`, completes the request, releases its KV blocks, and removes it from the returned remaining batch. Thus there is no repeated zero-token loop or charged forward pass for that request, although an otherwise-empty step incurs the simulator's minimum one-tick advance. |
| G2 | correctness | ✅ CONFIDENT | In Phase-2 (`FormBatch` PD decode sub-requests), when `outRoom := decodeTokensToCompletion(next)` is less than 1 (i.e., 0), the code explicitly overrides `decodeTokens = 1`. Why does Phase-2 floor the grant at 1 while Phase-1 allows `decodeTokens` to become 0, and could Phase-2 granting 1 token to an already-completed sub-request cause `ProgressIndex` to overshoot by 1 token on PD deployments? | Phase 1 can safely leave a zero-grant request in the already-running batch because completion processing runs immediately after execution. Phase 2 cannot enqueue a zero-token request: the admission loop would break before moving it from `WaitQ`, so it could remain stranded indefinitely. A realistic `outRoom == 0` occurs for a one-output-token PD sub-request, whose transferred state starts at `ProgressIndex == InputLen`, also its completion boundary. Phase 2 intentionally grants the same one token as the pre-feature PD path. That moves its `ProgressIndex` one past `completionProgressIndex()`, but `ProgressIndex-InputLen == 1`, exactly its one output token, and completion then releases it. This is an intentional PD accounting exception, not the multi-token overshoot that cancelled sessions. |
| G3 | invariants | ✅ CONFIDENT | INV-9 prohibits using oracle knowledge (`OutputTokens`) in scheduling servability decisions. The diff adds a bounded carve-out in `invariants.md` documenting that Phase-2 admission reads `completionProgressIndex()` (derived from `len(OutputTokens)`). Because capping `decodeTokens` reduces the KV cache reservation and batched-token debit for that step, could this oracle-derived reduction allow a subsequent waiting request to be admitted that would otherwise have been blocked by KV limits? | Yes. A smaller final-step grant consumes fewer accepted-token KV slots and less batched-token budget, so a later waiting request can be admitted in a step where it otherwise would not fit. The revised INV-9 explicitly acknowledges that consequence and treats the oracle read as execution-side step sizing, not a servability decision about the current request. The cap cannot reject or reorder the capped request itself and is floored at one in Phase 2. |
| G4 | invariants | ✅ CONFIDENT | Under INV-6 (determinism), speculative decoding uses `peekDecodeTokens` to accumulate fractional carry (`specDecodeCarry`). When `FormBatch` caps `decodeTokens` via `decodeTokensToCompletion`, `peekDecodeTokens` does not mutate the carry, but `executeBatchStep` later commits `req.NumNewTokens`. How does this final-step clamping affect the leftover `specDecodeCarry` on the request object upon completion, and could any uncommitted carry bleed into downstream metrics or session state? | The committed carry increases by `specTokensPerStep - NumNewTokens`, so a boundary-clamped final step can leave positive deferred carry. That carry has no effect after completion: metrics use `ProgressIndex`, `ITL`, and output lengths rather than the carry, and `recordRequestCompletion` resets `specDecodeCarry` to zero before the session completion callback runs. It therefore cannot bleed into a follow-up round or downstream metrics. |
| G5 | rules | ✅ CONFIDENT | Rule R23 requires single sources of truth. `completionProgressIndex()` in `sim/request.go` is introduced as the single authority for the completion boundary. However, in `sim/simulator.go` `recordRequestCompletion`, `decodeTokens` is still manually checked and clamped against `len(req.OutputTokens)`. Why wasn't `recordRequestCompletion` updated to use `completionProgressIndex()` or its exact delta math to avoid duplicate boundary definitions? | `completionProgressIndex()` is the authority for when execution is complete and how far a speculative grant may advance. The clamp in `recordRequestCompletion` serves a different invariant: converting potentially malformed or legacy `ProgressIndex` state into a conserved output-token metric. Its calculation also handles the PD one-token exception, where final `ProgressIndex` is one beyond the normal completion boundary. Replacing it mechanically with boundary-delta math would risk changing that established metric behavior. |
| G6 | rules | ✅ CONFIDENT | Rule R1 forbids silent degradation or unlogged fallback logic. In `sim/batch_formation.go` Phase-2, when `outRoom < 1`, the code silently executes `decodeTokens = 1` in the `else` branch without logging or metric counter increments. Under what exact conditions is `outRoom < 1` expected for a queued PD decode sub-request, and should this edge case emit a debug log or metric? | For producer-generated state, `outRoom < 1` is expected for a one-output-token PD decode sub-request: KV transfer initializes it at `ProgressIndex == InputLen`, while its completion boundary is also `InputLen`. The one-token grant preserves the pre-feature PD behavior and is required to dequeue it; this is not an error or fallback that drops data, so R1 does not require logging or a counter. A state already past the boundary would also enter the branch, but that would require invalid external mutation or corrupt state. |
| G7 | parity | ✅ CONFIDENT | Does clamping `decodeTokens` at batch formation affect the calculation of TPOT (`RequestITLs`) versus raw verification ITLs (`AllITLs`) across `blis run` and `blis replay`? Are trace round-trips under INV-13 guaranteed to produce byte-identical token timestamps when running multi-turn closed-loop sessions? | The clamp changes the final raw ITL entry only through per-token output processing: `executeBatchStep` still charges the full verification-step GPU time, then adds OTPT multiplied by the actually granted token count. `AllITLs` stores those per-verification-step entries. `RequestITLs` divides their sum by a token-derived denominator for speculative length-capped requests and by `len(OutputTokens)-1` for normal completions, so it remains amortized TPOT rather than per-step latency. Run and replay share the same simulator configuration and execution path, so the clamp itself introduces no parity divergence. INV-13 guarantees TTFT, E2E, and aggregate metric parity for exported traces replayed in fixed mode; it does not define a separate serialized per-token timestamp guarantee. |
| G8 | parity | ✅ CONFIDENT | When speculative decoding is disabled (`K=0`), `ctx.DecodeTokensPerStep` is `nil`, bypassing `decodeTokensToCompletion`. If a request has `MaxOutputLen` or `len(OutputTokens) == 0`, does a `K=0` run execute the exact same completion boundary path as a `K>0` run with `--speculative-acceptance-rate 0`? | They use the same completion test, because `processCompletions` always calls `completionProgressIndex()`, including for zero-output requests. They do not use the exact same batch-planning path: K=0 leaves `DecodeTokensPerStep` nil and preserves the legacy one-token behavior, while K>0 with acceptance zero enables speculative mode, proposes one accepted token, applies the new output and model-length caps, and still pays verify-width cost K+1. Zero-output requests never enter the decode gate in either case. `MaxOutputLen` is a client/admission budget; runtime completion uses the assigned `OutputTokens` length. |
| G9 | tests | ✅ CONFIDENT | New unit test `TestSpecDecode_PDDecodeSubRequest_AtBoundary_GrantsOneToken` verifies that a sub-request at the boundary gets `NumNewTokens == 1` during `FormBatch`. Is there a companion test verifying that after this 1-token grant is executed in the simulator, the PD sub-request is successfully completed and evicted from `RunningBatch` without stranding KV cache blocks? | There is no narrowly named companion unit test that executes exactly that fixture and directly asserts final KV usage. However, the new PD closed-loop integration test runs `outLen=1` sessions through the complete cluster simulation, requires all rounds and completion counts to match K=0, and would expose a stranded sub-request. The normal completion path unconditionally calls `ReleaseKVBlocks`. Core realistic behavior is therefore covered, though the resource-release assertion is indirect. |
| G10 | tests | ✅ CONFIDENT | In `sim/latency/trained_physics_model_test.go`, `TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens` checks cost insensitivity for `trained-physics` and `roofline` backends. Is there an invariant assertion or automated sanity check that will fail if a future third latency backend calculates decode FLOPs from `req.NumNewTokens` instead of verify width `w`? | No backend-independent runtime assertion enforces this. The test enumerates the two currently supported implementations, and the latency factory rejects every backend other than roofline and trained-physics. A future third backend would require extending the factory and should also be added to this test, but current production backends are fully covered. |
| G11 | rationale | ✅ CONFIDENT | The PR fixes the session cancellation bug by capping `decodeTokens` at generation time in `FormBatch`. What alternatives were considered—such as relaxing `SessionManager.OnComplete` in `sim/workload/session.go` to trim overshot tokens—and why was generation-time scheduler clamping chosen over completion-time trimming? | The repository does not record a complete alternatives discussion, but the implemented rationale is clear. Trimming in `SessionManager.OnComplete` would only hide the bad `ProgressIndex` for sessions, leave non-session request state overshot, distort detokenization/token accounting, and turn the session's over-cap corruption detector into an ambiguity. Generation-time clamping instead prevents invalid state at its source, applies equally to session and non-session requests, matches the existing completion predicate, and retains `OnComplete`'s over-cap branch for genuine corruption. The comments and guide also relate this to vLLM trimming accepted tokens once its stop condition fires. |
| G12 | rationale | ✅ CONFIDENT | In `sim/request.go`, `completionProgressIndex()` uses `req.InputLen() + max(int64(len(req.OutputTokens)), 1) - 1`. Why is the boundary defined with a `-1` offset relative to input plus output tokens, and how does this account for prefill charging token #1 versus decode steps charging subsequent tokens? | BLIS advances `ProgressIndex` through input prefill to `InputLen`; reaching that point also produces the first output token and records TTFT. Consequently only output tokens 2 through L require decode advances, totaling L-1 increments, so the terminal index is `InputLen + L - 1`. For zero output, `max(L,1)` keeps the boundary at `InputLen`, allowing completion immediately after prefill without a decode step. |
| G13 | scope | ✅ CONFIDENT | The diff refactors `processCompletions` in `sim/simulator.go` to use `req.completionProgressIndex()` instead of the inline `req.InputLen()+max(util.Len64(req.OutputTokens), 1)-1`. Does this refactoring introduce any subtle behavioral changes for zero-output-token requests or prefill-only requests outside of speculative decoding? | No. The helper is algebraically identical to the replaced expression: `InputLen() + max(int64(len(OutputTokens)),1) - 1`. Therefore zero-output and one-output requests still complete at `InputLen`, including prefill-only zero-output requests. The refactor centralizes the formula without changing its inputs, comparison operator, or timing. |
| G14 | scope | ✅ CONFIDENT | The documentation additions in `docs/guide/latency-models.md` and `invariants.md` note that boundary clamping sharpens the speculative occupancy contention gap (#1627) on the final step. Does this PR alter any KV reservation logic for speculative lookahead tokens, or is KV occupancy tracking strictly unchanged? | The PR does not add or modify a distinct speculative-lookahead reservation mechanism; BLIS still reserves KV and token budget using the accepted/granted count rather than unconditional K lookahead slots. It does, however, reduce the existing accepted-count allocation on a clamped final step because `preemptForTokens`/`AllocateKVBlocks` and token-budget debit receive the smaller grant. Thus the occupancy model is structurally unchanged, while its final-step reservation amount is intentionally reduced and the already-documented #1627 approximation becomes slightly more optimistic. |
| G15 | errors | ✅ CONFIDENT | If an invalid simulator state or external mutation causes `req.ProgressIndex` to exceed `req.completionProgressIndex()` before `FormBatch` runs, `decodeTokensToCompletion(req)` returns 0. Does the simulator detect or log this accounting anomaly, or could it lead to silent zero-token step loops? | It cannot cause a persistent zero-token loop for a running request. The cap returns zero, execution excludes that request from the scheduled forward-pass slice, and `processCompletions` removes it in the same simulator step because the completion comparison uses `>=`. A session request whose excess also makes `actualOutputLen > len(OutputTokens)` is then logged and its session cancelled by `SessionManager`; a non-session request has no dedicated anomaly log and its output metric is defensively clamped. |
| G16 | errors | ✅ CONFIDENT | In `sim/cluster/spec_decode_session_test.go`, `runAccumulateSession` uses `mustRun(t, cs)` to drive the cluster simulator. If an invalid spec-decode configuration or boundary error causes an infinite loop or deadlock in session completion, does the test harness fail fast with a clear error message or hit a test execution timeout? | `mustRun` only fails immediately when `ClusterSimulator.Run()` returns an error; it does not install a watchdog or context deadline. If a regression caused the event loop to deadlock or continually schedule work, this integration test would run until the enclosing `go test` timeout rather than fail fast with a boundary-specific message. The direct Phase-2 unit test reduces that risk for the known zero-grant regression, and the current suite completes successfully. |

### Items to fix

### Important to consider
- **F1** — The implementation deliberately targets the existing completion boundary `InputLen + max(outputLen,1) - 1`, rather than changing BLIS's established first-token-at-prefill accounting model.
  _sim/batch_formation.go:174-212; sim/batch_formation.go:262-326; sim/request.go:158-176; sim/cluster/spec_decode_session_test.go:111-219; sim/simulator_decode_test.go:107-170; `go test ./sim/... -count=1` passes_
- **F3** — Consider rewording `sim/simulator.go:1136` to say the helper treats zero output as one for boundary calculation.
  _sim/simulator.go:1133-1137; sim/request.go:158-175_
- **G3** — This makes final-step occupancy slightly more optimistic relative to vLLM's lookahead reservation, as documented under issue #1627.
  _sim/batch_formation.go:275-325; docs/contributing/standards/invariants.md:122-132_
- **G5** — The metric conversion remains separate arithmetic, but it is defense-in-depth rather than a competing definition of the scheduling/completion boundary.
  _sim/request.go:158-175; sim/simulator.go:726-748; sim/simulator.go:1133-1166_
- **G6** — A debug assertion for the impossible `ProgressIndex > boundary` case could improve diagnostics, but is not required for the valid one-token PD case.
  _sim/cluster/pd_events.go:127-144; sim/batch_formation.go:262-326; sim/request.go:158-175; docs/contributing/standards/rules.md:24-32; sim/simulator_decode_test.go:292-341_
- **G7** — There is no targeted trace-export/replay test combining speculative decoding with multi-turn sessions; this is an edge-combination coverage opportunity.
  _sim/simulator.go:1090-1101; sim/simulator.go:750-805; docs/contributing/standards/invariants.md:178-186; docs/guide/latency-models.md:228-239_
- **G9** — A direct `UsedBlocks()==0` assertion after the PD outLen=1 run would make KV-release coverage explicit.
  _sim/cluster/spec_decode_session_test.go:181-219; sim/simulator_decode_test.go:292-341; sim/simulator.go:1137-1166; `go test ./sim/... -count=1` passes_
- **G10** — If backends become dynamically registered, converting this check into a shared latency-backend contract test would reduce future omission risk.
  _sim/latency/trained_physics_model_test.go:589-633; sim/latency/latency.go:161-210; sim/latency_model.go:3-25_
- **G15** — A generic diagnostic for already-overshot non-session requests could improve observability, but the state is safely completed rather than livelocked.
  _sim/batch_formation.go:85-107,184-212; sim/simulator.go:1042-1068; sim/simulator.go:1133-1175; sim/simulator.go:737-748; sim/workload/session.go:280-294_
- **G16** — A bounded simulation-step guard in the test helper would provide faster and clearer diagnosis of future liveness regressions.
  _sim/cluster/cluster_test.go:37-43; sim/cluster/cluster.go:789-812; sim/simulator_decode_test.go:292-341; `go test ./sim/... -count=1` passes, with `sim/cluster` completing in 43.857s_

namasl:
@claude /blis-pr-review

Closing the review cycle on this head (`8fc1f52d`, unchanged since your READY-TO-MERGE re-review — no new commits). Posting so the `Claude Code Action (comment trigger)` status resolves on this commit: the `/qa-review` comment above triggered a skipped run that left it pending, which blocks merge.

**Gate status on `8fc1f52d`:**
- **archon** — `NO_CHANGE`, internal-only, fast-track eligible (0 violations / 0 package edges / 0 public-surface changes).
- **blis-pr-review** — READY TO MERGE, no blocking issues (twice: at `e6b87c8e` and incrementally at `8fc1f52d`).
- **qa-review** (cross-vendor Gemini questioner + OpenAI Sol answerer, 19 questions) — **PASS**, 19/19 CONFIDENT, **Items to fix: none**. It independently confirmed the Phase-1 zero-grant case cannot loop (a zero-grant request is excluded from the `scheduled` forward-pass slice and completed by `processCompletions` in the same step), that the Phase-2 floor-at-1 PD exception yields `ProgressIndex − InputLen == 1` for a one-output sub-request (its one output token, not a multi-token overshoot), that the leftover `specDecodeCarry` on a clamped final step cannot bleed anywhere (`recordRequestCompletion` zeroes it before the session callback), and that the `processCompletions` refactor is algebraically identical to the expression it replaced.
- **CI** — all shards green (build, lint, 7 test shards).

Its 10 non-blocking "important to consider" notes are recorded in that comment; none is a defect in this change. Two are cheap follow-ups I deliberately did **not** fold in, to avoid re-opening a review cycle on a green head: rewording the `processCompletions` "set it to 1 manually" comment (the normalization now lives in `completionProgressIndex()` — the described behavior is still correct), and adding an explicit `UsedBlocks()==0` assertion after the PD `outLen=1` run (KV release is already covered indirectly by the closed-loop PD test plus the unconditional `ReleaseKVBlocks` on the completion path). Happy to fold either in if you'd prefer them in this PR.

This PR needs a maintainer's approving review to merge (branch protection: `REVIEW_REQUIRED`); I cannot self-approve.

namasl:
Rebased onto `main` at `017fcd98` (#1660 / issue #1656, EP-aware weight-footprint sizing) — the PR was `CONFLICTING`/`DIRTY` and is now `MERGEABLE` again. New head `fefc04b7` (was `8fc1f52d`); the four commits are unchanged in content.

**Conflicts and how they were resolved** (both were pure insertion-adjacency, not content merges — kept BOTH sides):
- `CLAUDE.md` — #1660 inserted its `**EP-aware weight-footprint sizing (#1656)**` paragraph immediately above the `**Speculative decoding / MTP (#1528)**` paragraph that this PR edits. Kept #1660's paragraph verbatim (including its #1666 disclosure that BLIS's EP-off `DP>1` capacity is conservative rather than vLLM-faithful) alongside this PR's updated #1528 paragraph. `git diff main -- CLAUDE.md` is now exactly one changed line: the #1528 paragraph.
- `docs/reference/models.md` — #1660 had independently de-staled the same "MTP throughput is not modeled" bullet that this PR was rewriting, and added two new bullets (EP-sizing-is-an-average, MTP-module-weights-unmodeled). Kept #1660's two new bullets and its de-staled sentence, and folded in only what is genuinely from this PR: the section anchor link (`#speculative-decoding-mtp-1528` — single hyphen, since mkdocs' slugify collapses the run), the #1657 clamp clause, and the residual #1627 occupancy gap. No duplicated "IS modeled as of" sentence.

No Go file conflicted — this PR's production changes are in `sim/request.go`, `sim/batch_formation.go`, `sim/simulator.go`, none of which #1660 touched.

**Re-verified on the new base (a rebase voids prior verdicts, so nothing here is carried over by assumption):**
- `go build ./...` clean; `go vet ./...` clean; `go test ./... -count=1` all packages pass. `gofmt` clean on every file this PR touches (`sim/request.go` and `sim/simulator.go` show pre-existing alignment noise that reproduces on `main` at `017fcd98` itself — confirmed by running `gofmt -l` on `git show upstream/main:` copies).
- #1660's own guard `TestEveryKVCapacityCallSiteIsEPAware` **PASSES**, so no `CalculateKVBlocks` call site lost its `WithExpertParallelSize` option in the rebase.
- **Headline evidence reproduces exactly** (100 accumulate sessions × 5 rounds, `K=2, α=0.755`, 2 instances): `017fcd98` → **701/1000** rounds with **121** `cancelling session to contain corruption` log lines; this branch → **1000/1000** rounds, **0** log lines.
- **INV-6**: `K=0` stdout byte-identical to `017fcd98` on both a 200-request run and the 100-session multi-turn spec — so #1660's KV-capacity changes and this clamp do not interact.
- **INV-13**: `run --trace-output` → `replay` under `K=3, α=0.7` reproduces the same completions and total output tokens (50 / 28839 both legs).
- **Determinism**: three identical runs with spec-decode on (same md5).

/archon-pr-review

namasl:
@claude /blis-pr-review

Re-review requested because the branch was **rebased**, which voids your prior verdicts. New head `fefc04b7`, new base `017fcd98` (#1660 / issue #1656, EP-aware weight-footprint sizing). Archon on this exact pair returned **`NO_CHANGE`** — "No architectural change. Internal-only — fast-track eligible." All CI shards green (build, lint, 7 test shards).

**The four commits are unchanged in content.** The only rebase edits were two documentation conflicts, both pure insertion-adjacency where BOTH sides belong:
1. `CLAUDE.md` — #1660 inserted its `**EP-aware weight-footprint sizing (#1656)**` paragraph directly above the `**Speculative decoding / MTP (#1528)**` paragraph this PR edits. Kept #1660's verbatim (including its #1666 disclosure) plus this PR's updated #1528 paragraph; `git diff main -- CLAUDE.md` is now exactly one changed line.
2. `docs/reference/models.md` — #1660 had independently de-staled the very bullet this PR was rewriting, and added two new ones. Kept #1660's two new bullets and its de-staled sentence, folding in only this PR's own contributions: the section anchor (single-hyphen `#speculative-decoding-mtp-1528`, since mkdocs' slugify collapses the run), the #1657 clamp clause, and the residual #1627 occupancy gap — no duplicated sentence.

No Go file conflicted; this PR's production changes (`sim/request.go`, `sim/batch_formation.go`, `sim/simulator.go`) are disjoint from #1660's.

**What I re-verified on the new base rather than assuming** (#1660 changed MoE/EP KV-capacity accounting, so the interaction is exactly what a rebase can surface):
- #1660's own guard `TestEveryKVCapacityCallSiteIsEPAware` **passes** — no `CalculateKVBlocks` call site lost its `WithExpertParallelSize` option.
- **Headline evidence reproduces exactly**: 100 accumulate sessions × 5 rounds at `K=2, α=0.755` — `017fcd98` gives **701/1000** rounds with **121** corruption logs; this branch gives **1000/1000** and **0**.
- **INV-6**: `K=0` stdout byte-identical to `017fcd98` on both a 200-request run and the 100-session multi-turn spec, so the EP capacity changes and this clamp do not interact.
- **INV-13**: `run --trace-output` → `replay` under `K=3, α=0.7` — same completions and total output tokens (50 / 28839).
- **Determinism**: three spec-on runs, identical md5.
- `go build`, `go vet`, `go test ./... -count=1` clean; `gofmt` clean on every touched file (the `sim/request.go` / `sim/simulator.go` alignment noise reproduces on `017fcd98` itself).

Please focus on whether the rebase changed anything substantive, plus the same five areas as before: the clamp target (`completionProgressIndex()` as the R23 single source of truth shared with `processCompletions`), the Phase-1 plain-`min` vs Phase-2 floor-at-1 liveness asymmetry, the INV-9 bounded carve-out for the Phase-2 admission-loop read, cross-path parity, and the length-capped / preemption / timeout paths. The prior cross-vendor `/qa-review` also passed 19/19 CONFIDENT with zero items to fix; I will re-run it after your verdict since the rebase voided it too.

namasl:

## qa-review — PR #1659: ✅ PASS
_fix(sim): clamp spec-decode progress at the completion boundary (Fixes #1657) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR clamps speculative decode grants to the same terminal ProgressIndex used by completion detection, covers both ordinary and PD decode paths, and adds direct and closed-loop regression tests. The additions are narrowly related supporting documentation, tests, and the shared boundary helper. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. The user-facing latency guide, model reference, invariant documentation, and repository-level implementation notes describe the new completion clamp, metric consequences, run/replay parity, INV-9 carve-out, and the existing speculative occupancy limitation. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No materially stale comments were found. Comments now distinguish the proposed accepted-token count from the capped grant, explain zero-work Phase 1 behavior, document the PD floor-at-one exception, and mark the completion-time metric clamp as defense-in-depth. |
| G1 | correctness | ✅ CONFIDENT | In FormBatch Phase 1 (running decode requests), if decodeTokensToCompletion(req) returns 0 because req.ProgressIndex already equals completionProgressIndex(), decodeTokens becomes 0. How does executeBatchStep handle a decode step with NumNewTokens = 0, and does it avoid panics or invalid zero-token KV allocations? | A zero grant is safe. FormBatch skips preemption/allocation and leaves NumNewTokens at 0. executeBatchStep excludes that request from the latency batch, does not advance ProgressIndex or commit carry, and all latency backends are given an empty-or-reduced scheduled batch with a minimum one-tick clock advance. processCompletions then sees the request already at its boundary and completes it, releasing its KV state. There is no zero-token KV allocation. |
| G2 | correctness | ✅ CONFIDENT | In FormBatch Phase 2 (queued decode sub-requests in disaggregated PD), decodeTokensToCompletion(next) is checked with 'if outRoom >= 1'. If outRoom is 0, the else branch forces decodeTokens = 1. Why does Phase 2 grant 1 token when outRoom == 0 while Phase 1 grants 0 tokens, and can this cause a queued decode sub-request at its boundary to execute an extra decode step and overshoot completionProgressIndex? | Phase 1 can grant zero because the request is already in RunningBatch and processCompletions will finish it after the zero-work step. A Phase 2 request is still in WaitQ; granting zero would hit the break and repeat forever because it is never exposed to processCompletions. Therefore Phase 2 preserves the pre-feature PD behavior by admitting it with one token. For a one-output-token decode sub-request already at its boundary, this does advance that internal sub-request one past completionProgressIndex, but completion metrics avoid double-counting and the cluster-level parent/session accounting remains aligned with the baseline. |
| G3 | invariants | ✅ CONFIDENT | INV-9 (Oracle Knowledge Boundary) prohibits admission decisions in FormBatch from reading OutputTokens. Does calling decodeTokensToCompletion(next)—which reads len(OutputTokens)—inside the Phase 2 admission loop alter KV block reservation or token budget checks in a way that allows or denies admission of a queued decode request? | The oracle-derived cap can change per-step resource usage: it may request fewer KV blocks and debit fewer batched tokens, so under pressure it can allow this request, or additional requests behind it, to be scheduled in the current step. It cannot increase the grant, reduce it below one, or deny/reorder a request because of its output length. The documented INV-9 scope treats this as execution-step resource planning rather than a servability decision. |
| G4 | invariants | ✅ CONFIDENT | When FormBatch caps decodeTokens below the proposed value from peekDecodeTokens, does any unconsumed fractional carry remain on req.specDecodeCarry upon request completion, and could that residual carry leak into subsequent rounds or metric aggregations? | A capped grant leaves the ungranted amount in specDecodeCarry during a continuing request, because commitDecodeTokens adds the mean rate and subtracts only the actual grant. If the cap is the completion cap, any residual carry is explicitly reset to zero in recordRequestCompletion. Each follow-up session round is a new Request, so no carry leaks into it or into metrics. |
| G5 | rules | ✅ CONFIDENT | Rule R23 requires a single source of truth for shared calculations. The new req.completionProgressIndex() encapsulates the terminal index law; are there any remaining call sites in sim, workload/session.go, or metrics reporting that still compute target progress via manual InputLen + len(OutputTokens) arithmetic? | The terminal ProgressIndex law itself is centralized in Request.completionProgressIndex and used by both the grant cap and processCompletions. Remaining ProgressIndex-minus-InputLen arithmetic in session and metrics code computes actual generated-token counts, not the target completion boundary. Test code manually computes expected values, which is appropriate for independently checking the helper. |
| G6 | rules | ✅ CONFIDENT | Rule R1 forbids silent degradation or error swallowing. In decodeTokensToCompletion, max(req.completionProgressIndex()-req.ProgressIndex, 0) silently clamps negative differences to 0. If ProgressIndex ever strictly exceeds completionProgressIndex due to an upstream step accounting bug, does this silent clamp mask the corruption? | The clamp does not silently discard a request or metric. In normal execution an overshooting request is completed immediately after its generating step, before FormBatch can call the helper again. If corrupt state is injected into a running request, the zero grant prevents further damage and processCompletions still completes it; for an accumulate session, OnComplete independently detects and logs the over-cap value and cancels that session. Thus the helper is defensive saturation, not the sole corruption detector. |
| G7 | parity | ✅ CONFIDENT | Does the completion boundary clamping in FormBatch operate identically between blis run and blis replay when speculative decoding is enabled, or does trace replaying bypass peekDecodeTokens and decodeTokensToCompletion? | Run and replay both construct the same ClusterSimulator/InstanceSimulator pipeline from SimConfig. NewSimulator enables peekDecodeTokens from SpeculativeConfig, and scheduleBatch supplies it to FormBatch regardless of request source. Replay only changes how Requests are loaded; it does not bypass batch formation. A dedicated direct-run versus exported-trace replay test verifies identical per-request timing under speculative PD decoding. |
| G8 | parity | ✅ CONFIDENT | When comparing metrics between spec-decode (K > 0) and baseline (K = 0) runs on identical inputs, does clamping ProgressIndex at completionProgressIndex guarantee that total output token counts, TPOT, and session context growth reported by blis observe remain bit-identical? | No. The clamp makes ordinary completed requests' output-token accounting and accumulate-session context growth match K=0, but TPOT is intentionally not bit-identical: speculative decoding uses fewer, wider steps with different step costs, and the final clamp also removes previously overcharged per-token detokenization time. Consequently E2E, ITLs, and TPOT may differ even though token totals match. `blis observe` reports metrics from its observed input; the clamp alone cannot make timing observations identical across configurations. |
| G9 | tests | ✅ CONFIDENT | TestSpecDecode_PDDecodeSubRequest_AtBoundary_GrantsOneToken asserts that a 1-output-token PD decode sub-request gets granted 1 token at the boundary. If the Phase 2 fallback logic in FormBatch were removed, would any non-PD cluster or workload test fail, or is this behavior only asserted by this synthetic unit test? | The fallback is specific to speculative PD decode-sub-request admission, so removing it would not affect non-PD tests or the K=0 PD test. It is not covered only by the synthetic unit test, however: the new PD accumulate-session integration test includes outLen=1 and requires all rounds to complete; without the floor, that realistic PD session path would strand the sub-request. |
| G10 | tests | ✅ CONFIDENT | The new test TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens checks that step cost is invariant to granted tokens. Does this test exercise both the trained-physics and roofline latency models under tensor-parallel, expert-parallel, or disaggregated configurations, or only single-GPU defaults? | The test covers both trained-physics and roofline, but only with TP=1 and the default dense model configuration. It does not exercise tensor-parallel TP>1, expert parallelism, or a disaggregated cluster; it invokes each latency model's StepTime directly on a synthetic decode batch. |
| G11 | rationale | ✅ CONFIDENT | Why was completion boundary clamping placed inside FormBatch during decode token grant calculation rather than inside peekDecodeTokens or inside Simulator.executeBatchStep? | FormBatch is the point that turns a pure proposed rate into an actual grant after applying token-budget, output-boundary, MaxModelLen, and KV constraints. Keeping peekDecodeTokens oracle-blind preserves its pure carry calculation. Capping later in executeBatchStep would make allocation, ComputedTokens, and token-budget accounting reflect tokens that were never emitted and would complicate carry commitment. FormBatch therefore keeps granted progress and scheduler resource accounting consistent. |
| G12 | rationale | ✅ CONFIDENT | vLLM grants K+1 lookahead tokens and trims accepted tokens post-execution in _update_from_output, whereas BLIS caps the scheduler grant pre-execution in FormBatch. What alternatives were considered for handling spec-decode overshoots, and why was pre-execution grant capping chosen over post-execution trimming? | The codebase exposes three plausible locations: the existing completion-metric clamp, tolerating/clamping inside SessionManager, or trimming ProgressIndex after execution. The existing metric clamp was insufficient because SessionManager reads raw ProgressIndex before metric clamping, while relaxing SessionManager would propagate incorrect request accounting elsewhere. The chosen pre-execution cap ensures ProgressIndex itself is correct and keeps carry, emitted-token detokenization, KV allocation, and batched-token debit based on one granted count. The tradeoff is that BLIS does not reserve vLLM's full speculative lookahead footprint on the final step; that divergence is explicitly documented under #1627. |
| G13 | scope | ✅ CONFIDENT | The diff includes updates to invariants.md, latency-models.md, and reference/models.md documenting the INV-9 carve-out and speculative occupancy limitation (#1627). Does any of this documentation expansion introduce new requirements or contract changes beyond fixing issue #1657? | No unrelated runtime requirement is introduced. The documentation formalizes the execution-side oracle read needed by this fix and records its resource-accounting consequence. The #1627 occupancy behavior was already a known limitation; the PR only explains that pre-execution final-step capping sharpens the same limitation. |
| G14 | scope | ✅ CONFIDENT | In sim/latency/trained_physics_model_test.go, comment alignment for NumLocalExperts and NumExpertsPerTok was modified. Are there any other non-functional formatting changes or refactoring edits in this PR outside the spec-decode boundary fix? | The shown production changes are limited to the completion-boundary helper, its use in FormBatch/processCompletions, and updated explanatory comments. The trained-physics test file contains incidental gofmt alignment changes in addition to the new test. No unrelated production refactor was found. |
| G15 | errors | ✅ CONFIDENT | If a request has zero input tokens (InputLen == 0) and zero output tokens (len(OutputTokens) == 0), what does completionProgressIndex() evaluate to, and how does decodeTokensToCompletion handle this boundary case in FormBatch? | With zero input and zero output, completionProgressIndex evaluates to 0 + max(0,1) - 1 = 0. At an initial ProgressIndex of 0, decodeTokensToCompletion returns 0. In the ordinary Phase 1 path the decode branch is gated by len(OutputTokens) > 0, so no decode allocation occurs and processCompletions completes the request after the prefill pass. If a hand-constructed zero-output PD decode sub-request reaches Phase 2, the PD liveness fallback would grant one token. |
| G16 | errors | ✅ CONFIDENT | SessionManager.OnComplete cancels closed-loop sessions when actualOutputLen > len(OutputTokens). With FormBatch clamping in place, under what remaining conditions could over-cap output lengths still occur, and does Simulator log a warning if recordRequestCompletion ever observes an unclamped overshoot? | For ordinary requests, an over-cap can now occur only if another/future execution path bypasses FormBatch, state is manually corrupted, or an upstream accounting bug mutates ProgressIndex. The internal PD one-output-token path also intentionally advances its decode sub-request one step past its boundary for liveness, preserving pre-feature behavior. recordRequestCompletion clamps TotalOutputTokens but does not log when that clamp binds. Accumulate SessionManager does log and cancel if the parent request presented to it remains over-cap. |

### Items to fix

### Important to consider
- **F1** — The PD one-output-token internal sub-request intentionally retains its pre-existing one-step overshoot behavior for liveness; parent/session accounting remains correct.
  _sim/request.go:158-176; sim/batch_formation.go:174-212,262-326; sim/simulator.go:1137; sim/simulator_decode_test.go:104-163; sim/cluster/spec_decode_session_test.go:120-219. `go test ./sim/... -count=1` and `go test ./cmd -count=1` both pass._
- **G2** — The exact-boundary guarantee has an intentional internal PD sub-request exception inherited from the pre-feature PD lifecycle.
  _sim/batch_formation.go:184-212,277-326; sim/simulator.go:726-748; sim/simulator_decode_test.go:289-331; sim/cluster/spec_decode_session_test.go:184-219; sim/cluster/disaggregation_test.go:621-646._
- **G3** — This makes final-step occupancy slightly more optimistic relative to vLLM, as documented under the existing #1627 limitation.
  _sim/batch_formation.go:275-325; docs/contributing/standards/invariants.md:124-132._
- **G6** — Outside an accumulate-session callback, recordRequestCompletion clamps over-counting without logging the unexpected overshoot.
  _sim/batch_formation.go:88-108,184-212; sim/simulator.go:819-831,1123-1137; sim/workload/session.go:280-294; docs/contributing/standards/rules.md:24-32._
- **G10** — Additional topology-crossed coverage could be useful, but the property under test is local to both latency backends' treatment of NumNewTokens.
  _sim/latency/trained_physics_model_test.go:511-525,559-570,597-632; the helper configurations set TP: 1, and trainedPhysicsTestModelConfig uses NumLocalExperts: 0._
- **G12** — The repository does not record an exhaustive design-alternative discussion, so this answer reflects the alternatives and rationale evidenced by the implementation and documentation.
  _sim/simulator.go:726-745,1027-1040,1086-1101; sim/workload/session.go:201-211,280-294; sim/batch_formation.go:181-212,275-325; docs/guide/latency-models.md:235-239._
- **G15** — A zero-output PD decode sub-request is defensive/hand-constructed territory; normal zero-output requests complete at prefill.
  _sim/request.go:158-176; sim/batch_formation.go:107-108,174-175,267-314; sim/simulator.go:1127-1137._
- **G16** — Logging when the defense-in-depth metric clamp binds would improve diagnostics, but its absence is not data loss because output conservation is preserved.
  _sim/batch_formation.go:184-212,277-305; sim/simulator.go:726-748; sim/workload/session.go:280-294._

namasl:
@claude /blis-pr-review

Closing the post-rebase review cycle on `fefc04b7` (no new commits since your READY-TO-MERGE re-review — this comment exists so the `Claude Code Action (comment trigger)` status resolves on this commit; the `/qa-review` comment above triggered a skipped run that left it pending, which blocks merge on its own).

**All gates green on `fefc04b7` (rebased onto `main` `017fcd98`):**
- **archon** — `NO_CHANGE`, internal-only, fast-track eligible.
- **blis-pr-review** — READY TO MERGE (post-rebase re-review; `claude: success` in the job breakdown, not a skipped run). It independently confirmed the rebase changed nothing in the production Go path, that `completionProgressIndex()` is algebraically identical to the inline expression it replaced, that the accessor is absent from every servability file, that the Phase-2 double-floor has no ordering interaction, and that #1660's EP KV-capacity work shares no code path with this clamp.
- **qa-review** (cross-vendor: Gemini questioner + OpenAI Sol answerer, 19 fresh questions on the rebased head) — **PASS**, 19/19 CONFIDENT, **Items to fix: none**.
- **CI** — build, lint and all 7 test shards pass on this head.

The re-run qa-review raised one thing worth stating explicitly for the record, because it is a sharper reading than my PR body gave: the "lands exactly ON the boundary" guarantee has one **intentional internal exception** — a 1-output-token **PD decode sub-request** is admitted already at its boundary, and the Phase-2 floor-at-1 advances it one past `completionProgressIndex()`. That is required for liveness (it is not yet in `RunningBatch`, so a 0-token grant would strand it in the wait queue forever) and is exactly what the pre-feature `K=0` PD path did, so `ProgressIndex − InputLen == 1` equals its one output token and the parent/session accounting still matches the `K=0` baseline. Both reviewers reached the same conclusion independently. It is documented at the call site and in the INV-9 carve-out.

Its other non-blocking notes are recorded in that comment; none is a defect in this change. Three are diagnostics-only suggestions I did not fold in, to avoid re-opening a cycle on a green head: logging when the defense-in-depth `recordRequestCompletion` clamp binds (it cannot lose data — conservation is preserved either way, and an accumulate session already logs and cancels loudly), a direct `UsedBlocks()==0` assertion after the PD `outLen=1` run (KV release is covered by the closed-loop PD test plus the unconditional `ReleaseKVBlocks`), and TP>1 / EP / disaggregated variants of the verify-width cost test (the property under test is local to how each backend treats `NumNewTokens`, which topology does not vary). Happy to fold any of them in if you would rather they land here.

Still `BLOCKED` on `REVIEW_REQUIRED` only — this needs a collaborator's approving review; I cannot self-approve, and I have not attempted any workaround.

susiejojo (approved):
lgtm

Issue #1556, opened 2026-08-12
DP-as-placement parity for blis replay (--dp > 1 on MoE)


namasl:
## Summary

Follow-up to #1531. #1531 makes MoE `--dp N` real placement on `blis run` (N single-node engine replicas per `--num-instances`, each `DP=1`) and makes it **run-only**: `blis replay` `logrus.Fatalf`s on MoE `--dp>1` at startup (INV-13 — replay keeps the single-instance DP-math model, so run and replay would diverge for identical flags).

The three DP-as-placement steps (expand `numInstances`, divide the auto KV total to per-rank, set `DP=1`) are **mechanically identical** in the replay command body, which already supports multi-instance and PD pools. So replay parity is a deferred feature, not a fundamental limitation.

## Acceptance criteria

- [ ] `blis replay` supports MoE `--dp N` as placement (N `DP=1` replicas), producing metrics identical to `blis run` for the same trace + flags (restoring INV-13 parity for this config).
- [ ] The #1531 replay guard (`cmd/replay.go`) is lifted once parity holds.
- [ ] INV-1 / INV-6 hold on the replay path.

## Context

- Blocked-by: #1531. Related: #1532 (GLM-5.2 tracking), #1553 (DP + PD/autoscaler/node-pools on run).

Issue #1420, opened 2026-06-10
#D — KV-capacity DP scaling


sriumcp:
## #D — KV-capacity DP scaling

**Tracker:** #1416 · **Umbrella:** #1348 · **Source of truth:** `docs/plans/2026-06-10-dp-ep-latency-scoped-design.md` §6 · **vLLM parity:** `vllm-project/vllm@f6ec81c7`

`CalculateKVBlocks` (`sim/latency/kv_capacity.go:136`) sizes the KV budget for a single TP group. Each DP rank is a separate `EngineCore` with its own full KV budget on its own GPUs, and requests split disjointly across ranks, so aggregate usable KV blocks scale by `DP`.

### Why (vLLM proofs)
- **Each DP rank profiles & allocates its own KV cache independently** — one full budget per `DPEngineCoreProc` (MoE-only): [core.py:1243-1276](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/v1/engine/core.py#L1243-L1276).
- **Per-GPU KV sized by attention TP only, not DP** (`num_kv_heads = total_num_kv_heads // tp_size`): [qwen3_moe.py:229-242](https://github.com/vllm-project/vllm/blob/f6ec81c766c63e59216c0845b5cb3b1d05a19cfb/vllm/model_executor/models/qwen3_moe.py#L229-L242).
- **KV budget is independent of EP mode** — KV is for attention, and `enable_expert_parallel` touches only MoE expert sharding/all-to-all, never attention or KV. So EP-off (flattened MoE TP) and EP-on (expert parallel) both have one KV budget per DP `EngineCore`.
- **Dense `DP>1` is NOT one logical instance** in vLLM (N independent engines) → rejected upstream in #A; never reaches capacity sizing.

⟹ Aggregate budget = `MemoryGiB · util · TP · DP`; per-GPU KV bytes remain divided only by attention `TP`.

### Tasks
- Multiply the **final usable KV-block count** by `DP` when `isMoE && DP > 1`, regardless of `EnableExpertParallel`.
- **Do not** change per-GPU weight/activation overhead — those are per-replica and unchanged; only the block count scales. Update the `dp=1` comment.
- **Plumbing:** extend `CalculateKVBlocks` with an explicit positional `dp int` immediately after `tp int`, so all sites fail to compile until updated; do not multiply at scattered call sites.

[...]

PR #1660, opened 2026-09-03
feat(latency): EP-aware weight-footprint sizing — shard routed-expert weights across the EP group (Closes #1656)

### LINKED ISSUE #1656: EP-aware weight-footprint sizing: KV auto-calc replicates routed experts across TP, blocking large MoE+EP deployments

namasl:
## Summary

`CalculateKVBlocks` charged **every** routed expert's weights to **every GPU in the TP group** — weights were modeled as tensor-sharded (`/TP`) but never *expert*-sharded across an expert-parallel group. For a large MoE served with EP that is the dominant term (a Kimi-K3-class checkpoint carries ~99% of its bytes in routed experts), so the auto weight footprint came out a factor of `DP` too large and KV auto-calculation **failed outright** — a real 32-GPU `TP=16 / DP=2 / EP=32` deployment was reported as needing "21 GPUs per instance".

This PR makes the weight footprint EP-aware: the **routed-expert** term is charged to the EP group (`ep_size = TP·DP`, mirroring vLLM's `--enable-expert-parallel`, where each rank holds `num_experts/ep_size` *whole* experts), while attention, shared experts, dense-prefix MLP, router/gate, embeddings and norms stay on the TP-sharded path.

### The arithmetic basis (the trap this PR had to avoid)

The issue proposes `routed_expert_bytes /= EffectiveEP()`. Taken literally that is wrong by a factor of `TP`: `computeModelWeightBytes` returns a **total** over the rank's TP GPUs and is compared against `gpu_mem × util × TP`, so the total already carries an implicit `/TP`. Dividing the total by `EP` would charge `R/(TP²·DP)` per GPU. The implemented law is therefore

```
charged total = N + R · TP/EP        ⇒  per-GPU = N/TP + R/EP
```

and the BC-2 test asserts exactly that, *and* asserts the observed delta does **not** match the per-total misreading — so the tolerance can never be loose enough to accept both.

## Behavioral contracts

- **BC-1 (INV-6).** GIVEN any inputs WHEN no EP option is passed (or `0`/`1`, the "EP off" sentinels) THEN the block count and every error message are exactly the pre-change values. The EP-off path returns the original expression, evaluated in the original order — bit-identical, not merely equal.
- **BC-2.** GIVEN routed-expert bytes `R` and other weights `N` at `TP` GPUs/rank with group `E ≥ TP` WHEN capacity is computed THEN the charged overhead is `N + R·TP/E` (per-GPU `N/TP + R/E`), so at `E = TP` nothing changes and at `E = TP·DP` the routed term shrinks by exactly `DP`.
- **BC-3.** GIVEN the same `(mc, tp)` WHEN the EP group varies THEN per-token/per-block KV bytes are unchanged (EP shards experts, never attention) and DP block scaling stays exactly `×dp` at any fixed group.
- **BC-4.** GIVEN a dense model, or a sub-`MoEMinExperts` config, WHEN a group `> TP` is supplied THEN capacity is unchanged — no phantom gain where there are no routed experts to move.
- **BC-5.** GIVEN a user-requested `TP=t, DP=d>1` with EP on THEN the group is `t·d` (the **logical** topology). A per-replica config that DP-as-placement (#1531) reconfigured to `DP=1` yields `t` — the two values are different, and the CLI passes the logical one. Reading EP off a per-replica config would silently no-op the sharding for exactly the topology this issue is about.
- **BC-6 (R1/R3).** The resolved value is a *divisor* of the routed-expert bytes, so a larger value charges less memory and a smaller one charges more. The two bad-input directions are therefore handled differently on purpose: `ep > TP·DP` is a **hard error** (no safe substitute exists — the caller described more EP ranks than GPUs, and honouring it would charge only `TP·DP/ep` of the experts); `ep > num_routed_experts` is **clamped down** to the expert count with a `logrus.Warnf` (the loaded ranks still hold one *whole* expert, so that is the widest divisor the footprint supports, and clamping down charges *more* memory — the safe direction). Negative `ep` and an overflowing `TP·DP` also error; anything `<= TP` resolves to `TP`, a strict no-op, so the very common `ep == TP` (DP=1, EP on) case can never fail validation.
- **BC-7 (R23).** Every production `CalculateKVBlocks` call site passes a group size; a repo-walking test fails if a new site omits it, so a MoE+EP deployment cannot be sized correctly by one path and over-counted by another.
- **BC-8.** GIVEN the committed `glm-5.2-fp8` fixture on 80 GiB GPUs at `TP=8, DP=2` THEN EP off fails with the weight-overhead error and `EP=16` sizes successfully (675 GiB of routed experts over 16 GPUs instead of 8).
- **BC-9.** GIVEN a large MoE run as `--dp 2 --enable-expert-parallel` THEN the run fails on the honest reason (the #1548 EP-placement guard) instead of a bogus capacity fatal that told the operator to buy more GPUs.

## What changes for a user today (and what does not)

The reduction only bites when `EP > TP`, i.e. `DP > 1`. At `DP = 1` the EP group **is** the TP group, so `/EP ≡ /TP` — correct physics, and byte-identical output.

`--dp>1` with `--enable-expert-parallel` is still rejected on `run` by `planDPPlacement` (#1548) and MoE `--dp>1` on `replay` (#1556), so the end-to-end EP-sharded run needs those. As the issue frames it, this is the **capacity half** of #1548 ("without it, an EP-sharded MoE can't even be sized"). What changes today is that the bogus capacity fatal no longer **masks** those guards — both `run` and `replay` call the shared `resolveLatencyConfig` *before* their own guard, so the operator now sees the accurate unsupported-feature message:

```
# before
fatal  --latency-model: KV capacity auto-calculation failed: … model overhead (687.80 GiB …)
       exceeds available GPU memory (576.00 GiB …). Minimum GPUs required per instance: 10

# after
fatal  --dp > 1 with --enable-expert-parallel is not yet supported: … EP-on placement
       (experts sharded across the DP group + inter-node network cost) is tracked by #1548.
```

## Implementation

- `sim/config.go` — `EffectiveEPSize(isMoE, tp, dp, epOn)`, the canonical group-size formula (pure, so the *logical* group is computable where no per-instance config exists); `ModelHardwareConfig.EffectiveEP()` becomes its config-bound accessor (semantics unchanged).
- `sim/latency/kv_capacity.go` — `WithExpertParallelSize(ep)` on the existing variadic option seam (same mechanism as `WithAdapterReservedBytes`, so positional call sites keep their shape); `resolveExpertShardSize` validation; the routed-expert subtotal is isolated inside the existing MoE branch and scaled by `tp/ep`. The overhead-exceeded message gains one clause **only** when the reduction is active, naming the assumption its minimum-GPU estimate rests on (that the group grows with TP at fixed DP).
- Call sites plumbed: `cmd/root.go` (global + PD prefill/decode), `cmd/replay.go` (PD prefill/decode), `sim/cluster/kv_autocalc.go`. `cmd.epSizeForKVCapacity(isMoE, tp)` is the single CLI formula (per-pool TP honoured; per-pool EP is not expressible, exactly as per-pool DP is not — #1420).
- Step time stays EP-blind (#1548 owns the EP-mode step-time toggle).

## Testing

`go build ./...` clean · `go test ./... -count=1` all packages pass · `golangci-lint run ./...` **0 issues**.

New: `sim/latency/kv_capacity_ep_test.go` (BC-1/2/4/6/8 + monotonicity + the honest error message), `cmd/ep_kv_capacity_test.go` (BC-7 repo walk, BC-9 subprocess, plus a behavioral test that the shared resolver returns strictly more blocks with EP on), `sim/config_test.go` (BC-5 logical-vs-per-replica, R23 accessor parity), and `sim/latency/kv_capacity_test.go`'s EP-independence test rewritten to the law it was really protecting (per-block bytes EP-independent; `×dp` scaling at fixed group).

**INV-6 evidence** — stdout *and* exit code diffed between binaries built from `main` and this branch, byte-identical on all **eight**: dense qwen3-14b; MoE deepseek-v2-lite TP=1; **MoE + `--enable-expert-parallel` at DP=1** (the `EP == TP` no-op); MoE `--dp 2` EP off (DP-as-placement); mixtral TP=2; mixtral TP=8 EP-on; a 4-expert MoE at TP=8 EP-on (`num_experts < TP`); qwen3-30b-a3b EP-on. Wide EP (`--tp 4 --dp 64`, ep=256 over 128 experts) reports the #1548 guard exactly as `main` does, with the clamp warned on stderr.

**Review history worth knowing** (three commits, not squashed by me): the first commit rejected an EP group wider than the routed-expert count. Two parallel cross-vendor reviews showed that bound fired on the *inert* `ep == TP` case — breaking previously-working runs such as `--tp 16 --enable-expert-parallel` on an 8-expert Mixtral — and, for genuine wide EP, replaced the honest #1548 diagnostic with an expert-count fatal, i.e. re-created the masking BC-9 exists to remove. The second commit replaced it with the downward clamp and corrected the vLLM framing above; the third fixed the last doc contradiction, which the qa-review pass flagged twice. The load-bearing `ep > TP·DP` rejection was never relaxed.

## Documented approximations (added to `docs/reference/models.md`)

Honest about what the freed headroom rests on: the per-rank expert charge is a continuous **average** while vLLM sizes KV from the *most-loaded* rank (non-divisible `ep_size`, EPLB / redundant experts are optimistic); the flat `8.0 GiB` MoE activation term and `0.6 GiB/GPU` non-torch allowance become the **dominant residual** once the routed term shrinks by `DP`, and the latter does not cover DeepEP/all-to-all workspaces; MTP module weights (`num_nextn_predict_layers`) are unmodeled. All point the same way (optimistic), so post-EP KV headroom is an upper bound.

## Discovered and filed: #1666 — the EP-**off** baseline is BLIS's DP model, not vLLM's

Verified at source while reviewing this change: vLLM's `FusedMoEParallelConfig.make` calls `flatten_tp_across_dp_and_pcp` **unconditionally**, before the `use_ep` branch. At TP=2/DP=2 with EP **off** the MoE `tp_size` is 4 — so vLLM's per-GPU routed-expert bytes are `R/(TP·DP)` in **both** EP modes, and `--enable-expert-parallel` changes the sharding *style* (whole experts vs tensor slices) and the collective, not the footprint.

The EP-**on** value this PR computes matches vLLM exactly. The EP-**off** baseline it is differenced against is BLIS's own DP model (#1531: MoE `--dp N` = N independent engine replicas, each holding a full tensor-sharded copy), which makes BLIS's EP-off `DP>1` capacity **conservative** rather than vLLM-faithful. That is a divergence in the DP model, not in this term, and it is filed as **#1666** and recorded in `CLAUDE.md`, `docs/reference/models.md` and `docs/guide/latency-models.md`.

One consequence worth stating: the trained-physics step-time model's `moeGroup = TP·DP` and its "EP-mode-agnostic" comments (`sim/expert_placement.go`, `sim/latency/trained_physics_model.go`) are **CORRECT** — they already match vLLM. It is the capacity side's EP-off baseline that differs, so #1666 (to be settled with #1548, which owns the EP-mode step-time toggle) is where the two are reconciled.

Closes #1656


namasl:
@claude /blis-pr-review

The deterministic ARCHON review above reports **`ARCHITECTURAL_CHANGE`, 0 violations**: no package added/removed, no import edge added/removed, 0 schema changes. The two public-surface additions are intentional and are the feature's seam:

- `sim.EffectiveEPSize(isMoE, tp, dp, epOn)` — the canonical expert-parallel group-size formula, extracted as a **pure free function** because the consumers that need the LOGICAL (user-requested) group are the CLI KV auto-calc paths, which hold the raw `--tp/--dp/--enable-expert-parallel` and cannot read it off a per-instance config (DP-as-placement, #1531, sets `DP=1` per replica, which would collapse EP to TP and silently no-op the sharding). `ModelHardwareConfig.EffectiveEP()` is now its config-bound accessor with unchanged semantics.
- `latency.WithExpertParallelSize(ep)` — added on the existing variadic `KVCapacityOption` seam (same mechanism as `WithAdapterReservedBytes`), so no positional signature changed and no new package arrow exists.

Please focus the review on:

1. **The arithmetic basis.** `computeModelWeightBytes` returns a TOTAL over one DP rank's TP GPUs and is compared against `gpu_mem × util × tp`, so it already carries an implicit `/TP`. The routed-expert subtotal is therefore **scaled by `tp/ep`**, not divided by `ep` — the latter would charge `R/(TP²·DP)` per GPU. Please verify the per-GPU claim `N/TP + R/EP` independently, and that the `dp` block multiplier at Step 6 does not double-count.
2. **INV-6.** EP off (option absent, `0`, `1`, or a group equal to TP) must be bit-identical, which the code achieves by returning the original expression in the original order rather than by relying on the reduction degenerating. Evidence: stdout diffed between binaries built from `main` and this branch is byte-identical for dense, MoE TP=1, **MoE + `--enable-expert-parallel` at DP=1** (the `EP == TP` no-op), MoE `--dp 2` EP-off, and mixtral TP=2.
3. **Validation direction (R1).** Valid domain is `{0,1} ∪ [TP, min(TP·DP, num_routed_experts)]`. The **upper** bounds are the load-bearing ones — those are the directions in which a bad group hands back capacity for memory that does not exist. Is anything missing, and is the ordering right (the resolver runs after `KVBytesPerToken` so `tp > 0` is guaranteed before the bound messages are formatted)?
4. **INV-13 / cross-path parity.** `run` and `replay` both resolve capacity through the shared `resolveLatencyConfig`; all 6 production call sites are plumbed (run global + PD prefill/decode, replay PD prefill/decode, per-instance node-pool auto-calc) and a repo-walking test fails if a new site omits the option. Nothing is round-tripped through the TraceV2 header — deliberate, matching how `--gpu-memory-utilization` and `--kv-cache-dtype` are re-supplied on the replay CLI.
5. **Scope honesty.** The reduction only bites when `EP > TP` (i.e. `DP > 1`); at `DP = 1` the EP group *is* the TP group. `--dp>1 --enable-expert-parallel` is still rejected by #1548 on run and #1556 on replay, so this is the capacity half the issue describes. What changes today is that the bogus capacity fatal no longer **masks** those guards. Please check the PR body and CLAUDE.md do not over-claim beyond that.

CI: all 7 test shards, build, and lint pass.

namasl:
@claude /blis-pr-review

Re-review of `265aa935` (your earlier READY TO MERGE was on `88cf1e6f`). The deterministic ARCHON review immediately above is unchanged: **`ARCHITECTURAL_CHANGE`, 0 violations, 0 edge changes, 2 intentional surface additions**.

Two independent cross-vendor code reviews (run in parallel with yours) found a **regression I had introduced**, and this commit fixes it plus the framing issue behind it. Please focus there:

**1. The routed-expert-count bound is now a CLAMP, not a rejection.** The bound fired even when the EP reduction is not taken (`ep == tp`, i.e. *every* DP=1 `--enable-expert-parallel` run), so `--tp 16 --enable-expert-parallel` on an 8-expert Mixtral began failing a KV auto-calc whose block count the feature does not change at all. Worse, for genuine wide EP (`tp·dp > num_experts` — the DeepSeek EP320-over-256 class) it reported an expert-count error where `main` reported the honest #1548 guard, i.e. it re-created exactly the masking BC-9 exists to remove. Now: anything `<= tp` resolves to `tp` (a strict no-op, so it can never fail validation), and a group wider than the routed-expert count is clamped to that count with a `logrus.Warnf`. Please check that clamping is the right *direction* — the loaded ranks still hold one WHOLE expert, so `num_experts` is the widest divisor the footprint supports and the sub-one-expert average would be optimistic — and that `logrus.Warnf` in `sim/latency` is acceptable (precedent: `config.go` hybrid-attention fallback).

**2. A vLLM claim this PR was making is corrected.** `FusedMoEParallelConfig.make` calls `flatten_tp_across_dp_and_pcp` **unconditionally**, before the `use_ep` branch — at TP=2/DP=2 with EP **OFF** the MoE `tp_size` is 4. So vLLM's per-GPU routed-expert bytes are `R/(TP·DP)` in **both** EP modes. The EP-ON value this PR computes matches vLLM exactly, but the EP-OFF baseline it differences against is BLIS's own DP model (#1531: N independent replicas, each holding a full tensor-sharded copy), so BLIS's EP-off `DP>1` capacity is **conservative**. That is now stated in the option contract, CLAUDE.md and `docs/reference/models.md`, and filed as **#1666**. Consequently the step-time model's "EP-mode-agnostic" comments (`sim/expert_placement.go`, `trained_physics_model.go`) are **CORRECT** — they already use the flattened `TP·DP` group — and my earlier note telling a future PR to "correct" them is removed. Please sanity-check that reading of vLLM, since three docs now assert it.

**3. Coverage added for the two terms that must NOT move with the EP group.** BC-2's exact-bytes law now also runs on the `glm-5.2-fp8` fixture, which has BOTH a dense prefix (3 of 78 layers) and a shared expert — an implementation that scaled all layers, or folded the shared expert into the routed subtotal, passed the entire suite before. Plus a routed-weight **conservation** law (summed over `dp` ranks, routed bytes at `ep = tp·dp` total exactly `R`, versus `dp·R` with EP off) and a regression guard that a group at or below TP never fails validation.

Also addressed from your minor notes: the misleading `overflow_guard` case label (renamed — it exercises the upper bound; the overflow probe has its own test), `EP=off` instead of `EP=1` in the auto-calc log, the call-site guard tightened to require a topology-**derived** group (`epSizeForKVCapacity`/`EffectiveEP`, since a literal `WithExpertParallelSize(1)` would be EP-blind), and one shared `params.hasRoutedExperts()` predicate for the resolver and the arithmetic (R23). On your Q4 (PLAUSIBLE): `writeCompleteMoEFixture` is genuinely MoE — `num_local_experts: 8, num_experts_per_tok: 2` — and the test is non-vacuous by evidence, since the resolver logs `EP=off → 97912` vs `EP=4 → 183928` blocks on it.

**INV-6 re-verified** against a binary built from `main` across **8** configurations (stdout AND exit code), now including both regression shapes: dense; MoE TP=1; MoE EP-on at DP=1; MoE `--dp 2` EP-off; mixtral TP=2; mixtral TP=8 EP-on; a 4-expert MoE at TP=8 EP-on (`num_experts < tp`); qwen3-30b-a3b EP-on. Wide EP (`tp=4 dp=64`, ep=256 over 128 experts) once again reports the #1548 guard, with the clamp warned on stderr.

namasl:
### Clarifications for the re-review (three points a reviewer should not have to re-derive)

**1. Rejection was NOT removed — the load-bearing upper bound is intact.** The earlier review called the `ep > tp·dp` check load-bearing because a too-large group "would charge only `tp·dp/ep` of routed experts, handing back memory that doesn't exist." That check is unchanged and still a hard error. The split in `resolveExpertShardSize` is now:

| condition | outcome |
|---|---|
| `ep < 0`, or `tp < 1` | **error** |
| `ep <= tp` | resolve to `tp` — a strict no-op on the block count |
| `tp·dp` overflows int64 | **error** |
| `ep > tp·dp` | **error** (the load-bearing bound, unchanged) |
| `ep > num_routed_experts` | **clamp down** to `num_routed_experts`, warned |

**2. The clamp direction is conservative, with the arithmetic.** The charged per-rank routed total is `R·tp/expertShardSize`. Clamping `expertShardSize` **DOWN** from `ep` to `num_routed_experts` makes the divisor *smaller*, so the charge is *larger* — more memory charged, fewer KV blocks. Concretely at `tp=2, dp=4`, 4 experts: a requested `ep=8` resolves to 4, charging `R·2/4 = R/2` rather than the requested `R·2/8 = R/4`. Nothing anywhere divides by an unclamped larger `ep`, so the original hazard is not reintroduced — it is moved from "reject" to "charge the safe amount". `TestCalculateKVBlocks_EP_ClampsGroupToRoutedExpertCount` asserts exactly this (`ep=tp·dp` must return the *same* capacity as `ep=num_experts`, never more).

**3. The clamp is loud, never silent (R1).** `logrus.Warnf` names the requested group, the expert count, and what is charged instead. Verified on a real run (`qwen3-30b-a3b --tp 4 --dp 64 --enable-expert-parallel`):

```
level=warning msg="KV capacity: expert-parallel group size 256 exceeds the model's routed-expert
count 128; charging routed-expert weights over 128 GPUs (one whole expert each) instead.
Expert redundancy (--enable-eplb / --num-redundant-experts) is not modeled."
```

**Why the reversal** (a stated cause, not a loosened guard): the rejection had two concrete failure modes, both reproduced by the reviewers. (a) It fired on the **inert** `ep == tp` case — every DP=1 `--enable-expert-parallel` run — so `--tp 16 --enable-expert-parallel` on an 8-expert Mixtral began failing a KV auto-calc whose block count the feature does not change at all; that is a previously-working configuration. (b) For genuine wide EP (`tp·dp > num_experts`, the DeepSeek EP320-over-256 class) it emitted an expert-count fatal where `main` emitted the honest #1548 guard — re-creating the exact masking BC-9 exists to remove. And wide EP over fewer experts is not an impossible deployment: vLLM does not reject it (only `--enable-eplb` requires even distribution), and `--num-redundant-experts` is how it is built.

### On the archon invariant removal

The one removed invariant, `TestCalculateKVBlocks_DPScaling_EPIndependent`, was **renamed and refined, not dropped** — it is now `TestCalculateKVBlocks_DPScaling_KVSizingEPIndependent` (`sim/latency/kv_capacity_test.go:2048`). Its old comment asserted as a *contract* that "CalculateKVBlocks has no EP parameter at all", which is precisely what this PR changes. The refined test keeps the two halves that remain true and are worth protecting: per-token/per-block KV bytes are EP-independent (EP shards experts, never attention — `KVBytesPerToken` still takes no EP input), and DP block scaling stays exactly `×dp` at a fixed EP group. Net invariant count across the three packages is +20/−1.

### Fixture check (the prior review's one PLAUSIBLE answer)

`writeCompleteMoEFixture` is genuinely MoE, verified by reading it rather than inferring from CI — `"num_local_experts": 8, "num_experts_per_tok": 2`, with `vocab_size: 32000` and complete dims (`cmd/dp_placement_test.go`). So `TestResolveLatencyConfig_EPRaisesAutoKVCapacity` is not vacuous, and it is non-vacuous by evidence too: on that fixture the shared resolver logs `EP=off → 97912 blocks` vs `EP=4 → 183928 blocks`.

namasl:

## qa-review — PR #1660: ✅ PASS
_feat(latency): EP-aware weight-footprint sizing — shard routed-expert weights across the EP group (Closes #1656) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR implements the scoped KV-capacity half of the issue: routed-expert weights alone are scaled from the TP basis to the effective EP basis, all production auto-calculation paths pass the EP size, and EP-off behavior remains unchanged. The still-rejected end-to-end `DP>1 + EP` placement is explicitly owned by #1548 rather than this capacity change. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The user-facing and maintainer documentation was updated across the glossary, latency guide, model reference, roofline scope note, and CLAUDE.md. It explains the routed-only EP scaling, EP-off behavior, CLI reachability limitation, and known approximations. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale code comment affecting correctness was found. The relevant comments were updated to distinguish per-token KV sizing from EP-sensitive weight footprint sizing and to document the current placement limitation. |
| G1 | correctness | ✅ CONFIDENT | In `sim/latency/kv_capacity.go`, `resolveExpertShardSize` validates `ep` against `tp` and `tp*dp`. What happens if `tp` or `dp` passed into `CalculateKVBlocks` is 0 or negative, or if `p.NumLocalExperts` is less than `MoEMinExperts` when `WithExpertParallelSize` is passed? Could `expertShardSize` calculation result in a divide-by-zero or bypass validation? | There is no divide-by-zero path. `CalculateKVBlocks` rejects `dp < 1` before EP resolution, and `KVBytesPerToken` rejects `tp <= 0` before EP resolution. `resolveExpertShardSize` also independently rejects `tp < 1`. If the routed-expert count is below `MoEMinExperts`, `hasRoutedExperts` is false, the model follows the dense weight path, `routedExpertParams` remains zero, and no EP scaling is applied. An invalid negative EP size is still rejected, while an otherwise valid EP value is inert for that degenerate model. |
| G2 | correctness | ✅ CONFIDENT | In `computeModelWeightBytes`, routed-expert bytes are scaled by `float64(tp) / float64(expertShardSize)`. How does this calculation handle dense-prefix MoE models (`first_k_dense_replace`) where early layers are dense MLPs — are dense layer weights guaranteed to stay on the `TP`-sharded path rather than being sharded by `expertShardSize`? | Yes. Dense-prefix layers are calculated separately using `densePrefixMLPPerLayer`. Only the remaining MoE layers contribute to `routedExpertParams`, and only that subtotal receives the `TP/expertShardSize` scaling. Dense-prefix, shared-expert, router, attention, embedding, norm, and head weights remain in `nonRoutedParams` on the TP path. |
| G3 | invariants | ✅ CONFIDENT | How does `resolveLatencyConfig` guarantee run/replay parity (INV-13) when `cmd/replay.go` and `cmd/root.go` auto-calculate KV blocks for separate prefill and decode pools with `--prefill-tp` and `--decode-tp`? If prefill and decode use different TP degrees under `--enable-expert-parallel`, does `epSizeForKVCapacity` properly calculate pool-specific EP sizes without violating determinism (INV-6)? | Run and replay share the global resolver, and their separate prefill/decode paths use the same pool logic. Each path first resolves its pool-specific TP and then calls `epSizeForKVCapacity`, which deterministically computes `TP(pool) * global DP` for an EP-enabled MoE. Thus pools with different TP degrees get corresponding different EP group sizes without randomness or map-order dependence. |
| G4 | invariants | ✅ CONFIDENT | By scaling routed expert weights down by `TP/EP` in `CalculateKVBlocks`, per-GPU memory overhead decreases and total KV blocks increase. Does the calculation ensure INV-4 (KV conservation) is preserved when custom adapter reserved bytes or fixed activation memory constants are present alongside EP sharding? | Yes. EP changes only the routed portion of `modelWeightGiB`. Adapter reservation, activation memory, and non-torch overhead remain separately additive per-rank overheads; the remaining memory is converted once to blocks, and DP aggregation occurs afterward. This changes the configured total block count but not the allocator identity `allocated + free = total`, so INV-4 remains structurally preserved. |
| G5 | rules | ✅ CONFIDENT | In `sim/latency/kv_capacity.go`, when `ep` exceeds `num_routed_experts`, `resolveExpertShardSize` logs a warning via `logrus.Warnf` inside a core library package and clamps the value. Does writing directly to global `logrus` inside `sim/latency` violate library boundaries or rule R1 regarding silent modification of configuration parameters? | No stated rule is violated. R6 prohibits `Fatalf`/process termination in `sim`, not warnings; existing latency-library code already uses `logrus.Warnf`. R1 requires potentially lossy adjustment to be observable, and the clamp emits a warning explaining the requested and effective sizes. The clamp is conservative: it prevents charging less than one whole expert on the most-loaded rank. |
| G6 | rules | ✅ CONFIDENT | `EffectiveEPSize` is defined as a free function in `sim/config.go`, while `epSizeForKVCapacity` is defined in `cmd/root.go`. To adhere to canonical helper rules (R23), why is `epSizeForKVCapacity` placed in package `cmd` rather than exposing a unified accessor directly from `sim` for all CLI and engine callers? | `sim.EffectiveEPSize` is the unified canonical formula. The `cmd` helper is intentionally a thin adapter over it because CLI capacity sizing must use package-level logical user topology, whereas `ModelHardwareConfig.EffectiveEP()` describes the possibly rewritten per-instance topology after DP placement. Keeping the CLI-state access in `cmd` avoids making `sim` depend on CLI globals while still using one formula. |
| G7 | parity | ✅ CONFIDENT | In `sim/cluster/kv_autocalc.go`, `applyPerInstanceKVCapacity` passes `simCfg.EffectiveEP()` to `CalculateKVBlocks`. If DP-as-placement (#1531) reconfigures individual engine replicas to `DP=1`, will cluster auto-calculation compute an EP group of `TP` while CLI auto-calculation computes `TP*DP`, causing a disparity between cluster simulation and CLI execution? | No reachable disparity exists today. Node-pool deployments with MoE `DP>1` are rejected by `planDPPlacement`, and EP-on `DP>1` is also rejected before cluster execution. Therefore node-pool per-instance auto-calculation currently sees `DP=1`, where `EffectiveEP()==TP` and the EP reduction is a no-op. The source explicitly records that deployment-wide EP must be threaded through `KVAutoCalcConfig` when #1548 enables that future combination. |
| G8 | parity | ✅ CONFIDENT | Does `--latency-model roofline` use the exact same `CalculateKVBlocks` path with `WithExpertParallelSize` as `trained-physics`, or is there any flag combination where roofline and trained-physics derive divergent KV capacity values when EP is enabled? | Both analytical backends enter the same `resolveLatencyConfig` auto-capacity branch and call the same `CalculateKVBlocks` invocation with `WithExpertParallelSize`. Consequently they derive the same capacity for identical model, hardware, and topology inputs. Roofline with `DP>1` or EP enabled is then rejected because its step-time model is DP/EP-blind, so no successful roofline EP configuration can expose a divergent capacity. |
| G9 | tests | ✅ CONFIDENT | `TestEveryKVCapacityCallSiteIsEPAware` in `cmd/ep_kv_capacity_test.go` uses Go source code string matching (`CalculateKVBlocks(`, `WithExpertParallelSize(`) to enforce call-site compliance. Is this static AST/string check resilient against formatting/refactoring changes, and does it guarantee that the option passed at runtime evaluates to the expected group size? | The string check is not fully refactor-resilient and does not by itself prove runtime semantics. Its fixed byte window and textual source requirements can produce false positives or failures after harmless formatting or helper extraction. Runtime behavior is instead covered by `TestResolveLatencyConfig_EPRaisesAutoKVCapacity`, while direct formula and capacity tests verify the expected group-size arithmetic. |
| G10 | tests | ✅ CONFIDENT | Is there a companion invariant test verifying that when `EP == TP` (e.g., `DP=1` or `enableExpertParallel=false`), `CalculateKVBlocks` produces output that is bit-identical to the pre-#1660 code path across all supported model architectures (dense, MoE, hybrid attention) to validate INV-6? | There is a direct invariant test proving the absent option, explicit 0, explicit 1, and `EP==TP` all produce exactly the same block count for a MoE fixture. Dense and degenerate sub-threshold expert models are also tested as EP-inert. Hybrid attention is not separately crossed with the `EP==TP` no-op case, but the implementation guard returns the original expression whenever `expertShardSize <= TP`, independent of attention architecture. |
| G11 | rationale | ✅ CONFIDENT | The PR implements expert-parallel weight sharding by scaling routed-expert bytes by `TP / expertShardSize` in `computeModelWeightBytes`. Why was `WithExpertParallelSize` chosen as a variadic `KVCapacityOption` seam rather than incorporating EP directly into `KVCapacityParams` or `ModelHardwareConfig`? | The option seam preserves all existing positional callers and mirrors the existing adapter-reservation option. EP size is deployment topology, not an intrinsic model-architecture parameter, so it does not belong in `KVCapacityParams`. Passing a `ModelHardwareConfig` would also be unsafe for CLI sizing because DP-as-placement rewrites each replica to `DP=1`, losing the logical `TP*DP` group needed by the pre-placement calculation. |
| G12 | rationale | ✅ CONFIDENT | When `ep > num_routed_experts`, the PR clamps `expertShardSize` to `num_routed_experts` with a warning instead of returning an error. What was the rationale for auto-clamping here rather than failing hard, given that BLIS design rules generally favor strict validation over automatic adjustment? | The rationale is conservative most-loaded-rank sizing. If the EP group is wider than the expert count, loaded ranks still hold a whole expert; using the continuous `experts/EP` average would understate memory. Clamping to the expert count therefore avoids granting nonexistent KV capacity. It warns rather than silently adjusting, and avoids rejecting a planning topology whose effective weight footprint is well-defined. |
| G13 | scope | ✅ CONFIDENT | The PR updates `CLAUDE.md` and multiple doc files (`docs/concepts/glossary.md`, `docs/guide/latency-models.md`, `docs/reference/models.md`) with extensive notes on step-time models, vLLM baseline differences, and future issues (#1548, #1666). Are any of these documentation updates describing unmerged step-time behavior beyond the scope of weight-footprint capacity sizing? | The documentation discusses existing step-time behavior and explicitly separates it from this PR's capacity change; it does not implement or claim that the pending #1548 EP-mode step-time toggle has landed. References to #1548 and #1666 identify limitations and follow-up work rather than new runtime behavior in this PR. |
| G14 | scope | ✅ CONFIDENT | The PR introduces `formatEPForLog` in `cmd/root.go` and alters the `--latency-model` auto-calc log output format to include `EP=%s`. Does this change impact any downstream log parsers or existing CLI golden/integration tests that expect the legacy log line format? | No in-repository parser or golden assertion depends on the old global auto-capacity log format. The only exact occurrence is the new log statement; another occurrence is a test comment, not an assertion. Both `./cmd` and `./sim/latency` test suites pass. |
| G15 | errors | ✅ CONFIDENT | If an out-of-bounds EP group size (`ep > tp*dp`) is passed, `CalculateKVBlocks` returns a formatted error. Are all CLI call sites in `cmd/root.go` and `cmd/replay.go` (including prefill and decode pool paths) handling this error uniformly with appropriate user feedback rather than panicking or swallowing the error? | Yes. The global resolver and all four run/replay prefill/decode call sites check the returned error and terminate through contextual `logrus.Fatalf` messages. None panic or silently ignore an out-of-bounds EP error. |
| G16 | errors | ✅ CONFIDENT | In `epSizeForKVCapacity`, dense models return `1` for EP size even if `--enable-expert-parallel` is set. If a user specifies `--enable-expert-parallel` for a dense model, does `resolveLatencyConfig` reject the invalid flag combination before or after `CalculateKVBlocks` runs, and could `CalculateKVBlocks` running first mask the invalid CLI configuration? | The dense-model EP rejection occurs after auto-capacity calculation. However, `epSizeForKVCapacity` returns 1 for dense models, so the new EP option is a no-op and cannot itself make that calculation fail or alter capacity. A separate pre-existing model/capacity error could still be reported before the more specific invalid-EP diagnostic. |

### Items to fix

### Important to consider
- **F1** — The documentation expansion is large, but it mostly records important scope and approximation boundaries rather than adding unrelated runtime behavior.
  _sim/latency/kv_capacity.go:396-409,426-428,531-550,583-638,660-671; cmd/root.go:795-808,1752-1756,1789-1793; cmd/replay.go:474-480,512-517; sim/cluster/kv_autocalc.go:71-87; cmd/root.go:468-493_
- **F2** — The latency guide's statement that vLLM EP-off uses `numExperts/TP` conflicts with the new `WithExpertParallelSize` commentary and CLAUDE.md, which say vLLM flattens TP across DP unconditionally. This documentation inconsistency should be reconciled.
  _docs/concepts/glossary.md:188; docs/concepts/roofline.md:7; docs/guide/latency-models.md:178-188,221; docs/reference/models.md:18-40; CLAUDE.md:648_
- **F3** — The existing `ExpertLoad.PerGPUExpertCount` comment still calls the step-time representation EP-mode-agnostic; that accurately describes the current implementation, although the surrounding documentation now presents conflicting claims about vLLM's EP-off placement.
  _sim/latency/kv_capacity.go:310-339,396-405,531-550; sim/cluster/kv_autocalc.go:71-77; cmd/root.go:443-493_
- **G3** — The CLI currently rejects the only topology where the capacity reduction is observable (`DP>1 + EP`), so pool-specific EP reduction is primarily forward wiring until #1548 lands.
  _cmd/root.go:418-431,518-529,795-808,1734-1756,1772-1793; cmd/replay.go:286-300,458-480,496-517; sim/config.go:329-367; cmd/ep_kv_capacity_test.go:18-48_
- **G4** — Adapter reservation and EP are individually tested, but there is no dedicated cross-product test combining a nonzero adapter reservation with active `EP>TP`; this is an edge-case coverage opportunity rather than evidence of incorrect accounting.
  _sim/latency/kv_capacity.go:426-462,506-528; docs/contributing/standards/invariants.md:41-45_
- **G5** — Returning the effective shard size or accepting an injected logger would make this library behavior easier to observe programmatically, but that is a design preference, not a correctness defect.
  _docs/contributing/standards/rules.md:24-34,90-98; sim/latency/kv_capacity.go:293-305; sim/latency/config.go:163,608_
- **G7** — The documented follow-up threading is required before removing either placement guard; otherwise the disparity described in the question would become real.
  _cmd/root.go:443-493,2265-2277; sim/cluster/kv_autocalc.go:71-82_
- **G8** — Because capacity calculation precedes backend validation, a capacity error can be reported before the roofline unsupported-flag error; this ordering is documented and pre-existing in the shared resolver.
  _cmd/root.go:719-720,777-808,876-890_
- **G9** — An AST-based check would be more robust, and the behavioral companion exercises the shared global resolver rather than every individual prefill/decode call site.
  _cmd/ep_kv_capacity_test.go:50-145,237-286; sim/config_test.go:818-879; sim/latency/kv_capacity_ep_test.go:77-136_
- **G10** — A table-driven architecture matrix including hybrid MLA would strengthen owed coverage, but it would exercise the same guarded no-op branch rather than a distinct new behavior.
  _sim/latency/kv_capacity_ep_test.go:51-75,185-225,339-378; sim/latency/kv_capacity.go:662-671_
- **G12** — Non-divisible groups below the expert count still use a continuous average and are documented as optimistic rather than modeled by exact rank assignment.
  _sim/latency/kv_capacity.go:293-305; sim/latency/kv_capacity_ep_test.go:290-338_
- **G13** — The vLLM EP-off baseline description is internally inconsistent between the latency guide and the new source/CLAUDE commentary and should be corrected.
  _docs/guide/latency-models.md:178-188,221; docs/concepts/glossary.md:188; docs/reference/models.md:30-40; sim/expert_placement.go:24-43_
- **G14** — External consumers parsing human-readable logs could need adjustment, but the repository provides no evidence of such a supported contract.
  _cmd/root.go:805-808; sim/latency/kv_capacity_test.go:1992; reproducible: `go test ./cmd -count=1` and `go test ./sim/latency -count=1` both exit 0_
- **G16** — Moving unsupported flag validation before auto-sizing would improve diagnostic priority, but the ordering predates this change and the PR does not worsen it.
  _cmd/root.go:719-808,876-908; cmd/root.go:418-431; sim/config.go:353-362; sim/latency/kv_capacity.go:662-671_

namasl:
@claude /blis-pr-review

Final head `3c5e1535`. The ARCHON review immediately above is unchanged for the fourth round: **`ARCHITECTURAL_CHANGE`, 0 violations, 0 edge changes, 2 intentional surface additions, 3 packages with added invariants**.

**The delta since your READY TO MERGE on `00776a83` is COMMENT-ONLY** — no behavior, no tests, no docs. `git diff 00776a83..3c5e1535` touches exactly one file (`sim/latency/kv_capacity.go`) and adds only doc-comment lines. It moves the clamp-vs-reject reasoning you validated out of this PR thread and into the code, because a PR comment is not durable documentation and the next reader of `resolveExpertShardSize` should find it at the call site:

- The function doc now states that the resolved value is a **divisor** of the routed-expert bytes (`charge = R·tp/shard`), so a larger resolved value charges *less* memory and a smaller one charges *more* — and then names the two directions explicitly: `ep > tp·dp` is **rejected and deliberately not clamped** (no safe substitute exists for "more EP ranks than GPUs"; that bound is load-bearing and was never relaxed), while `ep > num_routed_experts` is **clamped down**, which shrinks the divisor and therefore charges *more* memory — the conservative direction. It closes by noting both directions land at or above the true footprint, so neither can inflate capacity.
- The clamp site records why rejecting was abandoned (it fired on the inert `ep == tp` case and masked the #1548 diagnostic) and that vLLM does not reject such a group either — only `--enable-eplb` requires an even distribution.

Also since your review: the **#1666** finding is now linked from the PR body itself (not only a code comment), along with the clamp rationale and the three-commit revision history, so the record survives the squash.

**qa-review status:** the cross-vendor pass returned ✅ **PASS**, 19/19 CONFIDENT, **no items to fix**. Its two non-blocking notes were both acted on in `00776a83` (the last vLLM EP-off contradiction in `docs/guide/latency-models.md`, which it flagged twice, and a missing adapter-reservation × active-EP cross-product test). The `--adjudicate` pass reports "No prior blocking findings to adjudicate" — vacuous by design, since the fresh pass had none.

**Gates on this exact head:** CI green (build ×2, lint, all 7 test shards), `golangci-lint run ./...` 0 issues, and the 8-configuration INV-6 stdout+exit-code diff against a binary built from `main` still byte-identical.

One small correction for the record: `TestEffectiveMoEGroupSize_EPModeIndependent` (`sim/config_test.go:181`), which you cited as new, is pre-existing — your reading of what it locks in is right, it just predates this PR. Nothing to change.

susiejojo (approved):
lgtm

PR #1664, opened 2026-09-03
feat(cmd): DP-as-placement parity for blis replay (MoE --dp > 1) (Closes #1556)

### LINKED ISSUE #1556: DP-as-placement parity for blis replay (--dp > 1 on MoE)

namasl:
## Summary

Follow-up to #1531. #1531 shipped DP-as-real-placement as a **`blis run`-only** feature and made `blis replay` `logrus.Fatalf` on MoE `--dp > 1`: replay kept the lumped single-instance DP-math model, so the two paths would have diverged for identical flags (INV-13). As #1556 observed, the three placement steps (expand `numInstances`, divide the auto-KV total to per-rank, set `DP=1`) are **mechanically identical** in the replay command body — replay already supports multi-instance and PD pools — so the parity was a deferred feature, not a fundamental limitation.

This PR lifts the guard and makes the parity **structural** rather than comment-enforced.

### Design: one shared code path, not a second copy

Duplicating the ~50-line runCmd DP block into replayCmd would recreate exactly the drift smell that `resolveLatencyConfig` was created to remove (R23). Instead, the decision + application + diagnostics move into one shared `cmd.resolveDPPlacement`, and — like `resolveLatencyConfig` and `resolvePolicies` — it reads and writes the package-level flag vars itself, so **neither command body carries any wiring that could drift**. Both call sites are now one line plus the `Fatalf`.

Layering (each piece independently testable):

| function | role | purity |
|---|---|---|
| `planDPPlacement` | the decision + the unsupported-combo guards | pure |
| `applyDPPlacement` | count expansion, per-rank KV division, `--max-model-len` re-cap, zero-block guard | pure |
| `resolveDPPlacement` | reads/writes the `cmd` deployment vars, emits operator diagnostics | shared by `run` + `replay` |

`replay` now threads `dpPlan.PerRankDP` into `NewModelHardwareConfig`, identically to `run`. `sim/` and `sim/cluster/` are unchanged.

## Behavioral contracts

- **BC-1 (replay expands):** GIVEN an MoE model and `--dp N` (N>1) with no EP / PD / autoscaler / node pools, WHEN `blis replay` runs, THEN the cluster contains `--num-instances × N` engine replicas, each configured per-rank (`DP=1`).
- **BC-2 (run/replay parity, INV-13):** GIVEN a trace exported by `blis run --dp N` (MoE) and the same flags, WHEN it is replayed, THEN stdout is byte-identical to the run's stdout.
- **BC-3 (per-rank KV, no dp² double-count):** GIVEN the auto-KV path, WHEN DP-as-placement is active, THEN each replica holds the per-rank budget (aggregate over the N replicas equals the pre-expansion lumped total) and `--max-model-len` is re-capped to the per-rank KV-feasible maximum — no per-replica `NewSimulator` panic.
- **BC-4 (guards preserved):** GIVEN MoE `--dp N` plus `--enable-expert-parallel` (or PD flags), WHEN `blis replay` runs, THEN it exits 1 naming #1548 / #1553 — never a silently mis-modeled run.
- **BC-5 (INV-6 no-op):** GIVEN `--dp 1` or a dense model, THEN nothing is mutated and stdout is byte-identical to the pre-change binary, on both commands.
- **BC-6 (INV-1):** GIVEN an expanded DP replay, THEN request conservation holds across the replicas.

## Hardening found while reviewing

- **Zero-block guard.** The per-rank division floors, so a `--dp` larger than the auto-derived block count would leave a 0-block replica — a downstream `NewSimulator` panic, and a derived `kvFeasibleMax` of `0` that would silently mean *unlimited* (the inverse of a cap). `applyDPPlacement` now returns a clean CLI error naming the remedies (R1). Defense in depth: the guarantee that the auto path cannot produce this lives in another package.
- **Diagnostic parity.** `runCmd`'s `Starting simulation with N KV blocks` log ran *before* the DP block (reporting the pre-division aggregate) while replay's ran *after* (per-rank), so the two commands printed different block counts for a config this PR advertises as identical. run's log moved below the DP block. stderr-only; stdout unaffected.
- **`autoscalerActive` on replay** now folds in the `--model-autoscaler-interval-us` CLI override, matching what `run` folds into `bundleAutoscalerIntervalUs`, so the #1553 guard stays correct if replay ever gains autoscaler support.

## Testing verification

`go build ./...` clean; `go test ./... -count=1` all packages pass; `golangci-lint run ./...` → **0 issues**.

New/changed tests (all in `cmd/`):

- `TestINV13_RunReplayParity_MoEDPPlacement` (**BC-2**) — CLI-level `blis run --dp 2 --trace-output` → `blis replay --dp 2`, asserting byte-identical stdout, with a non-vacuity gate (`completed_requests > 0`) and an INV-1 companion (R7). Runs at **TP=1 and TP=2**, so a hardcoded TP argument at either constructor site is caught behaviorally rather than only by a source grep.
- `TestReplayCmd_MoEDPPlacement_SpawnsReplicas` (**BC-1/BC-6**) — `1×2 → 2` and `2×2 → 4` replicas, exactly (no `instance_2` / `instance_4`), INV-1, plus repeat-run INV-6. Replaces the #1531-era `TestReplayCmd_MoEDPPlacement_Rejected`.
- `TestReplayCmd_MoEDPPlacement_AutoKV_Parity` (**BC-3**) — the fixture is nowhere near KV-bound, so stdout parity alone would be *vacuous* here (verified: a replay-only skipped division passed a stdout-only version of this test). It therefore asserts the numbers the resolver reports: `per-rank × dp == the auto-derived aggregate` (a skipped division or a dp² double-count fails), and `re-cap == per-rank blocks × block size` (capping to the aggregate instead panics per-replica), on **both** legs, plus run==replay on the derived per-rank budget.
- `TestReplayCmd_MoEDPPlacement_ForeignTrace_Mixtral` (**BC-1**) — restores two things the deleted test carried: a **second MoE architecture family** (synthetic non-MLA `num_local_experts` Mixtral, vs the MLA `n_routed_experts` deepseek fixture used elsewhere) and a **hand-authored TraceV2** this binary did not generate (the realistic operator flow).
- `TestReplayCmd_MoEDP1_ByteIdentical` (**BC-5**) — the INV-6 fence on the newly-supported path; `--dp 1` previously had a run-side fence only.
- `TestResolveDPPlacement_MutatesDeploymentVars` / `..._DenseModelIsInert` — the shared resolver's contract: what it writes to `numInstances`/`totalKVBlocks`/`maxModelLen`, and that **every guard error leaves all three untouched**.
- `TestApplyDPPlacement` — rewritten as a table over the new signature, covering the re-cap, the explicit-KV branch, the zero-block error, and the aggregate-conservation law.
- `TestRunReplay_DPEPFlags_ThreadedIntoConstructor` — updated: both files now wire `dpPlan.PerRankDP`.
- `TestReplayCmd_MoEDPPlacement_GuardedCombos_Rejected` (**BC-4**) — EP → #1548, PD → #1553, exit 1 each.

**Mutation-tested** (not just green): re-adding the #1531 guard, dropping the assign-back, wiring raw `dataParallelism`, warning instead of fataling on a guard error, and neutering the per-rank division each fail at least one test. INV-6 was verified by diffing against a **pre-change binary**: MoE `--dp 1` replay, dense run, and dense replay stdout are byte-identical.

Manual end-to-end parity (byte-identical stdout, run vs replay): `dp=2 ni=1`, `dp=4 ni=2`, `dp=3 ni=1`, weighted routing, the auto-KV path with a huge `--max-model-len`, and `--kv-offload-config` + `--flow-control`.

## Notes and boundaries

- **Match `--horizon` on both legs.** `blis run` defaults it to unlimited while `blis replay` auto-computes 2x the max arrival time — a pre-existing difference in *defaults* (orthogonal to DP) that would otherwise truncate the replay leg. INV-13's canonical statement now says so explicitly, and the replay guide documents it with a worked example.
- **`--dp` is re-supplied on the replay CLI**, like `--tp` / `--kv-cache-dtype`; the trace header is not consulted for it.
- A multi-turn/accumulate spec's `run` → *fixed* `replay` metrics differ, but identically at `--dp 1` on the pre-change binary — pre-existing closed-loop re-export behavior (#1621/#1630 territory), not DP-induced.
- Still deferred: EP-on placement (#1548) and DP + PD/autoscaler/node pools (#1553), both fail fast on **both** commands now.

### Documentation

Corrected the stale "Data parallelism (DP) and expert parallelism (EP) are not yet modeled" claims in `docs/reference/models.md` and `docs/guide/latency-models.md` (untrue since #1419/#1531, and they directly contradicted this PR's own claims), added the missing `--dp` / `--enable-expert-parallel` / `--moe-comm-backend` rows to the configuration reference, and updated CLAUDE.md, INV-13, and the glossary.

Closes #1556


namasl:
@claude /blis-pr-review

The deterministic ARCHON PR review above returned **`NO_CHANGE`** for `9af90b71 → 6dd2b427`: no packages added/removed, no import edges added/removed, zero surface changes, zero schema changes, zero violations — internal-only, fast-track eligible at package altitude. `sim/` and `sim/cluster/` are untouched; the whole change lives in `cmd/`.

Archon flagged exactly one guarded promise moving inside the existing boundary: the INV-13 invariant test set in `cmd/`. That is the intent of the PR, and it is worth your attention as the substantive part of the review:

- **Added:** `TestINV13_RunReplayParity_MoEDPPlacement` (run→replay byte-identical stdout at TP=1 and TP=2), `TestReplayCmd_MoEDPPlacement_SpawnsReplicas`, `TestReplayCmd_MoEDPPlacement_AutoKV_Parity`, `TestReplayCmd_MoEDPPlacement_ForeignTrace_Mixtral`, `TestReplayCmd_MoEDPPlacement_GuardedCombos_Rejected`, `TestReplayCmd_MoEDP1_ByteIdentical`, `TestResolveDPPlacement_MutatesDeploymentVars`, `TestResolveDPPlacement_DenseModelIsInert`.
- **Removed:** `TestReplayCmd_MoEDPPlacement_Rejected` — this asserted the #1531 run-only `logrus.Fatalf` that #1556 exists to lift, so it is obsolete by design. Its two unique fixtures were deliberately preserved elsewhere: the non-MLA Mixtral MoE config and the hand-authored (non-`blis`-generated) TraceV2 both live on in `TestReplayCmd_MoEDPPlacement_ForeignTrace_Mixtral`.
- **Modified:** `TestApplyDPPlacement` (rewritten as a table over the new signature, which absorbed the max-model-len re-cap and a new zero-block guard), `TestRunReplay_DPEPFlags_ThreadedIntoConstructor` (both `root.go` and `replay.go` now wire `dpPlan.PerRankDP`, where before they intentionally differed).

Specific things I would like challenged:

1. **INV-13.** Is byte-identical stdout over the full metrics dump the right statement of the parity contract here, and is the `--horizon` precondition I added to INV-13's canonical statement in `invariants.md` correct — i.e. is the run/replay horizon-default difference genuinely a difference in *defaults* rather than in behavior?
2. **The R23 shape.** `resolveDPPlacement` reads and writes the `cmd` package-level flag vars itself (following `resolveLatencyConfig` / `resolvePolicies`) precisely so neither command body carries wiring that could drift. Is that the right trade against the reduced unit-testability, given `planDPPlacement` and `applyDPPlacement` stay pure?
3. **Placement of the call in `replay.go`** — after PD-topology / autoscaler / node-pool validation, before the `DeploymentConfig` literal. Are there uses of `numInstances` / `totalKVBlocks` / `maxModelLen` on either side of that point that want the other value?
4. **INV-1 / INV-6.** Conservation across the expanded replicas, and the claim that `--dp 1` and dense models are byte-identical to the pre-change binary (verified by diffing against a pre-change build, but please check the reasoning).
5. **The guards.** #1548 (EP-on) and #1553 (PD / autoscaler / node pools) must still fail fast on *both* commands — these are unmodeled physics, not merely unreplayable features.

All CI shards pass, including `test (cluster (heavy), ./sim/cluster, 15m)`; `golangci-lint run ./...` reports 0 issues.

namasl:
@claude /blis-pr-review

Second cycle, head `a92e59ae`. The deterministic ARCHON review above returned **`NO_CHANGE`** again for `9af90b71 → a92e59ae` (internal-only, fast-track eligible, 0 violations, `cmd` the only changed package). All 10 CI shards are green on this head, including `test (cluster (heavy), ./sim/cluster)`; `golangci-lint run ./...` → 0 issues.

This push addresses all three non-blocking findings from your first review plus the test gap you flagged. Two of them are worth your scrutiny specifically, so I am volunteering both rather than letting them be found:

**1. The INV-13 wording change touches a source-of-truth file.** You were right that `invariants.md:178` understated the contract relative to what the tests enforce. I rewrote the opening to state the stronger CLI-boundary form — byte-identical stdout, which subsumes the per-request distributions, the cache/latency aggregates and the conservation counters — and to say it needs a non-vacuity gate plus a conservation companion, since byte-identity alone is satisfied by two identically-wrong runs. Because `docs/contributing/standards/invariants.md` is canonical per the source-of-truth map in `principles.md`, I also updated the working copy of that text in `CLAUDE.md` (self-audit dimension 10); that bullet additionally picks up the `--horizon` precondition, which it was missing. **Please check the two are actually consistent and that I have not over-claimed** — INV-13 governs far more than DP-as-placement, so a wording change there has blast radius beyond this PR, and "assert byte-identical stdout" is guidance for *new* parity tests, not a retroactive claim about every existing one.

**2. My first version of the new `--dp 1` auto-KV assertion was vacuous, and mutation testing is what caught it.** The gap you identified was real: `TestReplayCmd_MoEDP1_ByteIdentical` pinned `--total-kv-blocks`, so auto-KV × `--dp 1` × replay — the INV-6 byte-identity path that must never regress — was unfenced, and `AutoKV_Parity` only reaches `--dp 2`. It is now a two-row table over the explicit-KV and auto-KV states. Both rows also assert that **no DP-as-placement diagnostic is emitted at all**, because at `--dp 1` the division, the re-cap and the expansion are each `×1` or no-ops — so an inactive-plan early return that regressed into an *active* plan would be byte-identical on stdout, and the diagnostic is the only thing separating "looks the same" from "took the early return". I wrote that assertion first at the default log level, where it silently passed under the mutation that relaxes `planDPPlacement`'s `dp <= 1` gate, because the activation line is a `logrus.Infof` and the default level is `warn`. The rows therefore now pass `--log info`, and the mutation fails both. **Worth confirming that `--log info` does not perturb stdout** (logrus writes to stderr, and the byte-identity comparison is stdout-only) — that is the assumption the row rests on.

**3.** `replay.go` / `root.go`: the placement comment no longer claims the `DeploymentConfig` literal is the boundary for all three adjusted quantities. Both now name the per-pool KV auto-calc as an earlier reader of `maxModelLen`, say it is safe *only* because PD + `--dp>1` is guarded at #1553, and say what to do if #1553 is ever lifted.

**4.** Nit fixed: `wantPerRankDP` is documented as unchecked on the error rows, so their Go zero value is not an expectation. Also added the comment you asked for on why the PD guard row needs `--num-instances 2` (`ValidatePoolTopology` runs before the DP guard) while the EP row does not.

I did not act on one observation: you noted `TestResolveDPPlacement_MutatesDeploymentVars`'s no-`t.Parallel()` requirement is comment-enforced only. I left it as-is because that is the existing convention across this package (`captureCmdLevelVars`/`restore` with a comment), and introducing a serial lock for one test would diverge from ~a dozen sibling tests. Say the word if you would rather it be enforced.

namasl:
@claude /blis-pr-review

Third cycle, head `c412386b`. ARCHON above: **`NO_CHANGE`** for `9af90b71 → c412386b` (internal-only, fast-track eligible, 0 violations, `cmd` the only changed package). All 10 CI shards green on this head including `test (cluster (heavy), ./sim/cluster)`; `golangci-lint run ./...` → 0 issues.

**Your cycle-2 must-fix is applied.** `invariants.md:180` named `TestResolveDPPlacement`, which matches zero tests — I confirmed with `go test ./cmd/ -run 'TestResolveDPPlacement$'` ("testing: warning: no tests to run") *before* fixing rather than taking it on faith. It went stale because the function was split in two when I reshaped the resolver to the package-var form mid-cycle-1. Now names both real functions, `TestResolveDPPlacement_MutatesDeploymentVars` and `TestResolveDPPlacement_DenseModelIsInert`, and says what each pins. CLAUDE.md carries **no** working copy of that Verification paragraph (checked, rather than assumed — it mirrors only the Statement), so there was nothing to mirror. I then swept every `` `Test*` `` name in `invariants.md`, and repo-wide across all docs, against `func <name>(` in the test files: no other stale reference.

**All four nits taken**, including the two you left to my judgment:

- (B) The `9223372036854775807` literal is annotated as 2^63−1 / unlimited / run's own default, with a note that any value works so long as both legs share it.
- (C) `dpParityHorizon` documents the coupling to run's default `simulationHorizon`, and what would need re-checking if that default moved.
- (D) **Taken.** `dpLegOK` now separates the harness's own failure mode (exit 3 = cobra rejected the arg list *this file* built) from a simulator `logrus.Fatalf` (exit 1), with the magic 3 named `dpLegCobraErrorExit`. Verified by injecting a bogus flag into `dpMoEFixtureArgs`: the "test harness bug, not a simulator failure" message fires.
- (A) **Taken in a variant I think is better.** Rather than adding `--log info` to the parity legs, `TestINV13_RunReplayParity_MoEDPPlacement` now reports the replay leg's **stderr** on a divergence. Same diagnostic value — the capacity and re-cap warnings are already warn-level — without adding info-level noise to a test whose entire assertion is byte-identity.

I did **not** act on your observation that `TestResolveDPPlacement_MutatesDeploymentVars`'s no-`t.Parallel()` requirement is comment-enforced only: that is the existing convention across this package (`captureCmdLevelVars`/`restore` plus a comment), and adding a serial lock for one test would diverge from ~a dozen siblings. Happy to change it if you disagree.

**New since your last review — a pre-emptive doc-truthfulness sweep.** I swept every doc making a claim about DP, `--dp`, data parallelism, or run/replay parity for statements this PR falsifies, and found two beyond the two I had already corrected. Both are the class where a doc describes behavior the code would reject, or contradicts its own body:

1. **`docs/concepts/roofline.md:10`** said the roofline model "does not model data parallelism (DP) or expert parallelism (EP) scheduling overhead". True, but it understates: roofline does not silently *ignore* them, it **hard-rejects** them — I verified `--latency-model roofline --dp 2` on an MoE model exits 1 with "The roofline backend is DP/EP-blind for step time". As written a reader could conclude a roofline `--dp 2` run silently drops DP. Now says rejected, and notes DP-as-placement is trained-physics-only for the same reason.
2. **`docs/guide/latency-models.md:133`** (β_EP) said "Active only for MoE models with `--dp > 1`", which reads as if the dispatch/combine term fires whenever the CLI gets `--dp 2`. Under DP-as-placement every replica is configured at `DP=1`, so it is inert — which the DP-as-placement note 45 lines *below* already stated, making the two mutually contradictory read in isolation. Now distinguishes a **config** `DP > 1` from a **CLI** `--dp > 1` and points at that note.

**Please sanity-check both**, since they are the delta since your last pass and neither is covered by a test: (1) is the roofline rejection I quoted actually unconditional, or is there a path where roofline + `--dp > 1` proceeds? (2) is "config `DP > 1`" the right way to name the β_EP gate, given the gate is on `ModelHardwareConfig.EffectiveDP()` rather than on the flag?

Docs I checked and deliberately left alone as still accurate: `glossary.md:191`, `models.md:21`, `latency-models.md:224`, `configuration.md:149-150` (all corrected earlier in this PR), and `observe-replay-calibrate.md:508` / `CLAUDE.md:180` (both about `--num-gpus` and what the trace header records — unrelated to placement).

namasl:
@claude /blis-pr-review

Fourth cycle, head `7ce79647`. ARCHON above: **`NO_CHANGE`** for `9af90b71 → 7ce79647` (internal-only, fast-track eligible, 0 violations). All 10 CI shards green including `test (cluster (heavy), ./sim/cluster)`; `go test ./... -count=1` all packages pass locally; `golangci-lint run ./...` → 0 issues.

**Your cycle-3 must-fix was real and is fixed.** I ran the example verbatim before touching it: it exits 1 with `--latency-model trained-physics requires --hardware (GPU type) and --tp (tensor parallelism). No defaults found in defaults.yaml for model=deepseek-ai/deepseek-v2-lite`. You were right that it dies before any DP logic. I fixed it properly rather than minimally — the suggested `--hardware H100 --tp 1` alone is not sufficient, since the fixture also needs `--model-config-folder` and `--hardware-config` to resolve offline — and then verified the corrected commands end to end: run exit 0, replay exit 0.

**All four nits applied, each verified against the code first rather than taken on faith:**

- **β_EP bullet** — now names `ModelHardwareConfig.DP > 1`, states it is unreachable from the CLI under DP-as-placement and reachable only by constructing a config directly (e.g. tests), and carries the `TP > 1` condition on the all-reduce clause, noting that at `TP == 1` there is no collective at all.
- **Parity activation check** — confirmed absent, and I agree this is more than a nit: it is the same vacuity class this PR has now hit three times. Byte-identical stdout is equally satisfied by two legs that BOTH failed to expand, so a regression making `planDPPlacement` inactive for every MoE `--dp>1` config would have passed silently. Both legs must now show `instance_1`. **Mutation-verified**: forcing the inactive branch (`if true || !isMoE || dp <= 1`) fails the check, where before the mutation passed.
- **Mixtral foreign-trace** — literal `os.Exit(3)` at line 605 confirmed, replaced with `dpLegCobraErrorExit`, and given the same harness-bug-vs-simulator-failure separation as `dpLegOK`.
- **Guard assertions** — confirmed bare `"1548"`/`"1553"`, and confirmed production writes `#1548`/`#1553`. Tightened to the hashed form so a message merely containing the digits cannot satisfy them.

Also folded in one item I had been holding: `configuration.md` now says the TraceV2 header has **no** `data_parallel` field at all — verified, it carries `tensor_parallel` only and replay reads neither back — which is stronger and more actionable than the previous "not round-tripped through the trace header".

**Two things I would like you to check specifically**, since both are new text with no test behind them:

1. The corrected example in `observe-replay-calibrate.md` now hardcodes `--model-config-folder model_configs/deepseek-v2-lite` and `--hardware-config hardware_config.json` as **repo-root-relative** paths. That is right for someone running `./blis` from a clone, which is how every other example in this guide reads — but confirm it is consistent with the surrounding examples and not misleading for an installed binary.
2. The β_EP claim "at `TP == 1` there is no collective at all" is my inference from the `m.dp == 1 && m.tp > 1` gate at `trained_physics_model.go:484` being the only MoE-FFN all-reduce branch. Please confirm there is no other collective term that fires for an MoE model at `DP == 1, TP == 1` — if there is, that clause is wrong and I would rather narrow it than ship an over-claim.

For transparency on process: I have been verifying each finding against the source before acting on it, including the ones that turned out to be correct. Both cycle-3 items above were real. If anything in this review rests on a `git diff` observation, please quote the specific hunk so I can check it against the tree.

namasl:

## qa-review — PR #1664: ✅ PASS
_feat(cmd): DP-as-placement parity for blis replay (MoE --dp > 1) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. Replay now removes the old MoE DP rejection, invokes the same placement resolver as run, expands instances, uses per-rank DP=1, and carries the adjusted KV/max-model-length values into the canonical deployment configuration. The added tests cover run/replay parity, replica expansion, auto-KV behavior, guarded combinations, determinism, and request conservation. The supporting refactor and documentation are directly related to those acceptance criteria. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. User-facing CLI help, configuration reference, replay guide, latency-model documentation, model/roofline descriptions, glossary, and the formal INV-13 text were updated to say replay supports MoE DP-as-placement and to document its guards and horizon requirements. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No material stale comments were found. The former run-only/replay-rejection descriptions were removed or rewritten, and the remaining references distinguish #1531 as the original run implementation from #1556 replay support. |
| G1 | correctness | ✅ CONFIDENT | In `applyDPPlacement`, when `autoScaledKV` is true and `in.TotalKVBlocks` is divided by `plan.Replicas`, what happens if `TotalKVBlocks % Replicas != 0`? Does integer division truncation result in total KV capacity loss across ranks? | Integer division would truncate a remainder in isolation, but the shipped auto-KV producer cannot create that remainder: it first computes an integral per-rank block count and then multiplies that exact integer by DP. `applyDPPlacement` divides that aggregate by the same CLI DP, and an active plan has `Replicas == dp`, so the original per-rank count is recovered exactly. A non-divisible value would require a hand-constructed inconsistent call. |
| G2 | correctness | ✅ CONFIDENT | When `applyDPPlacement` calculates `kvFeasibleMax = perRankKV * blockSize` and re-caps `MaxModelLen`, what happens if `MaxModelLen` was 0 (unspecified/unlimited) vs explicitly set? Could re-capping set `MaxModelLen` when it was previously unconstrained? | A zero `MaxModelLen` remains zero. Re-capping is guarded by `out.MaxModelLen > 0`, so an unspecified/unlimited value is never turned into a finite limit. A positive value—whether explicitly supplied or auto-derived—is capped only when auto-KV placement is active and the value exceeds the per-rank KV capacity. Explicit `--total-kv-blocks` disables this re-cap because `autoScaledKV` is false. |
| G3 | invariants | ✅ CONFIDENT | Under INV-13 (Run/Replay Parity), `blis run` defaults `--horizon` to `math.MaxInt64` while `blis replay` auto-computes a drain-time horizon. Why does `resolveDPPlacement` not automatically harmonize horizon defaults for replay under MoE `--dp > 1`, instead requiring users and tests to explicitly pass identical `--horizon` flags on both legs? | `resolveDPPlacement` intentionally owns only placement quantities, not workload timing. The different horizon defaults predate this PR and apply to replay generally, not specifically to DP placement. Automatically changing replay's horizon only for MoE DP would make placement resolution unexpectedly alter trace-drain semantics and create a DP-specific default. Instead parity is defined over the same resolved configuration, and tests explicitly pass the same horizon to both commands. |
| G4 | invariants | ✅ CONFIDENT | Under INV-1 (Request Conservation) and INV-4 (KV Conservation), when MoE `--dp > 1` expands logical instances into N replicas with per-rank `DP=1`, how is KV block allocation and request routing verified to be identical between run and replay when requests arrive with distinct timestamps in replay? | Both commands construct the same `DeploymentConfig` after the shared resolver, including identical replica count, per-replica KV blocks, DP=1 model configuration, routing policy, and seed. Replay schedules each trace request at its recorded `ArrivalTime`; the cluster processes arrivals in timestamp/priority/sequence order and routes them through the same router as run. The parity test compares complete stdout, checks that the second replica exists, requires completed work, and independently checks cluster request conservation. Auto-KV tests also assert `perRank * dp == aggregate`. INV-4 itself remains enforced within each replica's KV cache independently of arrival timestamps. |
| G5 | rules | ✅ CONFIDENT | Under rule R1 (no silent degradation/mutation) and R2 (canonical state construction), `resolveDPPlacement` directly mutates package-level CLI flag variables (`numInstances`, `totalKVBlocks`, `maxModelLen`). Does mutating global package variables create hidden state coupling if tests or command invocations run in sequence within the same process? | The mutation is explicit and bounded: the function documents every package variable it reads and the only three it writes, performs planning and arithmetic before committing them, and mutates nothing on an error or inactive plan. Both command bodies immediately consume those resolved values to construct the canonical configuration. Normal CLI processes execute one command and exit; tests that call the resolver in-process capture and restore the globals and do not run in parallel. |
| G6 | rules | ✅ CONFIDENT | Under rule R8 / R20 (fail-fast / explicit input validation), when an explicit `--total-kv-blocks` is provided alongside MoE `--dp > 1`, `applyDPPlacement` skips division and leaves the value as per-instance. How does `resolveDPPlacement` guard against a user accidentally providing an aggregate lumped total rather than a per-instance total? | It cannot infer whether the user's numeric value was mentally intended as an aggregate. The CLI contract defines explicit `--total-kv-blocks` as a per-instance/per-replica value; the resolver preserves it and the help/documentation explain that interpretation. It still validates the value as positive upstream. Treating it as aggregate would contradict existing explicit-capacity semantics. |
| G7 | parity | ✅ CONFIDENT | The PR removes the `blis replay` startup guard for MoE `--dp > 1` by calling `resolveDPPlacement` in both `root.go` and `replay.go`. Are there any trace-header or trace-data fields loaded during replay that could cause auto-KV resolution in `resolveDPPlacement` to evaluate differently than during `blis run`? | No trace field changes DP auto-KV resolution. `resolveLatencyConfig` derives capacity from the replay CLI/defaults, model config, hardware config, TP, DP, block size, memory utilization, KV dtype, and LoRA reservation just as run does. `TraceHeader` has no DP, TP, or total-KV-block field. Replay does read header-authoritative KV-offload metadata and observed metrics, but those do not change `KVParamsOK` or the global auto-KV block calculation; run-generated offload data is reconciled separately for parity. |
| G8 | parity | ✅ CONFIDENT | `CLAUDE.md` notes that `blis observe` is unaffected as a black-box dispatcher. If a user runs `blis observe` with an MoE model and `--dp > 1`, does `observe` enforce the same placement resolution or reject `--dp > 1`, or could it silently misrepresent the cluster topology? | `blis observe` neither performs simulator placement nor accepts the shared simulation `--dp` flag. It has an independent flag set for dispatching HTTP requests to an already deployed real server, so passing `--dp` to observe is rejected by Cobra as an unknown flag rather than silently changing or representing topology. |
| G9 | tests | ✅ CONFIDENT | `TestINV13_RunReplayParity_MoEDPPlacement` re-execs the test binary in subprocesses to check for byte-identical stdout between run and replay. Is this byte-identity check vulnerable to harmless output reordering (e.g., metric key order or log timing), and how does the test distinguish between identical success output and identical failure/empty output? | Any stdout reordering would fail the byte comparison, but that is intentional because INV-6 defines deterministic output as byte-identical stdout. Wall-clock/log timing is emitted to stderr and excluded. The test prevents vacuous equality by requiring both subprocesses to exit successfully, requiring the run's cluster metrics to report completed requests, checking request conservation, and requiring `instance_1` in both outputs so two identically unexpanded runs cannot pass. |
| G10 | tests | ✅ CONFIDENT | In `TestResolveDPPlacement_MutatesDeploymentVars`, package-level state (`dataParallelism`, `enableExpertParallel`, etc.) is manually captured and restored using `defer origCmd.restore()`. If `resolveDPPlacement` panics before the defer is registered or mutates state not tracked in `dpResolveVars`, will subsequent tests be contaminated? | The restoration defers are registered before the test mutates any package state or calls `resolveDPPlacement`, and Go executes defers during normal panic unwinding. The two snapshots together cover all variables the resolver writes and all resolver-read variables that the test changes: deployment fields and block size are in `captureCmdLevelVars`, while DP, EP, pool counts, and communication backend are in `dpResolveVars`. The test is explicitly prohibited from using `t.Parallel`. |
| G11 | rationale | ✅ CONFIDENT | Why was `resolveDPPlacement` designed to mutate package-level CLI flag variables in place rather than returning an immutable structured deployment config passed into `NewModelHardwareConfig` and `NewSimulator`? | The stated rationale is to centralize all command-side wiring in one resolver so run and replay cannot drift: both commands already operate over package-level Cobra-bound variables, and the resolver commits the three adjusted values once after pure planning and arithmetic succeed. The final `DeploymentConfig` and `NewModelHardwareConfig` are still constructed canonically with those resolved values rather than being patched afterward. Pure `planDPPlacement` and `applyDPPlacement` preserve isolated unit-testability. |
| G12 | rationale | ✅ CONFIDENT | Why was DP-as-placement parity for replay implemented by expanding instances in `cmd/` rather than pushing instance expansion down into `sim/cluster` so all CLI commands automatically inherit uniform multi-instance semantics? | DP-as-placement is CLI interpretation of a model flag, while `sim/cluster` already has uniform multi-instance semantics through `DeploymentConfig.NumInstances`. Resolving it in `cmd/` converts `--dp N` into the existing generic cluster inputs and also handles CLI-specific auto-KV and max-model-length adjustment before construction. Pushing this into cluster would couple the generic simulator to MoE CLI semantics and require constructing then overriding per-instance DP/KV state. |
| G13 | scope | ✅ CONFIDENT | Does `resolveDPPlacement` introduce any extra re-capping or validation logic for dense models or single-instance configs that goes beyond the acceptance criteria of issue #1556 (MoE `--dp > 1` replay parity)? | No. For dense models or `dp <= 1`, planning returns an inactive identity plan and `resolveDPPlacement` returns before applying any re-cap, validation, logging, or mutation. Dense `--dp > 1` is rejected earlier by the existing latency configuration validation. |
| G14 | scope | ✅ CONFIDENT | The issue acceptance criteria specifically require lifting the replay guard and restoring INV-13 parity for MoE `--dp > 1`. Does the new auto-KV `--max-model-len` re-capping logic in `applyDPPlacement` modify behavior for existing `blis run` workflows, or is it strictly constrained to the DP placement path? | It is strictly gated to an active MoE DP-placement plan and successful auto-KV calculation. The run path already performed the same per-rank max-model-length re-cap before this PR; the change moves that behavior into the shared helper so replay receives identical treatment. `dp=1`, dense models, and explicit-KV configurations do not gain a re-cap. |
| G15 | errors | ✅ CONFIDENT | If `applyDPPlacement` returns an error (e.g. KV division results in 0 blocks), how do `blis run` and `blis replay` handle the error? Do both paths exit immediately via `logrus.Fatalf` before any simulator instantiation occurs? | Yes. `applyDPPlacement` returns the original deployment unchanged on failure; `resolveDPPlacement` propagates the error without mutating globals; and both run and replay immediately call `logrus.Fatalf` before logging startup or constructing `DeploymentConfig`/`NewClusterSimulator`. |
| G16 | errors | ✅ CONFIDENT | When `resolveDPPlacement` rejects guarded combinations (such as `--enable-expert-parallel` with `--dp > 1`), does `blis replay` produce the exact same error log, issue reference (#1548 / #1553), and exit code (1) as `blis run`? | For shared DP-placement guards such as EP-on and valid PD topologies, yes: both commands call the same `planDPPlacement` through `resolveDPPlacement`, format the returned error with the same `logrus.Fatalf("%v", err)`, and therefore emit the same substantive message and tracking reference and exit with code 1. Replay tests verify exit 1 and #1548/#1553 references. |

### Items to fix

### Important to consider
- **G1** — The pure helper does not reject a non-divisible hand-authored input, but production auto-KV values satisfy the divisibility invariant.
  _sim/latency/kv_capacity.go:370-387 computes `totalBlocks` per rank then multiplies by `dp`; cmd/root.go:428-453 sets active `Replicas: dp`; cmd/root.go:478-486 divides the aggregate by `dp`_
- **G3** — The differing run/replay horizon defaults remain a pre-existing usability sharp edge for all parity comparisons, not a regression introduced by DP replay.
  _cmd/root.go:529-534 lists the resolver's state and side effects; cmd/replay.go:250-254 implements replay horizon behavior before placement resolution; cmd/dp_replay_parity_test.go:70-81 defines the common parity horizon; docs/contributing/standards/invariants.md:175-181 documents the resolved-horizon requirement_
- **G4** — The new CLI tests check capacity conservation and observable cache metrics, while low-level block conservation is covered by the existing KV-cache invariant tests rather than rechecked directly in the DP parity test.
  _cmd/root.go:2376-2407; cmd/replay.go:630-659; sim/cluster/cluster.go:746-783; sim/cluster/cluster_event.go:21-42; cmd/dp_replay_parity_test.go:218-245; cmd/dp_replay_parity_test.go:363-400; cmd/dp_placement_test.go:342-377; sim/kv/cache.go:476-505_
- **G5** — Returning an immutable deployment object would reduce coupling for hypothetical multiple command executions in one process, but the current design is consistent with this package's existing flag-global architecture.
  _cmd/root.go:523-543; cmd/root.go:544-566; cmd/root.go:2379-2400; cmd/replay.go:633-652; cmd/dp_placement_test.go:718-726; cmd/dp_placement_test.go:802-844_
- **G6** — A more explicit flag name could reduce operator ambiguity, but there is no reliable validation capable of distinguishing an intended aggregate from a large valid per-replica value.
  _cmd/root.go:472-477; cmd/root.go:493-499; cmd/replay.go:305-311; docs/reference/configuration.md:146-151; cmd/dp_placement_test.go:792-799_
- **G7** — Capacity-affecting CLI flags still must be re-supplied identically on replay; the trace is intentionally not authoritative for them.
  _sim/workload/tracev2.go:17-67; cmd/replay.go:264-303; cmd/replay.go:286-287; cmd/root.go:788-872; cmd/root.go:1009-1017_
- **G11** — An immutable returned deployment structure would be a reasonable alternative design, but it is not required for correctness here.
  _cmd/root.go:518-543; cmd/root.go:544-566; cmd/root.go:2379-2400; cmd/replay.go:633-652_
- **G14** — The helper also adds a clean zero-block error on the active auto-KV path, but production auto-KV already guarantees a positive per-rank count.
  _cmd/root.go:479-485; cmd/root.go:507-514; cmd/root.go:551-562; cmd/dp_placement_test.go:755-760; cmd/dp_placement_test.go:792-799_
- **G16** — Autoscaler and node-pool configurations are rejected earlier by replay's pre-existing replay-specific guards, so those messages intentionally differ from run even though both reject the configurations.
  _cmd/root.go:432-451; cmd/root.go:2347-2350; cmd/replay.go:618-623; cmd/dp_replay_parity_test.go:405-466; `go test ./cmd -count=1` passes_

namasl:
@claude /blis-pr-review

Final head `ebf5fffe`. This closes the one leftover nit from your cycle-4 review; no other changes.

**Your cycle-4 nit is fixed — and you undercounted it, which I only caught by grepping instead of trusting the list.** You flagged `dp_placement_test.go:65, 768, 775, 782, 789` (five sites). There were **six**: line 522 also had a bare `strings.Contains(string(out), "1548")` in the `TestRunCmd_MoEDPPlacement_GuardedCombo_Rejected` subprocess assertion. All six are now hashed, matching the 4x `#1548` / 10x `#1553` in `cmd/root.go`. `grep -rn ."1548".\|."1553". cmd/` now returns nothing.

**Mutation-verified rather than assumed:** dropping the `#` from one production message (`is tracked by #1548.` → `is tracked by 1548.`) now fails `TestResolveDPPlacement_MutatesDeploymentVars/EP-on_guard_errors_and_mutates_nothing`, which the bare form accepted. So the tightening is real, not cosmetic.

**You were right about the underlying problem, and it is worth naming.** My cycle-3 report said the guard assertions were tightened, when I had only done it in `dp_replay_parity_test.go`. It read as complete because it was complete in the file I was looking at — I described my attention rather than the codebase. Same class as the `tensor_parallel` sentence you and I both let through four cycles. The rule I am now applying to my own status reports, not just to docs: **grep to confirm the count before reporting the count.** Doing that here is what surfaced your list being one short.

**One more self-correction since cycle 4**, unprompted, in `docs/reference/configuration.md`: I had written that the TraceV2 header "records `tensor_parallel` informationally". That is false. `TraceServerConfig` declares the field, but its only non-test construction site (`cmd/observe_cmd.go:701`) sets just `Type` and `Model`, and `blis run --trace-output` emits no `server:` block at all — verified against a real emitted header, which contains only `trace_version`, `time_unit`, `mode`, `warm_up_requests`, `workload_seed`. With `omitempty` the field never reaches the YAML. Five of the seven `TraceServerConfig` fields are dead this way. The row now says "replay reads no parallelism field back from it", which is true regardless. The pre-existing twin of that claim at `observe-replay-calibrate.md:226` (`gpu_memory_utilization` and an auto-`total-kv-blocks` described as recorded — the former dead, the latter has no header field at all) is filed as **#1670** rather than fixed here, per `pr-workflow.md`'s pre-existing-issue rule.

**Gate status on `ebf5fffe`:** all 10 CI shards pass including `test (cluster (heavy), ./sim/cluster)`; locally `go build ./...` clean, `go test ./... -count=1` all packages pass, `golangci-lint run ./...` 0 issues. `/qa-review 1664` returned **PASS** (19/19 `CONFIDENT`, zero blocking findings) on `7ce79647`.

**Disclosure on the Archon gate:** I could not obtain an Archon verdict for this exact head. Three runs were attempted; each had `check-permissions` succeed and then `archon-review` sit with **no runner assigned** (`runner=` empty) for ~3 minutes before being cancelled — runner starvation, unrelated to this PR. Archon returned `NO_CHANGE` on the four preceding heads (`94d442c9`, `c412386b`, `d9090e87`, `7ce79647`). I am not claiming a verdict I did not receive; I am noting that the delta since the last successful one is `git diff 7ce79647 ebf5fffe` = 8 insertions / 8 deletions over two files — six string literals inside existing test bodies plus one markdown table cell, **zero production Go** — which cannot move a package-altitude result (packages, import edges, exported surface, schema) or change the guarded-promise note, since the same test functions persist under the same names. Please treat that as reasoning to check, not a substitute for the gate.

susiejojo (approved):
lgtm

namasl:
@claude /blis-pr-review

**Rebased onto `main` at `017fcd98` (#1660, EP-aware weight-footprint sizing). New head `6fd47fbc`.** ARCHON above: **`NO_CHANGE`** against the new base — 0 packages/edges/surface/schema changed, 0 violations. A rebase voids prior verdicts, so this is a full re-review request, not a delta.

**The one interaction worth your scrutiny — #1660 depends on the transformation this PR extends to replay.** #1660 added `epSizeForKVCapacity(isMoE, tp)` → `sim.EffectiveEPSize(isMoE, tp, dataParallelism, enableExpertParallel)`, and its own doc comment says it must read the **logical** `--dp` because DP-as-placement reconfigures each replica to `DP=1`, which would collapse the EP group to `tp` and silently void the sharding. My `resolveDPPlacement` mutates package flag vars, so the hazard is: if the per-rank `DP=1` ever reached `dataParallelism`, #1660's just-merged fix would break — on the replay path specifically, which is the path this PR opens.

I verified it is safe, by two independent arguments, and I would like both checked:

1. **`dataParallelism` is never written anywhere in non-test code** after flag parsing (`grep -rn "dataParallelism = " --include=*.go` → no non-test hits). `resolveDPPlacement` writes exactly one line — `numInstances, totalKVBlocks, maxModelLen` — and the per-rank DP travels as the *returned* `dpPlan.PerRankDP` into `NewModelHardwareConfig`, never back into the flag. So the ordering question is **moot**, not merely favourable.
2. Ordering is favourable regardless: every `epSizeForKVCapacity` call site (`root.go:916` global, `root.go:1883/1920` and `replay.go:464/501` per-pool) precedes its `resolveDPPlacement` call (`root.go:2404`, `replay.go:620`).

Empirical confirmation rather than code-reading alone: the auto-calc log prints `DP=2` *from `dataParallelism`* inside `resolveLatencyConfig` on a `--dp 2` run, and #1660's EP annotation resolves correctly — `EP=off` when EP is off, `EP=2` for `--tp 2 --dp 1 --enable-expert-parallel` (= TP·DP, 227391 blocks). I added a durable note at `epSizeForKVCapacity` recording this invariant, so a future change that writes the per-rank DP back into the flag var is warned at the exact site it would break #1656.

**A stale reference #1660 could not have known about, now fixed.** Its doc cited "the MoE dp>1 guard on replay, **#1556**" as what makes EP+DP unreachable — but this PR *removes* that guard. Corrected to say the single `planDPPlacement` #1548 guard covers both commands since #1556. Please confirm that reading.

**EP-on at `--dp 1` on replay** (reachable, unlike EP+DP): `--tp 2 --dp 1 --enable-expert-parallel` gives byte-identical run/replay stdout, both resolving `EP=2`.

**Conflict resolutions — four were genuine content merges, not pick-a-side**, because both PRs edited the same sentences: `docs/concepts/glossary.md`, `docs/guide/latency-models.md`, `docs/reference/models.md`, `docs/concepts/roofline.md`. Worth spot-checking that I kept #1660's EP-sizing content intact in each rather than clobbering it. `roofline.md` is the interesting one: #1660 had independently added the same "`--dp`/`--enable-expert-parallel` are rejected on roofline" correction I made, so I kept my verified fatal-error framing plus their EP-capacity-sizing clause. `CLAUDE.md` was pure insertion-adjacency — both paragraphs kept (verified one occurrence each).

**Re-verification on `6fd47fbc`:** `go build` clean; `go test ./... -count=1` 14/14 packages pass; `golangci-lint run ./...` 0 issues. **#1660's `TestEveryKVCapacityCallSiteIsEPAware` is green** — no call site lost or literalised its option. Run→replay byte-identity re-run at `dp=2/1`, `dp=4/2`, `dp=3/1`: all identical. Auto-KV parity identical, both legs deriving 74312 blocks/replica and the same 1188992-token cap (that arithmetic changed under me, so it was re-checked rather than assumed). #1548 and #1553 guards still exit 1 naming the right issue on replay.

Cycle-4 findings from before the rebase are all still applied; the earlier `/qa-review` returned PASS 19/19 on the pre-rebase tree.

Issue #1663, opened 2026-09-03
Roofline backend models no TP communication at all, so cross-node cost is unpriced there


namasl:
## Summary

`rooflineStepTime` has **no communication term at all**: the tensor-parallel degree enters only as a sharding divisor (`tpFactor` on FLOPs and on KV/weight bytes). There is no all-reduce byte volume, no MoE dispatch/combine volume, and therefore no place for #1530's cross-node re-scale to apply.

Consequence: under `--latency-model roofline`, a multi-node instance (#1529) is priced exactly like a single-node one — TP communication is free both within and across nodes.

## Why #1530 did not fix it

#1530's mechanism is a *re-scale of existing communication bases*. Roofline has none, so the change there would be to **add** a communication term — a different and larger change, requiring its own analytical form and its own validation against the trained-physics term. It is out of scope for an issue whose acceptance criteria are about cross-node re-scaling.

BLIS surfaces the gap rather than hiding it: `ClusterSimulator.applyPlacementTopology` emits a one-time warning when a spanning placement is created under a backend that models no communication, naming `--latency-model trained-physics` as the way to price it.

## Scope of the underlying gap

This is broader than cross-node: roofline under-prices *intra*-node TP as well. Roofline is not the default backend (trained-physics is), and its stated purpose is FLOPs/bandwidth-based estimation without learned corrections, so an analytical ring-all-reduce term (`2·(tp-1)/tp · tokens · hidden · dtype / link_bandwidth`) would fit its style — it just needs a defensible link-bandwidth input, which `HardwareCalib.IntraNodeBwGBps` (added in #1530) now supplies.

## Acceptance criteria

1. At TP>1 the roofline step time includes a communication term that grows with the token population and with the number of layers.
2. At TP=1 the term is exactly zero, so every TP=1 roofline result is byte-identical (INV-6).
3. A spanning placement is charged more than a contained one, reusing #1530's span penalty (so the two backends agree in direction, R23).
4. The cross-node warning in `applyPlacementTopology` for a comm-less backend is removed or narrowed once this lands.

## Triage

`enhancement` + `accuracy`. Discovered while implementing #1530.

Issue #1662, opened 2026-09-03
Inter-node fabric bandwidth is keyed by gpu_type, but a fabric is a pool property (coupled to #1543)


namasl:
## Summary

#1530 reads the inter-node fabric bandwidth (`InterNodeBwGBps`) from `HardwareCalib`, i.e. **per GPU type**. But a fabric is a property of the *cluster/pool*, not of the GPU die: two H100 pools can legitimately differ (InfiniBand vs RoCE vs a single 100 GbE uplink), while one H100 cannot differ from itself.

## Why it works today

It is correct today only because of a constraint #1529/PR #1537 added: **every node pool must declare a distinct `gpu_type`**, enforced by a panic in `NewPlacementManager` (cost and GPU memory are resolved by first-match on `gpu_type`, so a shared type would be ambiguous). That makes `gpu_type` ↔ pool one-to-one, so a per-GPU-type fabric value *is* per-pool.

`#1543` is slated to relax exactly that constraint (pool-identity lookup instead of `gpu_type` first-match). When it does, two pools sharing a `gpu_type` would silently share one fabric calibration — mis-pricing cross-node cost for whichever pool has the other fabric, with no diagnostic.

Note the intra-node half does **not** have this problem: NVLink/xGMI generation genuinely is a die/board property, so `IntraNodeBwGBps` belongs where it is.

## What this would require

- Move (or add an override for) the inter-node bandwidth onto `NodePoolConfig`, which already owns every node-scoped attribute (`gpus_per_node`, `gpu_memory_gib`, `cost_per_hour`) and already gets strict YAML validation with NaN/Inf checks in `IsValid()`.
- Thread it through the same three placement sites that already resolve per-pool GPU memory (#1522) and cost (#1529) — the pool is already looked up there, so no new plumbing is needed.
- Precedence: an explicit pool value wins over the GPU-type value, mirroring how the placement-derived per-GPU KV capacity wins over the global `--total-kv-blocks`.

## Acceptance criteria

1. Two pools with the same `gpu_type` but different declared inter-node bandwidths price cross-node collective traffic differently for instances placed in each.
2. A pool that declares no inter-node bandwidth inherits the GPU-type value (byte-identical to #1530's behavior, INV-6).
3. Invalid pool values (negative, NaN, Inf) are rejected loudly by `NodePoolConfig.IsValid` (R3/R10).

## Dependency

[...]

Issue #1661, opened 2026-09-03
Cross-node collective latency (launch + RTT) not modeled — #1530 prices bandwidth only


namasl:
## Summary

The inter-node network cost landed in #1530 is **bandwidth-only**: a collective that spans a node boundary has its byte volume charged at a blended intra/inter-node bandwidth. It charges **nothing** for the fixed per-collective cost of crossing the fabric — NCCL kernel launch, IB/RoCE round-trip, and the cross-node synchronization that a hierarchical collective imposes at each of its two levels.

For decode steps that fixed cost is plausibly the *dominant* cross-node effect, because the messages are small.

## Evidence / magnitude

Order-of-magnitude estimate for a GLM-5.2-shaped model (92 layers, hidden 6144, bf16) at TP=16 spanning two nodes, decode batch 64:

- Per all-reduce byte volume ≈ 1.2 MB. Aggregated over 92 layers × 2 comm units, the **bandwidth** term after β₄ is on the order of tens of µs, and #1530's cross-node re-scale adds roughly half as much again (the TP comm term rises ~1.53× at a 9× fabric ratio).
- The same step performs ~186 cross-node collectives. At even ~2 µs of launch + RTT each, the **fixed** cost is ~370 µs — an order of magnitude above the bandwidth delta.

This is also the mechanism behind vLLM's own guidance to prefer pipeline parallelism across nodes and tensor parallelism within a node: per-layer TP all-reduce is punishing over inter-node links not only because the links are slower but because there are so many small collectives.

## Why it was deferred

#1530's scope was fixed by its issue: *re-scale the two existing communication bases*, swapping the on-package bandwidth for an effective inter-node bandwidth. A fixed per-collective latency is a new additive term, and it needs an empirical per-collective constant that BLIS has no calibration source for today — the repo's convention (see the Kimi-K3 KDA note in #1636 and the LoRA TTFT-leg honesty in #1470) is to document such a gap rather than ship a fabricated coefficient.

## What this would require

- A per-collective inter-node base latency, either as a new `HardwareCalib` field (alongside `IntraNodeBwGBps` / `InterNodeBwGBps`) or derived from the fabric type.
- A decision on the collective *count* per step: the natural unit is the same `units` the comm bases already use (layers × comm phases), charged only when the collective spans a node boundary.
[...]

PR #1667, opened 2026-09-03
feat(latency,cluster): price cross-node collective traffic from real placement (Closes #1530)

### LINKED ISSUE #1530: Inter-node network cost not modeled (cross-node TP/EP traffic is free)

namasl:
## Summary

Cross-node collective traffic was free. Both trained-physics communication bases divided
byte volume by the on-package `bwHbmUs`, so an instance spanning nodes — possible since
#1529 — paid NVLink prices for traffic crossing InfiniBand.

Both bases now divide by an **effective** link bandwidth: a re-scale in place, never an
additive term (which would double-charge a `DP>1` MoE instance whose all-to-all
`moeDispatchBasis` already prices). Topology is **derived from real placement**, never a
flag. Fabric speeds are per-GPU `HardwareCalib` fields.

## Mechanism

With `G` the group size, `p` the placed node size, `n = ceil(G/p)`, `g = min(G, p)`, and
`r = IntraNodeBwGBps / InterNodeBwGBps`:

| Collective | Applies to | Penalty |
|---|---|---|
| Ring | TP all-reduce (`G = tp`) and the `allgather_reducescatter`/`naive` MoE family | `1 + (r-1)·(n-1)/(G-1)` |
| All-to-all | `deepep_*`/`pplx`/`mori`/`flashinfer_all2allv` dispatch (`G = TP·DP`) | `1 + (r-1)·(G-g)/(G-1)` |

Plus a size-independent half, `InterNodeLatencyUs` — the fixed launch + fabric round-trip
of one cross-node collective, charged per collective (2 per layer for a dense TP step,
3 for an MoE step at `DP>1`: attention all-reduce + dispatch + combine).

### Why ~1.53× and not ~9×

The single most consequential number here, so the derivation is in the code and checked by
a test rather than asserted.

A **flat** ring all-reduce spanning nodes is throttled end-to-end by its slowest link,
giving ≈ `r` — 9× at the committed H100 ratio. NCCL does not do that across nodes. It runs
the collective **hierarchically**: intra-node reduce-scatter + all-gather over the `g`
co-located ranks, then an inter-node all-reduce of the **already-reduced** `S/g` chunk
across the `n` nodes. Only `1/g` of the volume — one eighth on an 8-GPU node — ever touches
the fabric.

At TP=16 across two 8-GPU H100 nodes with `r = 450/50 = 9`:

- fast link: `2(g-1)/g = 1.75 S`
- fabric: `2(n-1)/n · S/g = 0.125 S`, at 9× each
- flat baseline: `2(G-1)/G = 1.875 S`
- penalty = `(1.75 + 0.125·9) / 1.875 = 1.533`

and the closed form `1 + (r-1)(n-1)/(G-1) = 1 + 8/15` reproduces it exactly. The identity is
exact, not approximate, because hierarchical all-reduce moves the same per-rank bytes as a
flat ring: `2(g-1)/g + 2(n-1)/(n·g) ≡ 2(G-1)/G`.
`TestRingSpanScale_MatchesHierarchicalDecomposition` checks the closed form against that
per-phase decomposition over a grid of `(g, n, r)`.

Checkable rather than only arithmetic: measured two-node H100 all-reduce bus-bandwidth
degradation is ~1.3–1.5× (nccl-tests `all_reduce_perf` at 1–2 node scale) — the band the
hierarchical model lands in and the flat model misses by ~6×. The assumption and its failure
mode are recorded on `ringSpanScale`.

**This does not contradict vLLM's TP-within-node guidance.** vLLM prefers PP across nodes
because per-layer TP means *many small* collectives — dominated by fixed per-collective cost,
not bandwidth. That is the term below, and why leaving it at 0 is called out as a limitation
rather than presented as a complete model.

## The uncalibrated latency default, and AC-2

`InterNodeLatencyUs` **ships at 0 — not charged — and that is a real limitation, stated
rather than papered over.** AC-2 requires the charged cost not to decrease as bandwidth falls
*or as latency rises*. The bandwidth clause is exercised by the committed calibration. The
latency clause is exercised **only synthetically**: four tests cover the mechanism, but no
shipped configuration makes the term fire, so out of the box the cross-node cost is
bandwidth-only.

I could not supply a defensible default — per-collective latency depends on message size,
fabric and NCCL algorithm, and BLIS has no measurement to anchor it; a plausible-looking
constant would be indistinguishable from a calibrated one. The repo's convention is to
document such a gap rather than fabricate the coefficient (#1636's KDA note is the
precedent). So the mechanism is present and tested, `TestCommittedHardwareConfig_LatencyIsUncalibrated`
pins the 0 so any future value must arrive with a source, and **#1661** carries the
calibration bar.

Read `IntraNodeBwGBps` as *the on-node link speed β₄ was calibrated against* (β₄ already
absorbs the NVLink/HBM ratio); the latency rides the same coefficient, so it calibrates in
the frame `β · units · latency`.

## Behavioural contracts

- **BC-1 (AC-1, TP leg)** — an instance placed on nodes smaller than `TP` has strictly
  greater step time than one whose node holds the whole TP group. Verified through real
  placement and end-to-end through mean TTFT.
- **BC-2 (AC-1, DEP leg)** — an MoE group exceeding the placed node size is charged at the
  blended bandwidth. Latency-model level for both comm families; placement-driven assertion
  deferred to #1548 per the issue's per-leg coverage.
- **BC-3 (AC-2)** — step time is monotone non-decreasing as `InterNodeBwGBps` falls, strictly
  increasing over 450→5 GB/s.
- **BC-4 (AC-3)** — byte-identical (exact equality) at three independent gates: no
  `node_pools`; collective fits one node; no interconnect calibration. Dense `DP=1` still
  satisfies INV-BC-DP1.
- **BC-5** — the node span used for latency **pricing** equals the span used for cost
  **accounting** (`InstanceCostPerHour`), for every placement shape.
- **BC-6 (R23)** — startup, deferred `NodeReadyEvent` and autoscaler scale-up all stamp it.
- **BC-7 (INV-13)** — run-only, fenced on both paths (below).
- **BC-8** — every degenerate input (unknown/negative node size, group ≤ 1, bandwidth
  0/negative/NaN/Inf/overflowing) yields a penalty of exactly 1.0: no NaN, Inf, division by
  zero or panic.
- **BC-9 (INV-6)** — penalties frozen at construction from configuration only; no RNG, no map
  iteration, nothing re-read on the hot path. Zero heap allocation, asserted via
  `testing.AllocsPerRun`.

## Cross-path parity — both paths fenced (INV-13)

`blis replay` `logrus.Fatalf`s on `node_pools` and `observe` builds no simulator, so a
non-inert topology is unreachable off the run path.

That guard does **not** cover the trace path, and I checked rather than assumed: on `main`,
exporting a spanning node-pool run and replaying its own trace returns **identical** metrics
(0 of 27 differ), so the divergence this feature would introduce is **mine, not inherited** —
with the feature, 16 of 27 differ. So `blis run --trace-output` records the widest instance
node span (`max_nodes_spanned`) and `blis replay` hard-errors above 1.

**Still design (b), not (a).** The header field is a **refusal signal, not a reconstruction
input** — replay reads it only to fail loudly; reproducing cross-node timing would need
per-instance topology, not one fleet-wide maximum.

Behaviour change: replaying a **newly-exported spanning** trace is now a hard error where it
previously produced wrong numbers. Pre-#1530 traces recorded nothing, so they stay in the
accept bucket and still replay as single-node — verified, 0 of 27 metrics differ against an
old-binary replay. That is the guard's known limitation: it protects traces from this version
onward.

## Testing

`go build ./...` clean · `go test ./... -count=1` 14 packages ok, 0 failures ·
`golangci-lint run ./...` 0 issues.

Plus CLI-level INV-6 evidence no unit test gives: identical stdout between a `main` binary and
this branch across four configs, including `mixtral-8x7b --tp 4` and a spec-decode config.

## Discovered issues

Filed, not fixed here (`pr-workflow.md`, "Filing Pre-Existing Issues"):

- **#1661** — per-collective cross-node latency: mechanism ships here, calibrated value does not.
- **#1662** — fabric keyed by `gpu_type` rather than by pool.
- **#1663** — roofline backend models no communication term at all.
- **#1668** — `hw_config_by_gpu` documented as policy-config YAML, but `PolicyBundle` has no
  such field and parses with `KnownFields(true)`, so the doc produces a config that fails to
  parse. Verified pre-existing on `main`.
- **#1669** — `models.md` denied the shipped MTP feature (#1528).

Other documented approximations, each at its point of use: the all-to-all penalty sums rather
than overlaps the on/off-node portions and ignores DeepEP's per-node RDMA coalescing (both
pessimistic); the lumped `TP·DP` MoE group's span is extrapolated from the placed node size,
so the DEP leg ships **inert in every reachable config** today (`node_pools`+`--dp>1` is a
fail-fast, #1553) and becomes reachable with #1548; and at `dp>1` the latency half inherits a
`/dp` division that is right for volume but not for a launch cost — unreachable today, flagged
in the code for #1548.

Closes #1530


namasl:

namasl:

namasl:

## qa-review — PR #1667: ⛔ BLOCK
_feat(latency,cluster): price cross-node collective traffic from real placement (Closes #1530) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The PR implements both communication legs through one shared span-aware mechanism: TP collectives use the ring scale, while MoE dispatch selects ring or all-to-all scaling by backend family. It derives topology from actual placement, applies it at startup, deferred placement, and autoscaler scale-up, preserves the single-node path, and implements the permitted loud-replay-rejection design. The optional fixed-latency component is directly relevant to AC-2 rather than unrelated scope. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. Several committed documentation statements are stale or internally inconsistent. The latency guide says fixed per-collective launch/round-trip cost is not modeled, although this PR implements `InterNodeLatencyUs`. The cluster guide retains the heading “Cross-node interconnect not yet priced” while its body says it is priced. The configuration reference also presents `hw_config_by_gpu` as YAML policy configuration, but `PolicyBundle` has no such field and strict YAML decoding rejects it. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | There are stale comments left by the change. Most notably, `NetworkTopology` still says no TraceV2 round-trip is needed because replay cannot receive a non-inert topology, but the PR now adds `MaxNodesSpanned` specifically to round-trip a refusal signal and close the trace replay path. This is misleading commentary, not a runtime defect. |
| G1 | correctness | ✅ CONFIDENT | In `moeDispatchBasis`, when `DP > 1` and `TP > 1` are combined, how does the network scaling formula calculate `MembersPerNode` and `NodesSpanned` when placement is only recorded at the TP group level? Could a lumped `TP·DP` MoE group span nodes in a way that miscalculates the per-node rank count `g = min(G, p)` when `G = TP·DP`? | For the lumped MoE group, the code sets `G = EffectiveMoEGroupSize() = TP·DP`, then extrapolates topology from the TP placement's node size `p`: `NodesSpanned(G)=ceil(G/p)` and `MembersPerNode(G)=min(G,p)`. Thus it assumes a uniform flattened TP·DP placement even though only the TP group was physically placed. That can be inaccurate for a future non-uniform or differently arranged DP/EP placement, but it cannot misprice a currently reachable node-pool deployment because DP-as-placement changes each spawned replica to DP=1, leaving the cross-node DEP leg inert. The approximation is explicitly documented and deferred to EP placement work. |
| G2 | correctness | ✅ CONFIDENT | How does the ring network penalty formula `1+(r-1)(n-1)/(G-1)` behave when `G=1` (e.g., single rank or `TP=1`)? Does `NodesSpanned` or `MembersPerNode` safely return 1 or 0 to avoid a division-by-zero or negative penalty scaling term? | The formulas safely handle `G=1`. Both `ringSpanScale` and `all2AllSpanScale` return exactly 1 before evaluating a `G-1` denominator. Independently, `NodesSpanned` returns 1 when the group fits or is non-positive, and `MembersPerNode` returns 0 for non-positive groups and 1 for a one-rank group. There is therefore no division by zero or negative scaling. |
| G3 | invariants | ✅ CONFIDENT | INV-13 (run/replay parity) requires that unsupported replay scenarios fail loudly rather than degrade silently. If a trace generated on `main` before PR #1667 (lacking `max_nodes_spanned` in its header) from a multi-node run is replayed with the new binary, does it silently replay at single-node speed, and how does this impact INV-13 backwards compatibility guarantees? | An old trace with no `max_nodes_spanned` loads the field as zero and is accepted, so the new binary cannot detect that its source placement crossed nodes. However, a trace generated by `main` predating this PR was itself produced without the new cross-node charge; replaying it with unknown topology preserves that historical single-node-rate behavior rather than diverging from the source run. New spanning traces carry the refusal signal and fail loudly. This is a deliberate backward-compatibility limitation, not an INV-13 regression introduced by the PR. |
| G4 | invariants | ✅ CONFIDENT | INV-6 requires byte-identical determinism for identical seeds and configurations. Does `applyPlacementTopology` or any placement site state update depend on execution or event ordering across multiple instances in a way that could cause subtle step-time drift across runs? | No nondeterministic dependency was found. Each instance's topology is derived only from its concrete `gpuIDs` and immutable placement indexes, then the numeric span scales are frozen when its latency model is constructed. The only cluster-wide updates are monotonic integer maxima and warning latches, neither of which affects StepTime. Placement and event processing are single-threaded, and a spanning end-to-end test compares marshalled metrics byte-for-byte across identical runs. |
| G5 | rules | ✅ CONFIDENT | Rule R1 (fail fast on malformed config) and Rule R23 (load-boundary validation): `HardwareCalib.ValidateInterconnect()` validates `IntraNodeBwGBps` and `InterNodeBwGBps`. What happens if a policy config supplies `hw_config_by_gpu` with only one of the two interconnect fields set—does `GetHWConfig` immediately trigger a hard error, or does it fall back or silently ignore the partial setting? | A CLI policy bundle cannot currently supply `hw_config_by_gpu`: `PolicyBundle` has no such field and strict `KnownFields(true)` decoding rejects the key immediately. For hardware loaded through `--hardware-config`, `GetHWConfig` calls `ValidateInterconnect` and returns a hard error for a half-set pair, which the CLI converts to `logrus.Fatalf`. A programmatically populated `DeploymentConfig.HWConfigByGPU` bypasses `GetHWConfig`; under trained-physics it is still rejected by the latency-model constructor, while roofline does not consume or validate those interconnect fields. |
| G6 | rules | ✅ CONFIDENT | Rule R4 / R16 (canonical constructors and latency input immutability): `WithNetworkTopology` introduces a new variadic option for `ModelHardwareConfig`. Is `NetworkTopology` immutably stored or deep-copied inside `ModelHardwareConfig` to prevent external modifications after constructor execution? | `NetworkTopology` is a value type containing only an `int`. `WithNetworkTopology` assigns that value into `ModelHardwareConfig`, so it is copied and has no pointer, slice, map, or other mutable alias requiring a deep copy. The latency constructor further resolves it into scalar span scales, so later StepTime calls do not consult live placement state. |
| G7 | parity | ✅ CONFIDENT | Why is `MaxNodesSpanned` recorded in `TraceHeader` during `blis run` but omitted from `blis observe` trace outputs, and could replaying an observed multi-node trace bypass the fence in `cmd/replay.go`? | `blis observe` records timing from a real server and constructs no `ClusterSimulator` or BLIS node-pool placement, so it has no authoritative `MaxNodesSpanned` value to emit. An observed trace can therefore lack the field and pass the replay fence, but this does not bypass a modeled cross-node run restriction: the observed source never used this DES network term. Replay simulates the observed workload under its explicitly supplied simulation configuration; it does not claim to reconstruct the real server's topology. |
| G8 | parity | ✅ CONFIDENT | When `--hardware-config` is provided with non-default interconnect calibrations, how is this file path or resolved interconnect calibration preserved in `TraceHeader` or replay flags so that replay runs assert consistent hardware assumptions against the original run? | The `--hardware-config` path and resolved interconnect calibration are not stored in TraceHeader. Replay resolves hardware again from its own `--hardware`, `--hardware-config`, and defaults, consistent with the existing CLI-resupplied hardware/model configuration contract. For a newly generated spanning run trace, replay fails on `MaxNodesSpanned` before any topology-dependent timing can be reproduced; for accepted single-node traces, the interconnect fields are inert because no boundary is crossed. |
| G9 | tests | ✅ CONFIDENT | Is there a companion invariant/law test asserting that step time is strictly monotonic with respect to `r = IntraNodeBwGBps / InterNodeBwGBps` (AC-2) for any arbitrary model shape, independently of specific baseline golden test values? | Yes at the law/formula level, and also at StepTime level for representative model and batch shapes. `TestSpanScale_MonotoneAndBounded` verifies the shared algebra independently of model goldens; `TestStepTime_MonotoneInFabricQuality` holds placement fixed and checks nondecreasing StepTime as inter-node bandwidth falls, with a strict endpoint increase. The cluster end-to-end suite independently repeats the monotonic placement test. |
| G10 | tests | ✅ CONFIDENT | Given that the DEP all-to-all leg (`moeDispatchBasis`) is currently inert for cross-node placement until DP/EP placement lands (#1531/#1548), what synthetic unit tests verify that the all-to-all network penalty formula `1+(r-1)(G-g)/(G-1)` scales correctly when `G = TP·DP` across multi-node topologies? | Synthetic coverage directly checks the all-to-all formula across node sizes 1/2/4/8, spans of 1/2/4 nodes, and ratios 1/2/9 against an independently calculated per-peer decomposition. A second test proves all-to-all exceeds ring scaling for the same topology, and a StepTime test uses `TP=4, DP=4` with an all-to-all backend to verify the selected family produces a larger isolated cross-node penalty than the ring family. |
| G11 | rationale | ✅ CONFIDENT | Why was in-place scaling of `bwHbmUs` (`bwHbmUs / spanScale`) chosen in `trained_physics_model.go` over adding a distinct cross-node communication term, and what specific double-charging issue would an additive term introduce for `DP > 1` MoE configurations? | The existing TP and MoE communication bases already represent the full collective byte volume and are weighted by the trained coefficients. Replacing their effective bandwidth preserves that calibrated decomposition. Adding a second cross-node term for the same bytes would retain the original intra-node-priced collective and then charge its crossing again. This is especially wrong for `DP>1` MoE, where `moeDispatchBasis` already prices dispatch/combine over the flattened `TP·DP` group; an additive network term would double-count that traffic. |
| G12 | rationale | ✅ CONFIDENT | Why are `IntraNodeBwGBps` and `InterNodeBwGBps` defined per GPU type in `hardware_config.json` rather than per pool in the `node_pools` policy configuration? How does this design handle heterogeneously networked nodes sharing the same GPU type? | The implementation treats fabric bandwidth as hardware calibration and topology as placement data, using the existing `HardwareCalib` resolution path rather than adding a separate policy knob. This works while node pools use distinct `gpu_type` keys. It cannot faithfully represent two differently networked pools sharing the same GPU type; current pool validation already requires distinct GPU-type identifiers, and the remaining per-pool calibration limitation is explicitly tracked as #1662. Such deployments need distinct/aliased GPU calibration keys until per-pool fabric calibration exists. |
| G13 | scope | ✅ CONFIDENT | The PR introduces `InterNodeLatencyUs` for fixed launch and round-trip cost, but defaults it to 0 (uncalibrated) across all bundled configs. Does adding this uncalibrated parameter expand the scope beyond #1530's acceptance criteria, and why was it included ahead of calibration data (#1661)? | The parameter is within #1530's stated network model scope and directly supports AC-2's requirement that increasing inter-node latency not reduce cost. It was included as the size-independent half of collective cost, but defaults to zero because no defensible bundled calibration exists. Consequently it is inert for shipped hardware and does not alter current results; synthetic tests establish its semantics and monotonicity so measured values can be added later under #1661. |
| G14 | scope | ✅ CONFIDENT | The PR adds `MaxNodesSpanned` to `TraceHeader` and adds logic to `cmd/replay.go` to reject multi-node traces. Was modifying the TraceV2 header schema strictly necessary to close #1530, or could the replay fence have been implemented without altering exported trace metadata? | Some persistent trace metadata was necessary for the selected loud-rejection design. Rejecting `node_pools` in replay only fences the forward configuration path; users can replay a trace without supplying its original policy bundle, at which point replay has no other way to know the source run spanned nodes. `MaxNodesSpanned` is a compact refusal-only signal, omitted for non-spanning runs, rather than an attempted topology reconstruction. The alternative allowed by the issue was to serialize enough per-instance topology to reproduce timing, which would require more metadata, not less. |
| G15 | errors | ✅ CONFIDENT | Under what exact conditions does `warnIfCrossNodeUnpriced` emit a warning on stderr versus triggering a fatal error, and why is an uncalibrated interconnect for a multi-node placement handled via a log warning rather than a strict validation failure at startup? | `warnIfCrossNodeUnpriced` only runs for a real span greater than one. It warns, once per cause, when the backend has no communication term, node size could not be resolved, all interconnect calibration is absent, or a present ratio exceeds the unit-error threshold. Malformed calibration is different: negative/non-finite values or a half-set bandwidth pair are validation errors. Completely absent calibration is allowed for backward compatibility and single-node use, so startup cannot reject it globally; it becomes operationally problematic only if an actual placement spans nodes, where the warning makes the optimism visible. |
| G16 | errors | ✅ CONFIDENT | In `HardwareCalib.ValidateInterconnect()`, what types of inputs (e.g. negative values, NaN, infinite ratios) trigger standard Go `error` returns to library callers versus CLI `logrus.Fatalf` exit points? | `ValidateInterconnect` returns a normal Go error for any negative, NaN, or infinite bandwidth/latency value and for exactly one positive bandwidth with the other unset or unusable. Zero for both bandwidths and zero latency is valid and inert; positive latency may stand alone. `GetHWConfig` wraps the validation error, and CLI resolution converts that returned error to `logrus.Fatalf`. Programmatic trained-physics construction also returns the wrapped error from `NewLatencyModel`; `NewInstanceSimulator` converts that library error to a contextual panic, not a process-level fatal from inside `sim/`. |

### Items to fix
- **F2 · FLAW_FOUND** — No. Several committed documentation statements are stale or internally inconsistent. The latency guide says fixed per-collective launch/round-trip cost is not modeled, although this PR implements `InterNodeLatencyUs`. The cluster guide retains the heading “Cross-node interconnect not yet priced” while its body says it is priced. The configuration reference also presents `hw_config_by_gpu` as YAML policy configuration, but `PolicyBundle` has no such field and strict YAML decoding rejects it.
  _docs/guide/latency-models.md:384-388 versus sim/model_hardware_config.go:192-209 and sim/latency/trained_physics_model.go:362-408; docs/guide/cluster.md:67-68; docs/reference/configuration.md:464-477 and docs/reference/configuration.md:688 versus sim/bundle.go:14-27 and sim/bundle.go:111-124_

### Important to consider
- **F1** — The DEP leg remains unreachable with current production placement, as the issue explicitly permits until EP-as-placement lands; its formula and synthetic StepTime path are nevertheless covered.
  _sim/latency/trained_physics_model.go:298-359; sim/latency/trained_physics_model.go:376-408; sim/cluster/cluster.go:395-427; sim/cluster/infra_lifecycle_event.go:48-83; sim/cluster/direct_actuator.go:76-109; cmd/replay.go:98-112; sim/cluster/network_topology_e2e_test.go:119-214; sim/latency/network_cost_test.go:199-328_
- **F3** — The stale guide text identified in F2 should be corrected at the same time.
  _sim/network_topology.go:21-27 versus sim/workload/tracev2.go:68-91 and cmd/replay.go:98-112_
- **G1** — When #1548 introduces real EP placement, this topology signal will need to describe the actual expert group rather than extrapolate from TP.
  _sim/network_topology.go:53-83; sim/latency/trained_physics_model.go:328-352; sim/latency/trained_physics_model.go:996-1037; cmd/root.go:2253-2268; docs/guide/latency-models.md:396-400_
- **G3** — Old traces cannot retroactively communicate placement provenance, so they cannot benefit from the new fence.
  _sim/workload/tracev2.go:68-91; sim/workload/tracev2_crossnode_test.go:95-108; cmd/replay.go:98-112; sim/network_topology.go:21-27_
- **G5** — The configuration reference incorrectly describes `hw_config_by_gpu` as YAML policy input; see F2.
  _sim/bundle.go:14-27; sim/bundle.go:111-124; sim/latency/config.go:183-208; sim/model_hardware_config.go:227-249; cmd/root.go:690-694; sim/cluster/cluster.go:395-405; sim/latency/trained_physics_model.go:941-949_
- **G7** — If replay is later expected to reconstruct real-server topology from observation metadata, observe would need a separate topology input and provenance contract.
  _cmd/observe.go:540-580; sim/workload/tracev2.go:68-85; cmd/replay.go:98-112; cmd/root.go:2491-2498_
- **G8** — Replay can use different hardware assumptions generally because TraceV2 is workload-focused and hardware remains caller-supplied; this is pre-existing behavior rather than a new cross-node defect.
  _cmd/root.go:605-694; cmd/replay.go:91-112; cmd/replay.go:630-643; sim/workload/tracev2.go:68-91; sim/latency/trained_physics_model.go:996-1003_
- **G9** — The StepTime test uses one representative model rather than property-generating arbitrary model shapes, but monotonicity is pinned independently in the shared scalar formula.
  _sim/latency/network_cost_test.go:76-88; sim/latency/network_cost_test.go:236-255; sim/cluster/network_topology_e2e_test.go:142-160_
- **G12** — Per-pool fabric configuration would be more expressive, but the current restriction makes the implemented keying coherent for reachable CLI configurations.
  _sim/model_hardware_config.go:178-209; sim/cluster/network_topology.go:60-69; docs/guide/latency-models.md:389-391; docs/guide/cluster.md:61-65_

namasl:
Addressed the qa-review **BLOCK** (F2, documentation) in `afeb5b08`, plus the two items from the `@claude` re-review.

**F2 — the three stale statements, and a fourth the sweep found**

Two were self-contradictions this PR introduced, and both are fixed:

- `docs/guide/latency-models.md` said the fixed per-collective launch/round-trip cost is *not modeled*. It is now modeled but ships **uncalibrated** (0 ⇒ inert) — a third state distinct from both "not modeled" and "modeled". The bullet now says exactly that and ties it to the AC-2 limitation.
- `docs/guide/cluster.md`'s admonition was still titled *"Cross-node interconnect not yet priced"* while its own body explained that it is priced. Retitled to name what is and is not calibrated.

The answerer's third item is **pre-existing on `main`**, which I verified before deciding how to handle it: `hw_config_by_gpu` has been documented as policy-config YAML (with snake_case keys that would not bind either way) since before this PR, while `PolicyBundle` has no such field and parses with `KnownFields(true)`. Per `pr-workflow.md` "Filing Pre-Existing Issues" I **filed it as #1668 rather than fixing it here** — folding an unrelated doc defect into a latency PR is the scope creep that rule exists to prevent. What this PR does keep is the one-line disclosure its own adjacent `#1530` note needs in order to be truthful, pointing at #1668 for the rest.

Because doc drift had now been flagged twice on this PR, I also swept every claim about cross-node / inter-node / interconnect pricing across `docs/` and `CLAUDE.md` rather than fixing only the named files. That found **a fourth contradiction both reviews missed**: the three-inertness-gates list said gate 3 was "no interconnect bandwidths", which stopped being sufficient once the latency half existed — a declared latency with no bandwidths still charges. Fixed, including why a latency-dominated fabric is a legitimate configuration. It also surfaced an unrelated pre-existing staleness (`models.md` denying the shipped MTP feature), filed as **#1669**.

Also in `afeb5b08`: `sim/network_topology.go` still claimed no TraceV2 round-trip was needed, contradicting the `MaxNodesSpanned` field. Reworded to state what is actually true — parity is by *rejection*, and what round-trips is a refusal signal, not a cost.

**From the `@claude` re-review** (`9112cfb0`): the `placedGPUsPerNode` messages no longer say "will be unpriced" (the cost is charged at the on-node rate — optimistic, not absent) and now match `warnIfCrossNodeUnpriced`'s wording; `TestReplayCmd_SingleNodeSpanTraceAccepted` gained the sibling silence assertion; and the zero-allocation claim is now **asserted** via `testing.AllocsPerRun` across the inert, spanning-bandwidth and spanning-bandwidth-plus-latency paths rather than left as a comment next to a benchmark that only reported it.

**Verification on `afeb5b08`** — run locally, since both `@claude` review environments reported no Go toolchain: `go build ./...` clean, `go test ./... -count=1` all packages pass, `golangci-lint run ./...` 0 issues.

namasl:

namasl:
Round 3: the 🟡 Important finding was real and is fixed in `7b35948c`. The 🔴 Critical finding is a false positive — evidence below, and I did not act on it.

## 🟡 MoE dispatch/combine charged one collective instead of two — fixed

Correct catch, and it was a genuine internal inconsistency rather than a style issue. The dispatch **volume** already counts two collectives per MoE layer (the `2` in each `moeDispatchBytes` branch) because dispatch and combine are separate NCCL calls — all-gather then reduce-scatter, or two all-to-alls for a modular backend — each with its own launch and round-trip. The **latency** was charging one. Two halves of the same cost disagreeing 2× on the size-independent term.

Now charged via an explicit `moeDispatchCollectivesPerLayer = 2`, with the reasoning written down, because the per-layer collective count genuinely *differs* between the two paths and that asymmetry is what makes the bug easy to introduce:

- a **TP unit is one NCCL call** — a ring all-reduce is a single call even though its byte volume has two phases — and a dense step gets its `2L` from having two such units per layer (attention, FFN);
- an **MoE step at `dp>1` launches 3 per layer**: attention all-reduce + dispatch + combine.

The `HardwareCalib.InterNodeLatencyUs` doc stated the rule as a flat `L × 2`, which is only the dense case; generalized so the code and the field doc agree. CLAUDE.md likewise. `TestStepTime_MoEDispatchChargesTwoCollectivesPerLayer` pins it as a ratio against an identically-shaped dense model (3L vs 2L) — charging dispatch/combine once would make the MoE step *equal* the dense one, so the assertion fails on exactly the mutation being guarded.

It is inert twice over today (`InterNodeLatencyUs` is 0 in every bundled entry, and `moeDispatchBasis` only runs at `dp>1`, a fail-fast alongside `node_pools`), but it would have misled whoever calibrates #1661 — which is why it was worth fixing rather than deferring.

## 🔴 Critical — CLAUDE.md regression: false positive, not acted on

The finding reports that `afeb5b08` deleted the ~1,500-word `Inter-node network cost (#1530)` section and reverted the #1529 cross-reference to "deferred to #1530". I checked before changing anything, because restoring a section from memory would have been destructive if the claim were wrong. All three sub-claims are contradicted by the repository:

```
$ git diff afeb5b08~1 afeb5b08 --stat -- CLAUDE.md
(empty — afeb5b08 did not touch CLAUDE.md at all)

$ git show --stat --format= afeb5b08 -- CLAUDE.md
(empty)

$ grep -c 'Inter-node network cost (#1530' CLAUDE.md
1

$ for ref in 98b281db afeb5b08; do git show $ref:CLAUDE.md | awk '/^\*\*Inter-node network cost \(#1530/,/^$/' | wc -w; done
1218
1300
```

So the section is present and is **longer** than in `98b281db` — `9112cfb0` added the backward-compatibility sentence — and the cross-reference reads `Cross-node latency is now priced by #1530`, not the stale text. `afeb5b08` changed four files: `docs/guide/cluster.md`, `docs/guide/latency-models.md`, `docs/reference/configuration.md`, `sim/network_topology.go`.

This matches a known false-positive pattern for this reviewer: on #1537 a "working-tree CLAUDE.md clobbered" flag appeared in five separate rounds and was a CI-runner artifact every time, with HEAD always correct.

I have adopted the process suggestion regardless, since it is good practice independent of this false alarm: every doc commit now gets `git diff HEAD~1 -- '*.md' CLAUDE.md` read for **deletions**, not just additions. For `7b35948c` that check shows `CLAUDE.md | 1 insertion(+), 1 deletion(-)` — one paragraph rewritten, nothing removed.

## Verification on `7b35948c`

Run locally, since all three `@claude` review environments have reported no Go toolchain: `go build ./...` clean, `go test ./... -count=1` all packages pass, `golangci-lint run ./...` 0 issues.

namasl:

namasl:

## qa-review adjudication — PR #1667: ⛔ BLOCK

|----|----|----|----|
| F2 | FLAW_FOUND | ❌ STILL-OPEN | The latency guide and cluster heading were corrected, and open issue #1668 genuinely covers the pre-existing `hw_config_by_gpu` documentation defect. However, the cluster guide still says the fixed per-collective launch/round-trip cost is “unmodeled,” while the current implementation models it through `InterNodeLatencyUs` and merely ships it uncalibrated at zero. The configuration summary also still lists `hw_config_by_gpu` as YAML-only despite the adjacent section correctly saying it is programmatic-only. Update those remaining statements for internal consistency; the implementation itself need not change. |

### Still blocking
- **F2** — The latency guide and cluster heading were corrected, and open issue #1668 genuinely covers the pre-existing `hw_config_by_gpu` documentation defect. However, the cluster guide still says the fixed per-collective launch/round-trip cost is “unmodeled,” while the current implementation models it through `InterNodeLatencyUs` and merely ships it uncalibrated at zero. The configuration summary also still lists `hw_config_by_gpu` as YAML-only despite the adjacent section correctly saying it is programmatic-only. Update those remaining statements for internal consistency; the implementation itself need not change.
  _docs/guide/latency-models.md:389-393 correctly says the cost is modeled but uncalibrated. sim/model_hardware_config.go:192-210 defines and documents `InterNodeLatencyUs`, and sim/latency/trained_physics_model.go:376-388 charges it. In contrast, docs/guide/cluster.md:67-68 says the fixed cost is “unmodeled.” docs/reference/configuration.md:460-474 now marks `hw_config_by_gpu` programmatic-only, but docs/reference/configuration.md:690 still lists it as YAML-only. sim/bundle.go:17-27 has no such `PolicyBundle` field and sim/bundle.go:111-124 enables strict known-field decoding. GitHub issue #1668 is OPEN and explicitly covers this exact `hw_config_by_gpu` documentation/parsing mismatch._

### Cleared
- _None._

namasl:
Both adjudication items fixed in `1e17345b`, plus two more that the same class of check found.

**1. `docs/guide/cluster.md` said the fixed per-collective cost is "unmodeled".** Correct catch, and mine — it contradicted `latency-models.md` and the code that charges it. Now uses the same three-state phrasing as the other guide: modeled, mechanism present, **uncalibrated at 0** so nothing is charged out of the box.

**2. `docs/reference/configuration.md`'s sub-config table still listed `hw_config_by_gpu` as YAML-only** while the prose section corrected it to programmatic-only. Agreed this one is in scope and distinct from #1668: that issue excuses not fixing the underlying `PolicyBundle` gap, not leaving one file contradicting itself after my edit. Table row fixed and still points at #1668 for the parsing question.

Thank you for explicitly accepting the #1668 disposition — that was the judgement call I most wanted checked.

**Two more, found by making the check structural instead of lexical**

The reason items 1 and 2 survived two earlier sweeps is that those sweeps grepped for the phrases I expected ("not yet priced", "unpriced", "not modeled") and neither item contains them. So I did a pass comparing *claims across documents* rather than matching strings, which turned up:

- `cluster.md` said "`blis replay`/`observe` reject node pools". Replay rejects; **observe cannot express them at all** — no `--policy-config`, no simulator. `latency-models.md` and the code comments already said it correctly, so this was another cross-document disagreement.
- `cmd/network_topology.go` was **missing** from `docs/reference/project-structure.md` while the other two new files were listed, and the `sim/network_topology.go` entry still implied `validate()` was exported surface.

That second one is worth naming: a sweep looks for *wrong* text, so it structurally cannot find *missing* text. Both are now listed.

**Verification on `1e17345b`** — run locally, since all four `@claude` review environments reported no Go toolchain: `go build ./...` clean, `go test ./... -count=1` all packages pass, `golangci-lint run ./...` 0 issues. Doc-only commit; `git diff --stat -- '*.md'` shows 4 insertions / 3 deletions across three files, no section removals.

namasl:

Note on the archon gate for this head (`1e17345b`): its last **successful** run is on `7b35948c` (`ARCHITECTURAL_CHANGE`, **0 violations**). Three subsequent runs failed with `The job was not acquired by Runner of type self-hosted even after multiple attempts` — a runner-availability failure, not a finding. The delta `7b35948c..1e17345b` is documentation-only (`git diff --name-only | grep -c '\.go$'` → **0**; three `.md` files, 4 insertions / 3 deletions), so archon's package-altitude analysis — packages, edges, exported surface, schema, invariants — is provably unchanged from the verdict it already gave. I will re-run it when the runner returns.

namasl:

## qa-review adjudication — PR #1667: ✅ PASS

|----|----|----|----|
| F2 | FLAW_FOUND | 🟡 WAIVED (deferred) | The two PR-introduced contradictions are corrected in the current documentation. The remaining pre-existing `hw_config_by_gpu` documentation defect is explicitly disclosed and is genuinely covered by open follow-up issue #1668. |

### Still blocking

### Cleared
- **F2** (🟡 WAIVED (deferred)) — The two PR-introduced contradictions are corrected in the current documentation. The remaining pre-existing `hw_config_by_gpu` documentation defect is explicitly disclosed and is genuinely covered by open follow-up issue #1668.
  _docs/guide/latency-models.md:389-393 now says per-collective latency is modeled through `InterNodeLatencyUs` but uncalibrated at 0. docs/guide/cluster.md:67-68 now accurately distinguishes priced bandwidth from uncalibrated latency. docs/reference/configuration.md:460-474 and :690 explicitly mark `hw_config_by_gpu` as programmatic-only and not a policy-bundle key. sim/bundle.go:17-27 confirms `PolicyBundle` lacks that field, while sim/bundle.go:111-124 confirms strict decoding. GitHub issue #1668 is OPEN and specifically covers both the unavailable policy key and the unusable documented YAML field names._

namasl:

Re-run on the unchanged head `1e17345b` — the previous run ([33712883487](https://github.com/inference-sim/inference-sim/actions/runs/33712883487)) reported `failure` with no error annotation and no findings, so this is to displace that status with a completed run, not a response to anything. Round 4 on `7b35948c` was ✅ APPROVED — READY TO MERGE, and the delta since is documentation-only (three `.md` files, zero `.go` files).

**Archon status, stated plainly:** its last **successful** verdict is on `7b35948c` — `ARCHITECTURAL_CHANGE`, **0 violations**. Every attempt since has failed or been cancelled on infrastructure, never on a finding: three `The job was not acquired by Runner of type self-hosted even after multiple attempts`, two `cancelled` by workflow concurrency. Since archon derives its verdict from Go package altitude — packages, edges, exported surface, schema, invariants — and `git diff --name-only 7b35948c 1e17345b | grep -c '\.go$'` is **0**, its analysis cannot differ from the verdict it already gave. Flagging rather than papering over it.

**Local verification on `1e17345b`** (all review environments so far have reported no Go toolchain, so this is the only build evidence): `go build ./...` clean, `go test ./... -count=1` all packages pass, `golangci-lint run ./...` **0 issues**.

namasl:

Third attempt to get a completed run on this context for the unchanged head `1e17345b`. The two prior attempts did not report findings — [33712883487](https://github.com/inference-sim/inference-sim/actions/runs/33712883487) ended `failure` after 26s with no error annotation (only benign `warn: Failed to compute SHA for …` lines for the new files), and [33716149688](https://github.com/inference-sim/inference-sim/actions/runs/33716149688) was `cancelled` after being requeued. The self-hosted runner pool looks unstable at the moment: the Archon workflow on this same head has failed three times with `The job was not acquired by Runner of type self-hosted even after multiple attempts` and been cancelled twice.

Substantive position is unchanged and does not depend on this run: Round 4 on `7b35948c` was ✅ **APPROVED — READY TO MERGE**, the qa-review adjudication on `1e17345b` is ✅ **PASS**, and the delta between those two heads is documentation-only (three `.md` files, **zero** `.go` files). All ten substantive CI checks — 2× build, lint, 7 test shards — pass on `1e17345b`. Local verification on that head: `go build ./...` clean, `go test ./... -count=1` 14 packages ok / 0 failures, `golangci-lint run ./...` 0 issues.

namasl:
@claude See your response above, you got stuck. Try again

namasl:
Rebased onto `main` after #1660 (`017fcd98`) merged. New head `ce1f1505`, 7 commits, `mergeable=MERGEABLE`.

**One conflict, resolved as a union.** `docs/concepts/glossary.md` — #1660 added a clause about #1656 sharding routed-expert weights across the `TP·DP` EP group for KV sizing; I had added a sentence about the TP all-reduce being charged cross-node. Orthogonal, so I took #1660's text and re-inserted my sentence at the same anchor, then asserted both sides' distinctive strings survive. Everything else (`CLAUDE.md`, `docs/guide/latency-models.md`, `docs/reference/models.md`, `sim/config.go`) auto-merged; verified by grep that both #1660's and my content are present in each, and that `models.md` is byte-identical to `main`.

**Substantive interaction with #1660 — checked, no contradiction.** #1660 concluded that vLLM flattens TP across DP unconditionally, so per-GPU routed-expert bytes are `R/(TP·DP)` in both EP modes and the step-time model's flattened-group treatment is correct. My all-to-all penalty is scored over `G = TP·DP` (`docs/guide/latency-models.md:233`), i.e. the same flattened group — so the two PRs now agree rather than disagree. I asserted nothing about EP-mode agnosticism of my own; the only such text on my branch is the pre-existing #1531 sentence, identical on both sides.

**Re-verification on the new base** (the rebase voids the prior round-4 APPROVED and qa-review PASS):

- `go build ./...` clean · `go test ./... -count=1` **14 packages ok, 0 failures** · `golangci-lint run ./...` **0 issues**.
- The three inertness gates and INV-BC-DP1 re-run green on the new base: `TestStepTime_{NoTopology,ContainedGroup,UncalibratedFabric}IsByteIdentical`, `TestINVBCDP1_{DenseDP1UnaffectedByTopology,DenseStepTimeByteIdentical,DenseDP1Determinism}`, `TestPlacementTopology_{SingleNodeAndUncalibratedAreByteIdentical,SpanningRunIsByteIdenticalAcrossRuns}`, plus all six golden datasets.
- **CLI byte-identity against the NEW main**, which is the check worth doing rather than assuming — #1660 changed KV capacity, and capacity feeds batching: identical stdout for `qwen3-14b`, `qwen3-14b --tp 4`, `llama-3.3-70b --tp 8`, and `mixtral-8x7b --tp 4` (the MoE case its EP-aware sizing could have touched).
- Feature still works end to end on the new base: a `tp=8` instance across two 4-GPU nodes warns once, writes `max_nodes_spanned: 2`, and replay of that trace exits 1 with `cannot reproduce this trace`.

/archon-pr-review

namasl:

Fresh review requested on the rebased head `ce1f1505` — the rebase onto #1660 voids the prior round-4 APPROVED (which was on `7b35948c`) and the prior qa-review PASS, so this is a re-review rather than a re-trigger.

Archon has now run cleanly on this head: `017fcd98 → ce1f1505`, `ARCHITECTURAL_CHANGE`, **0 violations**, 0 packages added, 0 edges added (2 surface · 2 schema · 5 invariant, unchanged in kind from its earlier verdicts).

Worth your attention specifically: the one merge conflict was in `docs/concepts/glossary.md`, where #1660 and I had edited the same paragraph about the same thing (the routed-expert `TP·DP` group). I resolved it as a union rather than picking a side, and checked the substantive question behind it — #1660 concluded that vLLM flattens TP across DP unconditionally so the step-time model's flattened-group treatment is correct, and my all-to-all penalty is scored over that same `G = TP·DP` group. So the two agree. Please confirm you read it the same way.

Local verification on `ce1f1505` (this remains the only build evidence — all five review environments so far have reported no Go toolchain): `go build ./...` clean, `go test ./... -count=1` **14 packages ok / 0 failures**, `golangci-lint run ./...` **0 issues**. The three inertness gates, INV-BC-DP1 and all six golden datasets re-run green on the new base, and CLI stdout is byte-identical to the new `main` for four configs including `mixtral-8x7b --tp 4` — the MoE case #1660's EP-aware capacity change could have perturbed.

namasl:
Rebased again onto `main` after #1659 (`93dbdfb9`). New head `c2240eb0`, 8 commits, `mergeable=MERGEABLE`. No conflict this time — but I rebased anyway, because #1659 added a test about **step time** and this PR changes **step time**, so the combination was untested.

**The interaction was worth checking and it holds.** `TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens` passes on the combined tree. The two features meet on one quantity — spec-decode scales the decode token population by verify width `K+1`, and the comm bases I re-scale consume that same `totalTokens` — so I added a permanent cross-feature guard rather than checking once and moving on: `TestStepTime_CrossNodePenaltyComposesWithSpecDecode` asserts the verify-width relationship survives with a cross-node penalty active (both halves), and symmetrically that the cross-node penalty survives under spec-decode. Both directions matter: an additive cross-node term could have masked the verify-width effect, and a per-token rather than per-collective latency would have been inflated by spec-decode. That also closes a coverage gap an earlier round named.

Verified all three PRs' content now coexists: #1656's EP-sizing paragraph, #1659's spec-decode paragraph and my #1530 section are all present in `CLAUDE.md`; #1659's "Progress stops at the completion boundary" text and INV-9 carve-out survived; and `models.md`'s MTP anchor is still the **single**-hyphen `#speculative-decoding-mtp-1528` (a double hyphen 404s under mkdocs `--strict`, so I left it alone).

**Verification on `c2240eb0`** — `go build ./...` clean, `go test ./... -count=1` **14 packages ok / 0 failures**, `golangci-lint run ./...` **0 issues**. Three inertness gates, INV-BC-DP1, #1659's spec-decode guard and all six golden datasets green. CLI stdout byte-identical to the new `main` for four configs, now including `--num-speculative-tokens 4 --speculative-acceptance-rate 0.7` alongside `mixtral-8x7b --tp 4`.

/archon-pr-review

namasl:
@claude /blis-pr-review

Head `c2240eb0`, rebased onto `93dbdfb9` (#1659). Archon on this exact head: `ARCHITECTURAL_CHANGE`, **0 violations**, 0 packages/edges added.

Keeping this short deliberately — the previous attempt on `ce1f1505` failed with `"compact_result": "failed"` / `is_error:true` after 44s, i.e. the action exhausted its context, and this thread is long. Prior substantive verdict was ✅ APPROVED — READY TO MERGE on `7b35948c`; since then: two rebases and one added test.

What changed since that approval, in full:
1. Rebase onto #1660 — one conflict in `docs/concepts/glossary.md`, resolved as a union of both sides.
2. Rebase onto #1659 — no conflict.
3. `TestStepTime_CrossNodePenaltyComposesWithSpecDecode` — a new cross-feature guard, because #1659's step-time test and my step-time change meet on `totalTokens`.

Local on `c2240eb0`: `go build` clean, `go test ./... -count=1` 14 packages ok / 0 failures, `golangci-lint` 0 issues. Inertness gates, INV-BC-DP1, #1659's spec-decode guard and all six goldens green; CLI stdout byte-identical to `main` across four configs.

namasl:
Rebased a third time, onto `e3598b84` (#1664). Head `4be14710`, 8 commits, `mergeable=MERGEABLE`.

Same single conflict as the #1660 rebase — `docs/concepts/glossary.md`, the tensor-parallelism entry, which all three siblings and I edited. #1664 updated its DP-as-placement clause to cover `blis replay` (#1556). Resolved as a union again: took main's newer text, re-inserted my cross-node sentence at the β₄ anchor, and asserted programmatically that **all four** contributions survive — #1660's `#1656` EP-sizing clause, #1664's `blis replay` (#1556) text, and my #1530 sentence. Verified the same for `CLAUDE.md` (all three sibling paragraphs plus my section present).

Verification on `4be14710`: `go build ./...` clean, `go test ./... -count=1` **14 packages ok / 0 failures**, `golangci-lint run ./...` **0 issues**. Inertness gates, INV-BC-DP1, #1659's spec-decode guard, my replay-fence subprocess tests and all six goldens green. CLI stdout byte-identical to the new `main` across four configs (including `mixtral-8x7b --tp 4` and a spec-decode config).

/archon-pr-review

namasl:
@claude /blis-pr-review

Head `4be14710` on `e3598b84`. Archon this head: 0 violations.

Prior verdict was APPROVED (round 4, `7b35948c`). Since then: three rebases (#1660, #1659, #1664) and one added test.

Local: build clean, `go test ./... -count=1` 14 ok / 0 fail, lint 0 issues.

namasl:
Conflict map from #1664 (issue #1556, DP-as-placement parity for `blis replay`), which merged as squash **`e3598b84`** and put this PR back to `CONFLICTING`. **All line numbers below are post-merge, on `main` at `e3598b84`.**

My squash touched 13 files. Your Go overlap is `cmd/root.go` and `cmd/replay.go`; the rest is docs.

```
CLAUDE.md                                 |   4 +-
cmd/dp_ep_cli_test.go                     |  37 +-
cmd/dp_placement_test.go                  | 465 ++++++++---
cmd/dp_replay_parity_test.go              | 647 +++++++++++++++  (new file)
cmd/replay.go                             |  54 +-
cmd/root.go                               | 251 ++++---
docs/concepts/glossary.md                 |   2 +-
docs/concepts/roofline.md                 |   2 +-
docs/contributing/standards/invariants.md |   4 +-
docs/guide/latency-models.md              |   8 +-
docs/guide/observe-replay-calibrate.md    |  48 +-
docs/reference/configuration.md           |   3 +
docs/reference/models.md                  |   2 +-
```

## `cmd/root.go` — five regions

| Line | What is there now |
|---|---|
| **400** | `const dpPlacementInstanceWarnThreshold = 512` (unchanged anchor) |
| **439 / 445** | #1660's `epSizeForKVCapacity` / `formatEPForLog` — **not mine**, but I edited the doc comment above 439 (see below) |
| **478** | `planDPPlacement` — pure decision + the #1548/#1553 guards. Doc comment rewritten by me |
| **528** | `applyDPPlacement` — **signature changed** to `(plan, dp, dep dpPlacementDeployment, autoScaledKV bool, blockSizeTokens int64) (dpPlacementDeployment, error)`. It absorbed the per-rank `--max-model-len` re-cap and gained a zero-block guard |
| **594** | **`resolveDPPlacement` (new)** — the shared resolver. Reads the `cmd` package flag vars and writes exactly three: `numInstances, totalKVBlocks, maxModelLen` |
| **1006** | a `LATENT HOLE — READ THIS BEFORE LIFTING #1553` comment I added inside `resolveLatencyConfig`'s BC-ROOFLINE gate |
| **2404** | `dpPlan, dpErr := resolveDPPlacement(lr, bundleAutoscalerIntervalUs > 0, len(bundleNodePools) > 0)` in `runCmd` |
| **2414** | the `Starting simulation with %d KV blocks` log — **I moved this**, it now sits *after* the DP block so it reports the per-replica figure |
| **2451** | `sim.NewModelHardwareConfig(..., tensorParallelism, dpPlan.PerRankDP, enableExpertParallel, ...)` |

`resolveLatencyConfig` itself starts at **657**; the global auto-calc `epSize := epSizeForKVCapacity(...)` is at **916**.

## `cmd/replay.go` — two regions

- **Old ~295: the `--dp>1` MoE guard is GONE.** I deleted the `logrus.Fatalf("--dp %d on an MoE model is not supported in blis replay...")` block that sat right after `lr := resolveLatencyConfig(cmd)` (now **287**). If your rebase resurrects it, replay parity is silently reverted and four tests go red.
- **597–626: new DP block.** Comment at **597**, `resolveDPPlacement` call at **620**, `Starting replay with %d KV blocks` at **627**, and `NewModelHardwareConfig` at **648** now threading `dpPlan.PerRankDP` (was raw `dataParallelism`).
- #1660's per-pool `epSizeForKVCapacity` calls are at **464** and **501** — *before* my block, which matters (below).

## Docs — exact lines

- `docs/contributing/standards/invariants.md` **180** (INV-13 Statement — I strengthened it to byte-identical-stdout + added the `--horizon` precondition) and **182** (Verification — added four test names).
- `CLAUDE.md` **473** (INV-13 bullet) and **506** (the DP-as-real-placement paragraph). #1660's `#1656` paragraph is at **649** and #1659 touched **651** — all insertion-adjacent, so "keep every paragraph", never a content merge.
- `docs/reference/configuration.md` **149** (`--dp` row; I also added `--enable-expert-parallel` and `--moe-comm-backend` rows just below).
- `docs/guide/observe-replay-calibrate.md` **228** (new DP-on-replay admonition with a worked example).
- `docs/concepts/roofline.md` **10**, `docs/guide/latency-models.md` **133** (β_EP) / **180** / **224**, `docs/reference/models.md` **21**, `docs/concepts/glossary.md` **191**.

**Heads-up on the four doc sentences we both touch.** `roofline.md:10`, `latency-models.md:224`, `models.md:21` and `glossary.md:191` each already carry a *three-way* merge — #1660's EP-sizing content plus my DP-on-both-paths content, merged by hand rather than pick-a-side. Since you also edit these files, resolve by **combining**, not by taking either side: dropping the `#1656` EP-capacity clause or the `#1556` both-commands clause would each silently re-introduce a false statement. `roofline.md:10` is the subtle one — it now says roofline *rejects* `--dp`/`--enable-expert-parallel` with a fatal error (verified: exit 1) rather than "does not model" them, **and** carries #1660's parenthetical that EP still affects KV-capacity sizing.

## What a bad rebase silently breaks, and what catches it

**1. Partially dropping `resolveDPPlacement` so run and replay diverge again.** Three independent tests, all in `cmd/`:
- `TestINV13_RunReplayParity_MoEDPPlacement` (`dp_replay_parity_test.go`) — drives real `blis run --dp 2 --trace-output` then `blis replay --dp 2` in subprocesses and requires **byte-identical stdout**, at TP=1 *and* TP=2. Mutation-verified: dropping the `numInstances` assign-back, or wiring raw `dataParallelism` into `NewModelHardwareConfig`, each fail it. It also asserts `instance_1` on **both** legs, so two identically-unexpanded legs cannot pass vacuously.
- `TestRunReplay_DPEPFlags_ThreadedIntoConstructor` (`dp_ep_cli_test.go:255`) — source-level guard requiring the string `tensorParallelism, dpPlan.PerRankDP, enableExpertParallel,` in **both** `root.go` and `replay.go`. This is the one that fires fastest if a rebase resolves `replay.go:648` back to `dataParallelism`.
- `TestReplayCmd_MoEDPPlacement_GuardedCombos_Rejected` — replay must still exit 1 naming `#1548` / `#1553`.

**2. Re-adding the deleted replay guard** fails all four of the above (the resurrected message lacks `#1548`/`#1553`).

**3. The ordering hazard, which I verified is safe — and why it stays safe for you.** #1660's `epSizeForKVCapacity` must read the **logical** `--dp`; if it saw the per-replica `DP=1` the EP group would collapse to `tp` and #1656 would silently void. It is safe for two independent reasons: **`dataParallelism` is never written anywhere in non-test code** (`resolveDPPlacement` writes only `numInstances/totalKVBlocks/maxModelLen`; the per-rank DP travels as the *returned* `dpPlan.PerRankDP`), and every `epSizeForKVCapacity` site (`root.go:916`, `root.go:1883/1920`, `replay.go:464/501`) precedes its `resolveDPPlacement` call (`root.go:2404`, `replay.go:620`). I left a note at `epSizeForKVCapacity` recording this. **If your rebase introduces any write to `dataParallelism`, that invariant breaks** — and note the failure would be silent for you.

**4. Explicitly stated because it is a gap, not a safety net:** #1660's `TestEveryKVCapacityCallSiteIsEPAware` only checks that each `CalculateKVBlocks` site *has* a topology-derived `WithExpertParallelSize` option — it does **not** validate the EP degree's value, so it cannot catch a pre/post-expansion ordering regression. Since #1667 changes **step time**, not KV capacity, that is most likely moot for you — but if you do touch a capacity call site, hand-verify with a `--tp N --enable-expert-parallel --dp 1` run vs replay and compare resolved block counts (mine agree at 227391 blocks, `EP=2`).

For reference, my post-rebase verification on `6fd47fbc` was: `go build` clean, `go test ./... -count=1` 14/14 packages, `golangci-lint` 0 issues, `TestEveryKVCapacityCallSiteIsEPAware` green, run→replay byte-identity at `dp=2/1`, `4/2`, `3/1`, and auto-KV parity identical (74312 blocks/replica, 1188992-token cap). Those are the checks I would re-run after your rebase.

namasl:
Already rebased onto `e3598b84` (#1664) before the conflict map arrived — head is now `f4f6f2cc`, 9 commits, `mergeable=MERGEABLE`. My `cmd/` changes turned out to be purely additive (20 insertions, 0 deletions across `root.go`/`replay.go`), so #1664's content came through intact — I diffed its `LATENT HOLE — READ THIS BEFORE LIFTING #1553` comment against `main` and it is byte-identical, and the `node_pools` rejection my new replay guard's comment refers to is still at `replay.go:371`, so that claim stays accurate.

**The roofline interaction was a real find — thank you for flagging it.** `docs/concepts/roofline.md:49` listed *"Communication Overhead: If using Tensor Parallelism, adds All-Reduce latency per layer"* among the components of roofline step time. That is false: every `tpFactor` use in `rooflineStepTime` is a **division** on compute or memory traffic — four sharding divisors plus the weight-bytes one — so TP communication is free in that backend at any TP, within a node or across nodes. I verified it in the code rather than inferring it from #1663.

It is pre-existing, but this PR is what turns it into a contradiction: two guides I edited now say roofline models no communication, #1664 corrected **line 10 of this same file** to say roofline is "DP/EP-blind for step time", and #1663 tracks the gap. Line 49 asserting the opposite would ship a page that disagrees with itself and with a sibling PR. So I corrected it in place (`f4f6f2cc`) rather than filing — one line, squarely in this PR's subject, and the underlying capability gap already has an issue, so nothing is silently absorbed. The two statements now read as complementary rather than competing: #1664's is about **rejecting** `--dp>1`/EP, mine is about there being **no comm term**.

Also confirmed I make no claim anywhere about replay's old `--dp>1` MoE guard, which #1664 removed — so nothing of mine went stale there.

**Verification on `f4f6f2cc`:** `go build ./...` clean, `go test ./... -count=1` **14 packages ok / 0 failures**, `golangci-lint run ./...` **0 issues**. Three inertness gates, INV-BC-DP1, #1659's `TestSpecDecode_DecodeStepCostIsVerifyWidth_NotGrantedTokens`, my replay-fence subprocess tests and all six goldens green. CLI stdout byte-identical to `main` across four configs.

/archon-pr-review

namasl:
@claude /blis-pr-review — head `f4f6f2cc`; archon 0 violations; prior verdict https://github.com/inference-sim/inference-sim/pull/1667#issuecomment-5527155061 (round 4 APPROVED). Details in the PR body and the comment above.

namasl:
@claude /blis-pr-review — head `f6af53d7`.

namasl:
@claude /blis-pr-review — head `84080394`; the 3 should-fix items are addressed.

susiejojo:
### BLIS PR Review — #1667 · Commit-Resolution Audit (HEAD `84080394`)

**Question asked:** *Are all commits / all review findings addressed?* This pass audits the 6 prior `blis-pr-review` rounds, the archon run, and the two-agent `qa-review` **⛔ BLOCK** against the actual PR head — verified by checking out `pull/1667/head` and grepping the exact file:line each finding named (not just trusting the thread).

---

#### Timeline reconciliation

The `qa-review` **⛔ BLOCK** (02:09 UTC, `F2 FLAW_FOUND`) and the Round-5 should-fix list **predate** the last five commits. Commits `4b6037bc → 84080394` (all ~19:13–19:53 UTC) were authored *specifically* to close them — one is literally titled *"docs: fix stale cross-node pricing claims found by qa-review."* So the BLOCK is stale w.r.t. HEAD; it was not overruled, it was fixed.

#### qa-review `F2` (the only FLAW_FOUND / block reason) — ✅ all three sub-claims resolved at HEAD

| Sub-claim | Status at `84080394` | Evidence |
|---|---|---|
| Latency guide says fixed per-collective cost "not modeled" | ✅ Fixed | `docs/guide/latency-models.md:250-269, 389-391` now document `InterNodeLatencyUs` as *modeled but uncalibrated (0)* |
| Cluster guide heading "not yet priced" | ✅ Fixed | `docs/guide/cluster.md:67-68` heading now reads *"bandwidth priced, per-collective latency not calibrated"*; body agrees |
| `hw_config_by_gpu` presented as a policy-bundle YAML key | ✅ Fixed | `docs/reference/configuration.md:477, 693` now explicitly labels it *"Programmatic-only, NOT a policy-bundle key (issue #1668)"* |

#### qa-review `F3` (stale comment) — ✅ resolved

`sim/network_topology.go:25-30` no longer claims "no round-trip needed"; it now documents the `MaxNodesSpanned` refusal-signal that this PR added.

#### Round-5 should-fix items — ✅ all three landed

1. `TestStepTime_MoEReduceLegChargedCrossNode` — `assert.Equal → assert.InDelta(…, 1)` (`network_cost_test.go:292`), with a comment explaining truncation noise.
2. `MaxNodesSpanned` godoc — moved below `SetArrivalHook`'s closing brace; `SetArrivalHook`'s panic-contract godoc restored (`cluster.go:747-777`).
3. Three comments citing non-existent `moeDispatchBytes` → corrected to `moeDispatchBasis` (0 remaining refs to the bad name; 10 correct refs).

#### Correctness fix from Round 5 (`6e10b150`) — ✅ landed
MoE now charges **dispatch AND combine** (2 collectives/layer), aligning the per-collective latency half with the volume half. Mutation-verified in the prior round (`c=1.0 → FAIL`).

---

#### Independent gate checks (this pass)
- **CI:** all `build`, `test (×7 shards)`, and `lint` checks **pass** at HEAD.
- **Archon:** `ARCHITECTURAL_CHANGE`, 0 violations. Boundary moves (`sim/cluster→sim`, `sim/latency→sim`) mirror the existing `HardwareCalib` pattern — intentional, no new package arrow, no cycle.
- **Merge state:** `MERGEABLE`; `mergeStateStatus: BLOCKED` reflects the review-approval gate, not a failing check.

#### Remaining open items — all correctly deferred with tracking issues (none blocking)
- `InterNodeLatencyUs` ships at 0 (calibration) → **#1661**
- Fabric keyed by `gpu_type`, not pool → **#1662**
- Roofline backend prices no communication (warns at runtime) → **#1663**
- DEP/all-to-all leg inert until EP-as-placement → **#1548/#1553** (formula + synthetic StepTime path tested)
- `hw_config_by_gpu` programmatic-only surface → **#1668**
- qa-review's "Important to consider" items (G1, G3, G7, G8, G9, G12) are all `CONFIDENT` acknowledgements of documented approximations, not defects.

---

### Verdict: ✅ **Yes — all commits/findings are addressed. Ready to merge.**

Every concrete finding raised across the 6 review rounds, the archon run, and the qa-review **⛔ BLOCK** has a corresponding fix commit verified present at HEAD `84080394`. CI is green, archon is clean, the math is mutation-verified, cross-path parity (run full / replay fenced on both config + trace paths / observe untouched) is complete, and the inertness gates (INV-6) have exact-equality tests. The only outstanding items are explicitly-tracked follow-up issues, none of which gate this PR.

> Audit method: fetched `pull/1667/head`, confirmed the fix commits (`4b6037bc..84080394`) post-date the qa-review BLOCK by ~17h, and grepped each named file:line to confirm resolution rather than relying on the review thread's self-reported status.

susiejojo (approved):
lgtm

PR #1674, opened 2026-09-03
fix(ci): dedupe the archon comment, check out the PR head, guard report-status

jgchn:
## Summary

Three CI defects found while triaging why `@claude /blis-pr-review` died on #1667 ([run 33772637816](https://github.com/inference-sim/inference-sim/actions/runs/33772637816)).

**Scope note (revised after review):** this PR does **not** claim to fix the context-window overflow. Only **two** archon dumps ever reached that failing prompt — `grep -c "^### Component view"` over the assembled `FINAL PROMPT` returns `2` — and they were the byte-identical pair. The other 12 were never in it. The whole prompt block is ~74 KB of a **244,554-token** request, so the thread is a *minority* of the window. Deduping ~21 KB is strictly good and **not sufficient**. The overflow proper, including the `CLAUDE.md` growth data, is **#1676**.

## Fix 1 — archon appended a report on every trigger

`gh pr comment` appends. On #1667 archon dumps were **365,295 of the thread's 470,270 bytes (77.7%)**, and the Claude action folded two **byte-identical** copies into one prompt (verified with `diff`). `--edit-last --create-if-none` keeps one archon comment per PR.

Checked before proposing it:

- `gh 2.83.2` supports both flags.
- `archon.yml:182` was the only `gh pr comment` in `.github/workflows/` — no `createComment`, no `gh api .../comments`, no `peter-evans/create-or-update-comment` either.
- `--edit-last` scopes to `github-actions[bot]`; the Claude action comments under a separate App identity. Identity scoping is a known future fragility → **#1677**.
- History is retained **per run, not permanently** — `emit()` in `scripts/archon-review.sh:60` appends every body to `$GITHUB_STEP_SUMMARY`, which lives as long as the run is retained. Docs say so explicitly.

## Fix 2 — `claude.yml` checked out the default branch

For `issue_comment`, `github.ref` is `main`, so `actions/checkout` landed there. The action's prompt assembly then ran `git hash-object` against main and failed for all **12 files #1667 adds**, dropping those SHAs and emitting ~48 lines of `could not open ... No such file`.

**The review itself was never on main** — `claude-code-action` re-checks-out `refs/pull/N/head` 13 lines later, before the prompt is built. But the noise reads as though it were, and it produced exactly that wrong diagnosis during triage. The ref is **conditional** because the `issues:` trigger has no `refs/pull/N/head`; empty string falls back to checkout's default.

## Fix 3 — `report-status` overwrote real results with a stale `pending` *(added after review)*

An independent defect in a file this PR already edits. The job ran under a bare `always()`, so when `claude[bot]`'s own review comment fired `issue_comment` and the `claude` job correctly **skipped**, this job still wrote `pending / Claude Code: skipped` over the good status.

Verified on head `4be14710`:

```
15:17:20  failure  Claude Code: failure
15:21:11  pending  Claude Code: skipped   ← lands last, sticks
```

Plus run `33770188054`: `claude: skipped`, `report-status: success`, 34 s after the real run `33770122756`. Result is a required check that never resolves, masking a genuine failure. Guarded with `needs.claude.result != 'skipped'`. A denied permission check now also writes no status — better than a stuck `pending`.

## Out of scope by design

```
Restoring .claude, CLAUDE.md, .mcp.json, ... from origin/main (PR head is untrusted)
```

A fork PR must not execute its own config, so a PR editing `CLAUDE.md` (#1667 does, +26/-2) or `.claude/skills/blis-pr-review/SKILL.md` is always reviewed by main's copy. Noted so it is not rediscovered as a bug.

## Follow-ups filed rather than scope-crept

- **#1675** — archon triggers on prose **mentions** of `/archon-pr-review`. `contains()` cannot tell a mention from an invocation, so a comment *discussing* the command re-reviews the PR. Two comments on this very PR each fired a run ~4 s later, producing byte-identical reports for the same head. Explains #1667's 14 comments for 9 real triggers. `--edit-last` hides the bloat but not the runs.
- **#1676** — the context-window problem, with `CLAUDE.md` growth (65,478 → 80,500 B in 10 days; 79% of it is four reference sections that already name a canonical source elsewhere).
- **#1677** — marker-scoped comment lookup and a test for the event→ref mapping, plus `actionlint`.

## Testing

- `go build ./...` passes (no Go files changed).
- All five workflow YAMLs parse; the archon step body passes `bash -n`.
- Ref expression traced over all four trigger shapes: `pull_request_review_comment` → `refs/pull/<pr.number>/head`; `issue_comment` on a PR → `refs/pull/<issue.number>/head`; `issue_comment` on a plain issue → `''`; `issues` → `''`.
- `report-status` guard traced: `success`/`failure`/`cancelled` still report; `skipped` does not.

**Neither workflow fix is testable before merge.** `issue_comment` workflows always run from the default branch — both archon runs on this PR report `branch=main`, and the two byte-identical 5,548 B comments here are the **old appending code**. Merge, then watch the first `/archon-pr-review` and the first `@claude` run. Both changes are one-line reverts if they misbehave.

**Tier:** Small — 4 files, CI config + docs, no Go behavioral change.

### DIFF
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -179,13 +179,15 @@ jobs:
             echo "::warning::Archon produced no output. Skipping PR comment."
             exit 0
           fi
-          # --edit-last keeps ONE archon comment per PR instead of appending a ~27 KB
-          # report on every push. A long-running PR otherwise accumulates hundreds of KB
-          # of near-identical comments, which the Claude review action later folds into its
-          # prompt until it exceeds the model context window and the review dies with
-          # "Prompt is too long" (run 33772637816: 244,554 tokens against a 200,000 limit).
-          # Per-round history is preserved: emit() in scripts/archon-review.sh also appends
-          # every body to $GITHUB_STEP_SUMMARY, so each round stays on its own run page.
+          # --edit-last keeps ONE archon comment per PR instead of appending a ~21 KB report
+          # on every trigger (this workflow fires on a comment, never directly on a push).
+          # On #1667 archon dumps were 365,295 of the thread's 470,270 bytes (77.7%), and the
+          # Claude review action folded two BYTE-IDENTICAL copies into a single prompt.
+          # Scope: this bounds archon's contribution to the thread. It does NOT by itself
+          # prove a review fits the model context window — see #1676.
+          # History is retained per run, not permanently: emit() in scripts/archon-review.sh
+          # appends every body to $GITHUB_STEP_SUMMARY, which lives as long as the workflow
+          # run is retained.
           # --edit-last scopes to the authenticated user (github-actions[bot]); the Claude
           # action comments under a separate App identity, so this cannot edit those.
           gh pr comment "$PR_NUMBER" --body-file "$OUTPUT_FILE" \
--- a/.github/workflows/archon.yml
+++ b/.github/workflows/archon.yml
@@ -179,4 +179,14 @@ jobs:
             echo "::warning::Archon produced no output. Skipping PR comment."
             exit 0
           fi
-          gh pr comment "$PR_NUMBER" --body-file "$OUTPUT_FILE"
+          # --edit-last keeps ONE archon comment per PR instead of appending a ~27 KB
+          # report on every push. A long-running PR otherwise accumulates hundreds of KB
+          # of near-identical comments, which the Claude review action later folds into its
+          # prompt until it exceeds the model context window and the review dies with
+          # "Prompt is too long" (run 33772637816: 244,554 tokens against a 200,000 limit).
+          # Per-round history is preserved: emit() in scripts/archon-review.sh also appends
+          # every body to $GITHUB_STEP_SUMMARY, so each round stays on its own run page.
+          # --edit-last scopes to the authenticated user (github-actions[bot]); the Claude
+          # action comments under a separate App identity, so this cannot edit those.
+          gh pr comment "$PR_NUMBER" --body-file "$OUTPUT_FILE" \
+            --edit-last --create-if-none
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -60,6 +60,21 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
+          # For issue_comment/pull_request_review_comment, github.ref is the DEFAULT
+          # branch — GitHub does not associate the event with the PR head. Without an
+          # explicit ref the action's own prompt assembly runs `git hash-object` against
+          # main and fails for every file the PR ADDS, dropping those SHAs from the prompt
+          # and emitting dozens of misleading "could not open ... No such file" lines
+          # (12 files on PR #1667). claude-code-action re-checks-out refs/pull/N/head
+          # afterwards, so the review itself was never on main — but the noise reads as
+          # though it were. Empty string falls back to checkout's default, which the
+          # `issues:` trigger needs: a plain issue has no refs/pull/N/head to check out.
+          ref: >-
+            ${{ github.event_name == 'pull_request_review_comment'
+                && format('refs/pull/{0}/head', github.event.pull_request.number)
+             || (github.event.issue.pull_request
+                && format('refs/pull/{0}/head', github.event.issue.number))
+             || '' }}
 
       - uses: anthropics/claude-code-action@v1
         env:
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -96,7 +96,15 @@ jobs:
 
   report-status:
     needs: claude
-    if: always() && github.event_name == 'issue_comment' && github.event.issue.pull_request
+    # needs.claude.result != 'skipped' is load-bearing. claude[bot]'s own review comment
+    # fires issue_comment, so the claude job correctly skips — but with a bare always()
+    # this job still ran and wrote "pending / Claude Code: skipped" OVER the real result.
+    # Verified on head 4be14710: failure @ 15:17:20 then pending @ 15:21:11, stale last and
+    # sticking, masking a genuine failure behind a required check that never resolves
+    # (run 33770188054: claude skipped, report-status success, 34 s after the real run).
+    if: |
+      always() && needs.claude.result != 'skipped' &&
+      github.event_name == 'issue_comment' && github.event.issue.pull_request
     runs-on: ubuntu-latest
     permissions:
       statuses: write
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -55,6 +55,13 @@ jobs:
       pull-requests: write
       issues: write
       id-token: write
+    outputs:
+      # Both are EMPTY when the action returns early ("No trigger found, skipping remaining
+      # steps") — the outer trigger below is a broad substring test, so a comment that merely
+      # mentions @claude in prose admits the job and the action then no-ops. The job still
+      # concludes 'success', so report-status must not read the job result alone (see #1675).
+      execution_file: ${{ steps.claude.outputs.execution_file }}
+      conclusion: ${{ steps.claude.outputs.conclusion }}
 
     steps:
       - uses: actions/checkout@v5
@@ -77,6 +84,7 @@ jobs:
              || '' }}
 
       - uses: anthropics/claude-code-action@v1
+        id: claude
         env:
           ANTHROPIC_BASE_URL: ${{ secrets.LITELLM_BASE_URL }}
           CLAUDE_CODE_DISABLE_THINKING: "1"
@@ -96,15 +104,29 @@ jobs:
 
   report-status:
     needs: claude
-    # needs.claude.result != 'skipped' is load-bearing. claude[bot]'s own review comment
-    # fires issue_comment, so the claude job correctly skips — but with a bare always()
-    # this job still ran and wrote "pending / Claude Code: skipped" OVER the real result.
-    # Verified on head 4be14710: failure @ 15:17:20 then pending @ 15:21:11, stale last and
-    # sticking, masking a genuine failure behind a required check that never resolves
-    # (run 33770188054: claude skipped, report-status success, 34 s after the real run).
+    # Publish a verdict ONLY when Claude actually ran. Two no-op paths otherwise corrupt the
+    # status, and the job RESULT alone cannot tell them apart:
+    #
+    #  - claude SKIPPED (permissions denied, or claude[bot]'s own comment fired the event).
+    #    A bare always() wrote "pending / Claude Code: skipped" OVER the real result. Verified
+    #    on head 4be14710: failure @ 15:17:20 then pending @ 15:21:11 — stale lands last and
+    #    sticks, hiding a genuine failure behind a check that never resolves (run 33770188054).
+    #
+    #  - claude SUCCEEDED WITHOUT RUNNING. The outer trigger is a broad substring test, so a
+    #    comment mentioning @claude in prose admits the job; the action logs "No trigger found,
+    #    skipping remaining steps" and returns, and the job still concludes 'success'. That
+    #    published a FALSE GREEN: run 33785598570 put "Claude Code: success" on head 6c7c3a81
+    #    with no review having run. execution_file is empty on that early return, so it — not
+    #    the job result — is the signal that work happened.
+    #
+    # Stated positively so a new job result cannot silently fall through to a status. A real
+    # success that somehow reported no execution_file publishes NOTHING rather than a verdict:
+    # a missing status is visible and honest, a wrong one is neither.
     if: |
-      always() && needs.claude.result != 'skipped' &&
-      github.event_name == 'issue_comment' && github.event.issue.pull_request
+      always() &&
+      github.event_name == 'issue_comment' && github.event.issue.pull_request &&
+      (needs.claude.result == 'failure' || needs.claude.result == 'cancelled' ||
+       (needs.claude.result == 'success' && needs.claude.outputs.execution_file != ''))
     runs-on: ubuntu-latest
     permissions:
       statuses: write
@@ -137,11 +159,14 @@ jobs:
           script: |
             const result = process.env.CLAUDE_RESULT;
             const sha = process.env.PR_HEAD_SHA;
+            // No 'skipped' entry: the job condition above excludes skipped runs, so a
+            // skipped result reaching here means that gate regressed. Fall through to the
+            // warning + 'error' path rather than silently re-publishing a stale 'pending'
+            // over a real verdict.
             const stateMap = {
               success: 'success',
               failure: 'failure',
-              cancelled: 'error',
-              skipped: 'pending'
+              cancelled: 'error'
             };
             const state = stateMap[result];
             if (!state) {
--- a/docs/contributing/archon/pr-review-context.md
+++ b/docs/contributing/archon/pr-review-context.md
@@ -9,7 +9,9 @@ How to incorporate archon's PR review output during code review.
 ## What the comment contains
 
 One archon comment per PR — re-running `/archon-pr-review` edits it in place, so what you
-read is always the latest round. It comes in one of two shapes:
+read is always the latest round. Earlier rounds are not in the thread; they are in each
+workflow run's job summary, for as long as that run is retained. It comes in one of two
+shapes:
 
 - **three views** — component, witness delta, interface-contract
 - **three views plus plan checks** — the same, with `### G5 — Plan distance ratchet` and a plan verdict appended
--- a/docs/contributing/archon/pr-review-context.md
+++ b/docs/contributing/archon/pr-review-context.md
@@ -8,7 +8,8 @@ How to incorporate archon's PR review output during code review.
 
 ## What the comment contains
 
-One comment per `/archon-pr-review`, in one of two shapes:
+One archon comment per PR — re-running `/archon-pr-review` edits it in place, so what you
+read is always the latest round. It comes in one of two shapes:
 
 - **three views** — component, witness delta, interface-contract
 - **three views plus plan checks** — the same, with `### G5 — Plan distance ratchet` and a plan verdict appended
--- a/docs/guide/archon-pr-review.md
+++ b/docs/guide/archon-pr-review.md
@@ -12,8 +12,11 @@ Comment on any PR:
 
 The GitHub Action runs automatically and posts the results as a PR comment. Re-running
 `/archon-pr-review` **updates that same comment in place** rather than adding a new one, so
-a long-running PR keeps one current report instead of a stack of near-identical ones. Every
-round's full body is still kept on its own workflow run page (job summary).
+a long-running PR keeps one current report instead of a stack of near-identical ones.
+
+Each round's full body is also written to its own workflow run's job summary. That is
+retention-bound, not an archive: it disappears when the run is aged out under the
+repository's retention policy. If you need a round preserved, quote it in a PR comment.
 
 ## What It Reports
 
--- a/docs/guide/archon-pr-review.md
+++ b/docs/guide/archon-pr-review.md
@@ -10,7 +10,10 @@ Comment on any PR:
 /archon-pr-review
 ```
 
-The GitHub Action runs automatically and posts a comment with the results.
+The GitHub Action runs automatically and posts the results as a PR comment. Re-running
+`/archon-pr-review` **updates that same comment in place** rather than adding a new one, so
+a long-running PR keeps one current report instead of a stack of near-identical ones. Every
+round's full body is still kept on its own workflow run page (job summary).
 
 ## What It Reports
 
@@ -44,7 +47,7 @@ If the PR changes package boundaries, Archon reports three sections:
 2. Runs `archon-go delta --json` to detect structural changes
 3. If empty: posts fast-track message
 4. If non-empty: runs `archon-go delta` (human report), `archon-go impact` (blast radius per changed package), and `archon-go evidence` (contract test coverage)
-5. Posts combined output as a PR comment
+5. Posts combined output as a PR comment, editing the existing archon comment if one exists
 
 ## Key Properties
 


namasl:

## qa-review — PR #1674: ⛔ BLOCK
_fix(ci): stop the review prompt outgrowing the model context window · 13 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes, for the PR's stated CI-config scope. It changes Archon posting from append to edit-or-create, checks out the PR head for PR-associated Claude triggers while retaining default checkout for issue triggers, and updates the two relevant Archon guides. The documentation additions are directly related rather than unnecessary. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. Both existing Archon user/reviewer guides now describe the single current comment and per-run job-summary history. The Claude checkout change is internal workflow behavior and does not expose a user-facing interface requiring separate guide changes. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale comments were found. The new comments match the implemented behavior: the script appends the full body to GITHUB_STEP_SUMMARY, the checkout expression falls back to the event's default when it resolves to an empty string, and the posting command uses edit-or-create. |
| G1 | correctness | ❌ FLAW_FOUND | In `archon.yml`, `--edit-last` targets the most recent comment on the PR authored by the authenticated bot (`github-actions[bot]`). If another GitHub Actions workflow (such as a benchmark or coverage reporter) posts a PR comment between Archon runs, will `gh pr comment --edit-last` overwrite that unrelated bot comment instead of Archon's report? | Yes. `--edit-last` selects the authenticated user's most recent comment, not the most recent comment carrying an Archon-specific identity. All workflows using the repository GITHUB_TOKEN comment as github-actions[bot], so an unrelated workflow comment posted after Archon's prior report would be overwritten. Archon's older report would remain, meaning the documented one-current-comment behavior would also cease to hold. |
| G2 | correctness | ✅ CONFIDENT | In `claude.yml`, the expression checks `github.event.issue.pull_request` for `issue_comment` events. What is the exact schema type of `github.event.issue.pull_request` when an issue comment is posted on a plain issue versus a PR, and could an empty object `{}` evaluate as truthy in GitHub Actions context expressions? | For an `issue_comment` payload on a plain issue, the `issue.pull_request` member is absent and therefore resolves as null in a GitHub Actions expression. For a comment on a PR, it is an object containing PR-link metadata such as `url`, `html_url`, `diff_url`, and `patch_url`. GitHub Actions lists null, false, zero, and empty strings as falsy; objects are not falsy, so even `{}` would take the true branch. Normal webhook payloads do not use an empty object here: they omit the member for plain issues and supply a populated object for PRs. |
| G3 | tests | ⚠️ CANNOT_ANSWER | The PR author notes that checking out `refs/pull/N/head` results in a detached HEAD state that could not be tested offline against `claude-code-action`'s internal `git fetch` and `git switch` logic. How can we verify before merge that `claude-code-action` handles a detached HEAD checkout without failing branch detection? | The worktree cannot prove the external action's detached-HEAD compatibility. It references the moving `anthropics/claude-code-action@v1`, and the action implementation is not vendored here. Before merge, the reliable verification is a live disposable PR: trigger both an issue comment and a review comment, confirm checkout reports `refs/pull/N/head`, and confirm the Claude action completes its fetch/switch and reviews the expected head SHA. Pinning the action to a tested commit would make that result reproducible. |
| G4 | tests | ❌ FLAW_FOUND | Given that workflow YAML expressions cannot be unit tested with standard Go tests, what test steps or workflow dry-run strategies can be added to ensure future edits to `claude.yml` event triggers do not silently break the conditional `ref` construction? | There is no committed automated check for the new event-to-ref mapping, despite all four event shapes being realistic workflow inputs. Add a small resolver script or local JavaScript module accepting an event name and payload, table-test it with fixtures for PR issue comments, plain issue comments, review comments, and `issues` events, and feed its output into checkout. As an integration layer, run `act` with those fixture payloads or add a temporary diagnostic workflow that prints and validates the resolved ref without invoking Claude. `actionlint` should also be run for syntax/schema validation, though it cannot alone prove the branch semantics. |
| G5 | rationale | ✅ CONFIDENT | In `archon.yml`, why was `--edit-last` chosen over searching for a unique metadata tag (such as an embedded `<!-- archon-report -->` HTML comment) to identify and update Archon's comment specifically? | The apparent rationale is simplicity: `gh pr comment --edit-last --create-if-none` provides edit-or-create behavior in one command, and currently Archon is the only repository workflow posting PR comments through this bot identity. A metadata tag would require listing comments, locating the tagged comment, and then updating its ID through `gh api` or equivalent. That choice is adequate under the current workflow set, although it is less robust to future same-bot commenters. |
| G6 | rationale | ✅ CONFIDENT | In `claude.yml`, why was an inline nested ternary expression used directly inside `with.ref` rather than computing the target ref in a dedicated shell step with explicit logging for debugging CI failures? | The inline logical expression directly supplies checkout's input before any repository checkout is needed and keeps the three event cases next to the consumer. A dedicated shell step would have to be inserted before checkout, write a step output, and then pass that output into `with.ref`; it would be longer without changing the resolution semantics. Checkout logs provide the selected ref and fetch failure details. The expression is technically an `&&`/`\|\|` conditional chain rather than a ternary. |
| G7 | scope | ✅ CONFIDENT | The PR body explicitly leaves stripping the 14.5 KB `<details>review.json</details>` payload out of scope. If a PR has long discussion threads between reviewers and `@claude`, will updating a single Archon comment be sufficient to stay under the 200,000 token window, or will `review.json` still cause context exhaustion? | No, this change cannot guarantee staying below 200,000 tokens. It bounds Archon's contribution to one current comment, but the remaining review.json payload and unrelated long review/Claude discussion still consume context and can eventually exhaust the model window. A single 14.5 KB payload is not itself close to 200,000 tokens, but it reduces the remaining budget and may be duplicated by the action's prompt assembly. This PR correctly treats payload stripping or a larger-context model as separate follow-up levers. |
| G8 | scope | ✅ CONFIDENT | The updated documentation in `docs/guide/archon-pr-review.md` notes that per-round Archon history is preserved in `$GITHUB_STEP_SUMMARY`. Since GitHub step summaries expire after the repository's artifact retention period (e.g. 90 days), was the loss of long-term Archon review history on long-lived PRs considered? | The implementation intentionally trades permanent in-thread history for a smaller current PR context. Each full report is appended to that run's job summary, but the repository contains no archival mechanism beyond GitHub's workflow-run retention. Therefore long-term history can disappear when old runs are deleted; this does not affect the current-report requirement but should not be interpreted as permanent preservation. |
| G9 | errors | ✅ CONFIDENT | In `archon.yml`, if `gh pr comment ... --edit-last --create-if-none` fails due to a GitHub API rate limit or permissions error during comment edit, does the workflow step fail loudly or does it swallow the error? | It fails loudly. The `gh pr comment` invocation is the final unguarded command in a GitHub Actions `run` block. GitHub executes bash run steps with fail-fast behavior, so a nonzero exit from an API rate-limit or permission failure fails the step and exposes gh's stderr in the job log. Only the explicit empty-output case exits successfully without posting. |
| G10 | errors | ✅ CONFIDENT | In `claude.yml`, if `actions/checkout` attempts to fetch `refs/pull/N/head` for a PR whose head commit was force-pushed or whose fork branch was removed, what error output is produced, and does `claude-code-action` fail gracefully? | If GitHub cannot resolve the requested pull ref, actions/checkout's underlying fetch fails and logs a git error such as `fatal: couldn't find remote ref refs/pull/N/head`, followed by the action reporting a nonzero git exit. The checkout step then fails, and the Claude step is skipped by normal step conditions; there is no fallback to main. For PR issue comments, the separate `report-status` job runs under `always()` and reports the failed Claude job against the resolved PR head when that API lookup succeeds. Review-comment events do not run that status job. |

### Items to fix
- **G1 · FLAW_FOUND** — Yes. `--edit-last` selects the authenticated user's most recent comment, not the most recent comment carrying an Archon-specific identity. All workflows using the repository GITHUB_TOKEN comment as github-actions[bot], so an unrelated workflow comment posted after Archon's prior report would be overwritten. Archon's older report would remain, meaning the documented one-current-comment behavior would also cease to hold.
  _Concrete scenario: run Archon, then have a coverage workflow post a PR comment using GITHUB_TOKEN, then run Archon again. The command at .github/workflows/archon.yml:191-192 edits the coverage comment because there is no marker or stored comment ID; the comments at .github/workflows/archon.yml:189-190 only distinguish the Claude App, not other GitHub Actions workflows. The documentation promises the same Archon comment at docs/guide/archon-pr-review.md:13-16._
- **G3 · CANNOT_ANSWER** — The worktree cannot prove the external action's detached-HEAD compatibility. It references the moving `anthropics/claude-code-action@v1`, and the action implementation is not vendored here. Before merge, the reliable verification is a live disposable PR: trigger both an issue comment and a review comment, confirm checkout reports `refs/pull/N/head`, and confirm the Claude action completes its fetch/switch and reviews the expected head SHA. Pinning the action to a tested commit would make that result reproducible.
  _.github/workflows/claude.yml:60-77 performs the explicit PR-ref checkout; .github/workflows/claude.yml:79 invokes the unvendored floating `anthropics/claude-code-action@v1`. No local test or action source in the worktree exercises their interaction._
- **G4 · FLAW_FOUND** — There is no committed automated check for the new event-to-ref mapping, despite all four event shapes being realistic workflow inputs. Add a small resolver script or local JavaScript module accepting an event name and payload, table-test it with fixtures for PR issue comments, plain issue comments, review comments, and `issues` events, and feed its output into checkout. As an integration layer, run `act` with those fixture payloads or add a temporary diagnostic workflow that prints and validates the resolved ref without invoking Claude. `actionlint` should also be run for syntax/schema validation, though it cannot alone prove the branch semantics.
  _.github/workflows/claude.yml:9-15 declares three event families; .github/workflows/claude.yml:72-77 contains the new untested mapping. The repository's CI test matrix only runs Go packages and script tests at .github/workflows/ci.yml:25-66; no workflow-expression fixture test is present._

### Important to consider
- **F1** — The edit-or-create implementation is fragile if another workflow later comments using the same github-actions[bot] identity; see G1.
  _.github/workflows/archon.yml:171-192; .github/workflows/claude.yml:59-79; docs/contributing/archon/pr-review-context.md:9-17; docs/guide/archon-pr-review.md:13-16_
- **F2** — The phrase "same comment" assumes Archon remains the latest comment by github-actions[bot], and job-summary history is only available while the workflow run is retained.
  _docs/contributing/archon/pr-review-context.md:9-17; docs/guide/archon-pr-review.md:13-16; docs/guide/archon-pr-review.md:44-50_
- **F3** — The wording "on every push" in archon.yml:183 is imprecise because this workflow is triggered by a created issue comment, not directly by pushes.
  _.github/workflows/archon.yml:182-192; scripts/archon-review.sh:55-68; .github/workflows/claude.yml:63-77_
- **G5** — An embedded marker or persisted comment ID would eliminate the collision demonstrated in G1.
  _.github/workflows/archon.yml:182-192 documents identity scoping and uses the single-command implementation. The other current workflow definitions—.github/workflows/ci.yml, claude.yml, docs.yml, and release.yml—contain no repository-token PR-comment posting step._
- **G6** — Extracting the resolver would make fixture-based testing and explicit logging easier, as suggested in G4.
  _.github/workflows/claude.yml:59-79 places the mapping directly on the checkout input; .github/workflows/claude.yml:9-15 shows the three trigger families it distinguishes._
- **G7** — Prompt-size telemetry or truncation would be needed for a hard upper bound.
  _.github/workflows/archon.yml:182-192 only changes how the Archon comment is updated; .github/workflows/claude.yml:79-95 passes control to the external prompt-assembling action and .github/workflows/claude.yml:85 pins `claude-sonnet-4-6`. No code in this PR filters review.json or other discussion._
- **G8** — The docs could qualify this as history retained with the workflow run.
  _scripts/archon-review.sh:55-68 appends to GITHUB_STEP_SUMMARY; .github/workflows/archon.yml:187-188 and docs/guide/archon-pr-review.md:15-16 identify each workflow run page as the history location. No workflow step persists summaries to a branch, release, external store, or durable artifact._
- **G10** — GitHub normally maintains `refs/pull/N/head` for a PR and updates it after force-pushes, so branch deletion or force-push alone does not ordinarily make this ref unavailable.
  _.github/workflows/claude.yml:60-77 passes the pull ref to checkout with no fallback after a fetch failure; .github/workflows/claude.yml:79 places Claude after checkout; .github/workflows/claude.yml:97-124 gates failure status reporting to PR `issue_comment` events. The exact surrounding checkout wording depends on the external actions/checkout@v5 implementation._

namasl:
## Review — the two fixes are correct, but one claim is not established and one defect is missed

I independently verified this PR's premises against the run logs and the API. **Both fixes are right and should land.** Two things need addressing first.

### ✅ What I confirmed

- **The duplicate-blob claim is exactly right.** In run `33772637816`'s assembled prompt, the two archon comments are **byte-identical** (`diff` clean), 21,270 B each, both for head `280c84ea`, posted 86 s apart (00:34:30 / 00:35:56). The `review.json` `<details>` block appears twice verbatim.
- **The byte attribution supports Fix 1.** Over #1667's thread: total 470,270 B, of which archon dumps are **365,295 B (77.7%)** and all 20 trigger comments are **7,615 B (1.6%)**. 14 archon comments for 9 `/archon-pr-review` triggers.
- **Fix 2's premises all hold**, including the caveat: `checkout -B main` (log line 260, 15:31:03) → `refs/pull/1667/head` / `Switched to branch` (767–770, 15:31:19) → `FINAL PROMPT` (781, 15:31:19). The switch precedes prompt assembly, so the review was indeed never on main. 12 `Failed to compute SHA` = the 12 added files. ✔
- **The `--edit-last` clobber check is sound.** I re-ran it wider than `gh pr comment`: no `createComment`, no `gh api .../comments`, no `peter-evans/create-or-update-comment` anywhere in `.github/workflows/`. And empirically every `github-actions[bot]` comment on #1660/#1664/#1667 is an `## ARCHON PR review`. Nothing else for `--edit-last` to hit.

---

### 🔴 1. The title's claim is not demonstrated — please soften it or land the model change

**Only 2 comments reached the prompt, and they were the byte-identical pair.** The whole `FINAL PROMPT` block is **74,799 B** (`<pr_or_issue_body>` ~326 lines + `<comments>` holding those 2 archon dumps). The other 12 archon dumps were **never in the prompt**.

So the causal chain "58 comments / 470,270 B → overflow" is not what the log shows. The request was **244,554 tokens**; the entire folded thread was ~75 KB (~20K tokens at typical ratios). Deduping the pair removes ~21 KB — real, strictly good, but a small fraction of the request. **The majority of those 244,554 tokens is not thread content**, and this PR does not touch it.

Concretely: `-27 KB` of a `244,554`-token request does not establish that the next review fits in 200,000.

Please either:
- **(a)** retitle/reframe to what is proven — *reduce archon thread bloat and remove the duplicate blob* — and keep the overflow fix as the follow-up; **or**
- **(b)** land the `claude.yml:70` model change here (you already identify it as "the bigger lever"), since that is the change that actually closes the gap; **or**
- **(c)** post a successful `@claude /blis-pr-review` run on this PR as evidence, with its token count.

I'd favour (b). Your own analysis reaches the same conclusion — the measurement just makes Fix 1 clearly insufficient on its own rather than "a nice-to-have".

### 🟡 2. Missing: `report-status` still overwrites a real result with `pending`

This is an **independent merge-blocker in the file this PR already edits**, and it is not mentioned. `claude.yml`:

```yaml
report-status:
  needs: claude
  if: always() && github.event_name == 'issue_comment' && github.event.issue.pull_request
```

`always()` with no guard on `needs.claude.result == 'skipped'`. `claude[bot]`'s own review comment fires `issue_comment`, the `claude` job correctly skips, and `report-status` still runs and writes `pending / Claude Code: skipped` over the good result.

Verified:
- Run `33770188054` — `event=issue_comment`, `triggering_actor=claude[bot]`, `check-permissions: skipped`, `claude: skipped`, `report-status: success`. Created 15:00:51, **34 s after** the real run `33770122756` (15:00:17).
- On head `4be14710` the status sequence is `failure @ 15:17:20` then **`pending / Claude Code: skipped` @ 15:21:11** — the stale one lands last and sticks.

Suggested one-liner:

```yaml
if: always() && needs.claude.result != 'skipped' && github.event_name == 'issue_comment' && github.event.issue.pull_request
```

No author-side comment discipline can prevent this, so it will keep producing stuck required checks regardless of Fix 1.

### 🔵 3. Follow-up worth an issue: archon double-fires

14 archon comments for 9 triggers, and the pair in the prompt is the same head SHAs 86 s apart. `--edit-last` makes the *bloat* harmless, but the duplicate **runs** still execute. Worth an issue so it is not lost once the symptom is hidden.

---

**Net:** Fix 1 and Fix 2 are correct, well-scoped, and the `--edit-last` safety reasoning checks out. Add the `report-status` guard, and either take the model change or retitle — as it stands the PR does not show the overflow is fixed.

Method note: measurements from `gh run view 33772637816 --log` with log prefixes stripped, and the GitHub comments API. Happy to share the exact commands.

namasl (commented):
### Additional finding: a no-op Claude run can still publish a green status

The `needs.claude.result != 'skipped'` guard suggested above does not cover the other no-op path. Run [33785598570](https://github.com/inference-sim/inference-sim/actions/runs/33785598570) was admitted by the workflow's broad substring condition because my prose review mentioned the trigger. Inside the composite action it logged `No trigger found, skipping remaining steps`, but the containing `claude` job concluded `success`; `report-status` then published `Claude Code: success` on head `6c7c3a8`. That is currently the PR's green Claude status even though no Claude review ran.

Please give the action step an ID and gate/report from an output that is absent on this no-trigger return (or make the outer trigger test match the action's parser), rather than relying only on `needs.claude.result`.

Also, that run has `headBranch: main`, workflow `headSha: e3598b8`, and its initial checkout was `origin/main`. Comment-triggered workflows use the default-branch workflow definition, so this apparent success does not exercise the new `ref:` block and cannot serve as the requested live validation of it.

susiejojo:
## Review — Approve

Well-diagnosed, low-risk CI-config fix. Every factual claim in the description checks out against the repo, all workflow YAMLs parse, and the conditional-ref expression is correct across all trigger shapes. No Go behavior changes.

### Verified claims

| Claim | Status |
|---|---|
| `archon.yml:191` is the only `gh pr comment` in workflows | ✅ grep returns exactly one hit |
| History preserved via `emit()` appending to `$GITHUB_STEP_SUMMARY` | ✅ `scripts/archon-review.sh:60` |
| No other workflow posts comments as `github-actions[bot]` | ✅ `docs.yml` only does `git config`/commits; `report-status` posts a commit *status*, not a comment |
| Archon runs on `issue_comment`, so `github.event.issue.number` is the right PR number | ✅ `archon.yml:9-11`, `PR_NUMBER` at line 175 |
| All five workflow YAMLs parse | ✅ |

### Fix 2 expression traced across all four trigger shapes

Parses (via `&&` > `||` precedence) as `(A && v1) || (B && v2) || ''`:

1. `pull_request_review_comment` → `refs/pull/<pull_request.number>/head` ✅
2. `issue_comment` on a PR → `issue.pull_request` truthy → `refs/pull/<issue.number>/head` ✅
3. `issue_comment` on a plain issue → `issue.pull_request` null → `''` ✅
4. `issues: [opened, assigned]` → `''`, and `actions/checkout`'s empty-ref default takes over ✅

Property access on a null object returns null (not an error) in GHA expressions, so `github.event.issue.pull_request` is safe on the PRRC event where `issue` is absent.

### Non-blocking notes

1. **`--edit-last` is identity-scoped, not content-scoped** — it edits the last comment by the token identity, not the last *archon* comment. Safe today because no other producer posts as `github-actions[bot]` (audited). Worth remembering if a future action posts under `GITHUB_TOKEN`. Author already flagged this.
2. **Genuinely unverifiable offline** (as stated): `--edit-last` bot-identity resolution under `secrets.GITHUB_TOKEN`, and the detached-HEAD interaction with `claude-code-action`'s subsequent `refs/pull/N/head` re-fetch. Watching the first post-merge `/archon-pr-review` and `@claude` run is the right mitigation; both changes are trivially revertible.
3. **Docs consistent** — both updated files correctly describe the sticky single-comment model and note per-round bodies persist on the run page.
4. **Root cause vs symptom, correctly deferred** — the description is honest that the `--model claude-sonnet-4-6` 200K-window pin is the real lever and these two fixes are palliative. Right call for a Small-tier PR.

### Comment quality
Inline comments explain *why* (with run ID + token counts as evidence) and pre-empt the two "why not just X" questions a reviewer would ask. This directly prevents the rediscovery-as-bug problem noted in the description.

Nothing to change. Ship it, and watch the first live trigger of each workflow as planned.

jgchn:
Thanks — both of your points held up under independent checking, and one of them was my error rather than a difference of opinion. All addressed in `8c4e0b51`.

### 1. The over-claim — you're right, and I'd already contradicted myself

I re-counted the archon dumps that actually reached the failing prompt:

```
$ grep -c "^### Component view" prompt.txt
2
```

Two, and they're the byte-identical pair. The other 12 were never in it. So the "58 comments / 470,270 B → overflow" chain is not what the log shows, exactly as you said.

Worse, I had already written the correct version during triage — *"I can't tell you Fix 1 alone gets you under 200K"* — and then titled the PR as though it did. Retitled and reframed to what's proven, and the overflow proper is now **#1676**, carrying the measurement that I think is the real story: `CLAUDE.md` grew 65,478 → 80,500 B in **10 days** and is a fixed ~20K-token tax on every run, with 79% of it in four sections that already declare a canonical source elsewhere.

On your (b) — I'd take it too, but I can't see what the LiteLLM proxy serves, so I can't pick a model I know resolves. Left as lever 3 in #1676 for someone who can. (c) is impossible pre-merge, see below.

### 2. `report-status` — verified independently, fixed here

Confirmed the code has no skip guard, and reproduced your evidence:

```
run 33770188054   claude: skipped, check-permissions: skipped, report-status: SUCCESS
head 4be14710     failure @ 15:17:20  →  pending/skipped @ 15:21:11   (stale wins)
```

Took your one-liner. Note it also means a denied permission check now writes no status at all, which I think is strictly better than a stuck `pending`.

### 3. The double-fire has a sharper mechanism than duplicate runs — filed as #1675

It isn't duplicate delivery. The trigger is a substring match on the whole comment body, which cannot distinguish an invocation from a **prose mention**. Neither of your two comments here invoked archon — both merely discuss it — and each fired a full run ~4 s later, producing byte-identical reports for the same head `e3598b84`.

So the comment reporting that archon fires twice is what made archon fire twice. That's also the 14-for-9 ratio on #1667. Fix is to anchor the match to the comment's first line. (Writing this reply without the literal command for the same reason.)

### Also fixed from the qa-review pass

- **F3** — "on every push" was wrong; archon fires on a comment, never directly on a push.
- **G8** — job-summary history is retention-bound, not an archive. Both docs now say that instead of implying permanence.
- **G1 / G4** → **#1677**. Neither is a present defect: I re-verified `archon.yml` is the only comment poster (no `createComment`, no `gh api .../comments`, no `peter-evans` action), matching your wider sweep, and the ref expression was traced by hand over all four event shapes. But identity-scoped `--edit-last` and the untested mapping are both real fragility, so they're written down with a marker-lookup proposal and `actionlint`.
- **G10** — you were right that I overstated the force-push risk; GitHub maintains `refs/pull/N/head` across force-pushes.

### One thing neither of us stated outright

**Neither workflow fix is testable before merge.** `issue_comment` workflows always run from the default branch — both archon runs on this PR report `branch=main`, and the two byte-identical 5,548 B comments here are the *old appending code* executing. That's also why your option (c) can't be satisfied on this PR. It's in the body now: merge, then watch the first archon trigger and the first Claude run. Both are one-line reverts.

Happy to split Fix 3 into its own PR if you'd rather keep this one to the two original changes.

namasl:

## qa-review adjudication — PR #1674: ⛔ BLOCK

|----|----|----|----|
| G1 | FLAW_FOUND | 🟡 WAIVED (deferred) | The current identity-scoped behavior remains, but the author cited an open follow-up that directly covers replacing it with marker-scoped comment lookup and includes an acceptance criterion proving that an intervening github-actions[bot] comment is unaffected. |
| G3 | CANNOT_ANSWER | ❌ STILL-OPEN | The response concedes that the changed checkout/action interaction has not been exercised before merge and provides neither an implementation fix nor a covering follow-up issue. Logs showing the action switching from a normal default-branch checkout do not verify compatibility after the newly added explicit PR-ref checkout. Concrete verification of that exact path is still required, or the external action must be pinned to a revision for which that interaction is reproducibly verified. |
| G4 | FLAW_FOUND | 🟡 WAIVED (deferred) | The author cited an open follow-up issue that genuinely and specifically covers adding automated verification for all four event-to-ref mappings and running actionlint in CI. |

### Still blocking
- **G3** — The response concedes that the changed checkout/action interaction has not been exercised before merge and provides neither an implementation fix nor a covering follow-up issue. Logs showing the action switching from a normal default-branch checkout do not verify compatibility after the newly added explicit PR-ref checkout. Concrete verification of that exact path is still required, or the external action must be pinned to a revision for which that interaction is reproducibly verified.
  _The current workflow explicitly checks out the PR ref at `.github/workflows/claude.yml:60-77`, then invokes the external moving reference `anthropics/claude-code-action@v1` at `.github/workflows/claude.yml:79`. No committed test or vendored implementation verifies their interaction, and the author states that the workflow fix is not testable before merge._

### Cleared
- **G1** (🟡 WAIVED (deferred)) — The current identity-scoped behavior remains, but the author cited an open follow-up that directly covers replacing it with marker-scoped comment lookup and includes an acceptance criterion proving that an intervening github-actions[bot] comment is unaffected.
  _Current code still uses `gh pr comment ... --edit-last --create-if-none` and explicitly notes authentication-user scoping at `.github/workflows/archon.yml:191-194`. Issue #1677 is OPEN and its first problem, proposed fix, and acceptance criteria specifically require locating Archon's comment by marker rather than authenticated identity._
- **G4** (🟡 WAIVED (deferred)) — The author cited an open follow-up issue that genuinely and specifically covers adding automated verification for all four event-to-ref mappings and running actionlint in CI.
  _The untested conditional mapping remains at `.github/workflows/claude.yml:72-77`. Issue #1677 is OPEN; its second problem describes this exact expression and the four event shapes, and its acceptance criteria require a committed test covering all four plus actionlint in CI._

jgchn:
Both remaining items addressed in `1617801c`.

### Your review finding — the false green. You're right, and it's the worse of the two.

Reproduced it exactly as you described:

```
run 33785598570  event=issue_comment  actor=namasl  headBranch=main
  check-permissions: success   claude: success   report-status: success

log:  Context prompt: NO PROMPT
      No trigger found, skipping remaining steps

head 6c7c3a81:  17:37:21  success  "Claude Code: success"
```

So this PR's green Claude status had no review behind it. My `needs.claude.result != 'skipped'` guard cannot see this — the job concluded `success`. And a false green is strictly worse than the stale `pending` I did fix: the stale one is visible, this one isn't, and a merge gate reads it as a pass.

Took your suggested approach rather than trying to match the action's parser. `claude-code-action` declares `execution_file` and `conclusion`, and **both are empty on the no-trigger early return**, so they are the signal that work happened. The action step now has `id: claude`, the job exposes both as outputs, and `report-status` gates on `execution_file`:

```yaml
if: |
  always() &&
  github.event_name == 'issue_comment' && github.event.issue.pull_request &&
  (needs.claude.result == 'failure' || needs.claude.result == 'cancelled' ||
   (needs.claude.result == 'success' && needs.claude.outputs.execution_file != ''))
```

Stated positively so a future job result can't fall through to a status by default:

| `claude.result` | `execution_file` | publishes | was |
|---|---|---|---|
| `skipped` | — | nothing | stale `pending` |
| `success` | empty | **nothing** | **false green** |
| `success` | set | `success` | `success` |
| `failure` | any | `failure` | `failure` |
| `cancelled` | any | `error` | `error` |

One deliberate trade-off: a real success that somehow reported no `execution_file` now publishes **nothing** rather than a verdict. A missing status is visible and honest; a wrong one is neither.

Also dropped the now-unreachable `skipped: 'pending'` entry from `stateMap`. If a skipped result ever reaches there again, falling through to the warning + `'error'` path surfaces the regression instead of silently re-publishing a stale pending over a real verdict.

Worth noting the root cause is the same prose-matching trigger as #1675 — `claude.yml` has the identical `contains()` pattern I filed against `archon.yml`. This commit hardens the consequence; anchoring the trigger is tracked there. Your prose review mentioning the trigger phrase is what admitted that no-op run, which is a neat demonstration of both bugs at once.

### G3 — filed as #1678

The adjudicator's objection was fair: I conceded the interaction is unexercised and gave it no covering issue. #1678 now carries it, including your point that the run cannot serve as validation:

- The five post-merge checks (checkout reports the PR ref; the action's own fetch/switch completes without `refusing to fetch into branch ... checked out`; reviewed head == PR head; the 12 `Failed to compute SHA` lines are gone; repeat for `pull_request_review_comment` and a plain `issues` event).
- Your `headBranch: main` observation as the reason it's post-merge only — every run on this PR reports it, so nothing here exercises the new `ref:` block.
- The pinning question. `@v1` is a moving tag, so this interaction can change with no commit here. I left the decision to you rather than pinning in this PR, since it trades automatic fixes for reproducibility.
- An explicit revert path: the `ref:` block is self-contained, and the action's own checkout makes the workflow correct without it.

G1 and G4 are waived against #1677 per the adjudication.

---

Current state: `1617801c`. The one thing I can't close from here is G3's actual verification — by construction it needs a post-merge trigger. If you'd rather not merge on an argument, the alternative is pinning the action to a SHA now and I'll add that here.

namasl:

## qa-review adjudication — PR #1674: ✅ PASS

|----|----|----|----|
| G1 | FLAW_FOUND | 🟡 WAIVED (deferred) | The identity-scoped behavior remains, but the cited open follow-up specifically requires replacing it with marker-scoped lookup and testing an intervening github-actions[bot] comment. |
| G3 | CANNOT_ANSWER | 🟡 WAIVED (deferred) | The action has not yet been pinned or locally verified, but the author cited an open issue dedicated to verifying the exact detached-HEAD/external-action interaction and deciding whether to pin the action. |
| G4 | FLAW_FOUND | 🟡 WAIVED (deferred) | The event-to-ref mapping remains without committed automated coverage, but the cited open follow-up explicitly requires fixture coverage for all four event shapes and actionlint in CI. |

### Still blocking

### Cleared
- **G1** (🟡 WAIVED (deferred)) — The identity-scoped behavior remains, but the cited open follow-up specifically requires replacing it with marker-scoped lookup and testing an intervening github-actions[bot] comment.
  _Current code still uses `gh pr comment ... --edit-last --create-if-none` and acknowledges authenticated-user scoping at `.github/workflows/archon.yml:191-194`. Issue #1677 is OPEN and its first problem and acceptance criterion directly cover this defect._
- **G3** (🟡 WAIVED (deferred)) — The action has not yet been pinned or locally verified, but the author cited an open issue dedicated to verifying the exact detached-HEAD/external-action interaction and deciding whether to pin the action.
  _The current workflow checks out the explicit PR ref at `.github/workflows/claude.yml:67-84` and still invokes the moving `anthropics/claude-code-action@v1` reference at `.github/workflows/claude.yml:86`. Issue #1678 is OPEN and requires post-merge checks of checkout, fetch/switch, reviewed SHA, missing-file diagnostics, relevant event shapes, and a recorded pinning decision._
- **G4** (🟡 WAIVED (deferred)) — The event-to-ref mapping remains without committed automated coverage, but the cited open follow-up explicitly requires fixture coverage for all four event shapes and actionlint in CI.
  _The conditional mapping remains at `.github/workflows/claude.yml:79-84`; the PR changed-file list contains no resolver or workflow-test file. Issue #1677 is OPEN and its second problem and acceptance criteria directly require a committed four-shape mapping test and actionlint._

namasl (approved):

Issue #1666, opened 2026-09-03
MoE DP: capacity charges routed experts at /TP with EP off while vLLM (and BLIS step time) shard them across TP·DP


namasl:
> Surfaced by code review of #1660 (the #1656 EP-aware weight-footprint PR). Not a defect introduced there — #1660 leaves this direction conservative and documents it. Filed so the divergence is tracked rather than living only in a doc comment.

**What antipattern or issue does this address?**

BLIS's KV-capacity weight footprint and its trained-physics step-time model disagree about where a MoE model's **routed-expert weights live when expert parallelism is OFF and `DP > 1`**, and the capacity side does not match vLLM.

vLLM's `FusedMoEParallelConfig.make` calls `flatten_tp_across_dp_and_pcp` **unconditionally**, before the `use_ep` branch — so the MoE layers' `tp_size` is `dp_size · tp_size` regardless of `--enable-expert-parallel`:

- **EP off**, TP=2, DP=2 → MoE `tp_size = 4`, `ep_size = 1` (expert tensors sharded across 4 devices).
- **EP on**, TP=2, DP=2 → `tp_size = 1`, `ep_size = 4` (whole experts split across the same 4 devices).

Per-GPU routed-expert **bytes are therefore `R/(TP·DP)` in both modes**; `--enable-expert-parallel` changes the sharding *style* (whole experts vs tensor slices) and the collective (all-to-all vs all-reduce), not the footprint.

Where BLIS stands today:

| | routed-expert bytes per GPU | matches vLLM? |
|---|---|---|
| Step time (`sim/expert_placement.go`, `trained_physics_model.go`, `moeGroup = TP·DP` in both modes) | `R/(TP·DP)` | ✅ yes |
| KV capacity, EP **on** (#1656) | `R/(TP·DP)` | ✅ yes |
| KV capacity, EP **off**, `DP>1` | `R/TP` | ❌ conservative by a factor of `DP` |

The capacity side's EP-off baseline is BLIS's own DP model — #1531 models MoE `--dp N` as N *independent* single-node engine replicas, each holding a full tensor-sharded copy of the experts. That is a coherent model for dense DP, but vLLM's MoE DP ranks are **not** independent: they are coupled through the MoE collective (which is why they must step in lockstep with dummy batches), and they do not each hold a full copy of the expert weights.

**Which invariants are at risk?**
- [x] None (fidelity/consistency, not an invariant break)

[...]

Issue #1568, opened 2026-08-13
DeepEP all-to-all: high-throughput (HT) vs low-latency (LL) step-time profiles


namasl:
> Part of #1532 (GLM-5.2 support). One PR.
> **Blocked by:** #1548 (per-role all-to-all backend selector + latency-model wiring) and #1530 (inter-node all-to-all cost term these profiles differentiate against).

## Problem (confirmed in code)

BLIS already accepts `--moe-comm-backend deepep_high_throughput` and `deepep_low_latency` ([`sim/latency/moe_comm_backend.go`](https://github.com/inference-sim/inference-sim/blob/main/sim/latency/moe_comm_backend.go)), but **both map to the same `commFamilyAll2All`** — the family distinguishes only comm *volume* (all-gather vs all-to-all), **not** the HT vs LL *kernel* latency/throughput characteristics. So selecting LL vs HT yields **identical simulated step time today**.

DeepEP ships two distinct kernel sets: *high-throughput* (NVLink+RDMA, large batched dispatch/combine) and *low-latency* (pure RDMA, IBGDA+NVSHMEM, small-batch). GLM-5.2's disaggregated recipe (vLLM [`recipes.vllm.ai/zai-org/GLM-5.2`](https://recipes.vllm.ai/zai-org/GLM-5.2) and the llm-d GLM-5.2 agentic-serving guide) runs **prefill in `deepep_high_throughput`** and **decode in `deepep_low_latency`** on purpose (prefill = large-batch/throughput regime; decode = small-batch/latency regime). With HT/LL collapsed, the decode pool's all-to-all is mis-modeled → decode-side capacity/throughput for the real recipe is wrong.

## Scope

- Two parameterized all-to-all cost profiles for dispatch + combine:
  - **HT** — throughput-optimized, favorable in the large-batch (prefill) regime.
  - **LL** — latency-optimized, favorable in the small-batch (decode) regime.
- Each profile is a function of the drivers that actually move it: **tokens-per-step (batch size)**, **EP degree**, and **placement (intra- vs inter-node)** — the last consuming #1530's cost term rather than hard-coding a fabric.
- The backend selected via #1548 applies the corresponding profile in `sim/latency/trained_physics_model.go`.

## Out of scope (keep the PR focused)

[...]

PR #1683, opened 2026-09-03
feat(latency,cmd): EP placement + live EP-mode step-time toggle (Closes #1548)

### LINKED ISSUE #1548: EP placement + live EP-mode step-time toggle (split from #1531)

namasl:
Closes #1548.

## Scope decision

#1548 permits narrowing to *EP-placement only, toggle step-time-inert*. **Not taken** — the live toggle needs **no new coefficient and no calibration**, being a *gate + group* change read off vLLM's own MoE path, and **no new cost term** (a second additive comm term would double-charge the dispatch/combine #1530 already prices). Two *existing* things move:

1. **The MoE-FFN collective gate** partitions on `DP>1 || EP-on` instead of `DP>1`. EP-off tensor-shards the experts and reduces the FFN output (`tMoEReduce`); EP-on owns whole experts and must route tokens to their owner (`tMoEDispatch`). Still exactly one of the two — mutually exclusive and exhaustive.
2. **Routed-expert WEIGHTS** shard over a new *expert-shard group* (`EffectiveExpertShardGroupSize`), while routed-expert **COMPUTE** keeps `moeGroup`.

### Why weights and compute use different groups

| per GPU | EP off | EP on |
|---|---|---|
| routed-expert **compute** | `T·k/TP` | **`T·k/TP` — unchanged** |
| routed-expert **weights** | `numExperts/TP` slices | **`numExperts/G` whole experts** |

With EP on, the `G = TP·DP` GPUs jointly process the whole group's `n_dp·T_local` tokens ⇒ `T_local·k/TP` per GPU: what tensor-sharding already gives. EP re-organises expert *ownership*, not FLOPs; dividing compute by `G` too would under-charge it by `DP` — the defect this split prevents, with its own test. (Assumes lockstep DP ranks, which vLLM enforces with dummy batches; BLIS runs replicas independently, so the charge is exact at saturation and pessimistic below — documented.)

### The #1531 collapse trap, avoided

DP-as-placement rewrites each replica's `DP` to 1, so a config-bound EP size collapses to `TP`. The CLI carries the **logical DP _width_** (`sim.WithExpertParallelGroupDP`, held on `dpPlacementPlan.EPGroupDP`, spread by `EPGroupOptions()` at both `NewModelHardwareConfig` sites). A **width**, not an absolute group size: `ResolvePoolConfig` can override a pool's `TP`, which an absolute size stamped from the global TP would contradict (`TestEffectiveExpertShardGroupSize_ScalesWithPoolTP`). Step time and #1656's KV capacity now shard the same experts over the same group.

**Placement (AC-1):** EP reserves **no** GPUs beyond the `num_instances × N × TP` DP placement already takes — the EP group *is* those GPUs; asserted by the replica count matching the EP-off run.

### The headline number, attributed to a named term

**Measured** (deepseek-v2-lite, 40 req, seed 42, cluster aggregate): `--tp 2 --dp 4` mean ITL `7.84 → 3.17 ms`, `1820 → 3375 tok/s`; Wide-EP `--tp 1 --dp 16` mean ITL `14.17 → 2.68 ms`, `1110 → 3659 tok/s`.

A 3.3x swing from flipping a modeling flag needs a mechanism. **It is routed-expert weight sharding** — the per-step weight-read bandwidth term, `numExperts/1 → numExperts/16` full-expert-equivalents per GPU. `TP=1` forces that attribution: there is no TP collective at all (`tpAllReduceBasis` returns 0 at `tp <= 1`) and `tMoEReduce` additionally requires `tp > 1`, so the all-reduce leg is **0 in both modes** — the gate change cannot improve this shape, only *add* the dispatch term. Compute is scoped by `moeGroup = TP·DP = 1` in both. Weights are the only term left. Three tests pin it:

- **Sign flip:** with realistic experts the flag is a speed-up; shrink `MoEExpertFFNDim` to ~0 and **the same flag becomes a slow-down**, only dispatch remaining — proving both that the win is the weight term and that the added all-to-all is genuinely charged.
- **No compute double-count:** top_k sensitivity — which on the all-gather family reaches *only* the compute basis — is identical in both modes.
- **Degenerate case:** a one-GPU EP group owns every expert and has no peer, so the toggle is exactly neutral.

**Phase asymmetry + external sanity check.** Dispatch cost scales with tokens, the weight saving does not, so the win is decode-heavy: `7.1x` on a decode step vs `1.6x` on a large prefill step — which is why production Wide-EP is a *decode*-side strategy, reproduced here without being calibrated to it. [NVIDIA reports](https://developer.nvidia.com/blog/scaling-large-moe-models-with-wide-expert-parallelism-on-nvl72-rack-scale-systems/) EP32 vs EP8 (4x wider) giving *up to 1.8x* output tok/s/GPU — sub-linear in group size, as here. **Not a calibration target, not a like-for-like baseline:** BLIS's EP-off leg at `TP=1` puts all 64 experts on one GPU, which a real large MoE could not fit, so it is a modeling baseline that inflates the ratio. Treat as uncalibrated model output.

## Per-role all-to-all backend (AC-5/AC-6)

`--prefill-moe-comm-backend` / `--decode-moe-comm-backend` (empty = inherit the global) on **both** commands, via `cluster.PoolOverrides.MoECommBackend`. Both roles and both commands validate through one shared `applyPerRoleMoECommBackends` (R23); an unknown name is a hard error (R1).

The selection routes through a **per-mode step-time profile** — `latency.all2AllProfile{commScale}`, one entry per backend in `moeCommBackends`. **Every backend ships the exact `1.0` nominal placeholder**, so DeepEP HT and LL cost the same today; #1568 fills the table (and may add fields) without touching the selector. Two tests hold that honest: a sentinel asserting the exact 1.0 for every backend (it fails when #1568 lands), and one mutating the field on a built model to prove the multiplication site exists.

## One EP-on shape is deliberately neutral — stated, not hidden

At `DP=1` with vLLM's **default** `allgather_reducescatter`, EP-on and EP-off are **numerically equal** — correct physics, not an inert toggle: all-gather + reduce-scatter *is* the ring-all-reduce decomposition (identical wire volume), and β_EP defaults to β₄. Its test doubles as the double-charge guard: if both terms fired it would come out strictly greater, not equal. AC-2 is met by the two routes carrying real physics — a modular all-to-all backend (top_k volume) and an EP group wider than TP (weight sharding).

## Honesty boundary (warned on stderr, R1)

The EP group spans `N` **independently placed** replicas. Cross-node pricing is placement-derived (#1530) and `node_pools` + `--dp>1` remains a #1553 fail-fast, so the inter-replica leg is charged at the **on-node** rate and multi-node EP step time is optimistic; `blis run` says so explicitly. `InterNodeLatencyUs` also still ships at 0 (#1661).

## Folded in from #1530's deferred comments

- `tpAllReduceBasis(units, tokens, dpDivisor)` divides only the byte **volume** by `dp`; a per-collective launch latency is paid in full by each rank (DP groups launch concurrently). `dpDivisor` is exactly `1.0` in every CLI-reachable config ⇒ byte-identical.
- `warnIfCrossNodeUnpriced` scores the **widest** collective group, not `TP` alone, so a MoE-group-only span cannot be priced without being diagnosed. Still unreachable until #1553; documented as such rather than claimed.
- **Newly closed:** the latent hole `resolveLatencyConfig`'s roofline gate documents — it reads only the **global** backend, so `--decode-latency-model roofline` could put one pool on DP/EP-blind step time while the cluster runs EP. Harmless while EP was inert; a real mis-model now. `validatePerPoolLatencyBackends` rejects it. Note #1556 wrote that as a `LATENT HOLE — READ THIS BEFORE LIFTING #1553` comment and predicted **#1553** as the trigger; it turned out to be **this** PR — a latent-defect comment firing on an unforeseen trigger is the argument for writing them.

## Invariants

- **INV-6 / INV-BC-DP1** — EP-off, dense, and option-absent configs are **byte-identical**, verified empirically against `f4c8c619` on 5 shapes incl. MoE `--dp 4` and dense `--tp 4`. `commScale` and `dpDivisor` are exact identities; the two groups coincide for every pre-#1548 shape; `EffectiveEPGroupDP`'s `max` means the option can only *widen*.
- **INV-13** — the EP-group width is re-supplied from the CLI on both legs (a model-level input like `--kv-cache-dtype`, not a trace field); `TestINV13_RunReplayParity_MoEDPPlacement` gains a `tp2-ep` byte-identity case. Also verified manually on the Wide-EP topology.
- **INV-1** — conservation asserted in the new EP-on run test.

## Tests updated on purpose (not regressions)

The five guard tests whose premise was "EP-on + `--dp>1` is rejected": the plan tests now assert EP-on plans *identically* to EP-off plus the carried width; the run/replay system guards moved to a still-guarded combo (PD, #1553); and `..._CapacityErrorNoLongerMasksPlacementGuard` became `..._CapacityFitsAndRuns` — #1656's `TP=16/DP=2/EP=32` case now runs end to end.

## Not in scope

#1568 (differentiated HT vs LL curves — this delivers the selector + per-role wiring + one shared cost), #1553 (PD / autoscaler / node pools + `--dp>1`), #1666 (the EP-**off** `DP>1` capacity baseline is unchanged; the step-time/capacity *inconsistency* it also names is closed).

**Mutation-tested, so no key law is vacuous** — each fails with its defect injected: pointing the compute `Resolve` calls at the EP group breaks the top_k law with exactly the 4x width ratio; dropping `!m.epOn` breaks the EP-on/all-reduce equality; removing `ResolvePoolConfig`'s `MoECommBackend` line breaks the per-role e2e law. That e2e test compares **stdout only** plus a same-config control leg — stderr's logrus timestamps would have made an inequality pass vacuously.

Local: `go build ./...`, `go test ./...`, `golangci-lint run ./...` all clean.


namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:

## qa-review — PR #1683: ⛔ BLOCK
_feat(latency,cmd): EP placement + live EP-mode step-time toggle (Closes #1548) · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The acceptance criteria are implemented: EP-on DP placement reuses the existing N×TP replicas while carrying the logical DP width into each replica; EP changes trained-physics step time by replacing the MoE reduction with dispatch/combine and by widening expert-weight sharding; run/replay parity is tested; and per-role backends flow through per-mode profiles. HT and LL intentionally retain the allowed shared placeholder cost. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The substantive project documentation is updated: the configuration reference, glossary, roofline scope, latency-model guide, and INV-13 text describe live EP step time, EP placement, replay requirements, and per-role backends. One CLI help string was not refreshed. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | There are a few stale or imprecise comments/help strings, but no behavioral defect. Most implementation comments were updated accurately. |
| G1 | correctness | ✅ CONFIDENT | How does `latency.ClampExpertShardToExpertCount` handle edge cases where `numExperts` is smaller than `EffectiveEP()`, and is there any risk of an off-by-one error or division by zero when calculating per-rank expert weight footprints? | `ClampExpertShardToExpertCount` returns `numRoutedExperts` when that count is positive and smaller than the requested EP group, so a loaded rank is charged at least one whole expert. For zero or negative expert counts it deliberately leaves the value unchanged, but production only applies the step-time clamp when the model is genuinely MoE; dense/sub-threshold models take the inert path. Valid production groups and the resulting divisor are at least one, so there is no off-by-one or division-by-zero path. |
| G2 | correctness | ✅ CONFIDENT | When DP-as-placement rewrites replica `DP` to 1, what happens if `--enable-expert-parallel` is enabled on a non-MoE model or an MoE model with zero routed experts? Does `EPGroupDP` inadvertently alter `EffectiveExpertShardGroupSize`? | The CLI rejects `--enable-expert-parallel` on a non-MoE model before simulation. At the library/config level, `EffectiveEPSize` returns 1 whenever `isMoE` is false, so `EPGroupDP` is ignored and `EffectiveExpertShardGroupSize` falls back to the ordinary MoE/dense group. A model with zero routed experts is non-MoE under the canonical predicate and behaves the same way. |
| G3 | invariants | ✅ CONFIDENT | Since `EPGroupDP` is supplied via CLI options, how does `blis replay` ensure byte-identical execution (INV-13) when replaying a trace where `--enable-expert-parallel` and `--dp` flags are specified differently or defaulted on replay? | INV-13 applies only when the replay uses the same resolved configuration. Both command paths call the shared DP-placement resolver and pass `dpPlan.EPGroupOptions()` to the canonical hardware constructor, so identical `--dp` and `--enable-expert-parallel` flags reconstruct the same logical EP width. If replay uses different flags or defaults them, parity is not promised and the trace does not override those model-level inputs; the documentation explicitly tells users to re-supply them. |
| G4 | invariants | ✅ CONFIDENT | How is determinism (INV-6) preserved when `tMoEDispatch` replaces `tMoEReduce` under `--enable-expert-parallel`, given that dispatch/combine and reduce collectives compute token routing and step-time delays across ranks differently? | The new collective selection is deterministic arithmetic, not stochastic token routing. The two communication terms are mutually exclusive, `BalancedPlacement.Resolve` is a pure closed-form calculation, and backend lookup uses an ordered slice rather than map iteration. Thus the same batch and configuration produce the same floating-point operations and rounded step time. CLI coverage also repeats an identical run and asserts byte-identical stdout. |
| G5 | rules | ✅ CONFIDENT | How does `cmd.applyPerRoleMoECommBackends` enforce Rule R1 (fail fast on invalid inputs) when an unrecognized backend name is supplied to `--prefill-moe-comm-backend` versus `--decode-moe-comm-backend`? | `applyPerRoleMoECommBackends` handles both roles through the same loop. For each explicitly changed flag it validates the value before writing the override and returns an error naming that role's flag and invalid value. Both run and replay immediately convert that error to `logrus.Fatalf`; tests cover invalid prefill and decode values independently. |
| G6 | rules | ✅ CONFIDENT | Does `sim.WithExpertParallelGroupDP` adhere to Rule R4 (canonical constructor options) across all hardware configuration entry points, and are there any simulator creation paths where this option could be silently dropped? | Yes for all production paths that need the logical width. `WithExpertParallelGroupDP` is a canonical `ModelHardwareOption`; the shared placement plan creates it only for active EP-on DP expansion, and both run and replay pass it to `NewModelHardwareConfig`. Pool resolution copies the complete `SimConfig`, preserving `EPGroupDP`, and per-pool TP changes compose with the stored DP width. No other production simulator creation path performs this DP-to-1 transformation. |
| G7 | parity | ✅ CONFIDENT | How do `--prefill-moe-comm-backend` and `--decode-moe-comm-backend` maintain parity between `blis run` and `blis replay` when disaggregated PD pools or pool overrides are active? | Both commands register the same flags, use the same `applyPerRoleMoECommBackends` resolver, and write the results into the same `PrefillOverrides` and `DecodeOverrides` fields. `cluster.ResolvePoolConfig` then copies the selected backend into each role's `ModelHardwareConfig`, with empty overrides inheriting the global backend. Shared-role instances consistently use decode precedence. |
| G8 | parity | ✅ CONFIDENT | How does `blis observe` handle `--enable-expert-parallel` and per-role MoE comm backends, and does it remain byte-identical or bypass placement transformations consistently with `run` and `replay`? | `blis observe` is a black-box dispatcher to an external inference server and constructs no simulator placement or latency model. It does not register `--enable-expert-parallel` or the per-role backend flags, so such options are rejected as unknown rather than silently affecting observation. Consequently this PR does not change observe output or apply run/replay placement transformations there. |
| G9 | tests | ✅ CONFIDENT | Does `TestINV13_RunReplayParity_MoEDPPlacement` cover EP-on configurations with heterogeneous per-role backend overrides (e.g., HT for prefill, LL for decode), or only uniform global backends? | `TestINV13_RunReplayParity_MoEDPPlacement` covers EP-on with the inherited global backend, not heterogeneous prefill/decode overrides. Per-role behavior is tested separately with a PD deployment and cluster-resolution tests, while flag registration/default parity is checked for both commands. |
| G10 | tests | ❌ FLAW_FOUND | Is the expert weight sharding logic in `latency.ClampExpertShardToExpertCount` tested against structural implementation changes with a standalone companion invariant test, or only indirectly via CLI integration tests? | The shared clamp is directly exercised for KV-capacity behavior, but the newly added trained-physics step-time clamp path has no standalone behavioral test where `EffectiveEP()` exceeds `NumLocalExperts`. Existing EP step-time tests avoid the clamp or only test inert dense/EP-off cases. Because wide EP over fewer experts is explicitly described as a realistic supported input, this leaves new behavior without a discriminating test. |
| G11 | rationale | ✅ CONFIDENT | Why was logical EP DP width passed into engine replicas via `sim.WithExpertParallelGroupDP` rather than refactoring DP placement to preserve the original `DP` value in `ModelHardwareConfig`? | Preserving the original DP in each replica would reintroduce the `/dp` token-work scaling that DP-as-placement intentionally removes after the router has already split requests across N engines, causing double division. The logical EP width is a separate topology fact needed only for expert ownership and collective sizing, so an option carries that width while each replica correctly remains `DP=1`. Carrying a width rather than an absolute EP size also composes with per-pool TP overrides. |
| G12 | rationale | ✅ CONFIDENT | What alternative designs were considered for routing per-role MoE comm backends (`--prefill-moe-comm-backend`/`--decode-moe-comm-backend`) through `cluster.PoolOverrides` versus `ModelHardwareConfig`, and why was this approach chosen? | The implemented design treats the role choice as a deployment override and the resolved backend as a model-hardware latency input. `PoolOverrides` is the existing seam for values that differ between prefill and decode; `ResolvePoolConfig` produces a per-instance `SimConfig`, whose embedded `ModelHardwareConfig.MoECommBackend` is then consumed by the latency constructor. Putting the CLI role fields directly on a single global `ModelHardwareConfig` would either require duplicate role-specific fields or bypass the established pool-resolution mechanism. |
| G13 | scope | ✅ CONFIDENT | Does adding `cmd.validatePerPoolLatencyBackends` to guard against per-pool roofline latency models fall strictly within the scope of #1548, or does it extend into general latency backend validation? | It is a narrowly scoped guard made necessary by #1548. The pre-existing global check cannot see a per-pool roofline override; once EP affects step time at DP=1, such an override would make one active pool silently EP-blind. The helper only runs when DP/EP is active and permits absent or trained-physics overrides, so ordinary per-pool backend validation remains unchanged. |
| G14 | scope | ✅ CONFIDENT | Why were the `tpAllReduceBasis` volume-division changes and `cluster.warnIfCrossNodeUnpriced` scoring adjustments included in this PR rather than being deferred to #1530 or a dedicated cleanup PR? | The `warnIfCrossNodeUnpriced` group change is directly related to #1548 because EP can make the priced collective wider than TP; retaining TP-only diagnostics would miss that future spanning group. The `tpAllReduceBasis` adjustment fixes an existing accounting error exposed while touching the same network-aware collective path: DP divides byte volume, but not each rank's fixed collective launch latency. Neither adds a second communication term. |
| G15 | errors | ✅ CONFIDENT | When an invalid backend string is passed to `--prefill-moe-comm-backend` or `--decode-moe-comm-backend`, does the error trigger a immediate `logrus.Fatalf` in CLI parsing or return a library error in `sim/latency`? | At the helper/library boundary, validation returns an error rather than terminating. In the CLI command bodies, that returned error is immediately passed to `logrus.Fatalf`, so an invalid prefill or decode value fails during configuration processing before simulator construction. Library callers of `PoolOverrides.Validate` or the latency backend lookup likewise receive errors, consistent with the rule forbidding fatal exits in `sim/`. |
| G16 | errors | ✅ CONFIDENT | If a user configures a wide EP topology across inter-replica links where cross-node network costs are unpriced, does the system emit a clear `logrus.Warnf` diagnostic rather than silently under-reporting step latency? | Yes. For EP-on DP-as-placement, the resolver emits a warning that the EP group spans independently placed replicas, that inter-replica fabric cost is not priced, and that multi-node step time is optimistic. Separately, placement-derived cross-node diagnostics warn when a genuinely spanning instance cannot price its traffic. |

### Items to fix
- **G10 · FLAW_FOUND** — The shared clamp is directly exercised for KV-capacity behavior, but the newly added trained-physics step-time clamp path has no standalone behavioral test where `EffectiveEP()` exceeds `NumLocalExperts`. Existing EP step-time tests avoid the clamp or only test inert dense/EP-off cases. Because wide EP over fewer experts is explicitly described as a realistic supported input, this leaves new behavior without a discriminating test.
  _sim/latency/kv_capacity_ep_test.go:303-347 directly tests capacity clamping; sim/latency/trained_physics_model.go:1108-1122 adds the separate new step-time clamp; sim/latency/ep_mode_test.go:121-145 uses 64 experts with groups no wider than 16, and lines 253-285 test only inert cases. A realistic missing scenario is an MoE with 8 experts, TP=16, DP=1, EP enabled: the constructor should clamp the step-time weight divisor to 8 while leaving the 16-rank dispatch group unchanged._

### Important to consider
- **F1** — The tpAllReduceBasis launch-latency correction is adjacent cleanup rather than necessary EP functionality, but it is a concrete correctness correction and does not invalidate the PR scope.
  _cmd/root.go:397-422,507-534,652-684; sim/config.go:444-499; sim/latency/trained_physics_model.go:738-765,853-877; sim/latency/moe_comm_backend.go:37-92; cmd/dp_replay_parity_test.go:201-253; cmd/ep_mode_cli_test.go:19-33,134-217; `go test ./cmd`, `go test ./sim/latency`, and `go test ./sim/cluster` all pass._
- **F2** — Update the `--dp` help at cmd/root.go:1482; it contradicts the newly supported EP+DP behavior.
  _docs/reference/configuration.md:149-153; docs/guide/latency-models.md:178-251,253-278; docs/concepts/glossary.md:191; docs/concepts/roofline.md:10; docs/contributing/standards/invariants.md:180. The stale help remains at cmd/root.go:1482, which still says EP with DP is not supported._
- **F3** — These comments should be refreshed to avoid confusing users and future maintainers.
  _cmd/root.go:1482 still says `--dp` is not supported with expert parallelism; cmd/root.go:1089-1091 describes the global backend as trained-physics/DP>1-only despite EP now activating dispatch at DP=1; sim/latency/trained_physics_model.go:713-714 still calls the collective split a DP boundary before the later lines correctly describe the DP-or-EP boundary._
- **G3** — A trace cannot detect a user intentionally replaying with different DP/EP flags because those settings are not trace fields; this is the established identical-flags contract.
  _cmd/root.go:411-422,2512-2515; cmd/replay.go:670-673; cmd/dp_replay_parity_test.go:215-253; docs/reference/configuration.md:149-150; docs/contributing/standards/invariants.md:180._
- **G6** — The option itself does not reject a negative hand-written width, but production never emits one and `EffectiveEPGroupDP` safely falls back to the config DP.
  _sim/config.go:297-323,326-377; cmd/root.go:397-422,507-534,2512-2515; cmd/replay.go:670-673; sim/cluster/resolve.go:90-112; sim/config_test.go:952-965._
- **G9** — A run-to-replay test combining EP-on PD pools with heterogeneous role overrides would provide useful combination coverage, but the core EP parity and per-role paths are already tested independently.
  _cmd/dp_replay_parity_test.go:208-220 contains only tp1, tp2, and tp2-ep cases; cmd/ep_mode_cli_test.go:19-33 verifies run/replay flag parity and lines 108-216 exercise per-role PD behavior; sim/cluster/resolve_test.go:940-973 verifies independent HT/LL resolution._
- **G12** — The repository does not preserve a formal alternatives record, but the code structure clearly supports the chosen separation: role selection in deployment configuration, resolved backend in model hardware configuration.
  _sim/cluster/resolve.go:10-38,74-104; sim/cluster/deployment.go:179-201; sim/latency/moe_comm_backend.go:78-144; sim/cluster/resolve_test.go:940-973._
- **G14** — The launch-latency correction could have been a separate cleanup, but it is tested and is not a correctness regression.
  _sim/cluster/network_topology.go:77-89,140-156; sim/latency/trained_physics_model.go:724-735,887-910; sim/latency/ep_mode_test.go:288-320._

namasl:
Response to the qa-review gate (head `f08a8862`).

**G10 — the finding as stated is factually incorrect, but its example named a real gap. Both addressed.**

The claim is that no standalone step-time test exercises `EffectiveEP() > NumLocalExperts`, citing `sim/latency/ep_mode_test.go:121-145` and `253-285`. Those two ranges are `TestStepTime_EPShardsExpertWeightsAcrossTheGroup` and `TestStepTime_EPGroupDPIsInertWithoutEP` — unrelated tests. The clamp tests are at lines **465, 495 and 520** of that same file and were present at the reviewed SHA `a8f27e55`:

- `TestStepTime_EPWeightDivisorClampedToExpertCount` — `NumLocalExperts = 8`, EP groups `TP·2 = 16` and `TP·8 = 64`.
- `TestStepTime_EPOnClampLandsExactlyOnTheExpertCount` — 8 experts at EP=16, pinning the clamp *value*.
- `TestStepTime_EPOffFractionalExpertShareIsNotClamped` — the EP-off counterpart.

So the exact scenario the finding proposes ("8 experts, EP enabled, group wider than 8") was already covered.

**However**, the second half of its example — "*while leaving the 16-rank dispatch group unchanged*" — was genuinely untested. Nothing asserted the dispatch/combine collective stays **unclamped** past the expert count, so a change that clamped it too would have slipped through. Added `TestStepTime_EPDispatchGroupIsNotClampedToExpertCount`, which isolates it exactly: past the clamp the weight term is pinned, so any step-time difference between EP=16 and EP=64 can only be the dispatch term's rising `(group-1)/group`. Mutation-tested — pointing the dispatch calls at the clamped group makes both yield an identical `29607µs` and the test fails.

**Non-blocking notes F2/F3 — taken, all three verified stale:**
- `cmd/root.go` `--dp` help still read "Not yet supported with `--enable-expert-parallel` (#1548)" — the string a user actually sees.
- `cmd/root.go`'s `--moe-comm-backend` comment still said "DP>1" only.
- `trained_physics_model.go`'s term list still called the split a "DP boundary".

**G9** (EP-on × heterogeneous per-role backends in one run→replay parity case) is a fair coverage suggestion, declined here as combination coverage: EP-on parity (`tp2-ep`) and per-role PD behaviour are each tested, and PD + `--dp>1` remains a #1553 fail-fast, so the combination is narrower than it looks.

Local on `f08a8862`: `go build ./...`, `go test ./...`, `golangci-lint run ./...` all clean; no gofmt regressions vs `main`.

namasl:

## qa-review adjudication — PR #1683: ✅ PASS

|----|----|----|----|
| G10 | FLAW_FOUND | ✅ RESOLVED | The current code has discriminating trained-physics behavioral coverage for an EP group wider than the routed-expert count, including the previously missing assertion that only the weight divisor is clamped while dispatch/combine continues to use the full EP group. The latency package test suite passes. |

### Still blocking

### Cleared
- **G10** (✅ RESOLVED) — The current code has discriminating trained-physics behavioral coverage for an EP group wider than the routed-expert count, including the previously missing assertion that only the weight divisor is clamped while dispatch/combine continues to use the full EP group. The latency package test suite passes.
  _sim/latency/ep_mode_test.go:465-478 tests 8 experts with EP groups 16 and 64; sim/latency/ep_mode_test.go:520-531 pins the EP-on clamp value to the expert count; sim/latency/ep_mode_test.go:548-561 compares EP=16 against EP=64 after the 8-expert weight clamp and requires the wider, unclamped dispatch group to increase step time. The implementation separately clamps expertWeightShardGroup at sim/latency/trained_physics_model.go:1109-1123 while retaining expertShardGroup for collective topology at lines 1125-1126. `go test ./sim/latency` passes._

namasl:
@claude /blis-pr-review

namasl:
Round-3 findings addressed at head `7f8432c3`. **No further review lap requested** — the 3-cycle budget for this PR is spent, so this is a disposition record rather than a re-trigger.

**Finding B — real, fixed.** `trained_physics_model.go:739` had a *second* "partitions on the DP boundary" comment, 25 lines below the term-list comment `f08a8862` fixed at `:714`. Good catch: the two contradicted each other inside one function. Now "DP-or-EP boundary (vLLM, #1419, #1548)".

**Finding C — real, fixed.** The `--moe-comm-backend` **help string** still ended `+ --dp > 1)`. `f08a8862` fixed the `--dp` help and this flag's inline comment but not its own user-visible help text. Now `+ either --dp > 1 or --enable-expert-parallel)`.

**Finding A — not reproducible; rejected with evidence.** Both quoted sentences were removed in `a8f27e55`, one commit *before* the head this round reviewed:

```
$ git log --oneline -1
7f8432c3
$ grep -n "stays EP-blind\|still rejected by \`planDPPlacement\`" CLAUDE.md
(no match)
```

They were replaced in `a8f27e55` by "Since **#1548** the step-time model shards routed-expert WEIGHTS over the same expert group this capacity model uses…" and "…**#1548 lifted that**, so the EP-sharded deployment now runs end-to-end on both commands". The round-2 review confirmed Finding #4 as ✅ FIXED on that same content. Likely the diff's context lines were read rather than the file at HEAD.

**Both non-blocking coverage gaps closed**, since they were cheap and one had been raised twice:
- `TestPlanDPPlacement` gains the `dp=1 + epOn` short-circuit row (no expansion ⇒ no logical width to carry).
- `TestWidestCollectiveGroup` unit-tests the diagnostic's group directly. Its EP-wider branch is unreachable until #1553 — which is precisely why it needed a unit test, since the "inherit a correct diagnostic rather than a silent gap" intent otherwise rested on inspection alone.

The third gap (per-role heterogeneous backends through run→trace→replay byte-identity) I'm declining as combination coverage: EP-on parity (`tp2-ep`) and per-role PD resolution are each tested, and PD + `--dp>1` is still a #1553 fail-fast.

A method note, since this is the second round where help text slipped: my structural doc pass swept prose and summary tables but not cobra help strings. That is the copy a user actually reads, and it is now part of the sweep.

Local on `7f8432c3`: `go build ./...`, `go test ./...`, `golangci-lint run ./...` all clean; no gofmt regressions vs `main`; both residual strings confirmed absent.

namasl:
@claude /blis-pr-review

sim/latency/trained_physics_model.go
```diff
@@ -2,6 +2,7 @@ package latency
 
 import (
 	"fmt"
+	"github.com/sirupsen/logrus"
```
susiejojo: **Nit (style, non-blocking): misgrouped import.** The new `github.com/sirupsen/logrus` is placed inside the stdlib block (`fmt` / logrus / `math`) rather than the third-party group:

```go
import (
	"fmt"
	"github.com/sirupsen/logrus"   // ← third-party, wedged between stdlib
	"math"

	"github.com/inference-sim/inference-sim/sim"
)
```

`gofmt` does not reorder across blank-line groups so it passes, and the repo’s golangci-lint v2 default set does not enable `gci`/`goimports`, so CI is green — but it is inconsistent with the sibling file this PR mirrors (`sim/latency/kv_capacity.go`), which correctly groups `logrus` with the other third-party imports below the stdlib block. Suggest:

```go
import (
	"fmt"
	"math"

	"github.com/sirupsen/logrus"

	"github.com/inference-sim/inference-sim/sim"
)
```

susiejojo:
## BLIS PR Review — #1683 (fresh pass, focus: relationship to #1531 + correctness/quality)

Reviewed at head `7f8432c3`. Built and ran `go test ./sim/... ./cmd/...` clean; validated the headline behaviour and the two byte-identity invariants end-to-end (see below). This is a targeted lap on top of the four prior `/blis-pr-review` rounds and the qa-review gate — I did **not** re-litigate findings already dispositioned; I focused on what the requester asked: **how this relates to #1531**, plus an independent correctness/quality sweep with one inline comment.

---

### 1. Did we merge something related to #1531, and how does it relate?

**Yes — #1531 is already on `main` and this PR is its direct successor.** The lineage:

| PR / issue | State | What it did |
|---|---|---|
| **#1531** (`0546dfa2`, merged) | on `main` | DP-as-**real-placement** for MoE `--dp N` on `blis run` — spawns `N` engine replicas, each rewritten to **`DP=1`**. Modelled **EP-OFF** physics only, and **explicitly rejected `--dp>1 --enable-expert-parallel`** in `planDPPlacement`. |
| **#1556** (`e3598b84`, merged) | on `main` | Extended the same `resolveDPPlacement` to `blis replay`, so the `#1548` EP-on rejection covered both commands. |
| **#1656** (`017fcd98`, merged) | on `main` | EP-aware **KV-capacity** weight-footprint sizing; introduced `EffectiveEPSize` and documented the "DP-as-placement collapses the per-replica EP group to TP" trap on the capacity side. |
| **#1683** (this PR) | open | Split from #1531 per issue #1548. **Lifts** #1531's `--dp>1 + EP` rejection and makes the toggle **step-time-live**. |

The single most important interaction is **the #1531 collapse trap**, and this PR handles it correctly. Because #1531 rewrites each replica's own `DP` to `1`, any EP-group size read off a per-replica config collapses to `TP` and silently no-ops the sharding. This PR carries the **logical DP width** through the replica boundary via a new `WithExpertParallelGroupDP` option (held on `dpPlacementPlan.EPGroupDP`, spread by `EPGroupOptions()` at both `NewModelHardwareConfig` sites in `cmd/root.go` and `cmd/replay.go`). `EffectiveEPGroupDP()` takes `max(ownDP, EPGroupDP)` so the option can only *widen* a group — a stale/too-small value can never shrink one. This mirrors what #1656 already did on the KV-capacity side, and the PR wires **the same logical width** into `applyPerInstanceKVCapacity` (`sim/cluster/kv_autocalc.go`) so step-time and capacity now shard the identical experts over the identical group.

**Net:** #1683 completes the DP/EP-placement story #1531 began. #1531 = "place the replicas, EP-off physics"; #1683 = "flip EP on: reshard the expert *weights* across those already-placed replicas, and swap the MoE-FFN all-reduce for a dispatch/combine all-to-all — reserving **zero** extra GPUs." The GPU reservation and the `Replicas`/`PerRankDP` plan are untouched by EP (verified: `planDPPlacement` returns identical `Replicas`/`PerRankDP` with or without `epOn`; only `EPGroupDP` differs).

---

### 2. Correctness — verified empirically

- **EP toggle is genuinely live** (issue #1548 AC-2). `deepseek-v2-lite --tp 2 --dp 4`, 40 req, seed 42: mean ITL **7.58 ms → 2.99 ms** EP-off→EP-on. Attributed correctly in the PR body to routed-expert **weight** resharding, which I confirmed is the mechanism: the compute path stays on `moeGroup` (`Resolve(..., m.moeGroup, ...)` at `trained_physics_model.go:595,644`) while only the weight term and the dispatch collective use `expertShardGroup` (`:689,859,877`). The compute/weight group split is the crux and it is implemented as described — dividing compute by the EP group too would have under-charged it by `DP`.
- **INV-6 (determinism):** repeated identical EP-off runs byte-identical. ✅
- **INV-6 (EP-on-neutral at DP=1):** `--tp 4` EP-off vs EP-on with the default `allgather_reducescatter` backend → **byte-identical stdout**, confirming no double-charge (all-gather + reduce-scatter *is* the ring-all-reduce decomposition; β_EP defaults to β₄). ✅
- **INV-13 (run→replay parity):** `--tp 2 --dp 4 --enable-expert-parallel` run with `--trace-output`, replayed with identical flags (+ matching `--horizon`) → **byte-identical stdout**. ✅ Correctly treats the EP width as a re-supplied model-level CLI input (like `--kv-cache-dtype`), not a trace field.
- **Divisor-clamp agreement:** the "one whole expert per loaded rank" clamp (`ClampExpertShardToExpertCount`) is now shared between the KV-capacity model and the step-time model, applied to the **weight** divisor only and **not** to the dispatch collective — physically correct (EP-off tensor-sharding legitimately holds a fraction of every expert; the all-to-all really spans every rank). The G10 dispatch-group-unclamped coverage gap raised by qa-review is now closed by a discriminating test.
- **Latent-hole closures are real, not cosmetic:** `validatePerPoolLatencyBackends` genuinely closes a hole that only became live with this PR (a per-pool `--decode-latency-model roofline` override would put one pool on DP/EP-blind step time while the cluster runs EP). Good defensive catch.

### 3. Quality

- **`tpAllReduceBasis` NaN guard** (`!(dpDivisor > 0)`) is the right idiom and consistent with `tpCommBwUs`'s existing guard; `dpDivisor` is provably `1.0` on every CLI-reachable path (DP-as-placement gives every replica `DP=1`), so this is byte-identical and the guard is belt-and-suspenders for struct-literal test callers.
- **Shared resolvers** (`anyPerPoolHardwareFlagChanged`, `applyPerRoleMoECommBackends`, `EPGroupOptions`) correctly consolidate run/replay so the two commands cannot drift (R23). The per-role backend validates the name loudly at three layers (CLI, `PoolOverrides.Validate`, model constructor) — appropriate, since a roofline-resolving pool would otherwise swallow a bad value.
- **`nominalAll2AllProfile` placeholder** (`commScale = 1.0` for every backend) is an honest, exact multiplicative identity; the two sentinel tests keep it from silently diverging when #1568 lands. Good separation of "wire it up now, calibrate later."
- **Docs** (CLAUDE.md, latency-models guide, INV-13, glossary, flag help) are updated and internally consistent at this head; the stale-comment findings from rounds 3–4 are resolved.

### 4. Findings

- **[Non-blocking, inline]** Misgrouped `logrus` import in `sim/latency/trained_physics_model.go:5` — wedged into the stdlib block instead of the third-party group, inconsistent with sibling `kv_capacity.go`. gofmt passes and CI's linter set doesn't flag it, so it's a pure style nit. Inline comment posted.

No correctness issues found in this pass.

---

### Verdict: ✅ **Ready to merge.**

The feature does exactly what issue #1548 asks, cleanly completes the #1531/#1556/#1656 DP/EP-placement arc, and the three byte-identity invariants (INV-6 determinism, INV-6 EP-neutral-at-DP=1, INV-13 run/replay) hold empirically. Test coverage is behavioral and the key laws are mutation-tested per the PR body. The only finding is a cosmetic import grouping.

*Note: `gofmt -l` flags `sim/cluster/resolve.go`, but that is a pre-existing comment-block reformatting under local Go 1.24 — it is present on the merge base `f4c8c619` too and is not introduced by this PR (CI uses a pinned toolchain).*

sim/latency/trained_physics_model.go
namasl: Fixed in `d803e1c1`. Thank you — and thank you especially for spelling out *why* no gate caught it. That was the useful part: `gofmt` not reordering across blank-line groups, and this repo's golangci-lint v2 default set enabling neither `gci` nor `goimports`, means nothing automated in the repo can see this class of defect. I'd been treating a clean `gofmt -l` plus `golangci-lint` as covering import hygiene, and it doesn't.

**One deviation from your snippet, deliberately — please push back if you'd rather have the split.** Your reasoning cites `kv_capacity.go` as the convention, and that file (plus `config.go`, the other non-test `sim/latency` file importing logrus) puts `inference-sim/sim` and `sirupsen/logrus` in a **single** alphabetical third-party group rather than two:

```go
import (
	"fmt"
	"math"

	"github.com/inference-sim/inference-sim/sim"
	"github.com/sirupsen/logrus"
)
```

Your suggested form separates logrus into its own group above the local import. I went with the single-group form so the block is now byte-identical to `kv_capacity.go`'s — maximum consistency with the sibling file you cited, and it satisfies the principle you stated ("group logrus with the other third-party imports below the stdlib block") either way. Happy to switch to the two-group form if that's the convention you intended.

I also swept every `.go` file this PR touches for the same defect — a third-party import sharing a group with stdlib imports — since I added `logrus` in more than one place. `trained_physics_model.go` was the only one. `sim/cluster/kv_autocalc.go` and `network_topology.go` have no stdlib imports at all, so their leading third-party group is correct, and this PR changed no import line in either.

Head `d803e1c1` is being re-gated now; CI and the review will report on it.

namasl:
@claude /blis-pr-review

susiejojo (dismissed):
lgtm

namasl:
@claude /blis-pr-review

namasl:
@claude /blis-pr-review

namasl:
Round-6 APPROVE acknowledged. **Its one finding is correct, and I am declining it deliberately — not disputing it.**

**Finding: `TestRunCmd_PerRoleFlagErrors_AreFatal`'s comment claims "on both commands" but the subprocess only drives `run`.** Verified and accurate: the test calls `epPDRunArgs(...)` only, so replay's identical call sites (`cmd/replay.go:593` `validatePerPoolLatencyBackends`, `:599` `applyPerRoleMoECommBackends`) are never exercised. The comment overstates its own coverage by one command.

**Declined because the branch is frozen, not because the finding is wrong.** `main`'s ruleset carries `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`. This PR already had a human approval (`susiejojo`, `d803e1c1`) and **I destroyed it with a push** that fixed only non-blocking observations. That was a bad trade: the observations were optional, the approval was not. So `d67ec3d4` stays as it is. Any commit — including a four-word comment edit — would dismiss the next approval the same way.

For the record, so the correction isn't lost if someone picks it up later: the accurate comment would be *"Covered on `run`; replay's two call sites are structurally identical (same shared helpers, same `Fatalf` conversion) but are not exercised here."* That is a one-line change for whoever next has the branch open for an unrelated reason. It is not worth a commit of its own, and it is not worth an approval.

Nothing else outstanding from rounds 5–6: obs 1, 2, 3 and 5 addressed in `d67ec3d4`; obs 4 declined by the round that raised it.

**#1683 now needs two things, neither of which I can provide:**
1. Review thread `PRRT_kwDOO48g086fTjOW` (@susiejojo's import nit — fixed in `d803e1c1`) **resolved**. `required_review_thread_resolution: true` blocks on it, and `isOutdated=true` does not exempt it. Resolving another reviewer's thread is theirs to cede, so I have not.
2. **A fresh human approval on `d67ec3d4`** — which, under `require_last_push_approval`, will now stick, since there are no further commits coming.

All 13 checks are green on `d67ec3d4` (11 pass, 2 skipping). I have made no commits since `d67ec3d4` and will make none.

susiejojo (approved):
lgtm

PR #1695, opened 2026-09-09
feat(latency): model cross-node collective latency as analytic hop-count × α_hop × S (#1694)

### LINKED ISSUE #1694: Model cross-node collective latency as analytic hop-count × per-fabric α_hop (restructure the flat #1667 term)
> **Split off from #1661.** #1661 stays scoped to *calibrating* the flat per-collective term #1667 shipped: supply a measured latency and update the pin test. **This issue is the separate, larger restructure** — turning the flat cross-node latency constant into a topology-aware analytic **hop-count** form. See the empirical anchor in [#1661's discussion](https://github.com/inference-sim/inference-sim/issues/1661).

## Scope of this issue (read first)

This issue restructures the flat cross-node latency term into the full analytic form:

```
T_collective_fixed = n_steps(algorithm, topology) · α_hop · S
```

It delivers **two structurally-separate terms** — and the one hard rule is that they stay separate constants, never merged into one:

**Part A — analytic hop-count × per-fabric `α_hop` (the fabric/topology term):**
- Compute the cross-node **hop count** analytically from the collective algorithm + placed node span (already available via `NetworkTopology.NodesSpanned(groupSize)`).
- Charge `n_steps · α_hop` in the two size-independent halves that #1667 added (`tpCrossNodeLatencyUs`, `moeCrossNodeLatencyUs` in `sim/latency/trained_physics_model.go`).
- `α_hop` is a per-**fabric** `HardwareCalib` constant, calibrated once from an NCCL microbenchmark.

**Part B — eager/no-overlap serialization multiplier `S` (the scheduling-regime term):**
- A multiplier on `n_steps · α_hop` capturing whether the deployment runs CUDA graphs / comm-compute overlap (`S ≈ 1`) or `enforce-eager` + un-fused + no overlap (`S ≫ 1`, all collectives serialize behind a full barrier each).
- `S` is a **deployment/regime** property, so it is exposed as a **CLI flag or model-config field**, NOT a `HardwareCalib` value.
- Default `S = 1` (graphs-on / overlap) ⇒ Part B is inert and this issue's output equals Part A alone when the regime is unset.

**The invariant that governs both:** `α_hop` (fabric) and `S` (regime) are kept as **separate, independently-calibratable constants**. `S` must **not** be folded into `α_hop`, or a graphs-ON deployment would inherit a graphs-OFF fabric constant. Keeping them distinct is the whole point of the α–β form — fabric latency, algorithm hop-count, and eager-serialization each stay independently checkable.

**One PR attacks both parts.** Parts A and B are two factors of one term (`n_steps · α_hop · S`) and are delivered together in a single PR — not split into two PRs or sub-issues. "Two parts" means *two separate constants in the code*, not two units of delivery: Part A alone is neither the dominant factor nor end-to-end validatable (see the 438866 decomposition below), and Part B needs Part A's `n_steps · α_hop` term to multiply. The only thing that would justify splitting is if Part B's regime-detection design blows up into its own discussion during planning — in that case, flag it then; the default is one PR covering both.

## Where this sits in the 438866 calibration gap (necessary, not sufficient)

This issue targets a real term in the decode residual, but it does **not** close the `experiments/` calibration gap on its own, and it is **not** the dominant contributor. Anyone picking this up should know the full decomposition so they don't expect the 438866 numbers to move when this lands. The measured gap (from `experiments/README-calibrate-inferencex.md`, "Known fidelity boundaries"):

| metric | BLIS vs real | driver |
|---|---|---|
| TTFT | ~1.2–1.3× (good) | prefill compute-bound; roofline captures it |
| **TPOT / E2E** | **~5–10× optimistic (too fast)** | the residual below |
| cache-hit | under-predicted | single cold closed-loop session vs warm cross-request prefix cache |

The TPOT/E2E residual splits across the following pieces of work. This issue delivers the cross-node latency model (both the hop-count structure **and** the serialization multiplier); the other two are separate:

| contributor to the decode residual | tracked by | in this issue? |
|---|---|---|
| Cross-node collective latency **structure** (hop-count × `α_hop`) | **#1694 (here, Part A)** | ✅ yes |
| Eager / no-CUDA-graph **serialization `S`** | **#1694 (here, Part B)** | ✅ yes (separate term, not merged into `α_hop`) |
| Cross-node topology **detection on the DP/EP path** | separate (entangled with #1531/#1548/#1553) | ❌ blocker (see caveat below) |
| MTP speedup **without contention** | #1627 | ❌ separate |

Two consequences worth stating plainly:

1. **This term is inert on the 438866 replay path today.** That run replays as `--tp 16 --dp 2 --enable-expert-parallel` with **no `node_pools`** (`experiments/calibrate_inferencex.py`), so DP-as-placement (#1531) builds per-replica configs with no `NetworkTopology` ⇒ `NodesSpanned()` returns 1 ⇒ the cross-node latency never fires, whatever `α_hop` or `S` is set to. End-to-end validation on 438866 is **blocked** on the DP/EP topology-detection work, not on this issue. (Both parts are testable today on the multi-node **TP** path, #1529.)
2. **`S` is the dominant part of the cross-node term — which is exactly why it is IN this issue, as its own factor.** Back-solving the ~400 ms residual over the hop count gives `α_hop` ~100× a plausible NCCL step (see "Why a raw α_hop is not the whole story" below). That is not evidence for a large fabric constant — it is the signature of eager serialization. Modeling `α_hop` (Part A) without `S` (Part B) would tempt whoever calibrates it to inflate `α_hop` to close the gap, overfitting a regime penalty into a fabric constant. Delivering both together, as separate factors, is what prevents that.

So the honest framing: **#1694 delivers a complete, correctly-factored cross-node latency model (`n_steps · α_hop · S`). It is a genuine physics fix, but the `experiments/` TPOT/E2E gap only fully closes once the DP/EP topology plumbing (so the term fires on 438866) and #1627 also land.**

#1667 (closing #1530) added `HardwareCalib.InterNodeLatencyUs`: a fixed cost charged **once per cross-node collective**, so `units · latencyUs` for the TP path and `moeDispatchCollectivesPerLayer · latencyUs` per MoE layer. That multiplier is **node-span-invariant** — a collective is priced identically whether it spans 2 nodes or 8.

That is physically wrong for the collective *algorithms* real servers run. A hierarchical (two-level) ring all-reduce does its inter-node phase in `2·(nodes−1)` steps; an all-to-all over `nodes` does `(nodes−1)` per direction. So the fixed-cost multiplier should grow with the node span, and the current flat term **under-charges a wide span by ~(nodes−1)×** (e.g. ~3× on a 4-node placement).

### Empirical anchor (from #1661)

Calibrating BLIS against a real InferenceX run (id 438866: Kimi-K3, H200, TP16/DP2/EP32 across 4×8-GPU nodes, MTP, conc=1):

| | ms per decode step |
|---|---|
| Real (measured) | **~462** |
| BLIS models (weight-load + KV + FLOP + β overhead) | ~34–66 |
| **Unpriced residual** | **~400 (≈90% of the step)** |

A roofline check confirms decode is **not** weight-bound here (~2 ms/step of active-weight stream at conc=1), so the ~400 ms really is cross-node collective + serialization — exactly this issue's "dominant for decode because messages are small" prediction. Both halves of that residual — the fabric hop-count (`α_hop`) and the eager serialization (`S`) — are modeled by this issue, as the two separate factors of `n_steps · α_hop · S`.

## Proposed solution

Model the fixed cost as an **analytic hop count × per-fabric constant × regime multiplier**:

```
T_collective_fixed = n_steps(algorithm, topology) · α_hop · S
                     └──── analytic (no fit) ────┘  └fabric┘ └regime┘
                     └──────────── Part A ─────────────────┘ └Part B┘
```

This *replaces* the flat `latencyUs` multiplier inside the two size-independent halves #1667 already wired (`tpCrossNodeLatencyUs`, `moeCrossNodeLatencyUs`); the size-**dependent** bandwidth halves (`tpSpanScale` / `moeSpanScale`, priced by #1530) are unchanged.

### Part A — `n_steps · α_hop`

- **`n_steps` is analytic** — derived from the collective algorithm and topology, no fitting:
  - **hierarchical ring all-reduce** (the TP-group collectives — attention, dense-FFN, and the DP==1 MoE-FFN reduce): inter-node phase = `2·(nodes−1)` steps **per collective**.
  - **all-to-all** (the MoE expert dispatch/combine): `(nodes−1)` steps per direction **per collective**.
  - The **collective count per layer** is a separate, already-modeled factor (2 for TP: attention + FFN; `moeDispatchCollectivesPerLayer = 2` for MoE: dispatch + combine). `n_steps = collectives_per_layer · hops_per_collective(algorithm, nodes) · layers`.
- **`α_hop` is the single measurable** — an NCCL microbenchmark on a 2-node reference cluster, measured **once** and reused across every model/topology (parallel to MFU / Discussion #589). The hop count absorbs all model/topology variation, so the same `α_hop` predicts GLM-5.2 TP16×2nodes, Kimi-K3 EP32×4nodes, etc. This is what separates a **fabric calibration** from curve-fitting one run.

### ⚠️ The worked hop-count example is unreliable — re-derive it against the real comm structure

The original `738`-hops figure (TP `93 × 2×(2−1) = 186`; MoE `92 × 2 × (4−1) = 552`) should **not** be trusted as the target. It has two problems the implementation must resolve carefully:

1. **Convention ambiguity.** The collective-count-per-layer factor (2) and the ring's `2·(nodes−1)` hops-per-collective factor are easy to double-use or drop, and they **coincide at a 2-node span** (`nodes−1 = 1`) — so the TP arithmetic that reads as `186` silently drops the ring's factor of 2 without the coincidence exposing it. Pin **one** convention and test it at a **≥3-node span** so the two are distinguishable.

2. **The per-path layer counts differ, and EP changes which collective fires.** In `sim/latency/trained_physics_model.go`, the TP all-reduce latency is charged as `units · latencyUs`, but `units` is **not** a uniform "layers × 2":
   - **attention all-reduce** → `numLayers` (Kimi-K3: 93, though #1636's hybrid split means only the 24 full-attention layers do the O(context) work — confirm whether the *latency* units should follow that split too);
   - **dense-FFN all-reduce** → `numDenseLayers` (Kimi-K3: **1**, since `first_k_dense_replace=1`), **not** 93;
   - **MoE-FFN**: at 438866 (**EP32, EP-on**) the `tMoEReduce` all-reduce leg does **not** fire — MoE layers route through the `tMoEDispatch` all-to-all (`numMoELayers` = 92) under the `DP>1 || EP-on` gate. The `2×(nodes−1)` ring hop count applies to the *ring* legs; the all-to-all leg uses `(nodes−1)` per direction.

So the correct `n_steps` is per-leg (`attention_layers · hops_ring`, `dense_layers · hops_ring`, `moe_layers · collectives · hops_a2a`), not one lumped `layers × 2 × hops`. Deriving the exact 438866 total is **implementation work** — this issue deliberately does **not** ship a replacement number, to avoid trading one wrong constant for another. The run-derived numbers stay an order-of-magnitude sanity check only.

### Part B — the serialization multiplier `S` (why a raw α_hop is not the whole story)

Back-solving the ~400 ms residual over the hop count implies an `α_hop` ~100× a plausible eager NCCL step (raw wire RTT ~2–5 µs; eager launch ~10s of µs). That µs-vs-ms gap is **not** evidence that `α_hop` should be large — it is the signature of the **eager/no-overlap serialization** this recipe runs (`enforce-eager: true` + `fuse_allreduce_rms: false`: no CUDA graphs, un-fused, no comm/compute overlap), where all collectives serialize behind a full barrier each.

That serialization is the multiplier `S` in `n_steps · α_hop · S`, and it is **Part B of this issue** — delivered as a separate factor, never folded into `α_hop`:

- **Regime, not fabric.** `S` is a property of the *deployment* (graphs-on vs `enforce-eager`), so it is a **CLI flag or model-config field**, not a `HardwareCalib` value. (Contrast `α_hop`, which is per-fabric hardware calibration.)
- **Default `S = 1`** (graphs-on / overlap): Part B is inert, output equals Part A alone.
- **Never merged.** Sizing `α_hop` to absorb `S` would make a graphs-ON deployment inherit a graphs-OFF constant. The run-derived per-hop number is an **order-of-magnitude sanity check on the product, not the value of `α_hop`**.
- **Interaction to resolve during implementation:** the β overhead coefficients (β₄/β₅ etc.) were fitted against runs that partly assume graphs-on, so `S` must be defined so that `S = 1` leaves today's β-based step time byte-identical (INV-6). `S` scales the cross-node latency term specifically, not the whole step.

**Acceptance criteria for Part B:**
1. `S = 1` (default / unset regime) ⇒ byte-identical to Part A alone (INV-6).
2. Monotone: raising `S` never lowers the charged cross-node latency.
3. `S` is surfaced through a regime flag/field and recorded in provenance, never baked into `α_hop`.
4. The eager regime (`S > 1`) is only chargeable when the cross-node latency term itself fires (spans nodes AND `α_hop > 0`) — `S` multiplies that term, it does not create cost on its own.

## ⛔ Anti-overfitting guardrail (hard requirement for the PR)

The 438866 run is an **anchor, not a fitting target**. The entire α–β / hop-count structure exists so BLIS predicts cross-node latency it was *not* fitted to. A PR that tunes `α_hop` or `S` until 438866's TPOT/E2E matches — and calls that done — has defeated the point. The following are binding on the implementation:

1. **`n_steps` carries ZERO free parameters.** It is derived purely from the collective algorithm + `NetworkTopology.NodesSpanned()`. No fitted layer weights, no per-model fudge factor, no "×1.7 to make Kimi-K3 line up." If a number in the hop-count helper isn't traceable to the topology or the algorithm, it doesn't belong.
2. **`α_hop` is ONE per-fabric constant, calibrated OFF the 438866 numbers.** Its source must be an independent NCCL microbenchmark (a 2-node reference cluster, the MFU / Discussion #589 pattern), not a back-solve of the 438866 residual. The run-derived ~542 µs/hop stays an **order-of-magnitude sanity check on the product**, explicitly *not* the committed value. The same `α_hop` must be reused unchanged across fabrics/topologies (H100, H200, A100…) — a per-run `α_hop` is the overfitting failure mode and is rejected.
3. **`α_hop` and `S` ship at their inert defaults (`α_hop = 0`, `S = 1`) in this PR** unless a genuinely independent measurement lands with them. Committing a value chosen to fit 438866 is worse than committing 0 — 0 is honest about the gap; a fitted value hides it. This mirrors #1667's deliberate ship-at-0 decision and keeps the pin test meaningful.
4. **No single-run calibration claim.** The PR must **not** report "438866 now matches" as evidence of correctness. 438866 is one point, on a path where the term is currently inert (see the plumbing caveat), and its residual is dominated by `S` — matching it proves nothing about the fabric constant. Validity is argued structurally (right algorithm, right hop count, right monotonicity), tested on the multi-node **TP** path (#1529), and cross-checked against the microbenchmark — not by curve-fitting one deployment.
5. **Generalization is the acceptance bar, not fit.** If a reviewer asks "would this same `α_hop` predict GLM-5.2 TP16×2nodes and a hypothetical 8-node run without re-tuning?", the answer must be yes by construction. A coefficient that only works for 438866 fails review.

**Tests that enforce this (must be in the PR):**
- The hop-count helper is unit-tested against hand-derived topology values at **≥3-node spans** (so `2(nodes−1)` vs `(nodes−1)` are distinguishable — see the convention warning above), with **no reference to any InferenceX run**.
- The committed-config pin test (`TestCommittedHardwareConfig_LatencyIsUncalibrated`, retargeted) asserts `α_hop` ships at 0, so a fitted value cannot land silently.
- Monotonicity + `S=1`/`α_hop=0` byte-identity tests (INV-6), none of which reference a fitted target.

**What 438866 IS legitimately used for:** motivating the *structure* (that a fixed per-hop cost exists and is large for eager decode), sizing the *order of magnitude* to catch unit errors, and — once the DP/EP topology plumbing lands — an *end-to-end sanity check* that the modeled cross-node time is the right order, with `S` set to the run's known eager regime and `α_hop` from the microbenchmark. That is a check, never a fit.

## The `α_hop` calibration field — decision

**Recommendation: a single, renamed per-hop field.** Repurpose #1667's `HardwareCalib.InterNodeLatencyUs` into a per-**hop** constant, and **rename it** (e.g. `InterNodeHopLatencyUs`) to make the units change explicit.

Rationale:
- The unit genuinely changes: `InterNodeLatencyUs` is µs **per collective**; `α_hop` is µs **per hop**. They differ by the hop count (×2 at a 2-node ring span, more at wider spans). **Keeping the name and silently changing the meaning would misinterpret any calibrated value by the hop count** — a latent trap.
- A **new coexisting field** (two 0-valued latency knobs) invites "which one drives?" confusion and leaves dead surface. One knob is cleaner.
- Renaming forces recalibration rather than a silent rescale, which is the honest outcome given the unit change.
- Safe today: every committed config ships the latency at 0 (pinned by `TestCommittedHardwareConfig_LatencyIsUncalibrated`), and no calibrated value has landed, so the rename breaks nothing. `α_hop` ships at **0** too (uncalibrated ⇒ not charged), preserving INV-6 / INV-BC-DP1 exactly as #1667's flat term does.

### Interaction with #1661 (must be sequenced)

#1661 calibrates the **flat per-collective** `InterNodeLatencyUs`; this issue **supersedes** that term. If #1661 lands a committed value first, this issue **removes** it in favor of the per-hop `α_hop` — a **recalibration** (different units), not a rescale. The pin test (`TestCommittedHardwareConfig_LatencyIsUncalibrated`) should be renamed/retargeted to the new field, staying a "0 unless a measured source lands" guard.

- [x] Core simulator (`sim/`) — **Part A:** `HardwareCalib` field rename (`InterNodeLatencyUs` → `InterNodeHopLatencyUs`) + accessor. **Part B:** the `S` regime value (a `ModelHardwareConfig`/latency-model input, mirroring how TP/DP/`MoECommBackend` are plumbed — R16). `NetworkTopology` unchanged (only read via `NodesSpanned`)
- [x] Latency model (`sim/latency/trained_physics_model.go`) — **Part A:** the `n_steps` hop-count helper feeding `tpCrossNodeLatencyUs` / `moeCrossNodeLatencyUs`; `crossNodeLatencyUs`, `moeCrossNodeLatency`, `tpAllReduceBasis`'s size-independent half. **Part B:** apply `S` as a multiplier on the size-independent cross-node latency term only
- [x] CLI (`cmd/`) — **Part B:** a regime flag (e.g. `--eager-serialization` / `--comm-serialization-factor`) on both `blis run` and `blis replay` (shared flag registration), threaded into the latency config; default leaves `S = 1`
- [x] `hardware_config.json` — **Part A:** key rename only (all values stay 0/absent)
- [x] Tests — `sim/latency/network_cost_test.go` (Part A: hop-count monotonicity-in-node-span at ≥3-node spans; Part B: `S = 1` byte-identity + monotone-in-`S`), `sim/latency/interconnect_calib_file_test.go` (retarget the pin test)
- **Not affected:** the bandwidth halves (#1530), the roofline backend (prices no comm term)

## Extension friction check

- **Files:** ~5–7 (`sim/model_hardware_config.go`, `sim/latency/trained_physics_model.go`, `hardware_config.json`, `cmd/` flag registration for Part B, the two test files above).
- **New interface?** No — extends the existing `HardwareCalib` + `NetworkTopology` surface #1667 established. `n_steps` reads the already-plumbed `NetworkTopology.NodesSpanned(groupSize)`; `S` follows the existing latency-input plumbing (TP/DP/`MoECommBackend`).
- **Invariants:**
  - **INV-6 / INV-BC-DP1** — `α_hop` ships at 0 (uncalibrated ⇒ not charged) and `S` defaults to 1; the term only changes the *value* charged when the group already spans nodes AND `α_hop > 0`, so every existing config stays byte-identical, exactly as #1667's flat term does.
  - **INV-13** — run/replay parity: `α_hop` rides the same rejection guard (`blis replay` rejects multi-node traces via `TraceHeader.MaxNodesSpanned`); `S` is a model-level input re-supplied on both `run` and `replay` CLIs (like `--kv-cache-dtype`), not round-tripped through the trace header, so a run and its replay with identical flags stay byte-identical.

## Alternatives considered

- **Flat per-collective (status quo, #1667):** node-span-invariant ⇒ under-charges wide spans by ~(nodes−1)×. Part A fixes exactly that.
- **Single fitted ms/collective per run:** overfits one deployment; won't transfer across topologies. Rejected in favor of the analytic hop count.
- **Fold serialization `S` into `α_hop` (one constant):** contaminates a fabric constant with a scheduling-regime property; a graphs-ON deployment would inherit a graphs-OFF value. Rejected — `S` and `α_hop` are delivered as **separate factors** of `n_steps · α_hop · S`, not one number.
- **Defer `S` to a separate issue:** considered, rejected. `S` is the *dominant* factor of the same cross-node term and the only real anchor (438866) needs it to reproduce the measured step, so Part A alone is neither the big term nor end-to-end validatable. They are two factors of one product; splitting the issue would leave `α_hop` un-anchorable and invite overfitting `S` into it. (If `S`'s regime-detection design turns out to need its own discussion, spin it out then — but the default is to deliver both here.)
- **New coexisting `α_hop` field alongside `InterNodeLatencyUs`:** two 0-valued knobs, redundant surface. Rejected in favor of the rename.

- **Builds on / supersedes #1661** (calibration of the flat term) and **#1667 / #1530** (the mechanism this restructures).
- **Plumbing caveat (blocks end-to-end validation on the DP/EP path):** a populated `α_hop`/`S` is **inert on the path that motivated this**. `crossNodeLatencyUs` gates on `NetworkTopology.NodesSpanned() > 1`, but DP-as-placement (#1531) builds per-replica configs with **no** `NetworkTopology`, so the span is never detected on the DP/EP path, and `blis replay` rejects multi-node traces. This term and the DP/EP topology-detection are **two separate pieces of work** — both Part A and Part B are testable today on the multi-node **TP** path (#1529), but can't be validated end-to-end on the 438866 (TP16/DP2/EP32) shape until the topology plumbing on the DP/EP path exists (entangled with #1548/#1553).
- **Calibration frame:** the latency rides the learned communication coefficient (β₄ for TP, β_EP for MoE dispatch), so the charge is `β · n_steps · α_hop · S`. Calibrate `α_hop` in that frame (per-fabric, `S = 1`), then calibrate `S` separately against a known eager run — not as one raw wall-clock number.

susiejojo:
## Summary

Restructures the flat per-collective cross-node latency term (`InterNodeLatencyUs`, #1667) into the full analytic form

```
T_collective_fixed = n_steps(algorithm, topology) · α_hop · S
                     └──── analytic, no fit ────┘  └fabric┘ └regime┘
                     └────────────── Part A ────────────────┘ └Part B┘
```

delivered as **two structurally-separate constants that are never merged**, per the issue's hard requirement.

- **Part A — `n_steps · α_hop`:** `n_steps` is the analytic **per-collective** cross-node hop count, derived purely from the collective algorithm + `NetworkTopology.NodesSpanned()` with **zero free parameters**. One whole ring all-reduce (reduce-scatter + all-gather) is `2·(nodes−1)`; one single-phase collective (an all-to-all direction, or one phase of the all-gather family) is `(nodes−1)`. `α_hop` is the per-fabric `HardwareCalib.InterNodeHopLatencyUs` (renamed from `InterNodeLatencyUs`, unit per-collective → per-hop). The flat #1667 term was node-span-*invariant* and under-charged a wide span by ~`(nodes−1)×`; this makes it span-aware without touching the existing per-layer collective-count multipliers.
- **Part B — `S`:** the eager/no-overlap serialization multiplier, a deployment-regime input on `ModelHardwareConfig.CommSerializationFactor` (`--comm-serialization-factor`), kept strictly separate from `α_hop`. `--enforce-eager` declares the regime and requires an explicit `S > 1` (BLIS ships no fitted eager magnitude).

**Both ship inert** (`α_hop = 0`, `S = 1`) ⇒ byte-identical output (INV-6/INV-BC-DP1), verified against `main`.

### Anti-overfitting guardrails honored
- `n_steps` carries zero free parameters (derived from algorithm + topology).
- `α_hop` ships at 0; `S` ships at 1. No value chosen to fit run 438866 lands here.
- The pin test `TestCommittedHardwareConfig_LatencyIsUncalibrated` is retargeted to `InterNodeHopLatencyUs` and still asserts 0.
- Hop-count helper is unit-tested at ≥3-node spans, so `2(n−1)` vs `(n−1)` are distinguishable, with no reference to any run.

> **Scope note (from the issue):** this term is inert on the DP/EP replay path (no `node_pools` ⇒ no `NetworkTopology`), so it does **not** close the `experiments/` 438866 TPOT/E2E gap on its own — that is blocked on separate DP/EP topology-detection work. Both parts are exercised today on the multi-node **TP** path (#1529).

## Behavioral contracts

- **BC-1 (rename inert):** no committed config declares `InterNodeHopLatencyUs`, so `α_hop = 0` ⇒ byte-identical (INV-6).
- **BC-2 (ring hop count):** each TP all-reduce comm unit is charged `2·(nodes−1)·α_hop·S`.
- **BC-3 (MoE hop count is family-INDEPENDENT):** MoE dispatch/combine is charged `(nodes−1)·α_hop·S` per collective for BOTH comm families, i.e. `2·(nodes−1)·α_hop·S` per MoE layer once the existing `moeDispatchCollectivesPerLayer` = 2 factor applies. Dispatch (all-gather) and combine (reduce-scatter) are each ONE single-phase collective, so charging the all-gather family a whole ring would double-count to `4·(nodes−1)`/layer: the ring's two phases ARE the two collectives that factor already counts. The family split is real for the bandwidth VOLUME (`spanScalesFor`) but spurious for the hop count. Collectives-per-layer factor unchanged. The TP leg keeps the whole-ring `2·(nodes−1)` of BC-2, because one TP comm unit IS one all-reduce. *(Corrected during review: this contract originally said the all-gather family was charged the ring form, which 2× over-charged the DEFAULT MoE backend. Fixed in 57174a08, test-locked by absolute per-leg hop anchors in 983ada24.)*
- **BC-4 (monotone in span):** more nodes ⇒ never-lower cross-node latency (4-node span = 3× a 2-node span in the latency term).
- **BC-5 (S default inert):** `S = 1` (and `S ≤ 1` clamped) ⇒ output equals Part A alone; `α_hop = 0` ⇒ `S` charges nothing.
- **BC-6 (monotone in S):** raising `S` never lowers the charged cross-node latency.
- **BC-7 (`--enforce-eager` guard):** `--enforce-eager` without an explicit `--comm-serialization-factor > 1`, or a factor `< 1`, is fatal at the CLI (R3/R6).
- **BC-8 (run/replay parity):** `S` re-supplied on both CLIs, not header-round-tripped; `α_hop` under the existing `MaxNodesSpanned` replay-rejection guard (INV-13).

## Testing

- `go build ./...` — clean
- `go test ./... -count=1` — all packages pass
- **INV-6 verified empirically:** default trained-physics run is byte-identical between `main` and this branch; `--comm-serialization-factor 1` is byte-identical to no flag.
- **CLI guards verified:** `--enforce-eager` without a factor → exit 1; `--comm-serialization-factor 0.5` → exit 1; `--enforce-eager --comm-serialization-factor 30` → exit 0.
- New tests: `TestCrossNodeHopCounts_MatchAlgorithmDerivation`, `TestStepTime_PerHopLatencyScalesWithNodeSpan`, `TestStepTime_CrossNodeLatencyHopConvention` (absolute per-leg hop anchors — mutation-verified to fail if the MoE leg is reverted to `crossNodeRingHops`), `TestStepTime_{SerializationFactorDefaultIsInert,MonotoneInSerializationFactor,SerializationFactorInertWithoutLatency}`, `TestEffectiveCommSerializationFactor`, `TestWithCommSerializationFactor_Option`, `TestLegacyInterNodeLatencyKeyIsRejected` (the pre-#1694 key is rejected at load, not silently zeroed), and the `cmd` suite in `comm_serialization_test.go` (flag parity, `observe` absence, the `--enforce-eager`/factor guards, `TestResolveCommSerializationFactor_WarnsWhenInert`, and byte-identity).

Fixes #1694


susiejojo:
@claude /blis-pr-review. Check for code correctness and quality, and take into account the Archon architecture changes through /archon-pr-review

susiejojo:
Thanks for the thorough review. Addressed in 5f64e833:

**Q6 (Minor, required) — `moeHops` all-to-all branch uncovered.** Added `TestStepTime_MoEDispatchLatencyHopCountByBackend`: isolates the latency half (`fabricHW(1)`, `InterNodeHopLatencyUs=20`) on a 4-node MoE span and asserts the ring (all-gather) latency penalty is ~2× the all-to-all one (6 hops vs 3). A silent swap of `crossNodeAll2AllHops`→`crossNodeRingHops` now fails this test, closing the fidelity risk you flagged for when α_hop is calibrated (#1661).

**Q8 (informational) — no observe-absence guard.** Added `TestCommSerializationFlags_AbsentOnObserve`, asserting neither flag is registered on `observe` (no step time ⇒ dead knob) — the S twin of `TestNetworkTopology_HasNoCLIFlag`'s observe row, matching the `--kv-cache-dtype` / `--kv-offload-config` boundary.

**Q5 (informational) — replay comment understated the fence.** The `cmd/replay.go` comment now states the stronger fact: the S term is *structurally unreachable* on replay (spanning traces are fenced via `node_pools` + `max_nodes_spanned > 1` rejection), so the flag is surface parity and re-supplying S keeps the non-spanning path byte-identical.

**CLAUDE.md action item.** No change needed — the transient working-tree revert you saw during the review is not in the branch. The pushed `CLAUDE.md` (commit db0c82f3) carries the #1694 changes (`InterNodeHopLatencyUs` paragraph + `--enforce-eager` example); `git status` on the branch is clean.

**latency-models.md:346 nit.** Left as-is: that inline `units · n_steps · α_hop` restatement is describing the per-hop *charge structure*, with the full `· S` form given as the calibration frame 11 lines later — the omission is intentional scoping, not an inaccuracy. Happy to add `· S` there if you'd prefer full consistency.

`go build`, `go test ./...`, and `golangci-lint run ./...` all clean.

susiejojo:
@claude /blis-pr-review

sim/latency/trained_physics_model.go
```diff
@@ -1135,6 +1188,19 @@ func NewTrainedPhysicsModel(coeffs sim.LatencyCoeffs, hw sim.ModelHardwareConfig
...
 
+	// Cross-node latency (#1694): each leg's fixed per-comm-unit cost is
+	// n_steps·α_hop·S. The TP-group collectives are hierarchical rings; the MoE
+	// dispatch/combine follows the SAME algorithm split as its bandwidth half
+	// (spanScalesFor) — a ring for the all-gather family (whose volume is ring-shaped),
+	// a true all-to-all for the modular backends. S (the eager/no-overlap serialization
+	// multiplier) is a global per-run input; it is applied identically to both legs and
+	// is exactly 1.0 (inert) unless calibrated (INV-6).
+	commSerialization := hw.EffectiveCommSerializationFactor()
+	moeHops := crossNodeRingHops
```
jgchn: **Correctness (blocking): this charges the all-gather family 2x the analytic hop count, and that family is the default backend.**

`crossNodeRingHops` is documented at `:477-487` as the hop count of ONE hierarchical ring **all-reduce**, i.e. `2*(nodes-1)` because it reduce-scatters *then* all-gathers. But the all-gather family's two collectives are each **one** of those phases, per this file's own `moeDispatchBasis` doc at `:899-900`: dispatch all-gathers, combine reduce-scatters. Then `moeCrossNodeLatency` (`:445-450`) multiplies by `moeDispatchCollectivesPerLayer = 2`, and `StepTime` multiplies by `numMoELayers` (`:828`).

So the charge becomes `2 collectives * 2*(nodes-1) = 4*(nodes-1)` hops per MoE layer, where the analytic count is `2*(nodes-1)`.

Measured with alpha_hop = 1 us so the returned value *is* the hop count, 16-GPU group over 4 nodes:

```
allgather_reducescatter (DEFAULT)  hops/collective=6 x collectives/layer=2 => charged 12 hops/layer
deepep/all-to-all                  hops/collective=3 x collectives/layer=2 => charged  6 hops/layer

analytic: all-gather(n-1)=3 + reduce-scatter(n-1)=3  =>  6 hops/layer
analytic: a2a dispatch(n-1)=3 + a2a combine(n-1)=3   =>  6 hops/layer
TP all-reduce ring: crossNodeRingHops(4)=6 per collective   (correct: rs+ag)
```

The correct conclusion is that **both** families charge `2*(nodes-1)` hops per MoE layer, because both are two single-phase collectives. The family split is real for the **bandwidth** half (`spanScalesFor`, where the volume shapes genuinely differ) but spurious for the hop count.

This matters because `allgather_reducescatter` is `DefaultMoECommBackend` (`sim/latency/moe_comm_backend.go:35`), so it is the path a user gets without asking. Magnitude on the shape that motivated the issue (92 MoE layers, 4-node span, alpha_hop ~542 us): ~598 ms/step charged vs ~299 ms analytic.

This is precisely the double-use the issue flagged ("the collective-count-per-layer factor (2) and the ring's `2*(nodes-1)` hops-per-collective factor are easy to double-use... Pin **one** convention"), and it is baked into **BC-3** as written, so the contract needs correcting too, not just the code.

Suggested shape: keep `crossNodeRingHops` (`2*(nodes-1)`) exclusively for one whole all-reduce, which is what the TP legs at `:1239` correctly consume, and give the all-gather MoE family a single-phase count. That makes the two MoE families equal in the latency term, which is the physically defensible outcome.

sim/latency/network_cost_test.go
```diff
@@ -443,6 +510,43 @@ func TestStepTime_MoEDispatchLegChargedCrossNode(t *testing.T) {
...
+	a2aPenalty := penalty("deepep_high_throughput")   // true all-to-all: (nodes−1) hops
+
+	assert.Greater(t, a2aPenalty, int64(0), "precondition: the all-to-all backend must pay a latency penalty")
+	assert.Greater(t, ringPenalty, a2aPenalty,
+		"the ring (all-gather) latency penalty must exceed the all-to-all one (2·(n−1) vs (n−1) hops); "+
+			"equality means the moeHops selection ignored the comm family (ring=%d µs, a2a=%d µs)",
+		ringPenalty, a2aPenalty)
+	// Ring 2·(4−1)=6 hops vs all-to-all (4−1)=3 hops ⇒ the ring penalty is ~2× the all-to-all.
+	ratio := float64(ringPenalty) / float64(a2aPenalty)
+	assert.InDelta(t, 2.0, ratio, 0.15,
```
jgchn: This test is aimed at exactly the right seam and is well constructed (>=3-node span so the factor of 2 is unambiguous, `fabricHW(1)` to isolate the latency half, contained baseline to cancel), but the expected value currently **pins the defect** described on `trained_physics_model.go:1199`.

Both the `assert.Greater` at `:540` and this `InDelta(2.0, ...)` encode "ring family costs 2x the all-to-all family in the latency term". If the hop count is derived per collective rather than per all-reduce, the two families are **equal** here (`2*(nodes-1)` hops per MoE layer either way) and the difference lives entirely in the bandwidth half.

Worth keeping the test and inverting the assertion: assert the two families' latency penalties are equal at a >=3-node span, and move the family-discrimination assertion onto the bandwidth half where `spanScalesFor` genuinely differentiates them. That would catch a future regression in either direction.

sim/model_hardware_config.go
```diff
@@ -189,31 +189,45 @@ type HardwareCalib struct {
...
 	// Calibration frame: like the bandwidth half, this rides the learned communication
 	// coefficient (β₄ for TP collectives, β_EP for MoE dispatch), so the charge is
-	// β·units·InterNodeLatencyUs. Calibrate it in that frame, not as a raw wall-clock
-	// number.
-	InterNodeLatencyUs float64 `json:"InterNodeLatencyUs"`
+	// β·units·n_steps·α_hop·S. Calibrate α_hop in that frame (per-fabric, S=1), not as a
+	// raw wall-clock number. S (the eager/no-overlap serialization multiplier) is a
+	// SEPARATE deployment-regime input on ModelHardwareConfig.CommSerializationFactor,
+	// never folded into this fabric constant.
+	InterNodeHopLatencyUs float64 `json:"InterNodeHopLatencyUs"`
```
jgchn, line 230: **Correctness (blocking): the rename silently discards an operator's calibrated value, rather than forcing recalibration.**

`parseHWConfig` uses plain `json.Unmarshal` with no `DisallowUnknownFields` (`sim/latency/config.go:170-181`). So a `--hardware-config` carrying the pre-#1694 `InterNodeLatencyUs` key, which #1667 and the previous `docs/reference/configuration.md` both told operators to "supply a measured value here" for, parses without error and yields 0:

```
parse error            = <nil>
InterNodeHopLatencyUs  = 0
EffectiveInterNodeHop  = 0
HasInterconnectCalib   = true      <- bandwidths present, so the warn path stays silent
```

(reproduced via `GetHWConfig` on a file with `IntraNodeBwGBps`/`InterNodeBwGBps` plus the legacy `"InterNodeLatencyUs": 542`, which is exactly the three-key shape the docs' example shows.)

Because `HasInterconnectCalibration()` is already satisfied by the bandwidths, `warnIfCrossNodeUnpriced` (`sim/cluster/network_topology.go:105-115`) never fires, so the whole size-independent cross-node term disappears with **no diagnostic at all**. That is the R1 "never silent" case this feature's own diagnostics exist to prevent, and it inverts the issue's stated rationale for renaming ("Renaming forces recalibration rather than a silent rescale"): with a permissive decoder it forces neither.

The committed `hardware_config.json` is genuinely unaffected, so this is purely about operator-supplied files. A cheap fix is a legacy-key check at the load boundary in `parseHWConfig`: decode into a `map[string]json.RawMessage` first, or add an explicit `InterNodeLatencyUs *float64` shadow field, and reject or warn naming the new key and the unit change (per-collective to per-hop, so the value needs dividing by the hop count, not copying).

cmd/root.go
```diff
@@ -1718,6 +1733,36 @@ func resolveSpeculativeConfig(cmd *cobra.Command) sim.SpeculativeConfig {
...
+	// fitted eager S to supply, and inventing one would put a fabricated constant in front
+	// of every eager multi-node estimate (#1694 guardrail #2). Require the operator to
+	// supply the calibrated factor explicitly, mirroring the --speculative-acceptance-rate
+	// Changed-gated required-companion idiom.
+	if enforceEager && (!cmd.Flags().Changed("comm-serialization-factor") || s <= 1.0) {
+		logrus.Fatalf("--enforce-eager requires an explicit --comm-serialization-factor > 1: BLIS ships no measured " +
+			"eager serialization magnitude, so the multiplier must be supplied from a calibrated source (#1694). " +
+			"Set it explicitly, e.g. --enforce-eager --comm-serialization-factor 30")
+	}
+	if s > 1.0 {
```
jgchn, line 1768: **`S > 1` is accepted, mandated by `--enforce-eager`, and then silently does nothing on every out-of-the-box config.**

```
$ blis run --model qwen/qwen3-14b --num-requests 20 --enforce-eager --comm-serialization-factor 30
exit=0
stderr: no mention of serialization, eager, or cross-node
stdout: byte-identical to the same run with no flags
```

Two causes:

1. This is `Infof`, but the default log level is `warn` (`cmd/root.go:1464`), so the one record of `S` is suppressed unless the user thinks to pass `--log info`.
2. Nothing checks whether the term can fire at all. It needs trained-physics **and** `node_pools` **and** `alpha_hop > 0`, and `alpha_hop` ships at 0 in every committed entry, so out of the box the answer is always no.

The codebase's precedent for the mirror case is a latched `logrus.Warnf` (`sim/cluster/network_topology.go:88-115`). A `Warnf` here when `s > 1` and the selected latency backend is not trained-physics, or when no `--policy-config` node pools are present, would close the R1 gap cheaply. The placement-dependent part could alternatively ride the existing `warnIfCrossNodeUnpriced` switch.

This also bears on the issue's Part-B acceptance #3, "`S` is surfaced through a regime flag/field **and recorded in provenance**". Today `S` appears in no persisted artifact: no trace-header field, no `MetricsOutput` field, and its only stderr line is off by default. Given the PR deliberately does not round-trip it through the header (which is sound, since replay rejects spanning traces anyway), some other record would be needed to satisfy that criterion.

cmd/root.go
```diff
@@ -1589,6 +1598,12 @@ func registerSimConfigFlags(cmd *cobra.Command) {
...
 
+	// Cross-node collective serialization S (#1694, Part B). trained-physics only; fires
+	// only for a multi-node span with a calibrated α_hop (InterNodeHopLatencyUs). Default
+	// 1.0 is inert (byte-identical, INV-6). Re-supply identically on replay (INV-13).
+	cmd.Flags().Float64Var(&commSerializationFactor, "comm-serialization-factor", 1.0, "Cross-node collective serialization multiplier S on the size-independent inter-node latency term (#1694). 1.0 (default) = CUDA graphs / comm-compute overlap (inert). >1 = enforce-eager / no-overlap regime, where collectives serialize behind a full barrier. Must be >= 1. Multiplies an existing cross-node latency term (spanning nodes AND a calibrated per-hop α_hop); never creates cost on its own.")
+	cmd.Flags().BoolVar(&enforceEager, "enforce-eager", false, "Mirror vLLM --enforce-eager for latency purposes: declares the deployment runs without CUDA graphs / comm-compute overlap. Provenance + guard for --comm-serialization-factor: when set, requires an explicit --comm-serialization-factor > 1 (BLIS ships no fitted eager magnitude — #1694). Does not itself pick a value.")
```
jgchn: Minor, but the help text over-promises: "Mirror vLLM `--enforce-eager` for latency purposes" reads as though the flag models eager execution, whereas it only gates `--comm-serialization-factor`.

vLLM's `enforce_eager` removes CUDA-graph capture from **every** kernel launch, which inflates per-step overhead on single-node deployments too. Here the flag is a pure guard: on any config without `node_pools` and a calibrated `alpha_hop` it changes nothing at all (verified: byte-identical stdout with `--enforce-eager --comm-serialization-factor 30`).

Suggest narrowing the wording to say it declares the regime for the **cross-node collective term only**, and that the broader eager kernel-launch overhead is not modeled, so a reader does not conclude BLIS now models `enforce_eager` generally.

cmd/root.go
```diff
@@ -2495,6 +2540,11 @@ var runCmd = &cobra.Command{
 			}
 		}
 
+		// Cross-node collective serialization S (#1694, Part B): a model-level latency
+		// input appended to the EP-group options, resolved once and threaded identically on
+		// replay (INV-13). Inert at the default S=1.
+		mhwOpts := append(dpPlan.EPGroupOptions(), sim.WithCommSerializationFactor(resolveCommSerializationFactor(cmd)))
```
jgchn: Style/R23 nit, not a behavior bug. `dpPlacementPlan.EPGroupOptions()` exists specifically so that "`blis run` and `blis replay` cannot disagree about it (R23, INV-13)" (`cmd/root.go:419-425`). Hand-assembling the combined option list at both call sites re-opens the seam that helper was created to close: the next option added has to be remembered in two places.

No aliasing hazard, for the record: `EPGroupOptions()` returns either `nil` or a fresh cap-1 slice literal (`:427-431`), so the `append` always reallocates.

Consider moving the composition onto the plan (or a small shared `modelHardwareOptions(cmd, dpPlan)` helper called from both) so the two paths keep sharing one construction site.

cmd/replay.go
```diff
@@ -654,6 +654,15 @@ Example:
...
 		startTime := time.Now()
 
+		// Cross-node collective serialization S (#1694, Part B): resolved and threaded
+		// identically to the run path so a run and its replay under the same flags stay
+		// byte-identical (INV-13). Not round-tripped through the trace header. Note the S
+		// term is STRUCTURALLY UNREACHABLE on replay today: it only fires for a multi-node
+		// span, and replay logrus.Fatalf's on node_pools and on any trace with
+		// max_nodes_spanned > 1 (#1530) — so registering the flag here is flag-surface
+		// parity, and re-supplying S keeps the non-spanning path byte-identical regardless.
+		mhwOpts := append(dpPlan.EPGroupOptions(), sim.WithCommSerializationFactor(resolveCommSerializationFactor(cmd)))
```
jgchn: Mirror of the note on `cmd/root.go:2546`: the two call sites now each hand-assemble the option list. The comment above this line is genuinely useful, though, and the reasoning is right that `S` is structurally unreachable on replay, so re-supplying it keeps the non-spanning path byte-identical either way.

jgchn (changes requested):
## Correctness review (#1695)

Verified on `pr1695` at 5f64e83: `go build ./...` clean, `go test ./sim/... ./cmd/...` all pass, and all four CLI guards fatal correctly (bare `--enforce-eager`, `S=0.5`, `S=NaN`, `--enforce-eager --comm-serialization-factor 1.0`). INV-6 reproduced: default run is byte-identical to `main`. The findings below are physics/plumbing issues the green suite does not catch.

### Blocking

1. **The MoE cross-node hop count is 2x over-charged on vLLM's default backend** (`sim/latency/trained_physics_model.go:1199`). `crossNodeRingHops` is `2*(nodes-1)`, the count for one *whole* ring all-reduce (reduce-scatter + all-gather). The all-gather family's two collectives are each one of those phases, and `moeCrossNodeLatency` then multiplies by `moeDispatchCollectivesPerLayer = 2`. See the inline comment for the measured proof (12 charged hops/layer vs 6 analytic). This also means **BC-3** as written is physically wrong, and the new `TestStepTime_MoEDispatchLatencyHopCountByBackend` pins the defect.

2. **The `InterNodeLatencyUs` -> `InterNodeHopLatencyUs` rename silently discards an operator's calibrated value** (`sim/model_hardware_config.go:230`). `parseHWConfig` does not set `DisallowUnknownFields`, so the legacy key parses cleanly and drops to 0, and `warnIfCrossNodeUnpriced` stays quiet when the bandwidths are declared alongside it. Proof inline. The issue's rationale was that renaming "forces recalibration rather than a silent rescale"; with a permissive decoder it forces neither.

### Non-blocking but worth fixing in the same pass

3. `S > 1` is accepted, mandated by `--enforce-eager`, and then silently does nothing on every out-of-the-box config (`cmd/root.go:1758`). Also bears on the issue's Part-B acceptance #3, "recorded in provenance".
4. `--enforce-eager`'s help text over-promises vLLM parity (`cmd/root.go:1605`).
5. The option list is now hand-assembled at both call sites, re-opening the seam `EPGroupOptions()` exists to close (`cmd/root.go:2546`, `cmd/replay.go:664`).

### Verified good

- Hop-count helpers carry zero free parameters and are tested at n = 3, 4, 8 (guardrail #1). The pin test is correctly retargeted to `EffectiveInterNodeHopLatencyUs` and still asserts 0.
- The TP ring leg (`:1239`) is right: one comm unit is one full all-reduce, so `2*(nodes-1)` is the correct count there. BC-4's 3x (4-node vs 2-node) checks out.
- Flags on `run` + `replay`, absent on `observe`, is the correct parity boundary. Only two `NewModelHardwareConfig` sites and both are updated. `S` survives the cluster's post-construction topology stamping.
- `EffectiveCommSerializationFactor`'s NaN/Inf handling is correct (NaN fails the `<= 1.0` compare and is caught by the explicit `IsNaN`).
- Docs across `latency-models.md`, `configuration.md`, `cluster.md` and `CLAUDE.md` are accurate and thorough, except that they inherit finding 1's family split and carry no migration note for finding 2.

Happy to be wrong on finding 1 if there is a reading of the all-gather dispatch under which it really is two full all-reduces per layer, but I could not construct one from the code's own definitions.

susiejojo:
Thanks — this is a genuinely sharp review, and both blocking findings were correct. Fixed in 57174a08.

**Blocking 1 — MoE hop count 2× over-charged on the default backend.** You're right, and your measured proof (12 charged hops/layer vs 6 analytic for `allgather_reducescatter`) was exactly the bug. The MoE dispatch/combine leg now uses `crossNodeAll2AllHops` (`(nodes−1)`, one single-phase collective) for **both** families, since `moeCrossNodeLatency` already multiplies by `moeDispatchCollectivesPerLayer = 2` → `2·(nodes−1)`/layer. The family split is real for the bandwidth VOLUME (`spanScalesFor`) but spurious for the hop count, precisely as you diagnosed. The TP leg keeps `crossNodeRingHops` because one TP comm unit *is* one whole all-reduce. Verified empirically: both families now charge 6 hops/layer (was 12 for all-gather).

I took your suggested test shape: `TestStepTime_MoEDispatchLatencyHopCountByBackend` is renamed to `..._IsFamilyIndependent` and now asserts the two families' latency penalties are **equal** at a 4-node span (it would fail if the all-gather family were charged the full ring). Family discrimination stays on the bandwidth half. BC-3 in the PR description is corrected accordingly.

**Blocking 2 — rename silently discarded a calibrated value.** Correct — `parseHWConfig` is permissive, so a legacy `InterNodeLatencyUs` key dropped to 0 with no diagnostic (the bandwidths alone satisfy `HasInterconnectCalibration`, so `warnIfCrossNodeUnpriced` stays quiet). Added `rejectLegacyInterNodeLatencyKey` at the load boundary: a fatal, per-GPU error naming the new key and stating it's a *recalibration* (unit changed per-collective → per-hop, so divide by the hop count, don't copy). `TestLegacyInterNodeLatencyKeyIsRejected` covers rejection + the new-key control. The committed config is unaffected (verified).

**#3 — `S > 1` silently does nothing.** Added a `Warnf` in `resolveCommSerializationFactor` for the two cheaply-knowable inert cases (non-trained-physics backend; no `--policy-config` node pools), so a requested penalty never vanishes silently. The placement-time α_hop>0 condition remains covered by `warnIfCrossNodeUnpriced`. `TestResolveCommSerializationFactor_WarnsWhenInert` covers all three cases (incl. the live shape that must NOT warn). Left the Infof as-is; the Warnf is the load-bearing diagnostic. On the provenance point (Part-B acceptance #3): I agree S is in no persisted artifact today; since the header round-trip is deliberately omitted (replay rejects spanning traces), a `MetricsOutput` field would be the right home — I'd prefer to do that as a focused follow-up rather than widen this PR, unless you'd like it here.

**#4 — help text over-promises.** Narrowed: the flag now states it declares the regime for the **cross-node collective term only** and explicitly does not model vLLM `enforce_eager`'s general kernel-launch overhead.

**#5 — R23 seam re-opened.** Hoisted the option composition into a shared `modelHardwareOptions(cmd, dpPlan)` helper called from both `run` and `replay`, so the next option is added once. (Thanks for confirming the `append` had no aliasing hazard — agreed.)

Docs (`latency-models.md`, `configuration.md`, `CLAUDE.md`) updated for the family-independent hop count and a migration note for the rejected legacy key. `go build`, `go test ./...`, `golangci-lint` all clean; INV-6 re-verified byte-identical to `main`.

susiejojo:
@claude /blis-pr-review

sim/latency/network_cost_test.go
```diff
@@ -443,6 +510,46 @@ func TestStepTime_MoEDispatchLegChargedCrossNode(t *testing.T) {
...
+			newNetModel(t, mc, hw, 4, 4, true, backend, 16).StepTime(batch)
+	}
+	ringPenalty := penalty("allgather_reducescatter") // all-gather family
+	a2aPenalty := penalty("deepep_high_throughput")   // modular all-to-all family
+
+	assert.Greater(t, a2aPenalty, int64(0), "precondition: the all-to-all backend must pay a latency penalty")
+	assert.Greater(t, ringPenalty, int64(0), "precondition: the all-gather backend must pay a latency penalty")
+	// Both families charge (nodes−1) hops per collective ⇒ the latency penalties are EQUAL.
+	// (A regression that charged the all-gather family the full ring would make ring ≈ 2×a2a.)
+	assert.Equal(t, ringPenalty, a2aPenalty,
```
jgchn, line 566: **This assertion no longer detects the bug it replaced.** Re-opening finding 1's test half.

The fix correctly deleted the `commFamily` branch, so the MoE leg now passes `crossNodeAll2AllHops` **unconditionally**. An equality-between-families assertion is therefore invariant to any *uniform* change in the per-collective hop count: double both and it still passes.

Mutation proof. Reverting exactly the bug this PR fixed (MoE leg back to `crossNodeRingHops`, restoring 12 charged hops/layer instead of 6):

```
=== FULL SUITE UNDER MUTATION (MoE leg = ring hops) ===
ok  sim            ok  sim/cluster      ok  sim/internal/hash   ok  sim/internal/kvkey
ok  sim/internal/util  ok  sim/kv       ok  sim/kvtransfer      ok  sim/latency
ok  sim/lora       ok  sim/saturation   ok  sim/trace           ok  sim/workload
ok  cmd
```

All 13 packages green with the 2x over-charge reinstated, including this test.

The comment on `:546` says "A regression that charged the all-gather family the full ring would make ring ~= 2x a2a". That was true only while the family branch existed. With the branch gone, the realistic regression (someone edits `:1261` to pass `crossNodeRingHops`, or swaps the two helpers' bodies) moves both families together and this test is blind to it. Worth noting the *pre-fix* test did discriminate: it asserted 2.0, and the corrected code makes the ratio 1.0, so it would have failed. The replacement lost that property.

What is needed is an **absolute anchor**, not a cross-family comparison. This test is in-package, so asserting the constructed model's `moeCrossNodeLatencyUs` field equals `float64(crossNodeAll2AllHops(nodes)) * alphaHop * s` is a value assertion, not a structural one, and it pins the convention directly. Something like:

```go
// The MoE leg's per-comm-unit cost is ONE single-phase collective: (nodes-1) hops.
// Pinned absolutely, because a cross-family comparison cannot see a uniform change.
m := newNetModel(t, mc, hw, 4, 4, true, "allgather_reducescatter", 4).(*TrainedPhysicsModel)
assert.Equal(t, float64(crossNodeAll2AllHops(4))*hw.InterNodeHopLatencyUs, m.moeCrossNodeLatencyUs,
    "MoE dispatch/combine is per-collective (nodes-1); the full ring 2*(nodes-1) here would "+
        "double-count, since moeDispatchCollectivesPerLayer already counts both collectives")
assert.Equal(t, float64(crossNodeRingHops(4))*hw.InterNodeHopLatencyUs, m.tpCrossNodeLatencyUs,
    "the TP comm unit IS one whole all-reduce, so it keeps the full-ring count")
```

Asserting both legs in one place is the useful part: it pins the asymmetry that the whole finding was about, and it fails under the mutation above.

I would keep the existing equality assertion alongside it. It still earns its place as a guard against reintroducing a family branch on the latency half, which is a different regression. Just not as the only guard.

cmd/root.go
```diff
@@ -1718,6 +1743,62 @@ func resolveSpeculativeConfig(cmd *cobra.Command) sim.SpeculativeConfig {
...
+		}
+	}
+	return s
+}
+
+// backendDisplayNameForWarn renders the latency backend for a diagnostic, spelling out the
+// empty-string default (which resolves to roofline) rather than printing "".
+func backendDisplayNameForWarn(backend string) string {
+	if backend == "" {
+		return sim.LatencyBackendRoofline + " (default)"
```
jgchn: Minor, introduced by this commit: this fallback is unreachable and its string is wrong.

`--latency-model` is registered with default `"trained-physics"` (`cmd/root.go:1505`), and `registerSimConfigFlags` runs for both `run` and `replay`, so `latencyModelBackend` is never `""` in production. If it somehow were, this would report `roofline (default)`, which contradicts the actual default.

For the record I did check whether the raw flag var can diverge from the resolved backend, since that would have been a real bug in the new warning: the only override is at `:764-766`, and it sets `trained-physics` when the flag was *unchanged*, which is already the default value. So reading `latencyModelBackend` at `:1779` is correct, and the warning fires on exactly the right condition.

Suggest dropping the helper and printing `latencyModelBackend` directly, or if you want the defensive branch, having it say `trained-physics (default)`.

jgchn (changes requested):
## Re-review of 57174a08

Thanks for the fast turnaround. I re-verified all seven comments against the new commit. **Six are genuinely fixed. One is not, and it is the test that is supposed to lock the blocking finding.**

### Fixed, verified

| # | Evidence |
|---|---|
| 1 | `MoE nodes=4: hops/collective=3 x collectives/layer=2 => 6 hops/layer` (analytic 6). TP leg correctly keeps `crossNodeRingHops` at 6 per all-reduce unit. The constructor comment at `:1198-1220` states the convention accurately, including why the ring's two phases ARE the two collectives. |
| 2 | Load-time error, per-GPU, `sort.Strings`-ordered: `hardware config declares the removed per-collective key "InterNodeLatencyUs" on GPU(s) [A100 H100] ... This is a RECALIBRATION, not a rename: divide the old value by the collective's cross-node hop count`. New key still parses (alpha_hop = 5). Putting the guard in `parseHWConfig` rather than on the struct was the better call. |
| 3 | Real run now warns: `--comm-serialization-factor 30.000 will have NO effect: ... without --policy-config there are no node_pools`. Roofline case warns on the backend instead. The table test covers all three cases including the must-NOT-warn live shape, which is the half that usually gets skipped. |
| 4 | Help text narrowed, and explicitly says the broader eager kernel-launch overhead is not modeled. |
| 5 | `modelHardwareOptions(cmd, dpPlan)` at `cmd/root.go:439`, consumed by both `root.go:2579` and `replay.go:662`. |

Also re-verified: `go build ./...` clean, `golangci-lint run ./...` 0 issues, full suite green, default run byte-identical to `main`. (INV-6 is safe by construction here regardless, since alpha_hop = 0 zeroes the term whatever the hop count is.)

### Still open

**The replacement test for finding 1 does not detect finding 1.** Details inline on `sim/latency/network_cost_test.go:547`, with a mutation proof: reverting the MoE leg to `crossNodeRingHops` leaves the **entire** suite green across all 13 packages. The code fix is right; nothing in the repo holds it in place.

Plus one small thing the fix introduced: a dead-and-wrong fallback in `backendDisplayNameForWarn` (`cmd/root.go:1797`).

Nothing else from the original review outstanding.

susiejojo:
Both open items fixed in 983ada24. Thank you for the mutation proof — you were exactly right that the equality assertion couldn't see a uniform change.

**Finding 1 test — replaced with absolute anchors.** `TestStepTime_MoEDispatchLatencyHopCountIsFamilyIndependent` is now `TestStepTime_CrossNodeLatencyHopConvention`, asserting each frozen field against its analytic value directly (your suggested shape):

```go
assert.Equal(t, float64(crossNodeAll2AllHops(moeNodes))*hw.InterNodeHopLatencyUs, m.moeCrossNodeLatencyUs, ...)
assert.Equal(t, float64(crossNodeRingHops(tpNodes))*hw.InterNodeHopLatencyUs,   m.tpCrossNodeLatencyUs,  ...)
```

One tweak from your snippet: I used `gpusPerNode=2` (TP=4, DP=4) so the two legs span **different** node counts — TP group 4 → 2 nodes, MoE group 16 → 8 nodes — so each anchor exercises its helper on its own group rather than a shared 4-node span (and the MoE span of 8 is well clear of the 2-node coincidence). I kept the family-equality assertion alongside as the secondary guard against reintroducing a family branch, per your note.

Verified against your mutation: reverting the MoE leg to `crossNodeRingHops` now makes this test **FAIL** (`expected 3·20=60, actual 6·20=120` on the MoE anchor), and it passes on the correct code. The regression is held in place.

**`backendDisplayNameForWarn` — dropped.** You're right it was unreachable and its string wrong. Removed the helper; the warning now prints `latencyModelBackend` directly, with a comment recording your finding that reading the raw var is correct (the only override sets trained-physics only when the flag was unchanged, i.e. already the default). Thanks for chasing down the flag-var-vs-resolved-backend divergence question — good to have that confirmed.

`go build`, full `go test ./...`, `golangci-lint` all clean; default run still byte-identical to `main`.

jgchn:
### One stale item in the PR description: BC-3 still states the defect as the contract

Everything in the code is verified good on `983ada24` (details in my re-review below). The only thing left is the PR **description**, which I did not want to edit on your behalf.

BC-3 currently reads:

> - **BC-3 (all-to-all hop count):** MoE dispatch/combine on an all-to-all backend charged `(nodes−1)·α_hop·S`; all-gather family charged the ring form. Collectives-per-layer factor unchanged.

The "all-gather family charged the ring form" clause is exactly the 2× over-charge that 57174a08 fixed. Since the PR body is the contract-verification source of truth for this workflow, merging as-is would put that into the permanent record as an intended contract. (Your reply on the first round said BC-3 was corrected, but the edit does not seem to have landed.)

Suggested replacement, drop-in for line 30:

```markdown
- **BC-3 (MoE hop count is family-INDEPENDENT):** MoE dispatch/combine is charged `(nodes−1)·α_hop·S` per collective for BOTH comm families, i.e. `2·(nodes−1)·α_hop·S` per MoE layer once the existing `moeDispatchCollectivesPerLayer` = 2 factor applies. Dispatch (all-gather) and combine (reduce-scatter) are each ONE single-phase collective, so charging the all-gather family a whole ring would double-count to `4·(nodes−1)`/layer: the ring's two phases ARE the two collectives that factor already counts. The family split is real for the bandwidth VOLUME (`spanScalesFor`) but spurious for the hop count. Collectives-per-layer factor unchanged. The TP leg keeps the whole-ring `2·(nodes−1)` of BC-2, because one TP comm unit IS one all-reduce. *(Corrected during review: this contract originally said the all-gather family was charged the ring form, which 2× over-charged the DEFAULT MoE backend. Fixed in 57174a08, test-locked by absolute per-leg hop anchors in 983ada24.)*
```

**Also stale, same section:** the Testing bullet predates the last two commits. Missing `TestStepTime_CrossNodeLatencyHopConvention` (which replaced `TestStepTime_MoEDispatchLatencyHopCountIsFamilyIndependent`), `TestLegacyInterNodeLatencyKeyIsRejected`, and `TestResolveCommSerializationFactor_WarnsWhenInert`.

**One heads-up unrelated to you:** the `claude[bot]` Round 3 **APPROVE** is stale. It reviewed 57174a08 and called the family-equality assertion "the right behavioral test", which the mutation proof disproved (reverting the MoE leg to `crossNodeRingHops` left the whole suite green). Its round-1 recommendation to assert the all-to-all penalty is "approximately half the `allgather_reducescatter` one" now contradicts the corrected physics, so it should not be actioned later.

No code changes needed. This is a description edit, so no CI re-run. Happy to approve once BC-3 is updated.

susiejojo:
Good catch — you're right, my round-1 reply claimed BC-3 was corrected but the edit never landed in the PR body. Fixed now (description-only, no code change):

- **BC-3** replaced with your drop-in (family-INDEPENDENT wording, the double-count explanation, and the "corrected during review" provenance line pointing at 57174a08 + 983ada24). Also fixed the **Part A** summary bullet, which had the same stale "all-to-all `(nodes−1)`; ring for all-gather" framing — it now states the per-collective convention correctly.
- **Testing bullet** refreshed: added `TestStepTime_CrossNodeLatencyHopConvention` (noting it's mutation-verified), `TestLegacyInterNodeLatencyKeyIsRejected`, and `TestResolveCommSerializationFactor_WarnsWhenInert`; dropped the renamed `TestStepTime_MoEDispatchLatencyHopCountIsFamilyIndependent`. Verified every cited test name exists on `983ada24`.

Thanks also for the heads-up on the stale `claude[bot]` Round-3 APPROVE — noted that its "family-equality is the right behavioral test" call was disproved by your mutation proof, and its round-1 "assert the all-to-all penalty is ~half the all-gather one" suggestion contradicts the corrected physics, so neither should be actioned.

jgchn (approved):

PR #1696, opened 2026-09-09
feat(replay): --session-mode fixed-accumulate for faithful high-concurrency agentic replay (#1692)

### LINKED ISSUE #1692: Faithful closed-loop replay of high-concurrency agentic runs: need fixed-arrival replay of accumulate corpora

BLIS cannot faithfully replay a high-concurrency agentic benchmark (e.g. an InferenceX AgentX run) because there is **no replay mode that combines two things that are individually available but mutually exclusive today**:

1. **Accumulate-delta input reconstruction** — real agentic sessions grow their context to ~300k tokens per round; the trace stores per-round `input_tokens` as *deltas* and reconstructs the true growing prefix only via the closed-loop accumulate-buffer path.
2. **A fixed, recorded arrival schedule** — the timing that produces cross-request queueing contention at concurrency > 1.

You can have (1) XOR (2), never both:

- `--session-mode closed-loop` gives correct accumulate inputs, but **regenerates** arrivals as `completion(round N) + think` (INV-10). Because arrivals are chained to the sim's own completion speed, a self-throttling feedback loop forms: requests drain instead of piling up, the queue never reaches its real depth, and TTFT collapses to just compute.
- `--session-mode fixed` replays the recorded `ArrivalUs` (open-loop, breaks the feedback loop) but is **hard-rejected on an accumulate corpus** at [`cmd/replay.go:145-147`](https://github.com/inference-sim/inference-sim/blob/main/cmd/replay.go#L145-L147) — because fixed mode reads `input_tokens` as absolute and would misinterpret the deltas. Forcing it via `--context-growth independent` loses the growing context (inputs become the tiny per-round deltas), so the model processes ~1.9M tokens instead of the real ~26M-equivalent.

Net effect: **high-concurrency runs are unfaithful by construction**, and the failure is silent (plausible-looking but 30–100× off).

**Evidence (Kimi-K3 H200, run 438872, conc=8, TP16/DP2/EP32):**

| Replay mode | TTFT | vs real | E2E | vs real |
|---|---|---|---|---|
| **REAL** (InferenceX) | 68.95 s | 1.00× | 227.4 s | 1.00× |
| closed-loop (accumulate, ~300k in) | 2.22 s | 0.03× | 11.9 s | 0.05× |
| fixed (recorded arrivals, `independent` deltas) | 0.64 s | **0.01×** | 9.2 s | 0.04× |

A full concurrency sweep (conc 1→32) shows TTFT fidelity degrading monotonically: ~1.0× at conc=1 (queue-free, prefill compute-bound — BLIS is accurate here), falling to 0.02–0.05× by conc≥8. The real high-conc TTFT is almost entirely **queue wait** (437 s at conc=12), which no current replay mode can produce.

**Proposed solution**

A replay mode that injects each round at its **recorded `ArrivalUs`** (open-loop, per `--session-mode fixed`) **while still reconstructing accumulate-delta inputs** from the running per-session buffer (as closed-loop does). The recorded arrivals carry the real cross-session overlap, so N large prefills arrive per the real clock and pile into BLIS's scheduler — producing genuine queueing delay instead of a self-throttled trickle.

Sketch (either a new mode value or a flag that lifts the guard):

```bash
./blis replay --model kimi-k3 --hardware H200 --tp 16 --dp 2 --enable-expert-parallel \
  --session-mode fixed-accumulate \
  --trace-header corpus.yaml --trace-data corpus.csv --max-model-len 1000000
```

The arrival timestamps are **already in the corpus** (`weka_convert.go` writes `ArrivalUs = t × 1e6` per round) — this feature consumes data that exists today; it does not need a trace-format change.

- [x] Core simulator (`sim/`) — session arrival scheduling: apply accumulate-buffer input reconstruction on an injection path driven by recorded arrival times rather than SessionManager's completion+think
- [x] Workload generation (`sim/workload/`) — the accumulate reconstruction currently lives on the closed-loop path only
- [ ] KV cache (`sim/kv/`)
- [x] CLI (`cmd/`) — the guard at `cmd/replay.go:145-147` is exactly where the new mode is admitted
- [ ] New package needed

**Extension friction check**
- Files: ~3 (`cmd/replay.go` mode wiring/guard, the injection path that builds requests from the trace, and the workload accumulate-reconstruction helper so it's reachable from a fixed-arrival path). Plus tests.
- Interface: extends the existing session-replay machinery; no new interface. The accumulate-buffer reconstruction already exists — it needs to be reachable from a fixed-arrival injection path instead of only the closed-loop one.
- Invariants: must preserve INV-6 (determinism) and INV-13 (run/replay parity where applicable). It deliberately does **not** enforce INV-10 (session causality: `arrival[N+1] >= completion[N] + think`), since the whole point is to inject at recorded times regardless of sim completion — so INV-10 should be scoped to closed-loop mode, or the new mode documented as exempt. Conservation (INV-1) and clock monotonicity (INV-3) unaffected.

**Second-order caveat (should be stated in the mode's docs, not a blocker):** even with faithful arrivals, BLIS's decode/step model currently runs ~5–10× fast on this workload (unmodeled `--enforce-eager` regime + MTP-speedup-without-contention, #1627), so the queue depth would still be under-predicted until decode is calibrated. This feature is a **necessary precondition** for high-conc fidelity, not sufficient on its own. It should be landed and validated alongside (or ahead of) the decode-side work.

**Alternatives considered**
- *Use `--session-mode fixed` with `--context-growth independent`* — tested; loses the growing context (inputs shrink to per-round deltas), so it trades the timing error for a token-count error. Not faithful.
- *Use `--concurrent-sessions N` (pooled closed-loop)* — maintains a fixed in-flight pool, closer to how the benchmark client drives load, but still regenerates arrivals closed-loop, so it inherits the same self-throttling feedback loop. Useful as a *prediction* tool, not a faithful *replay* of a recorded run.
- *Do nothing / document conc=1 as the only faithful point* — the current state. Acceptable short-term, but leaves the majority of published InferenceX runs (conc 2→70) unreplayable.

- Discovered via the `experiments/calibrate_inferencex.py` harness (posted in discussion #1653) that calibrates BLIS against InferenceX runs end-to-end.
- The accumulate-delta encoding + closed-loop reconstruction is the WekaTrace/OTel corpus path (`sim/workload/weka_convert.go`, closed-loop `SessionManager`).
- Decode-side fidelity gap is #1627 (speculative occupancy contention) plus the unmodeled `enforce-eager` regime — complementary, not a duplicate.

susiejojo:
## Summary

Adds `blis replay --session-mode fixed-accumulate` (#1692): a replay mode that injects each session round at its **recorded arrival time** (open-loop, like `--session-mode fixed`) while **reconstructing the growing accumulate-delta input** (like `--session-mode closed-loop`). It is the only faithful way to replay a high-concurrency agentic corpus (Weka/OTel).

### The problem

An accumulate corpus (`session_context_growth: accumulate`) couples two independent axes that are only individually available today:

- **closed-loop** reconstructs the growing context but *regenerates* arrivals as `completion + think` — a self-throttling feedback loop that drains the queue and collapses TTFT to compute-only (30–100× below real at concurrency ≥ 8).
- **fixed** replays recorded arrivals but is hard-rejected on an accumulate corpus (it reads per-round `input_tokens` as absolutes, not deltas).

The key realization: the growing buffer needs **no live sim output**. The trace's delta chain (`abs_N = abs_{N-1} + out_{N-1} + delta_N`) plus `input_tokens_reset` compaction markers fully determines every round's absolute input from trace data alone. So the loader walks that delta law to pre-bake every round as a request at its recorded arrival — preserving the real cross-session overlap so N large prefills pile into the scheduler at the real clock and produce genuine queue wait.

## Behavioral contracts

- **BC-1** Fixed-arrival injection + accumulate reconstruction: each round injected at recorded arrival; input length = recorded absolute `in_N`, not the delta.
- **BC-2** Growing prefix is content-consistent: round N's input is a strict prefix of round N+1's (prefix-cache fidelity).
- **BC-3** Compaction reset honored: an `input_tokens_reset` round re-seeds the buffer to the recorded absolute.
- **BC-4** Determinism (INV-6): same trace + seed → byte-identical stdout.
- **BC-5** Guards (R1): rejects a non-accumulate trace, `--concurrent-sessions` (open-loop, no pool), and keeps the think-time-requires-closed-loop guards.
- **BC-6** Non-session / single-shot records pass through unchanged.

## Files

- `sim/workload/fixed_accumulate.go` (new) — `LoadTraceV2FixedAccumulateRequests`
- `cmd/replay.go` — new mode value, guards, request-building branch (if/else → tagged switch)
- `sim/workload/fixed_accumulate_test.go`, `cmd/replay_fixed_accumulate_test.go` — property + determinism + guard tests
- `docs/contributing/standards/invariants.md` — INV-10 scoped to closed-loop; fixed-accumulate exempt by design
- `CLAUDE.md`, `docs/guide/observe-replay-calibrate.md` — mode docs + #1627 decode caveat

## Testing

`go build ./...`, `go test ./... -count=1` (all pass), `golangci-lint run ./...` (0 issues). End-to-end verified on an 8-session simultaneous-arrival corpus: warm rounds show ~2.3× the TTFT of cold rounds (growing context) and TTFT is in the seconds range from real queueing — behavior impossible in closed-loop. Conservation holds (INV-1: injected = completed + still_running).

## Notes

- **INV-10 exemption:** documented as scoped to closed-loop — chaining arrivals to sim completion is exactly the feedback loop this mode breaks.
- **Necessary, not sufficient (#1627):** BLIS's decode/step model still runs ~5–10× fast on this workload, so queue depth stays under-predicted until decode is calibrated. This mode is a precondition for high-conc fidelity, not a standalone fix — documented in the guide.

Fixes #1692

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -297,8 +297,10 @@ go build -o blis main.go
 # (unmodeled enforce-eager + MTP-without-contention, #1627), so queue depth is under-predicted
 # until decode is calibrated — this mode is a NECESSARY precondition for high-conc fidelity, not
 # sufficient alone. Same huge-ISL caveat: raise --max-model-len and scale --total-kv-blocks.
+# (Example uses a committed MoE config + a hardware key present in hardware_config.json so
+# it runs out of the box; the motivating shape is a large MLA MoE like Kimi-K3 on H200.)
 ./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
-  --model kimi-k3 --hardware H200 --tp 16 --dp 2 --enable-expert-parallel \
+  --model qwen/qwen3-30b-a3b --hardware H100 --tp 2 --dp 2 --enable-expert-parallel \
   --session-mode fixed-accumulate --max-model-len 1000000
 
 # Observe corpus-mode: drive the SAME corpus as a fixed session pool against a
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -283,6 +283,23 @@ go build -o blis main.go
 # to closed-loop). Same huge-ISL caveat — raise --max-model-len and scale --total-kv-blocks.
 ./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
   --model qwen/qwen3-14b --concurrent-sessions 8 --total-sessions 200 --max-model-len 1000000
+# FAITHFUL replay of a high-concurrency agentic run (#1692): --session-mode fixed-accumulate
+# injects each round at its RECORDED arrival time (open-loop, like fixed) WHILE reconstructing
+# the growing accumulate-delta input (like closed-loop). Closed-loop regenerates arrivals as
+# completion+think, a self-throttling feedback loop that drains the queue and collapses TTFT to
+# compute-only (30–100× low at conc≥8); fixed mode is hard-rejected on an accumulate corpus (it
+# would misread the deltas as absolutes). fixed-accumulate is the only mode that reproduces the
+# real cross-session overlap, so N large prefills pile into the scheduler at the real clock and
+# produce genuine queue wait. Requires an accumulate corpus; mutually exclusive with
+# --concurrent-sessions (open-loop, no pool) and --think-time-* (arrivals are recorded, not
+# regenerated). INV-10 (session causality) is scoped to closed-loop and does not apply here.
+# Second-order caveat: BLIS's decode/step model still runs ~5–10× fast on this workload
+# (unmodeled enforce-eager + MTP-without-contention, #1627), so queue depth is under-predicted
+# until decode is calibrated — this mode is a NECESSARY precondition for high-conc fidelity, not
+# sufficient alone. Same huge-ISL caveat: raise --max-model-len and scale --total-kv-blocks.
+./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
+  --model kimi-k3 --hardware H200 --tp 16 --dp 2 --enable-expert-parallel \
+  --session-mode fixed-accumulate --max-model-len 1000000
 
 # Observe corpus-mode: drive the SAME corpus as a fixed session pool against a
 # LIVE server (observe-side twin of `blis replay --concurrent-sessions`), so
--- a/sim/workload/fixed_accumulate.go
+++ b/sim/workload/fixed_accumulate.go
@@ -25,12 +25,20 @@ import (
 //
 // The insight: the growing buffer needs no live sim output. The trace's delta chain
 // (abs_N = abs_{N-1} + out_{N-1} + delta_N) plus input_tokens_reset compaction markers
-// fully determines every round's absolute input from trace data alone. So we walk the
-// delta law here — appending each round's generated output then its delta (or Reset-ing
-// to a recorded absolute on a compaction round) into a session-scoped buffer, exactly as
-// SessionManager.OnComplete does — and pre-bake every round as a request at its recorded
-// arrival. The result carries the real cross-session overlap, so N large prefills pile
-// into the scheduler per the real clock and produce genuine queueing delay.
+// fully determines every round's absolute input from trace data alone. So we invert the
+// ENCODER's delta law here (EncodeSessionToTraceRecords) — appending each round's
+// generated output then its delta (or Reset-ing to the recorded absolute on a compaction
+// round) into a session-scoped buffer — and pre-bake every round as a request at its
+// recorded arrival. The result carries the real cross-session overlap, so N large
+// prefills pile into the scheduler per the real clock and produce genuine queueing delay.
+//
+// This mirrors SessionManager.OnComplete's buffer mechanics but feeds it the RECORDED
+// output (rec.OutputTokens), whereas OnComplete appends the ACTUAL sim output
+// (ProgressIndex − InputLen, which a length cap can truncate). The two are the same
+// absent capping; feeding the recorded output is the correct choice here because the
+// encoder's law is written over the recorded output, so this is its exact inverse (no
+// INV-13 claim spans the two modes, and token CONTENT also differs from closed-loop on
+// compaction rounds due to RNG draw order — harmless).
 //
 // # Contract with the encoder (INV-13 with the accumulate closed-loop path)
 //
@@ -119,7 +127,14 @@ func LoadTraceV2FixedAccumulateRequests(trace *TraceV2, seed int64) ([]*sim.Requ
 		// Seed the growing buffer with round 0's input: prefix (if any) + a generated
 		// suffix of effectiveInputTokenCount tokens. Layout mirrors the closed-loop
 		// buffer: [prefix | r0_conversation | r0_output | r1_delta | r1_output | ...].
-		buf := newSessionTokenBuffer()
+		//
+		// Pre-allocate to the session's EXACT peak buffer length (computed from the delta
+		// chain) so the growth phase never reallocates — every round's Slice() view then
+		// aliases ONE backing array, and the session retains O(peak) tokens, not the
+		// O(Σ absolute inputs) an eager per-round copy would retain. On the motivating
+		// Weka corpus (ISL p50 ≈ 110K over ~37K rounds) that is the difference between
+		// ~0.4 GB and ~16–24 GB. Matches the reasoning.go accumulate pattern (#1445).
+		buf := newSessionTokenBufferWithCapacity(accumulatePeakBufferLen(rounds, len(prefix)))
 		if len(prefix) > 0 {
 			buf.Append(prefix)
 		}
@@ -152,10 +167,16 @@ func LoadTraceV2FixedAccumulateRequests(trace *TraceV2, seed int64) ([]*sim.Requ
 				inputTokens = buf.Slice(0, inputEnd)
 			} else {
 				// Normal growth: append prev round's output, then this round's delta.
+				// A round>0 in an accumulate corpus stores its per-round DELTA directly in
+				// rec.InputTokens — read it as-is. (Do NOT route through
+				// effectiveInputTokenCount here: that helper returns the ABSOLUTE
+				// ServerInputTokens when set, which would be a large silent over-count when
+				// consumed as a delta. ServerInputTokens/PrefixGroup are round-0 concerns;
+				// the prefix is already folded into the buffer.)
 				if len(prevOutputTokens) > 0 {
 					buf.Append(prevOutputTokens)
 				}
-				deltaToks := sim.GenerateRandomTokenIDs(sessionRNG, effectiveInputTokenCount(rec.InputTokens, rec.ServerInputTokens, rec.PrefixGroup))
+				deltaToks := sim.GenerateRandomTokenIDs(sessionRNG, rec.InputTokens)
 				_, inputEnd := buf.Append(deltaToks)
 				inputTokens = buf.Slice(0, inputEnd)
 			}
@@ -163,14 +184,17 @@ func LoadTraceV2FixedAccumulateRequests(trace *TraceV2, seed int64) ([]*sim.Requ
 			outputTokens := sim.GenerateRandomTokenIDs(sessionRNG, rec.OutputTokens)
 			prevOutputTokens = outputTokens
 
-			// Copy the buffer view into an independent slice: distinct rounds must not
-			// alias one growable backing array (a later Append/Reset would shift or
-			// orphan an earlier round's view). Copying makes each request's InputTokens
-			// stable and content-frozen at its recorded absolute length.
-			frozen := make([]sim.TokenID, len(inputTokens))
-			copy(frozen, inputTokens)
-
-			requests = append(requests, buildFixedAccumulateRequest(rec, sessionID, frozen, outputTokens, originShift))
+			// inputTokens is a VIEW into the shared session buffer (no copy). This is the
+			// #1445 shared-buffer pattern (see reasoning.go): the simulator treats
+			// Request.InputTokens as read-only (Request.FullInputTokens/InputTokenSlice
+			// hand out capped three-index slices, so a stray append can't overwrite the
+			// buffer), and the buffer's reallocation hazard is documented-safe — Append only
+			// grows (indices [0,end_i) of an earlier view never move) and Reset allocates a
+			// fresh array (earlier views keep pointing at the old, content-correct array).
+			// Because the buffer was pre-sized to the peak above, the no-compaction path
+			// never reallocates, so all views alias one array; a compaction Reset shrinks it
+			// and the pre-reset views remain valid on the old array.
+			requests = append(requests, buildFixedAccumulateRequest(rec, sessionID, inputTokens, outputTokens, originShift))
 		}
 	}
 
@@ -201,6 +225,41 @@ func LoadTraceV2FixedAccumulateRequests(trace *TraceV2, seed int64) ([]*sim.Requ
 	})
 
 	return requests, nil
+}
+
+// accumulatePeakBufferLen returns the maximum shared-buffer length the growth loop in
+// LoadTraceV2FixedAccumulateRequests will reach for a session, so the buffer can be
+// pre-allocated once and never reallocate (keeping every round's Slice() view aliased to
+// one backing array).
+//
+// The buffer length after round i's input is that round's ABSOLUTE input in_i, and it
+// never exceeds in_i mid-round (the transient append of out_{i-1} brings it to
+// in_{i-1}+out_{i-1} = in_i − delta_i ≤ in_i). So the peak is max_i(in_i), reconstructed
+// from the same delta law the loop applies:
+//   - round 0: in_0 = prefixLen + delta_0 (the seeded prefix + round-0 suffix)
+//   - reset round: in_i = *InputTokensReset (the recorded absolute)
+//   - normal round i>0: in_i = in_{i-1} + out_{i-1} + delta_i
+// rounds is assumed sorted ascending and consecutive (validated by the caller).
+func accumulatePeakBufferLen(rounds []TraceRecord, prefixLen int) int64 {
+	var peak, prevAbs, prevOut int64
+	for i, rec := range rounds {
+		var abs int64
+		switch {
+		case i == 0:
+			abs = int64(prefixLen) + int64(effectiveInputTokenCount(rec.InputTokens, rec.ServerInputTokens, rec.PrefixGroup))
+		case rec.InputTokensReset != nil:
+			abs = *rec.InputTokensReset
+		default:
+			// round>0 delta is rec.InputTokens directly (see the growth branch in
+			// LoadTraceV2FixedAccumulateRequests — not effectiveInputTokenCount).
+			abs = prevAbs + prevOut + int64(rec.InputTokens)
+		}
+		if abs > peak {
+			peak = abs
+		}
+		prevAbs, prevOut = abs, int64(rec.OutputTokens)
+	}
+	return peak
 }
 
 // buildFixedAccumulateRequest assembles a sim.Request from a trace record with the
--- a/sim/workload/fixed_accumulate.go
+++ b/sim/workload/fixed_accumulate.go
@@ -128,16 +128,30 @@ func LoadTraceV2FixedAccumulateRequests(trace *TraceV2, seed int64) ([]*sim.Requ
 		// suffix of effectiveInputTokenCount tokens. Layout mirrors the closed-loop
 		// buffer: [prefix | r0_conversation | r0_output | r1_delta | r1_output | ...].
 		//
-		// Pre-allocate to the session's EXACT peak buffer length (computed from the delta
-		// chain) so the growth phase never reallocates — every round's Slice() view then
-		// aliases ONE backing array, and the session retains O(peak) tokens, not the
-		// O(Σ absolute inputs) an eager per-round copy would retain. On the motivating
-		// Weka corpus (ISL p50 ≈ 110K over ~37K rounds) that is the difference between
-		// ~0.4 GB and ~16–24 GB. Matches the reasoning.go accumulate pattern (#1445).
-		buf := newSessionTokenBufferWithCapacity(accumulatePeakBufferLen(rounds, len(prefix)))
+		// Pre-allocate to the FIRST EPOCH's peak (rounds up to the first compaction reset),
+		// computed from the delta chain, so the growth phase never reallocates within the
+		// epoch — every round's Slice() view then aliases ONE backing array. At each reset
+		// we re-size to the NEXT epoch's peak (see ResetWithCapacity below). Per-epoch (not
+		// global-peak) sizing is deliberate: sizing to the global peak would pin every
+		// earlier round's small view inside one giant array, retaining more than an eager
+		// copy on a shrinking-then-growing corpus (PR #1696 review F1). Retained memory is
+		// therefore O(Σ per-epoch peaks): for a monotone (no-compaction) session that is a
+		// single O(peak) array (~0.4 GB vs ~16–24 GB of eager copies on the Weka corpus);
+		// for a compacting session it is one right-sized array per live epoch. Matches the
+		// reasoning.go shared-buffer pattern (#1445).
+		buf := newSessionTokenBufferWithCapacity(accumulateEpochPeakLen(rounds, 0, len(prefix)))
 		if len(prefix) > 0 {
 			buf.Append(prefix)
 		}
+		// Round 0 seeds from effectiveInputTokenCount (server-reported count wins when set),
+		// but rounds>0 read rec.InputTokens raw as the DELTA. CAVEAT (latent): if a corpus
+		// ever carried a round-0 ServerInputTokens that differed from InputTokens, seeding
+		// abs_0 from the server count while the encoder wrote deltas over the InputTokens
+		// column would shift every later reconstructed absolute by (server − input).
+		// Unreachable today — no converter (weka/otel) or re-exporter sets ServerInputTokens
+		// on an accumulate corpus, and observe (the only ServerInputTokens producer) never
+		// emits session_context_growth=accumulate, so such a trace is rejected by the CLI
+		// guard before reaching here. Left as a documented boundary rather than a guard.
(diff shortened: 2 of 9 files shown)


susiejojo:
@claude /blis-pr-review. Check for code correctness and quality, and take into account the Archon architecture changes through /archon-pr-review

susiejojo:
Thanks for the review! The verdict was **APPROVE** with two non-blocking follow-ups, both now addressed in 40c250f4.

**1. Document `--trace-output` re-export behavior** ✅
Added a note to the fixed-accumulate section of `observe-replay-calibrate.md` and to the `--trace-output` flag help on `blis replay`: because fixed-accumulate reconstructs the growing context in memory, the re-export records **reconstructed absolute** per-round inputs with **no** `session_context_growth` header — a faithful absolute-mode corpus. Re-replay it with `--session-mode fixed` (not fixed-accumulate). Verified empirically: `s1` re-exports as absolutes 100/180/245, and the existing accumulate-corpus guard produces a clear error if fixed-accumulate is attempted on it.

**2. Add a test for it** ✅
`TestReplayFixedAccumulate_TraceOutput_AbsoluteCorpus` pins the behavior: exported header has no `session_context_growth`, `s1` inputs are recorded as absolutes (not deltas), and no `input_tokens_reset` markers are emitted.

While wiring the test I also caught and fixed a **test-var leak**: `captureCmdLevelVars()` snapshots the *run* command's `traceOutput`, not replay's `replayTraceOutput`, so the new test's temp-dir path leaked into a later test. The helper now restores `replayTraceOutput` explicitly alongside the other replay-only vars.

I left the other two observations as-is since the review flagged them as pre-existing / by-design and non-blocking:
- **RequestSource ordering**: the output slice isn't arrival-sorted, but the cluster pushes all requests to a min-heap before the event loop, so ordering is correct at execution — same as `LoadTraceV2Requests`/`LoadTraceV2SessionBlueprints`. Pre-existing doc inconsistency, not introduced here.
- **`--think-time-ms` with fixed-accumulate**: the existing guard already rejects `--think-time-ms`/`--think-time-dist` outside closed-loop, so fixed-accumulate + think-time is a hard error (arrivals are recorded, not regenerated) — no separate path to test.

Full verification gate green: `go build ./...` (0), `go test ./... -count=1` (all pass), `golangci-lint run ./...` (0 issues).

susiejojo:
@claude /blis-pr-review

susiejojo:
Thanks — the blocking finding is valid and now fixed in d76307ec.

**🔴 Arrival-time ordering (blocking)** ✅
`LoadTraceV2FixedAccumulateRequests` emitted requests in session-major order, which violates the `RequestSource` non-decreasing-ArrivalTime contract for any interleaved (high-concurrency) corpus. Added a stable arrival-time sort before returning, matching `generator.go:307/513`. It's safe because every request is fully materialized with a **frozen** `InputTokens` slice before the sort, so reordering can't disturb the per-session accumulate reconstruction. `sort` was already imported.

**Tests** ✅
- `TestFixedAccumulate_MultiSessionArrivalOrder` (workload): asserts non-decreasing arrivals AND that reconstruction survives interleaving (s1 → 100/180/245, s2 → 60/95 regardless of slice position).
- `TestReplayFixedAccumulate_TraceOutput_AbsoluteCorpus` now also asserts the re-exported records are arrival-ordered.

**One correction on the panic claim:** the `--trace-output` re-export on the *replay* path does **not** go through `fireArrivalHook` — `SetArrivalHook` is a `blis run`-only wiring; replay re-exports from the captured `requests` slice (`ReExportClosedLoopRecords`/`RequestsToTraceRecords`) after `Run()`. That's why the existing trace-output test genuinely passed rather than getting lucky. The contract fix is still correct and necessary — it's what removes the same-timestamp tie ambiguity (INV-6) and matches every sibling loader — I just want the record to be precise about the failure mode.

**Non-blocking gaps** — addressed the cheap ones:
- `--think-time-ms` + fixed-accumulate rejection → `TestReplayFixedAccumulate_RejectsThinkTime` (subprocess).
- Single-round session → `TestFixedAccumulate_SingleRoundSession`.
- Reset content-after-reset → `TestFixedAccumulate_CompactionReset_Content` (verifies the reset segment is fresh, not a prefix continuation of round 0).

Left the `RequestSource` doc-contract observation and the export-time UX warning as-is (pre-existing / by-design; the flag help + guide note already cover the export semantics).

Full gate green: `go build ./...` (0), `go test ./... -count=1` (all pass), `golangci-lint run ./...` (0 issues).

susiejojo:
@claude /blis-pr-review

susiejojo:
Thanks — F1 is a real blocker and is fixed in fbb781ed, along with F2–F6.

**🔴 F1 — O(Σ inputs) copy → OOM (blocking)** ✅
Replaced the per-round anti-aliasing copy with the #1445 shared-buffer **view** pattern (as `reasoning.go`): each round's `InputTokens` is now a `buf.Slice(0, end_i)` view into one backing array, and the buffer is pre-sized to the session's **exact peak** via a new `accumulatePeakBufferLen` (reconstructed from the delta chain), so the no-compaction path never reallocates. Retained memory drops from O(Σ absolute inputs) to O(peak per session) — the ~16–24 GB → ~0.4 GB you computed. The copy was never needed for correctness: `InputTokens` is read-only in the sim (capped three-index slices), `Append` only grows (earlier indices never move), and `Reset` allocates a fresh array so pre-reset views stay content-correct. Pinned by `TestFixedAccumulate_ViewsStableAfterGrowthAndReset`, which re-reads every round's live view after the full session (including across a compaction reset) and asserts no mutation.

**F2 — third construction site, no guard** ✅
Added `TestFixedAccumulate_FieldParityWithLoadTraceV2Requests`: a fully-populated non-session record is run through both `LoadTraceV2Requests` and the fixed-accumulate loader, and the two `sim.Request` structs must be `reflect.DeepEqual` (minus RNG-dependent token contents). Any field added to one literal but not the other fails the test. (I went with the reflection guard over extracting a shared helper — the three loaders differ in prefix handling, so a helper would need conditional params; the guard is lower-risk and is what R4/R23 want structurally.)

**F3 — `effectiveInputTokenCount` on rounds>0** ✅
Round>0 now reads `rec.InputTokens` directly (the delta), both in the growth loop and in `accumulatePeakBufferLen`. Documented why (the helper returns the *absolute* `ServerInputTokens`, a latent over-count if consumed as a delta).

**F4 — doc comment** ✅
Reworded: the loader inverts the **encoder's** law using the **recorded** output; it mirrors but is not identical to `SessionManager.OnComplete` (which uses actual `ProgressIndex − InputLen`), and token content also diverges from closed-loop on compaction rounds (RNG draw order) — harmless, no INV-13 claim spans the two modes.

**F5 — INV-1 conservation** ✅
`TestReplayFixedAccumulate_EndToEnd` now also asserts `injected_requests == 5`.

**F6 — origin-shift test** ✅
`TestFixedAccumulate_NormalizesEpochSendOrigin` — epoch-scale `send_time_us` corpus injects on the arrival origin with send-delta spacing preserved (#1606), parallel to the `LoadTraceV2Requests` case.

Full gate green: `go build ./...` (0), `go test ./... -count=1` (14 pkgs ok, no FAIL), `golangci-lint run ./...` (0 issues).

susiejojo:
@claude /blis-pr-review

sim/workload/fixed_accumulate.go
```diff
@@ -0,0 +1,313 @@
...
+		// suffix of effectiveInputTokenCount tokens. Layout mirrors the closed-loop
+		// buffer: [prefix | r0_conversation | r0_output | r1_delta | r1_output | ...].
+		//
+		// Pre-allocate to the session's EXACT peak buffer length (computed from the delta
+		// chain) so the growth phase never reallocates — every round's Slice() view then
+		// aliases ONE backing array, and the session retains O(peak) tokens, not the
+		// O(Σ absolute inputs) an eager per-round copy would retain. On the motivating
+		// Weka corpus (ISL p50 ≈ 110K over ~37K rounds) that is the difference between
+		// ~0.4 GB and ~16–24 GB. Matches the reasoning.go accumulate pattern (#1445).
+		buf := newSessionTokenBufferWithCapacity(accumulatePeakBufferLen(rounds, len(prefix)))
```
jgchn: **F1 (Medium): this claim is false under compaction, and the pre-sizing can make memory 2x worse than the per-round copy it replaced.**

The mechanism breaks at every compaction round. `sessionTokenBuffer.Reset` (`sim/workload/session_buffer.go:103-108`) does `make([]sim.TokenID, len(toks))`, so **cap == len**. After the first reset the pre-sizing is gone, the next `Append` reallocates, and each pre-reset round's view keeps its old array alive.

Measured against this loader (live views grouped by backing-array address):

| Shape | Live arrays | Retained | Eager per-round copy |
|---|---|---|---|
| No compaction, 40 rounds | 1 | **1.00x peak** | 22.3x (claim holds here) |
| 9 compaction epochs, 37 rounds | 28 | **25.20x peak** | 27.1x (**a 7% saving, not O(peak)**) |
| Tiny first epoch, huge later epoch | 3 | **1,001,805 tokens** | 500,055 (**2x worse than a copy**) |

Row 3 is the sharp one: round 0 holds 10 tokens inside the `cap=500,025` array pre-sized to the *global* peak on line 137, and round 0's view keeps that array alive for the whole run.

This is not an edge case for the motivating workload. CLAUDE.md records that roughly 30% of rounds on the full Weka 051926 dataset are compaction rounds, which puts the Weka corpus squarely in rows 2 and 3, so the "~0.4 GB and ~16-24 GB" figures here and in the PR body are not supported.

Suggested fix: size the post-`Reset` array to the *remaining* peak (`accumulatePeakBufferLen(rounds[i:], 0)`), or add a `ResetWithCapacity`. At minimum, correct this comment and the PR body's GB numbers.

sim/workload/fixed_accumulate.go
```diff
@@ -0,0 +1,313 @@
...
+				// exactly resetTarget tokens (mirrors SessionManager.OnComplete). The
+				// just-completed round's output is intentionally NOT carried forward — it
+				// was folded into the compaction the trace recorded as this round's abs.
+				resetToks := sim.GenerateRandomTokenIDs(sessionRNG, resetTarget)
+				_, inputEnd := buf.Reset(resetToks)
+				inputTokens = buf.Slice(0, inputEnd)
+			} else {
+				// Normal growth: append prev round's output, then this round's delta.
+				// A round>0 in an accumulate corpus stores its per-round DELTA directly in
+				// rec.InputTokens — read it as-is. (Do NOT route through
```
jgchn, line 188: **F4 (Low, latent): the round-0 asymmetry this comment describes silently breaks the delta chain when `server_input_tokens` is set.**

Round 0 goes through `effectiveInputTokenCount` (line 122), so `ServerInputTokens` wins when set. Round>0 reads `rec.InputTokens` raw, as this comment correctly explains. But `EncodeSessionToTraceRecords` wrote every delta over the `input_tokens` column, so a corpus carrying `server_input_tokens` on round 0 shifts every later round's reconstructed absolute by `server - input`.

Unreachable today, so this is a note and not a bug: `observe` never emits an accumulate corpus, `reexport.go` sets no `ServerInputTokens`, and neither converter sets it. Worth either an inline caveat here or a round-0 guard.

For the record, closed-loop is worse in this respect. `sim/workload/replay.go:320` runs *every* round through the same helper, so a round>0 delta becomes an absolute there. This loader is an improvement, not a regression.

sim/workload/fixed_accumulate.go
```diff
@@ -0,0 +1,313 @@
...
+// buildFixedAccumulateRequest assembles a sim.Request from a trace record with the
+// already-reconstructed input/output token slices. Field set matches
+// LoadTraceV2Requests (R4). PrefixGroup/PrefixLength are deliberately NOT carried on
+// session rounds: accumulate folds the prefix into round 0's reconstructed input, so
+// re-prepending it would double-count. Non-session records keep them (their input is
+// suffix-only, exactly as fixed mode).
+func buildFixedAccumulateRequest(rec TraceRecord, sessionID string, inputTokens, outputTokens []sim.TokenID, originShift int64) *sim.Request {
+	req := &sim.Request{
+		ID:               fmt.Sprintf("request_%d", rec.RequestID),
+		ArrivalTime:      injectionTime(rec) - originShift, // #1606: on the arrival/deadline origin
```
jgchn, line 301: **F2 (Medium): this line lets two rounds of the SAME session be in flight concurrently, which the docs do not mention.**

`ArrivalTime` comes purely from the trace, with zero dependence on the previous round's completion. Verified directly against the loader: three rounds of one session all injected at tick 0, with round 1's input containing round 0's not-yet-generated output tokens.

This is reachable on real corpora rather than synthetic ones. `sim/workload/weka_convert.go:147` writes `ArrivalUs` as the turn's wall-clock time, so the recorded inter-round gap includes the *real* api_time. The moment sim service time exceeds real api_time, which is exactly the queueing regime this mode exists to create, round N is injected while round N-1 is still decoding.

Two consequences worth documenting:

1. Intra-session concurrency that no real agentic client produces, inflating measured concurrency above real.
2. Round N's prefix-cache probe covers tokens round N-1 has not emitted yet, so part of the hit rate is fictional.

`docs/guide/observe-replay-calibrate.md:503` sells the mode as reproducing "the real cross-session overlap". It also manufactures unreal *intra*-session overlap. The INV-10 exemption note covers the arrival *source* but not this second-order effect, so it needs its own caveat alongside the #1627 one.

sim/workload/fixed_accumulate_test.go
```diff
@@ -0,0 +1,457 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	return trace
+}
+
+// TestFixedAccumulate_ReconstructsAbsoluteInputs (BC-1): a session whose recorded
+// absolute inputs grow across rounds must produce requests whose InputTokens count
+// equals the recorded absolute per round, injected at the recorded arrival time.
+func TestFixedAccumulate_ReconstructsAbsoluteInputs(t *testing.T) {
```
jgchn, line 33: **Test gap worth closing:** every test in this file hand-computes the deltas (`180-100-50 = 30`), so an encoder/loader law drift would pass all of them. There is no test that goes through the real `EncodeSessionToTraceRecords`.

I ran the missing property test and it passes cleanly, so this is cheap to add rather than a defect: build `[]NormalizedRound` with known absolutes (including shrinking rounds to trigger compaction), encode with `EncodeSessionToTraceRecords`, load with `LoadTraceV2FixedAccumulateRequests`, and assert `r.InputLen() == rounds[i].InputTokensAbs`. Randomized over 300 trials with roughly a third compaction rounds it found no divergence, which is the evidence that pins BC-1 and BC-3 to the encoder rather than to the test author's arithmetic.

CLAUDE.md
```diff
@@ -283,6 +283,23 @@ go build -o blis main.go
...
+# real cross-session overlap, so N large prefills pile into the scheduler at the real clock and
+# produce genuine queue wait. Requires an accumulate corpus; mutually exclusive with
+# --concurrent-sessions (open-loop, no pool) and --think-time-* (arrivals are recorded, not
+# regenerated). INV-10 (session causality) is scoped to closed-loop and does not apply here.
+# Second-order caveat: BLIS's decode/step model still runs ~5–10× fast on this workload
+# (unmodeled enforce-eager + MTP-without-contention, #1627), so queue depth is under-predicted
+# until decode is calibrated — this mode is a NECESSARY precondition for high-conc fidelity, not
+# sufficient alone. Same huge-ISL caveat: raise --max-model-len and scale --total-kv-blocks.
+./blis replay --trace-header corpus.yaml --trace-data corpus.csv \
+  --model kimi-k3 --hardware H200 --tp 16 --dp 2 --enable-expert-parallel \
```
jgchn: **F3 (Low): this example fails out of the box on two counts.**

1. `model_configs/kimi-k3/` is not committed to the repo. `git ls-tree -r <head> --name-only | grep -i kimi` returns nothing.
2. `H200` is not a key in `hardware_config.json` (present: `H100`, `A100-SXM`, `A100-80`, `L40S`), so `GetHWConfig` hard-errors at `sim/latency/config.go:190-197` before the new mode is ever exercised.

Same line appears at `docs/guide/observe-replay-calibrate.md:515`. Switching to a committed pair such as `--model qwen/qwen3-30b-a3b --hardware H100` would make both copy-pasteable, or commit the kimi-k3 config and an H200 calibration entry if the K3-on-H200 shape is the point of the example.

jgchn (commented):
## Correctness review (head `fbb781ed`)

**The core reconstruction is correct.** Build, `go test ./... -count=1`, and `golangci-lint run ./...` (0 issues) all pass. Two findings I would fix before merge (both claim/doc accuracy on code that is functionally right), three notes, and one process item.

### The central claim is verified, not assumed

Every loader test hand-computes the deltas, so they would all pass even if the loader and the encoder disagreed. I ran the round trip that is missing: known absolutes -> the real `EncodeSessionToTraceRecords` -> `LoadTraceV2FixedAccumulateRequests` -> compare. **300 randomized sessions, 1 to 8 rounds, roughly a third compaction rounds: every round reconstructs the exact recorded absolute.** BC-1 and BC-3 hold against the encoder. Predicate alignment is exact too: `fixed_accumulate.go:152` uses `i > 0 && rec.InputTokensReset != nil`, identical to `sim/workload/replay.go:336`.

### Findings

| # | Severity | Finding |
|---|---|---|
| F1 | Medium | Shared-buffer memory claim is false under compaction, and can be 2x worse than the copy it replaced (inline on `fixed_accumulate.go:132-137`) |
| F2 | Medium | Same-session rounds can be in flight concurrently, undocumented (inline on `fixed_accumulate.go:274`) |
| F3 | Low | Both new examples fail out of the box (inline on `CLAUDE.md:301` and the guide) |
| F4 | Low, latent | Round-0 `ServerInputTokens` would break the delta chain (inline on `fixed_accumulate.go:171`) |
| F5 | Low | No observe-side twin, no follow-up filed (below) |
| F6 | Process | Both existing review comments are stale (below) |

### F5: cross-path parity gap, no follow-up filed

`blis observe` corpus-mode is closed-loop-pool only (`cmd/observe_corpus.go:39,56`), so you cannot drive the same corpus at recorded arrivals against a live server. That is precisely what `blis calibrate` needs to validate this mode's premise. The earlier CI review marked parity "N/A", which I do not agree with given this PR's own sim2real motivation. Filing the follow-up would satisfy the project's cross-path parity rule.

### F6: both existing review comments predate the current code

The CI `blis-pr-review` comment's central section, "Correctness: Aliasing Safety", reviews a `frozen := make(...); copy(...)` at `fixed_accumulate.go:170-171` that commit `fbb781ed` **deleted**. Its "output slice is NOT sorted by ArrivalTime" section was fixed by `d76307ec` (the sort now lives at `fixed_accumulate.go:223`), and its `--trace-output` recommendation was implemented by `40c250f4`. Archon ran base->head `9cd8205d`, two commits behind head `fbb781ed`; verdict `ARCHITECTURAL_CHANGE`, 1 surface added, 0 violations, and no plan-check section, so no plan gate ran.

Net effect: **nothing on this PR has reviewed the current no-copy view-sharing code**, which is exactly what F1 is about. Worth re-running both before merge.

### Verdict

Not blocking on correctness. F1's comment and F2's caveat are the two I would want fixed first: F1 is either a one-line `Reset` capacity change or an honest comment rewrite, F2 is a paragraph. F3 is a two-token doc fix.

susiejojo:
Thanks — this was a genuinely deep pass. All six addressed in b440eff2, plus the follow-up filed. And you're right that nothing had reviewed the no-copy code; the memory finding is real.

**F1 (memory claim false under compaction / worse than a copy)** ✅
You're correct on both counts: `Reset` did `make(..., len)` (cap==len), so after a compaction the pre-sizing was lost and the next `Append` reallocated, keeping each pre-reset epoch's peak array alive — and pre-sizing the *initial* buffer to the **global** peak pinned round 0's tiny view inside a giant array (your row 3, 2× worse than a copy).

Fix: added `sessionTokenBuffer.ResetWithCapacity` and switched sizing from global-peak to **per-epoch** peak (`accumulateEpochPeakLen`, computed over each run of rounds between compaction boundaries). The initial buffer is sized to epoch 0's peak; each reset re-sizes to the *next* epoch's peak. Retained memory is now O(Σ per-epoch peaks) — one right-sized array per live epoch (a single O(peak) array for a monotone session). Corrected the code comment and dropped the unsupported GB framing for the compaction case. Pinned by `TestFixedAccumulate_EpochScopedCapacity` (verified it fails on the old global-peak alloc: round-0 view cap 5008 vs ~10) and a `ResetWithCapacity` unit test.

**F2 (intra-session overlap undocumented)** ✅
Added a dedicated warning admonition in the guide: recorded arrivals ignore sim completion, so once sim service time exceeds the recorded api_time (the queueing regime), same-session rounds overlap — inflating measured concurrency and making part of the prefix-cache hit fictional (round N's prompt contains round N−1 output that hasn't been emitted yet). Called out as distinct from the faithful *cross*-session overlap and not covered by the INV-10 exemption.

**F3 (examples fail out of the box)** ✅
Both examples now use committed `qwen/qwen3-30b-a3b` + `H100` (verified runnable from the repo root with no overrides — 2 requests complete across the dp=2 replicas). Kept the K3-on-H200 shape as prose since it's the motivating case.

**F4 (round-0 ServerInputTokens latent break)** ✅
Added the inline caveat you suggested (unreachable today; documented boundary rather than a guard).

**Test gap (encoder round-trip)** ✅
Added `TestFixedAccumulate_EncoderRoundTrip` — exactly the property you ran: known absolutes → real `EncodeSessionToTraceRecords` → loader → assert exact reconstruction, 200 randomized trials with ~1/3 compaction rounds, deterministic (seeded LCG, no wall-clock). This pins BC-1/BC-3 to the encoder rather than to hand-computed deltas.

**F5 (cross-path parity, no follow-up)** ✅
Agree the "N/A" was wrong given this mode's sim2real premise. Filed #1701: an observe-side fixed-arrival corpus dispatch (the twin of `--session-mode fixed-accumulate`) so `blis calibrate` can close the loop. Left it as a follow-up per the "file, don't fix in-PR" rule.

**F6 (stale bot reviews)** — acknowledged; the CI `blis-pr-review` and Archon comments predate the no-copy code (they reviewed the deleted `frozen`/copy and an unsorted slice, both since fixed). Re-running them before merge makes sense.

Full gate green: `go build ./...` (0), `go test ./... -count=1` (14 pkgs ok, no FAIL), `golangci-lint run ./...` (0 issues).

jgchn (approved):

PR #1702, opened 2026-09-10
fix(kv-offload): bill CPU/secondary cache hits as cheap prefill, not recompute (Closes #1699)

### LINKED ISSUE #1699: KV offload to CPU changes hit rate but not latency
(I observed certain behaviors in BLIS, which need further investigation. Issue is created using the template for bug report.)

**Describe the problem**

Modifying the values of `kv_offload.cpu_bytes_to_use` changes the reported `cache_hit_rate` by 4.8x (0.0772 → 0.3674) while **every** timing and throughput metric stays identical. TTFT, ITL, E2E at every percentile, `responses_per_sec`, `tokens_per_sec`, and `vllm_estimated_duration_s` are unchanged to the last digit.

**CLI command to reproduce**

```bash
cat > /tmp/wl_mp2.yaml <<'EOF'
version: "2"
seed: 42
aggregate_rate: 4.0
num_requests: 600
cohorts:
  - id: tenants
    population: 100
    prefix_group: doc
    prefix_sharing: per_member
    prefix_length: 1024
    rate_fraction: 1.0
    arrival: {process: poisson}
    input_distribution:  {type: constant, params: {value: 1280}}
    output_distribution: {type: constant, params: {value: 16}}
EOF

cat > /tmp/kvoff-cpu.yaml <<'EOF'
kv_offload:
  cpu_bytes_to_use: 53687091200   # 50 GiB  | value changes in the table below
  block_size: 16
  eviction_policy: lru
  offload_prompt_only: true
EOF

./blis run --model qwen/qwen3-14b --workload-spec /tmp/wl_mp2.yaml \
           --total-kv-blocks 3000 --kv-offload-config /tmp/kvoff-cpu.yaml
```

Re-run with each `cpu_bytes_to_use` below, plus one run with `--kv-offload-config` omitted entirely.

**Expected behavior**

Larger CPU offload capacity retains more prefix blocks evicted from GPU, so prompt tokens are served from cache instead of recomputed -  mean TTFT should change with it. I don't expect latency to be identical when `cache_hit_rate` changes by 4.8x. 

**Actual behavior**

| `cpu_bytes_to_use` | | `cache_hit_rate` | ttft_mean_ms | e2e_mean_ms | responses_per_sec |
|---|---|---|---|---|---|
| `1099511627776` | 1 TiB | **0.3674** | 52.55036 | 250.30256166667 | 4.163978529361 |
| `53687091200` | 50 GiB | **0.3351** | 52.55036 | 250.30256166667 | 4.163978529361 |
| `2621440` | 1 block | **0.0772** | 52.55036 | 250.30256166667 | 4.163978529361 |
| *(flag omitted)* | offload off | **0.0772** | 52.55036 | 250.30256166667 | 4.163978529361 |

susiejojo:
## Problem (#1699)

With `--kv-offload-config`, changing `cpu_bytes_to_use` swung the reported `cache_hit_rate` (0.077 → 0.367) while **every** latency/throughput metric stayed byte-identical to the last digit. A cache hit that should have made prefill cheaper had no timing effect.

## Root cause

In `sim/batch_formation.go` Phase 2, a new request's billed prefill work (`numNewTokens`, the sole input to `StepTime`) is computed from the **GPU-only** `GetCachedBlocks` **before** the CPU→GPU reload runs inside `AllocateKVBlocks`. The reload lands prefix blocks on the GPU and bumps `CacheHits` (moving the metric) but `NumNewTokens` is never recomputed — so a genuine offload hit is billed as a **full recompute**.

Both offload paths funnel through `allocateThroughChain`, so the defect hit **both** the CPU-only path and the resolved-secondary-deferral (H3) path. Verified: a landed secondary hit also left TTFT flat.

## How vLLM does it

vLLM (v1 `scheduler.py`) reduces the tokens sent through the prefill forward pass by the external hit:

```python
num_computed_tokens = num_new_local_computed_tokens + num_external_computed_tokens  # GPU + offload
num_new_tokens      = request.num_tokens - num_computed_tokens
```

The `+ num_external_computed_tokens` term is exactly what BLIS lacked — it only subtracted the GPU prefix.

## Fix

Add an optional `ReloadReportingKVStore` interface (same inert-for-others pattern as `DeferrableKVStore`), implemented only by `OffloadCache`:

- `OffloadCache` records the post-reload cached-prefix boundary in `allocateThroughChain`, **only on admit-success paths** and **only for genuinely-new prefill admissions** (running continuations bill incrementally against `ProgressIndex`; PD decode sub-requests never read it) — so no unconsumed entry can leak.
- Phase 2 type-asserts the interface and re-bills `NumNewTokens` to the non-reloaded tail **within the committed chunk**: `endIndex − min(newStart, endIndex)`.
- **Progress (`ComputedTokens`) is unchanged** — the reload changes how much *compute* a chunk costs, not how far the request advances. A reload extending past a capped chunk never over-credits progress.

### Scope

Prefill-shrink only, applied on **both** offload stores — the `--kv-offload-config` `OffloadCache` chain (CPU-only and resolved-secondary-deferral paths) **and** the legacy `--kv-cpu-blocks` `TieredKVCache`, its structural twin (added after review; it had the identical bug). The CPU-only path stays a same-step optimistic reload; the secondary path keeps its existing H3 one-step load cost. Routing CPU-tier hits through a one-step load (vLLM `WAITING_FOR_REMOTE_KVS`) is a documented follow-up.

**Known limitation (#1706) — the credit is capped at one prefill chunk.** Under chunked prefill (`--long-prefill-token-threshold` / `--max-num-batched-tokens`), a reload that extends the cached prefix past a request's per-step chunk boundary credits only the first chunk; the reloaded remainder is re-billed as recompute on later Phase-1 steps (which consult no cache). So the realized TTFT benefit is roughly `chunk_size / prompt_length` — **near-full for prompts that fit one chunk, but small (~2% at the 2048-token default) for the long-context prompts that are the main reason to offload KV**. vLLM instead advances `num_computed_tokens = local + external` *before* the chunk cap, crediting the whole reloaded prefix. Tracked in #1706.

## Effect (issue's repro workload)

| `cpu_bytes_to_use` | hit rate | ttft_mean_ms |
|---|---|---|
| 1 TiB | 0.367 | **48.91** (was 52.55, frozen) |
| 1 block | 0.077 | 52.55 |

TTFT now tracks CPU capacity **for this repro** (1280-token prompts, which fit one prefill chunk). Note this workload sits entirely in the uncapped regime; under chunked prefill of long prompts the realized benefit is bounded by the one-chunk cap above (#1706).

## Invariants

- **INV-6**: the non-offload single-tier `KVCacheState` doesn't implement the interface → re-bill is dead code → a run without offload is **byte-identical to `main`**. A `--kv-cpu-blocks` run with no reloads is also byte-identical (the record is never written); only a run that actually reloads changes, deterministically. (Both offload stores implement the interface — that's the intended, output-changing case.) Verified via a `main`-built binary across single-tier, no-reload legacy, and reload legacy configs.
- **INV-13**: the re-bill is a pure function of deterministic allocation state (no RNG / wall-clock) → replay reproduces identically; e2e determinism asserted.

## Tests

- `ReloadedPrefixEnd` unit: boundary reporting, one-shot consumption, no-reload → `ok=false`, and **failed-tail-alloc records no boundary** (no leak).
- `FormBatch` re-bill: shrink, full-hit zero-prefill, MaxModelLen-capped chunk, reload-beyond-capped-chunk (progress not over-credited), non-reporting-store no-op (INV-6 guard).
- e2e: CPU-hit and secondary-hit TTFT reduction + determinism.


susiejojo:
@claude /blis-pr-review. Check carefully for code correctness, quality and whether the original issue's concerns are addressed. Also take into consideration architectural changes pointed out by /archon-pr-review.

susiejojo:
Thanks for the thorough review. Addressed the one non-blocking item in df6cffb: added a CLAUDE.md entry for `ReloadReportingKVStore` (#1699) in the KV-offload section, matching the `DeferrableKVStore` (#1591) documentation style — covering the billing bug, the interface + `recordReloadedPrefix` gating + Phase-2 re-bill, the progress-vs-compute distinction, the INV-6/INV-13 story, and the `ReloadsPerformed()` diagnostic.

On the noted design choice (§8, full-hit requests admitting up to `MaxNumSeqs` without consuming budget): agreed it's intended — zero compute ⇒ zero budget, matching vLLM, and `StepTime` charges 0 for `NumNewTokens=0`. Left as-is.

susiejojo:
@claude /blis-pr-review

susiejojo:
Addressed both non-blocking observations in e5d7f81:

1. **CLAUDE.md wiki-link** — replaced `[[the DeferrableKVStore above]]` with plain prose ("the `DeferrableKVStore` step-boundary deferral above (#1591)").
2. **`newStart > startIndex` re-check** — kept the guard (cheap defense against a future `ReloadReportingKVStore` implementer relaxing the contract and over-billing) but reworded the comment to state plainly that it's defensive, not load-bearing.

Both comment/doc-only, no behavior change.

sim/batch_formation.go
```diff
@@ -361,9 +369,28 @@ func (v *VLLMBatchFormation) FormBatch(ctx BatchContext) BatchResult {
...
+			if newStart, ok := reloadReporter.ReloadedPrefixEnd(next.ID); ok && newStart > startIndex {
+				billedTokens = endIndex - min(newStart, endIndex) // reloaded portion within the chunk is free
+			}
+		}
+
+		tokenBudget -= billedTokens
 		next.State = StateRunning
-		next.NumNewTokens = int(numNewTokens)
+		next.NumNewTokens = int(billedTokens)
 		ctx.ComputedTokens[next.ID] = numNewTokens + util.Len64(cachedBlocks)*ctx.KVCache.BlockSize()
```
jgchn, line 400: The credit is capped at one chunk, so this only ports half of the vLLM term the PR body quotes.

`ComputedTokens` stays at `endIndex`, so when the reload extends past a capped chunk the remainder is re-billed as **full recompute** on later steps: Phase 1 computes `numNewTokens = InputLen - req.ProgressIndex` and consults no cache at all, and `recordReloadedPrefix` skips running continuations.

vLLM sets `request.num_computed_tokens = num_new_local_computed_tokens + num_external_computed_tokens` (scheduler.py:401-402, 526) **before** the `long_prefill_token_threshold` / budget cap, then adds the capped chunk on top. So vLLM's progress after the step is `newStart + chunk`; this is `startIndex + chunk`.

Reachable with the real store: a 4-block all-CPU-resident input allocated with `endIndex=2` reports `newStart=8`. A two-step `FormBatch` probe (64-token input fully reloaded, `PrefillTokenThreshold=16`) bills 0 on step 1 then **48 of 64 tokens as compute** over 3 more steps.

Measured end to end, 4096-token prompts / 3072 shared prefix, instrumented billed-prefill-token counter:

| `--max-num-batched-tokens` | offload off | this PR, 1TiB CPU | all-GPU ideal | credit realized |
|---|---|---|---|---|
| 8192 (prompt fits one chunk) | 1,238,496 | 882,928 | 880,640 | 99.4% |
| 2048 (default) | 1,238,496 | 1,015,915 | 880,640 | 62.2% |

The credit is roughly `chunk/prompt`, so for the long-context offload workloads this feature targets (weka Kimi ~110K ISL at the 2048 default) it is about 2%. #1699's repro is 1280 tokens < 2048, which is why the PR's effect table sits entirely in the uncapped regime.

Not asking for the fix here, crediting to `newStart` also needs the block commit extended past `endIndex`. But please note the limitation in the PR body and CLAUDE.md and open a follow-up. Also `TestFormBatch_ReloadBeyondCappedChunk`'s comment "the tail beyond it is billed on a later step" reads as billed *correctly* later; it is billed as recompute.

sim/cluster/offload_deferral_e2e_test.go
```diff
@@ -117,3 +117,126 @@ func TestInstanceSimulator_Offload_EndToEnd_DrainsAndDeterministic(t *testing.T)
...
+	offload := runOffloadE2E(7) // has a secondary tier; workload exercises deferral
+	gpuOnly := runGPUOnly(7)
+
+	// Non-vacuity: the offload run must actually take the secondary-tier deferral path.
+	oc, ok := offload.sim.KVCache.(*kv.OffloadCache)
+	if !ok || oc.DeferralsStarted() == 0 {
+		t.Fatalf("secondary-tier reload path must be exercised (DeferralsStarted>0)")
+	}
+
+	if offload.Metrics().TTFTSum >= gpuOnly.Metrics().TTFTSum {
```
jgchn: This test is vacuous as a #1699 guard: it **passes on unmodified base**. I applied this file to `52007aa6` with only a `ReloadsPerformed()` shim and got:

```
=== RUN   TestInstanceSimulator_Offload_CPUHitReducesTTFT
--- FAIL: TestInstanceSimulator_Offload_CPUHitReducesTTFT
=== RUN   TestInstanceSimulator_Offload_SecondaryHitReducesTTFT
--- PASS: TestInstanceSimulator_Offload_SecondaryHitReducesTTFT
```

The CPU test is a genuine regression guard. This one compares offload-with-secondary-tier against GPU-only, which conflates the H3 deferral penalty with the prefill shrink, and that comparison already holds on `main`. So the PR's "guards the resolved-deferral branch" claim has no discriminating test behind it.

The branch is reached (shared `allocateThroughChain`), but weakly end to end: with a 20 MiB CPU tier plus one fs tier, PR vs base moves `ttft_mean` only 57.5399 -> 57.5252 ms (0.026%). The 52.55 -> 48.91 delta in the big-CPU + fs-tier config comes from CPU->GPU reloads, not secondary->CPU->GPU.

Suggest holding the offload config fixed and asserting on billed prefill work (or a per-request TTFT) for a request known to take the resolved-deferral admit path.

jgchn (commented):
Reviewed for correctness with code proof plus local testing (built PR and base `52007aa6`, ran the #1699 repro, invariant checks, and instrumented probes). The fix is correct and lands the bug: INV-6 byte-identical with offload off and with offload on but no hits; INV-13 byte-identical run vs replay on both CPU-only and secondary-tier configs; INV-1 conservation holds under 239 preemptions plus 576 timeouts; determinism holds; the `reloadedPrefixEnd` map is empty at end of run across 4 configs (no leak); `go test ./...` and `golangci-lint run ./...` clean; 4 of 5 new unit tests plus the CPU e2e test fail on base, so they are genuine regression guards. Two findings inline.

sim/batch_formation.go
susiejojo: Confirmed and agree with the analysis — thank you for the measurements. This ports only the single-chunk share of the vLLM term; a reload extending past the capped chunk leaves the remainder re-billed as recompute (Phase 1 consults no cache, and `recordReloadedPrefix` skips running continuations).

Addressed in f965420 (not fixing the behavior here, per your note — crediting to `newStart` needs the block commit extended past `endIndex`):
- Documented the one-chunk limit + the `chunk/prompt` realized-credit magnitude (~2% at the 2048 default for long-context offload) in the `batch_formation.go` comment and the CLAUDE.md entry.
- Fixed the misleading `TestFormBatch_ReloadBeyondCappedChunk` comment — it now says the remainder is re-billed as **recompute**, not "billed on a later step" (which read as billed correctly).
- Filed the follow-up: #1706.

sim/cluster/offload_deferral_e2e_test.go
susiejojo: You're right — that test compared offload-with-secondary vs GPU-only, which conflates the H3 deferral penalty with the prefill shrink and passes on base, so it wasn't a discriminating #1699 guard.

Replaced in f965420 with `TestDeferral_ResolvedAdmitReportsReloadedPrefix` (`sim/kv/offload_deferral_test.go`): it drives a cold-secondary request through the resolved-deferral admit via `roundsToAdmit`, then asserts `ReloadedPrefixEnd` reports the reloaded boundary (`newStart=4`) and is one-shot. That's the exact signal the fix adds — it holds the offload config fixed and isolates the prefill-shrink from the deferral penalty, and it does not exist on base. The `TestInstanceSimulator_Offload_CPUHitReducesTTFT` CPU e2e guard is unchanged.

jgchn:
## Correctness review

**Verdict: no correctness bug in the new code.** Two items to fix before merge, both about the fix's *reach*, not its logic.

### Verified

- **`billedTokens ∈ [0, numNewTokens]` always** (`sim/batch_formation.go:393`): the `:392` guard gives `newStart > startIndex`, so `newStart ≥ endIndex ⇒ 0`; `startIndex < newStart < endIndex ⇒ 0 < endIndex−newStart < numNewTokens`; `endIndex == startIndex ⇒ 0 == numNewTokens`. Never negative, never over-debits the budget clamped at `:343`.
- **`billedTokens == 0` cannot stall a request** (INV-8/INV-11): prefill progress comes from `ComputedTokens` — `sim/simulator.go:1082` assigns `ProgressIndex = reqNumComputedTokens[req.ID]` unconditionally.
- **The record can never be written unconsumed.** Writes only at `sim/kv/offload_chain.go:426`/`:443`, both on success, both gated to `!running && !IsDecodeSubRequest`. Exhaustive over callers: Phase 2 (`batch_formation.go:353`) always reads at `:392`; Phase-1 `preemptForTokens` (`:428`) has `running == true` (a Phase-2 admission always leaves a `RequestMap` entry — and the one path that leaves none, `startIndex == endIndex` with the fill loop at `sim/kv/cache.go:256` never executing, forces `ProgressIndex == InputLen`, so the `newStart > startIndex` guard then fails); `ReserveTransferredKV` (`sim/cluster/instance.go:463`) has `IsDecodeSubRequest` set at `sim/cluster/pd_events.go:143`. Probe: after a running continuation performing 2 real reloads, the map is empty and `ok == false`.
- **INV-6**: 5/6 configs byte-identical vs base (offload off; offload off + chunked prefill; `--kv-cpu-blocks 4000`; default synthetic; offload on w/ 1-block CPU tier). Only the intended case differs.
- **INV-13**: `run --trace-output` → `replay` with the offload config → byte-identical stdout.
- **INV-1/INV-4**: 6 offload×KV-pressure configs (up to 47 preemptions) conserve, 600/600 complete, no panics.
- **Non-vacuity**: the e2e test fails on base (`big TTFTSum=3178584 / 93 reloads` vs `small=3063929 / 0`) — on base the reload actually *raised* TTFT.
- `go test ./...` + `golangci-lint run ./sim/... ./cmd/...` clean.

### H1 (High) — the legacy `--kv-cpu-blocks` path still reproduces #1699

`sim/kv/tiered.go:222-231` (not in this diff) is the structural twin of the fixed code — `reloadPrefixFromCPU` → `newCached := GetCachedBlocks(...)` → `newStart` → `newStart > startIndex` / `>= endIndex` — and `TieredKVCache` doesn't implement `ReloadReportingKVStore`. Measured on **this PR's** binary, #1699's own workload:

| `--kv-cpu-blocks` | hit rate | ttft_mean_ms |
|---|---|---|
| 0 | 0.0772 | 52.55036 |
| 100 | 0.0772 | 52.55036 |
| 4000 | **0.0986** | **52.55633** |

Hit rate moves, TTFT doesn't (marginally *worse* — the legacy store charges transfer latency but never credits the prefill). `newStart` is already in hand at `tiered.go:225`, so it's ~4 lines; otherwise state in the PR body that `--kv-cpu-blocks` is knowingly left broken and file the follow-up.

### H2 (High) — inaccurate doc in the diff: `sim/kv_store.go:54-56`

> "…the single-tier and legacy-tiered stores do not, so batch formation type-asserts and the re-bill is inert (byte-identical, INV-6) **when offload is off**."

The legacy tiered store *is* an offload store: with `--kv-cpu-blocks > 0` offload is **on** and the re-bill is still inert (H1). Same wording at `sim/batch_formation.go:239-241` and in the `CLAUDE.md` addition.

### M1 (Medium) — the disclosed one-chunk cap leaves ≈0 benefit in the primary offload regime

20480-token prompts (16384 shared), 2600 GPU blocks, default `--max-num-batched-tokens 2048`:

| | offload off | offload 1 TiB |
|---|---|---|
| base ttft_mean_ms | 269.4693 | 289.3082 |
| PR ttft_mean_ms | 269.4693 | **269.7395** |

The PR removes base's 7.4% regression, but offload-on is still marginally *worse* than offload-off — for chunked long-context prompts (the main reason to offload KV) the user-visible symptom of #1699 persists. Accurately disclosed in `CLAUDE.md` + the comment at `sim/batch_formation.go:379-386` and tracked in #1706. Two asks:

1. Put the caveat in the **PR body** — its table says "TTFT now tracks CPU capacity" unqualified, and that text lands in the merge commit.
2. Reconsider #1706's proposed fix. Rather than "extend the block commit past `endIndex`", the vLLM-faithful shape is to apply the credit *before* the clamps at `sim/batch_formation.go:340-343` (vLLM computes `num_computed_tokens = local + external`, *then* applies `long_prefill_token_threshold`/budget). Today the chunk boundary is set by a budget the request then doesn't spend (`:343` clamps on `numNewTokens`, `:397` debits `billedTokens`), so a fully-cached prompt is still split into `⌈InputLen/budget⌉` steps with every chunk after the first billed as full recompute.

### M2 (Medium) — user-facing docs not updated

Only `CLAUDE.md` changed. `docs/guide/kv-cache.md:132` still reads "…instead of recomputing it — *so the cache hit-rate reflects the policy*", phrasing that was literally true only because of the bug. That guide is the `--kv-offload-config` reference.

### Low

- **L1 — no `num_new_tokens > 0` floor.** `sim/batch_formation.go:399` can set `NumNewTokens = 0` on an admitted request whose prefill completes that step (asserted by this PR's own `TestFormBatch_FullReloadZeroPrefill` / `…ReloadBeyondCappedChunk`). `sim/simulator.go:1056` then drops it from `scheduled`, and `StepTime(nil) == 1 tick` (measured, roofline) — so a solo fully-reloaded prefill costs 1µs. vLLM can't: `get_computed_blocks` caps the hit at `num_tokens − 1` and the waiting loop asserts `num_new_tokens > 0`. Pre-existing for a block-aligned 100%-GPU-cached prompt (`:338` has no floor either), but this PR widens reachability to any fully-reload-covered chunk. No near-zero TTFT observed in a realistic run (min 33.4ms / 600 requests), so: note, not blocker.
- **L2 — coverage gap.** Nothing tests a non-zero `startIndex` composed with a reload (the `FormBatch` fake returns `nil` from `GetCachedBlocks`; both `TestOffload_ReloadedPrefixEnd_ReportsBoundary` cases allocate from 0). I wrote it (warm 2 GPU blocks, block 2 on CPU, allocate `[4,8)`): `newStart=6`, `billed=2`, 4 blocks owned, INV-4 holds — a gap, not a bug. Worth adding: `newStart` being **absolute** rather than relative to `startIndex` is the one thing an implementer of this interface could plausibly get wrong.
- **L3 (nit)** `sim/kv/offload_chain.go:87` says "Written only for `!running` requests"; the gate at `:236` is `!running && !req.IsDecodeSubRequest` (the function doc at `:230-231` has both).

Nothing to report on: cross-path parity (run/replay share `FormBatch`; `observe` derives no KV state), preemption safety (`ReleaseKVBlocks` clears the record, and Phase 2 is skipped entirely in a preempting step via `!result.PreemptionHappened` at `:249`), timeout/deadline semantics, R2/R8, INV-9.

susiejojo:
Thanks @jgchn — thorough as always, and the local repro tables made these easy to action. All addressed in 568bb92 (code/tests/docs) plus PR-body and #1706 updates.

**H1 (legacy `--kv-cpu-blocks` still reproduces #1699) — fixed.** `TieredKVCache` now implements `ReloadReportingKVStore`: it records the post-reload boundary at its two commit-success points under the same `!running && !IsDecodeSubRequest` gate, clears it in `ReleaseKVBlocks`, and exposes a one-shot `ReloadedPrefixEnd`. Verified end-to-end on #1699's workload — matches your table's direction:

| `--kv-cpu-blocks` | hit rate | ttft_mean_ms (base) | ttft_mean_ms (fixed) |
|---|---|---|---|
| 4000 | 0.0986 | 52.55633 | **52.32122** |

INV-6 held: single-tier and no-reload `--kv-cpu-blocks` runs stay byte-identical to base; only a run that actually reloads changes, deterministically. Added `TestTieredKVCache_ReloadedPrefixEnd_ReportsBoundary`.

**H2 (inaccurate "when offload is off" doc) — fixed** in `kv_store.go`, `batch_formation.go`, and CLAUDE.md. Reworded to: the inert property is a store-**type** property — only the non-offload single-tier `KVCacheState` doesn't implement the interface; the legacy tiered store *is* an offload store and now implements it.

**M1 (one-chunk cap ≈ 0 benefit in the primary regime):**
1. Added the caveat to the **PR body** (Scope + a qualifier on the effect table's "tracks CPU capacity") so it lands in the merge commit, and to `docs/guide/kv-cache.md`.
2. Reconsidered #1706's fix shape per your suggestion — commented there: fold the external credit in **before** the `long_prefill_token_threshold`/budget clamps (vLLM order) rather than "extend the block commit," which should move the 2048-default case from ~62% realized credit toward ~100%.

**M2 (user-facing docs)** — `docs/guide/kv-cache.md:132` updated: the reloaded prefix is billed as a cache hit (lower prefill/TTFT), with the #1706 chunked-prefill caveat as an admonition.

**L2 (coverage gap)** — added a reload-composed-with-non-zero-`startIndex` test asserting `newStart` is **absolute** (=6), 4 blocks owned, INV-4 conservation. **L3** — field comment corrected to the full `!running && !IsDecodeSubRequest` gate.

**L1 (no `num_new_tokens > 0` floor)** — left as-is per your "note, not blocker": it's pre-existing (a block-aligned 100%-GPU-cached prompt already bills 0 at `:338` with no floor), matches vLLM's outcome via a different mechanism, and adding a floor here would risk the INV-6 byte-identity of the existing GPU-hit path. Noting it explicitly rather than silently widening scope.

susiejojo:
@claude /blis-pr-review check for code correctness thoroughly.

jgchn (approved):

Issue #1675, opened 2026-09-03
archon.yml triggers on prose mentions of /archon-pr-review, not just invocations


jgchn:
## Problem

`archon.yml` triggers on a substring match of the whole comment body:

```yaml
contains(github.event.comment.body, '/archon-pr-review') &&
!contains(github.event.comment.body, '/archon-pr-review-claude')
```

`contains()` does not distinguish an **invocation** from a **prose mention**. Any comment that merely discusses the command — a review that says "14 archon comments for 9 `/archon-pr-review` triggers", a doc quote, a postmortem — fires a full archon run.

## Evidence

On #1674 two comments *about* archon each fired a run ~4 s later. Neither was an invocation:

```
comment 17:31:58  mentions_archon_trigger=true  "> 🤖 **qa-review** — experimental two-age…"
comment 17:36:49  mentions_archon_trigger=true  "## Review — the two fixes are correct, b…"

run 33785123327  17:32:02  branch=main  event=issue_comment
run 33785598745  17:36:53  branch=main  event=issue_comment
```

Both produced **byte-identical** 5,548 B comments for the same head `e3598b84` (verified: bodies compare equal).

This also explains the ratio noted on #1667 — **14 archon comments for 9 actual triggers**. The extra runs are discussion, not double delivery.

Self-referentially, the comment reporting that archon fires twice *caused* archon to fire twice.

## Impact

- Wasted self-hosted runner time — a full `archon-go pr-review` per mention.
- Duplicate ~21 KB comments in the thread. #1674's `--edit-last` hides the bloat but not the runs.
- Any discussion of the trigger in a PR thread re-reviews that PR.

## Suggested fix

Require the command as a directive rather than a substring — e.g. match only when it appears on the comment's **first line**, or anchor with `startsWith`. Both preserve normal invocation while ignoring prose. `check-permissions` in `archon.yml:15-19` is the single place to change.

Note the same pattern is in `claude.yml:19-22` for `@claude`, though `@claude` in prose is likelier to be a genuine address.

## Acceptance

- A comment whose body mentions `/archon-pr-review` in prose (not as the leading directive) triggers no run.
- A normal `/archon-pr-review` invocation still triggers exactly one run.

Found while reviewing #1674.

---

## Scope note (added for delivery)

[...]

PR #1704, opened 2026-09-10
hardening(ci): run /blis-pr-review with a read-only token

### LINKED ISSUE #1697: hardening(ci): run /blis-pr-review with a read-only token

`/blis-pr-review` runs an AI agent whose job is to **judge** a pull request, but it runs with permission to **change** that pull request. The reviewer and the thing being reviewed are not separated.

On `main` (`2ebef6a`), `.github/workflows/claude.yml` grants the `claude` job:

- [`contents: write`](https://github.com/inference-sim/inference-sim/blob/2ebef6a17b95c3c41480f0f90529bb95cf9a0e5b/.github/workflows/claude.yml#L53-L54) — the token can push to branches
- [`--allowed-tools Skill Agent Bash Read Edit Write`](https://github.com/inference-sim/inference-sim/blob/2ebef6a17b95c3c41480f0f90529bb95cf9a0e5b/.github/workflows/claude.yml#L93) — file-editing tools

A review never needs either. This is a least-privilege / separation-of-duties gap: the component that decides whether code is acceptable should not be able to edit the code, and a mistake in a *review* should not be able to produce a *commit*.

This is not hypothetical — there has been at least one observed case of the review flow triggering Actions that made code changes to the PR under review. It was believed resolved; the permissions above show it was not.

Same reasoning as the existing rule that the delivery loop's verdict comes from `scripts/deliver-gate.sh` rather than from the reviewing agent: judgement should be structurally separated from the ability to act on it.

- [x] None (pure refactoring)

No simulator behaviour is affected. CI/workflow-only.

**Proposed fix**

The constraint that shapes this: `claude.yml` has **one** `claude` job serving *every* `@claude` trigger, and per `CLAUDE.md` non-review triggers may legitimately be asked to make changes. So the workflow cannot simply be downgraded wholesale — that would break normal `@claude` usage.

Split by command instead:

1. Add a review path selected when the comment contains `/blis-pr-review`, with a **read-only** token:
   - `contents: read`
   - keep `pull-requests: write` and `issues: write` — the agent must still post its review comment
   - drop `id-token: write` unless the action requires it for the model call
2. Leave the existing write-capable path for all other `@claude` usage, unchanged.
3. Optionally drop `Edit`/`Write` from `claude_args` on the review path.

**One thing to be clear about, so the change is not oversold:** `Bash` remains necessary for a review (`gh`, reading the diff, running the toolkit), and `Bash` can create files. So removing `Edit`/`Write` is intent-signalling, not a capability boundary. **The token scope is the actual control** — with `contents: read` the agent cannot push, whatever it writes locally in the ephemeral checkout.

**Open question for review:** does `/pr-review-toolkit:review-pr` (or any subagent it spawns) rely on `Write`/`Edit` for scratch or report files? If so, dropping those tools may break the review while `contents: read` alone would not. Recommend landing `contents: read` first and treating the tool list as a separate, verified step.

**Files affected**

- `.github/workflows/claude.yml` — job/permission split. Note `contents: write` appears **twice**: once as the [workflow-level default](https://github.com/inference-sim/inference-sim/blob/2ebef6a17b95c3c41480f0f90529bb95cf9a0e5b/.github/workflows/claude.yml#L3-L4) and again [on the job](https://github.com/inference-sim/inference-sim/blob/2ebef6a17b95c3c41480f0f90529bb95cf9a0e5b/.github/workflows/claude.yml#L53-L54); both need attention or the job-level grant will keep overriding.
- `docs/contributing/` — document that a review runs read-only, so the next person does not restore the write scope to fix an unrelated failure.

**Acceptance criteria**

- [ ] `/blis-pr-review` runs with `contents: read` and cannot push to any branch
- [ ] It can still post its review comment and set its status
- [ ] Non-review `@claude` usage is unaffected
- [ ] A review that attempts a push fails visibly rather than silently doing nothing
- [ ] Verified on a real PR before closing

**Design document**

Not needed — a permissions change with no architectural impact.

mtoslalibu:
A PR review is a judgement. Until now the agent that made that judgement also held the token needed to push to the branch it was judging: `claude.yml` had a single `claude` job serving every `@claude` trigger with `contents: write`. This splits the review out onto a read-only token, and leaves normal `@claude` usage alone.

Closes #1697.

## What changed

`.github/workflows/claude.yml`:

- `check-permissions` gains an `is_review` output — does the triggering comment/issue body contain `/blis-pr-review`. One definition of the test; the two job gates differ only in which literal they compare to.
- New **`claude-review`** job runs that path with `contents: read`, `pull-requests: write`, `issues: write`, `id-token: write`, and no `statuses: write`.
- The existing **`claude`** job is unchanged apart from its `if` — every other `@claude` trigger still gets `contents: write`, because per CLAUDE.md those may legitimately be asked to make a change.
- **Workflow-level** `contents` drops from `write` to `read`. The issue flagged this second occurrence, and it matters for a reason independent of today's jobs: a job-level block replaces the default wholesale, so the only thing this default can still do is hand push access to a future job that forgets its own block. The other three scopes are left as they were, so `check-permissions`' token is unchanged apart from `contents`.
- `report-status` now `needs: [claude, claude-review]` and reads whichever job ran.

Docs: `docs/contributing/standards/agent-trust.md` (new section — the tiers say which agent *outputs* to verify; this is the companion point that a judge should not hold the capability to act), plus the trigger docs in `docs/guide/skills-and-plugins.md` and `CLAUDE.md`.

## Why two jobs rather than one

`permissions:` accepts no expressions, so it cannot be varied per trigger inside one job. The two jobs' `steps` are byte-identical (asserted below) and carry a keep-in-sync comment; a reusable workflow would remove that duplication but would also move the PR-head `ref` resolution and the job outputs that `report-status` depends on — a worse trade against a workflow whose comments record several real false-green incidents. Happy to revisit if you'd rather have the indirection.

## On not overselling this

Per the issue's own warning, and stated in the workflow and in agent-trust.md:

- **The token is the control, not the tool list.** `Edit`/`Write` are deliberately *unchanged* on the review path. `Bash` is required for a review (`gh`, reading the diff, running the toolkit) and `Bash` can create files, so removing `Edit`/`Write` would signal intent without adding a boundary — while risking the review itself, since the toolkit's subagents write scratch and report files. What `contents: read` buys is that nothing written in the ephemeral checkout has a route to a branch, and a push fails with a 403 at the point of attempt instead of appearing to work.
- **The issue's open question is already answered empirically.** `deliver-verify.yml` has been running this same review, with this same tool list, under `contents: read` since the L1 delivery loop landed. So the review is known to work read-only, and the tool list did not need to be a separate verified step.
- **The guarantee covers the workflow token only.** These jobs run on a self-hosted runner; ambient credentials on that machine are out of scope, and agent-trust.md says so.

## Verification

- `actionlint` clean on the whole file (it also confirms `needs.claude-review.*` resolves and that `execution_file` is a real action output).
- `jobs.claude.steps == jobs['claude-review'].steps` — exactly `True`, checked by parsing the YAML.
- `mkdocs build --strict` passes; both new doc links resolve.
- No Go changes, so `go build` / `go test` / `golangci-lint` are untouched by this diff.
- Walked all reachable `(claude.result, claude-review.result)` pairs through the `report-status` gate and through `needs.claude.result != 'skipped' && needs.claude.result || needs.claude-review.result`: the skipped sibling matches no clause of the gate (and its outputs are null, which cannot satisfy `!= ''`), so exactly one correct status is published. Both-skipped yields `'skipped'`, which `stateMap` deliberately omits — a warning plus `error`, not a green.

## Review findings folded in

An adversarial pass plus a code review found five things, all fixed before this PR:

1. The non-review gate was first written `is_review != 'true'`, which would route an *unset* value to the write path. Both gates now compare to an explicit literal, so unset runs neither job.
2. Command matching is now case-insensitive. Both errors here are one-sided — over-matching costs a caller write access they can recover by re-commenting; under-matching runs a review that can push.
3. I had initially reduced the workflow-level default to `contents: read` alone. `check-permissions` is the one job with no permissions block, and if its collaborator probe 403'd under a narrower token the `catch` would report "not a collaborator" and *every* `@claude` trigger would silently no-op. Restored the other three scopes.
4. "Can post its review comment and its commit status" was wrong — the status comes from `report-status`. Corrected in all three docs, and the review job has no `statuses: write`, so it cannot set its own verdict.
5. Recorded the residual re-arming path: the reviewer keeps `pull-requests: write`, so it can post a comment, and a comment containing `@claude` is itself a trigger. Two things stop that reaching the write-capable job — the collaborator check, and the fact that `claude.yml` deliberately sets no `allowed_bots` (only the deliver-* workflows do, since those phases are dispatched bot-to-bot by design). agent-trust.md now names that as load-bearing.

Also removed the `conclusion` job output from the agent jobs rather than duplicating it: `actionlint` shows `claude-code-action@v1` defines no such step output, so it was always empty, and nothing in `.github/`, `docs/`, or `scripts/` reads it. Called out here since it is adjacent to, not required by, the issue.

## Acceptance criteria

- [x] `/blis-pr-review` runs with `contents: read` and cannot push to any branch
- [x] It can still post its review comment; the commit status is still set (by `report-status`)
- [x] Non-review `@claude` usage is unaffected — the `claude` job's permissions and steps are unchanged
- [x] A review that attempts a push fails visibly (403 at the point of attempt) rather than silently doing nothing
- [ ] **Verified on a real PR before closing** — needs a live `@claude /blis-pr-review` on this PR. That is the one criterion I cannot check locally: the routing, the read-only token, and the two-job `report-status` collapse only exercise on GitHub. Worth watching that the review comment posts and that exactly one "Claude Code Action (comment trigger)" status appears.

---

## Update after the `/blis-pr-review` pass (commit 05fa7c3d)

The review came back **ready to merge**, no CRITICAL or IMPORTANT findings, with three non-blocking items. All three are now addressed:

**1. No machine enforcement of `KEEP THE STEPS IN SYNC` — fixed.** This was the fair hit on the two-job design, and it was also the weakness I'd flagged myself when choosing duplication over a reusable workflow. `scripts/claude_workflow_test.go` now pins the arrangement, in the package whose own doc comment already says CI-only artifacts get pinned here because "nothing else can exercise them". Six invariants:

| Invariant | Mutation used to verify it fails |
|---|---|
| review job is `contents: read` | flipped to `write` |
| general path is still `contents: write` (stays a *split*) | covered by the same test |
| review job declares no `statuses: write` | added `statuses: write` |
| workflow-level `contents` is not `write` | flipped to `write` |
| both gates compare `is_review` to an explicit literal | rewrote one as `!= 'false'` |
| the two agent jobs' steps are identical | drifted the review job's `--model` |
| `report-status` needs both agent jobs | reduced to `needs: claude` |

Each mutation was applied, the suite run, and the file restored — every one failed in exactly the intended test and nothing else. Worth stating plainly since these assertions read as structural: for a workflow file the declared structure *is* the behaviour, because GitHub reads nothing else.

**2. `pull-requests: write` is not append-only — documented.** It permits `PATCH`/`DELETE` on existing comments, so the review *record* is mutable by the reviewer. Posting a review at all requires that scope, so this is inherent to the token model rather than something the split could have avoided; `agent-trust.md` now carries it as a third limit, with the consequence that the audit trail worth trusting is the run log, not the comment thread.

**3. Diff-line-comment reviews publish no commit status — documented.** Pre-existing. `report-status` gates on `issue_comment`, and `pull_request_review_comment` carries no `issue` object, so the second clause cannot hold. Now commented in place, including what extending it would actually take (reading the PR number from `github.event.pull_request`). I checked and corrected my own first attempt at that comment: I had initially written that `context.issue.number` is unavailable for that event, which is false — `@actions/github` falls back to `payload.pull_request`.

**Not fixed here: the SHA race → #1705.** `report-status` re-resolves the head SHA itself rather than using the one the agent checked out, so a push during a 60-minute run can attach a green verdict to a commit that was never reviewed. It's pre-existing on both agent jobs, so per the pr-workflow rule on pre-existing findings it's filed as its own issue rather than folded in. It's the right call on process grounds, but I'd flag it as the more consequential of the two findings — it weakens the review guarantee from the *other* direction than this PR closes, so it's worth prioritising.

### Verification (re-run after the second commit)

- `go test ./...` — exit 0, 14 packages ok, 0 failures (`scripts` 41.1s).
- `go build ./...` and `go vet ./scripts/` — clean.
- `actionlint` — clean.
- `mkdocs build --strict` — 0 errors/warnings.
- **`golangci-lint` caveat:** my local copy is 1.53.3 built with go1.20.5, too old to typecheck this tree — it reports 4 bogus `undefined:` errors on the *unmodified* tree, including `max` in the pre-existing `scripts/archon_review_test.go`. CI's pinned v2.9.0 is the authority; `go vet` is clean and `gofmt -l` is empty.

### Live verification status

The review run itself (`34524923783`) confirms the flow is unbroken, but note its job list: `check-permissions`, `claude`, `report-status` — **no `claude-review`**. That is expected and is the honest reading of the last acceptance criterion: `issue_comment` workflows always resolve from the default branch, so this run exercised `main`'s single-job `claude.yml`, not the split. The read-only routing can only be exercised after merge. First `/blis-pr-review` post-merge should show a `claude-review` job in place of `claude`, and one commit status — that's the check to watch.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -228,6 +228,12 @@ jobs:
     #
     # The test is applied to EACH agent job rather than to a pre-collapsed result, so the
     # skipped one cannot admit the job on its own: 'skipped' matches no clause.
+    #
+    # Scope, pre-existing and unchanged here: this fires for issue_comment ONLY, so a review
+    # invoked from a DIFF-LINE comment (pull_request_review_comment) runs and posts its
+    # review but publishes no commit status — that event carries no `issue` object, so the
+    # second clause cannot hold. Extending it means reading the PR number from
+    # github.event.pull_request instead, not just loosening the event test.
     if: |
       always() &&
       github.event_name == 'issue_comment' && github.event.issue.pull_request &&
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -47,6 +47,10 @@ jobs:
             // one-sided. Over-matching costs the caller write access they may have wanted
             // (re-comment without the command); under-matching runs a review with a token
             // that can push, which is the thing #1697 closes. So when in doubt, route here.
+            //
+            // Includes the `issues:` trigger, where an issue body quoting the command routes
+            // to the review job and checkout falls back to the default branch. Harmless —
+            // there is no PR to review — and still the safe side of the one-sided choice.
             const body = context.payload.comment?.body ?? context.payload.issue?.body ?? '';
             const isReview = body.toLowerCase().includes('/blis-pr-review');
             core.setOutput('is_review', isReview ? 'true' : 'false');
@@ -149,9 +153,13 @@ jobs:
   # granted). `Bash` is required for a review (gh, reading the diff, the review toolkit)
   # and `Bash` can create files anyway, so removing Edit/Write would signal intent
   # without adding a capability boundary — while risking the review itself, since the
-  # toolkit's subagents write scratch/report files. deliver-verify.yml already runs this
-  # same review with this same tool list under `contents: read`, which is the evidence
-  # that the token alone is sufficient and the tool list need not change.
+  # toolkit's subagents write scratch/report files.
+  #
+  # The evidence that the token alone suffices is deliver-verify.yml, which already runs this
+  # same review read-only in production: `contents: read` (its line 31), the identical
+  # `--allowed-tools Skill Agent Bash Read Edit Write` (line 419), invoking the blis-pr-review
+  # skill (line 448). Cited by line because an AI reviewer on #1704 asserted that workflow no
+  # longer invokes Claude at all — it does; verify before acting on a claim either way.
   claude-review:
     needs: check-permissions
     if: |
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -1,7 +1,13 @@
 name: Claude Code Action
 
+# Default for any job that declares no permissions block of its own (today just
+# check-permissions). `contents` is READ here, not write: a job-level block replaces this
+# one wholesale, so the only thing this default can do is hand push access to a future job
+# that forgets its own block — which is precisely how the read-only review path below would
+# be silently re-armed (#1697). The other three scopes are left as they were, so
+# check-permissions' token is unchanged apart from contents.
 permissions:
-  contents: write
+  contents: read
   pull-requests: write
   issues: write
   id-token: write
@@ -23,12 +29,27 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       allowed: ${{ steps.check.outputs.allowed }}
+      # Routes the trigger to ONE of the two mutually-exclusive agent jobs below.
+      # Computed here, once, so the substring test has a single definition — the two
+      # job-level `if`s then differ only in the sense of the comparison.
+      is_review: ${{ steps.check.outputs.is_review }}
     steps:
       - name: Check invoker permissions
         id: check
         uses: actions/github-script@v8
         with:
           script: |
+            // Before the try, so a bug in these two lines surfaces as a failed job rather
+            // than being absorbed by the catch below and misreported as "not a collaborator".
+            //
+            // The match is deliberately OVER-inclusive: case-insensitive, and satisfied by a
+            // body that merely quotes the command rather than invoking it. Both errors are
+            // one-sided. Over-matching costs the caller write access they may have wanted
+            // (re-comment without the command); under-matching runs a review with a token
+            // that can push, which is the thing #1697 closes. So when in doubt, route here.
+            const body = context.payload.comment?.body ?? context.payload.issue?.body ?? '';
+            const isReview = body.toLowerCase().includes('/blis-pr-review');
+            core.setOutput('is_review', isReview ? 'true' : 'false');
             try {
               const { data } = await github.rest.repos.getCollaboratorPermissionLevel({
                 owner: context.repo.owner,
@@ -45,9 +66,27 @@ jobs:
               core.info(`User ${context.actor} is not a collaborator — skipping.`);
             }
 
+  # Two agent jobs, identical steps, DIFFERENT tokens. They exist as two jobs only because
+  # `permissions:` accepts no expressions — it cannot be varied per trigger inside one job.
+  # report-status reads whichever one was not skipped.
+  #
+  # Both `if`s test is_review against an EXPLICIT value, so they are exclusive without being
+  # exhaustive: an unset is_review (which check-permissions cannot produce today — it is set
+  # before the permission probe that produces `allowed`) runs NEITHER job. That is the safe
+  # direction. Writing the non-review gate as `!= 'true'` would instead make an unset value
+  # route a review down the WRITE path, reintroducing exactly what #1697 closes.
+  #
+  #   claude        — every non-review @claude trigger. contents: write, since CLAUDE.md
+  #                   allows these to be asked for a code change.
+  #   claude-review — /blis-pr-review only. contents: read (#1697).
+  #
+  # KEEP THE STEPS IN SYNC. A fix applied to one belongs in the other; the only intended
+  # difference is the permissions block.
   claude:
     needs: check-permissions
-    if: needs.check-permissions.outputs.allowed == 'true'
+    if: |
+      needs.check-permissions.outputs.allowed == 'true' &&
+      needs.check-permissions.outputs.is_review == 'false'
     runs-on: [self-hosted]
     timeout-minutes: 60
     permissions:
@@ -56,12 +95,11 @@ jobs:
       issues: write
       id-token: write
     outputs:
-      # Both are EMPTY when the action returns early ("No trigger found, skipping remaining
-      # steps") — the outer trigger below is a broad substring test, so a comment that merely
+      # EMPTY when the action returns early ("No trigger found, skipping remaining steps")
+      # — the outer trigger below is a broad substring test, so a comment that merely
       # mentions @claude in prose admits the job and the action then no-ops. The job still
       # concludes 'success', so report-status must not read the job result alone (see #1675).
       execution_file: ${{ steps.claude.outputs.execution_file }}
-      conclusion: ${{ steps.claude.outputs.conclusion }}
 
     steps:
       - uses: actions/checkout@v5
@@ -102,8 +140,73 @@ jobs:
             pr-review-toolkit@claude-plugins-official
             superpowers@superpowers-marketplace
 
+  # The /blis-pr-review path: a reviewer must not be able to change what it reviews.
+  # `contents: read` is the ACTUAL control here — the token cannot push, so nothing the
+  # agent writes in the ephemeral checkout can reach a branch, and a `git push` fails
+  # loudly with a 403 at the point of attempt instead of appearing to work.
+  #
+  # The tool list is deliberately UNCHANGED from the write path (`Edit`/`Write` still
+  # granted). `Bash` is required for a review (gh, reading the diff, the review toolkit)
+  # and `Bash` can create files anyway, so removing Edit/Write would signal intent
+  # without adding a capability boundary — while risking the review itself, since the
+  # toolkit's subagents write scratch/report files. deliver-verify.yml already runs this
+  # same review with this same tool list under `contents: read`, which is the evidence
+  # that the token alone is sufficient and the tool list need not change.
+  claude-review:
+    needs: check-permissions
+    if: |
+      needs.check-permissions.outputs.allowed == 'true' &&
+      needs.check-permissions.outputs.is_review == 'true'
+    runs-on: [self-hosted]
+    timeout-minutes: 60
+    permissions:
+      contents: read          # cannot push — the whole point of this job (#1697)
+      pull-requests: write    # post the review comment
+      issues: write           # post on an issue-shaped trigger
+      id-token: write         # required by claude-code-action, as in every other workflow
+      # No statuses: write. The commit status is published by report-status below, not by
+      # the agent — so the review cannot set its own verdict either.
+    outputs:
+      # Same early-return semantics as the write path above — see that comment.
+      execution_file: ${{ steps.claude.outputs.execution_file }}
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          # Same PR-head ref resolution as the write path — see the comment there for why
+          # an explicit ref is required.
+          ref: >-
+            ${{ github.event_name == 'pull_request_review_comment'
+                && format('refs/pull/{0}/head', github.event.pull_request.number)
+             || (github.event.issue.pull_request
+                && format('refs/pull/{0}/head', github.event.issue.number))
+             || '' }}
+
+      - uses: anthropics/claude-code-action@v1
+        id: claude
+        env:
+          ANTHROPIC_BASE_URL: ${{ secrets.LITELLM_BASE_URL }}
+          CLAUDE_CODE_DISABLE_THINKING: "1"
+        with:
+          anthropic_api_key: ${{ secrets.LITELLM_API_KEY }}
+          claude_args: '--model claude-sonnet-4-6 --allowed-tools Skill Agent Bash Read Edit Write'
+          show_full_output: true
+          plugin_marketplaces: |
+            https://github.com/anthropics/claude-plugins-official.git
+            https://github.com/obra/superpowers-marketplace.git
+          plugins: |
+            code-review@claude-plugins-official
+            code-simplifier@claude-plugins-official
+            frontend-design@claude-plugins-official
+            pr-review-toolkit@claude-plugins-official
+            superpowers@superpowers-marketplace
+
   report-status:
-    needs: claude
+    # Both agent jobs, because either could be the one that ran. They are mutually
+    # exclusive, so the other is always 'skipped' with empty outputs — which is exactly
+    # what the "did work happen?" test below already knows how to reject.
+    needs: [claude, claude-review]
     # Publish a verdict ONLY when Claude actually ran. Two no-op paths otherwise corrupt the
     # status, and the job RESULT alone cannot tell them apart:
     #
@@ -122,11 +225,16 @@ jobs:
     # Stated positively so a new job result cannot silently fall through to a status. A real
     # success that somehow reported no execution_file publishes NOTHING rather than a verdict:
     # a missing status is visible and honest, a wrong one is neither.
+    #
+    # The test is applied to EACH agent job rather than to a pre-collapsed result, so the
+    # skipped one cannot admit the job on its own: 'skipped' matches no clause.
     if: |
       always() &&
       github.event_name == 'issue_comment' && github.event.issue.pull_request &&
       (needs.claude.result == 'failure' || needs.claude.result == 'cancelled' ||
-       (needs.claude.result == 'success' && needs.claude.outputs.execution_file != ''))
+       (needs.claude.result == 'success' && needs.claude.outputs.execution_file != '') ||
+       needs.claude-review.result == 'failure' || needs.claude-review.result == 'cancelled' ||
+       (needs.claude-review.result == 'success' && needs.claude-review.outputs.execution_file != ''))
     runs-on: ubuntu-latest
     permissions:
       statuses: write
@@ -153,7 +261,10 @@ jobs:
         if: steps.pr-info.outcome == 'success'
         uses: actions/github-script@v8
         env:
-          CLAUDE_RESULT: ${{ needs.claude.result }}
+          # Whichever agent job ran. They are mutually exclusive, so at most one is not
+          # 'skipped'; if both are (the gate above regressed) this yields 'skipped', which
+          # stateMap deliberately does not carry — see the comment on it.
+          CLAUDE_RESULT: ${{ needs.claude.result != 'skipped' && needs.claude.result || needs.claude-review.result }}
           PR_HEAD_SHA: ${{ steps.pr-info.outputs.pr_head_sha }}
         with:
           script: |
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -155,11 +155,24 @@ jobs:
   # without adding a capability boundary — while risking the review itself, since the
   # toolkit's subagents write scratch/report files.
   #
-  # The evidence that the token alone suffices is deliver-verify.yml, which already runs this
-  # same review read-only in production: `contents: read` (its line 31), the identical
+  # Partial evidence that the token alone suffices is deliver-verify.yml, which already runs
+  # this same review read-only in production: `contents: read` (its line 31), the identical
   # `--allowed-tools Skill Agent Bash Read Edit Write` (line 419), invoking the blis-pr-review
   # skill (line 448). Cited by line because an AI reviewer on #1704 asserted that workflow no
   # longer invokes Claude at all — it does; verify before acting on a claim either way.
+  #
+  # That precedent is only PARTIAL, and the gap is worth knowing before anyone "simplifies"
+  # this job. Every other read-only invocation in the repo passes `prompt:` (agent mode);
+  # this is the repo's only TAG-mode one, and tag mode is what runs the action's `setupBranch`
+  # — whose docs justify Contents write as "for reading repository files and creating
+  # branches". Reading the action source (recorded here so the next editor need not redo it):
+  # on an OPEN PR it fetches and checks out, returning no branch to create; on an issue or a
+  # closed/merged PR it does `repos.get` + `git.getRef` (both reads) and then creates the
+  # branch LOCALLY ONLY, because `use_commit_signing` defaults false and is not set here — the
+  # remote `createRef` path is unreachable. The single `git push` is gated on that branch also
+  # existing remotely, which a locally-created one never does. So no reachable path attempts a
+  # push or a ref write, including the `issues:` over-match above — which is what makes
+  # "harmless" there true rather than hopeful.
   claude-review:
     needs: check-permissions
     if: |
--- a/docs/contributing/standards/agent-trust.md
+++ b/docs/contributing/standards/agent-trust.md
@@ -79,6 +79,15 @@ Two limits worth stating, so nobody reads the guarantee as wider than it is:
   collaborator, and `claude.yml` deliberately does not set `allowed_bots` (the delivery
   workflows do, because their phases are dispatched bot-to-bot by design). Do not add
   `allowed_bots` to `claude.yml` without replacing that barrier.
+- `pull-requests: write` is also not append-only — it permits editing and deleting
+  existing comments, so the review *record* is mutable by the reviewer. Posting a review
+  at all requires that scope, so this is inherent to the token model rather than
+  something the split could have avoided. It is the reason the audit trail worth trusting
+  is the workflow run log, not the comment thread.
+
+The permission split is pinned by `scripts/claude_workflow_test.go`, which fails if the
+review job gains write access, if the workflow-level default returns to `contents: write`,
+if the routing gates stop failing closed, or if the two agent jobs' steps drift apart.
 
 ## Known Failure Modes
 
--- a/docs/contributing/standards/agent-trust.md
+++ b/docs/contributing/standards/agent-trust.md
@@ -42,6 +42,44 @@ orchestrator using different evidence than the agent's claim.
 Examples: "all construction sites updated," "0 CRITICAL issues," "review
 converged," "tests cover all contracts," "coverage is complete."
 
+## Structural Separation of Judgement and Action
+
+The tiers above say which agent outputs to verify. Separately, an agent whose job is
+to *judge* should not hold the capability to *act* on its own judgement — verification
+that can be skipped is weaker than a capability that was never granted.
+
+Two places implement this:
+
+- **The delivery loop's verdict** comes from `scripts/deliver-gate.sh`, not from the
+  reviewing agent.
+- **`/blis-pr-review` runs with a read-only token** (`contents: read`), so the reviewer's
+  `GITHUB_TOKEN` cannot push to the branch it is reviewing (#1697).
+  `.github/workflows/claude.yml` routes the `/blis-pr-review` command to a separate
+  `claude-review` job for this reason; every other `@claude` trigger keeps
+  `contents: write`, because those may legitimately be asked to make a change. The
+  review job also has no `statuses: write` — the commit status is published by a
+  separate `report-status` job, so the reviewer does not set its own verdict either.
+
+**If a review fails, do not restore `contents: write` to fix it.** The token scope is
+the control, and a review has no legitimate need to push. Note what the scope does and
+does not buy: `Bash` is required for a review (`gh`, reading the diff, running the
+toolkit) and `Bash` can create files, so the read-only *token* — not the tool list — is
+what makes the boundary real. Files written into the ephemeral checkout simply have no
+route to a branch, and an attempted push fails visibly with a 403 rather than appearing
+to succeed.
+
+Two limits worth stating, so nobody reads the guarantee as wider than it is:
+
+- It covers the **workflow token only**. These jobs run on a self-hosted runner, so
+  ambient credentials on that machine (a logged-in `gh`, a PAT in `~/.gitconfig`, an SSH
+  key) are outside it. Keep the runner free of push credentials.
+- The reviewer keeps `pull-requests: write`, so it can post a comment — and a comment
+  containing `@claude` is itself a trigger. Two independent things stop that from
+  reaching the write-capable job: `check-permissions` finds the bot is not a
+  collaborator, and `claude.yml` deliberately does not set `allowed_bots` (the delivery
+  workflows do, because their phases are dispatched bot-to-bot by design). Do not add
+  `allowed_bots` to `claude.yml` without replacing that barrier.
+
 ## Known Failure Modes
 
 Each failure mode below was discovered in a real PR. The tier system exists
--- a/docs/contributing/standards/agent-trust.md
+++ b/docs/contributing/standards/agent-trust.md
@@ -68,7 +68,7 @@ what makes the boundary real. Files written into the ephemeral checkout simply h
 route to a branch, and an attempted push fails visibly with a 403 rather than appearing
 to succeed.
 
-Two limits worth stating, so nobody reads the guarantee as wider than it is:
+Three limits worth stating, so nobody reads the guarantee as wider than it is:
 
 - It covers the **workflow token only**. These jobs run on a self-hosted runner, so
   ambient credentials on that machine (a logged-in `gh`, a PAT in `~/.gitconfig`, an SSH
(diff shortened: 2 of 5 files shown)


mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

namasl:

## qa-review — PR #1704: ⛔ BLOCK
_hardening(ci): run /blis-pr-review with a read-only token · 15 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ⚠️ CANNOT_ANSWER | Does this PR fully implement the issue it closes, without unnecessary additions? | The repository changes implement the intended permission split: review commands route to a job with `contents: read`, ordinary commands retain `contents: write`, and status publication remains in a separate job. The added tests and documentation are directly related rather than unnecessary. However, the acceptance criteria explicitly require verification on a real PR, and no repository-local evidence can establish that this external verification occurred. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. The canonical agent-trust standard documents the token boundary, routing split, status separation, and operational limitations. The agent-facing instructions and user-facing skills guide were also updated with consistent summaries and links back to the canonical standard. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | The operative workflow comments generally match the implementation. One evidentiary comment is stale: it says `deliver-verify.yml` already runs the same review with the same Claude tool list under `contents: read`, but that workflow currently performs an Archon review and contains no `claude-code-action` or `--allowed-tools` configuration. |
| G1 | correctness | ✅ CONFIDENT | In `check-permissions`, `body` is extracted from `context.payload.comment?.body ?? context.payload.issue?.body ?? ''`. If a `/blis-pr-review` command is submitted in a PR description or a PR review body (`context.payload.review?.body`), `is_review` will evaluate to `false`. Could this cause a review command to be mistakenly routed to the write-capable `claude` job? | No supported trigger exposes either of those bodies. The workflow subscribes only to `issue_comment`, `pull_request_review_comment`, and `issues`; the first two put the command in `payload.comment.body`, while the issues trigger uses `payload.issue.body`. It does not subscribe to `pull_request` or `pull_request_review`, so a command merely placed in a PR description or submitted-review body does not start this workflow and therefore cannot be routed to the write job. |
| G2 | correctness | ✅ CONFIDENT | In `claude-review`, `ref` is evaluated using `github.event_name == 'pull_request_review_comment'` or `github.event.issue.pull_request`. If `/blis-pr-review` is invoked on an event type where neither condition matches, `ref` evaluates to `''`. What commit does `actions/checkout@v5` fetch when `ref` is empty, and could it cause the agent to review the `main` branch instead of the PR head? | With an empty `ref`, checkout uses the event's default ref. For the configured `issues` event this is the repository default branch. For every configured PR-comment path, however, the expression is nonempty: `pull_request_review_comment` uses `github.event.pull_request.number`, and a PR `issue_comment` uses `github.event.issue.number`. Thus a supported PR review command checks out `refs/pull/N/head`; only a plain issue falls back to the default branch. |
| G3 | rules | ✅ CONFIDENT | In `scripts/claude_workflow_test.go`, `loadClaudeWorkflow` unmarshals YAML into a struct using `yaml.Unmarshal`. Does this unmarshaler comply with project YAML strictness requirements (such as enforcing known fields with `Decoder.KnownFields(true)`), or could unmapped keys in `.github/workflows/claude.yml` silently pass without validation? | The test intentionally does not enforce known fields. It uses `yaml.Unmarshal` into a partial struct, so fields outside that struct are ignored and unmapped keys can pass this particular test. No YAML strictness requirement appears in the project standards, and this helper explicitly states that it is a targeted structural assertion rather than a general workflow-schema validator. |
| G4 | rules | ✅ CONFIDENT | In `scripts/claude_workflow_test.go`, `wf.Jobs[reviewJob].Permissions["contents"]` directly accesses the `Permissions` map. If a future editor omits the `permissions:` block entirely from the `claude-review` job, will this trigger a nil-map access or panic rather than returning a structured assertion error? | It will not panic. Reading from a nil Go map is safe and returns the value type's zero value. If the job-level `permissions:` block is omitted, `Permissions["contents"]` returns `""`, which fails the comparison against `"read"` and produces the intended structured `t.Errorf`. |
| G5 | tests | ✅ CONFIDENT | In `TestClaudeWorkflow_AgentJobStepsStayInSync`, step nodes are decoded into `any` and compared via `reflect.DeepEqual`. Will `reflect.DeepEqual` catch subtle step discrepancies if unmarshaled YAML map keys or scalar types deserialize into different concrete Go types between the two jobs? | Yes. Both nodes are decoded by the same yaml.v3 implementation into `any`, and `reflect.DeepEqual` compares concrete types as well as nested keys and values. A scalar-type discrepancy, missing key, extra key, or changed value therefore makes the decoded structures unequal; a decode failure is reported fatally before comparison. |
| G6 | tests | ✅ CONFIDENT | Does `TestClaudeWorkflow_RoutingGatesFailClosed` verify that the GitHub Actions expressions in `claude.yml` actually evaluate to `false` when `is_review` is missing or set to an invalid value, or does it only verify the presence of literal substring patterns in the workflow source code? | It only performs source-level substring checks. It verifies that each gate contains the expected explicit comparison and does not contain `is_review !=`, but it does not parse or execute GitHub Actions expressions with missing or invalid values. The current expressions themselves compare against exact strings, so an empty or unexpected value matches neither job. |
| G7 | rationale | ✅ CONFIDENT | Why split `claude.yml` into two separate jobs (`claude` and `claude-review`) with duplicated steps rather than invoking a reusable workflow or job matrix, and what design choice prevents step configuration from drifting out of sync over time? | The split exists because job `permissions` cannot be selected dynamically from an expression, while review and non-review invocations need different `contents` scopes. The implementation uses two mutually exclusive jobs with identical steps and different permission blocks. Drift is prevented by `TestClaudeWorkflow_AgentJobStepsStayInSync`, which decodes and deeply compares every step in the two jobs; a length or effective-configuration difference fails the test. |
| G8 | rationale | ✅ CONFIDENT | Why are `--allowed-tools Skill Agent Bash Read Edit Write` still granted to the `claude-review` job in `claude_args` if the review is intended to be read-only, and what prevents a local workspace file edit from altering review outputs prior to `report-status` execution? | `Edit` and `Write` remain because removing them would not create a capability boundary while `Bash` can still create or modify files, and the toolkit may need scratch files. The security control is the review job's `contents: read` token, which prevents those local changes from being pushed through that token. Nothing prevents a local edit from influencing the agent's own review reasoning or comment, but `report-status` does not consume workspace files: it independently resolves the PR head SHA and derives status only from the selected job result. |
| G9 | scope | ✅ CONFIDENT | The PR changes the workflow-level default `permissions.contents` from `write` to `read` in `.github/workflows/claude.yml`. Does this change impact any other jobs in `claude.yml` (such as `check-permissions`), and was modifying the workflow default required beyond splitting the job permissions? | The only current job affected is `check-permissions`, because all three other jobs declare their own job-level permissions. Its `contents` permission changes from write to read while its other inherited scopes remain unchanged. Changing the default was not necessary merely to make `claude-review` read-only, but it was required by the issue's defense-in-depth requirement to address both existing `contents: write` declarations and prevent a future job without its own block from inheriting push access. |
| G10 | scope | ✅ CONFIDENT | The PR adds documentation in `docs/contributing/standards/agent-trust.md` regarding ambient runner credentials on self-hosted runners. Does documenting these self-hosted runner boundaries add operational constraints beyond issue #1697, and are those runner constraints verified elsewhere? | The documentation adds an operational constraint—self-hosted runners must not expose ambient push credentials—but presents it as a limitation of the workflow-token guarantee, not as a capability implemented by this PR. The workflow verifies only declared GitHub token permissions; no repository test or workflow shown audits the runner's credential store. Documenting that boundary is appropriate because ambient PAT, SSH, or logged-in `gh` credentials could bypass the intended token restriction. |
| G11 | errors | ✅ CONFIDENT | In `report-status`, `CLAUDE_RESULT` is evaluated as `${{ needs.claude.result != 'skipped' && needs.claude.result \|\| needs.claude-review.result }}`. How does GitHub Actions expression evaluation behave if `needs.claude.result` is `'skipped'` while `needs.claude-review.result` is `'failure'`, and does the ternary evaluation short-circuit correctly? | It evaluates to `failure`. GitHub Actions uses short-circuit `&&`/`\|\|` behavior for this ternary idiom: when `needs.claude.result != 'skipped'` is false, the `&&` side yields a falsy value and the `\|\|` expression selects `needs.claude-review.result`. The report gate separately admits review-job failures, so that value reaches the status mapping and becomes a failure commit status. |
| G12 | errors | ✅ CONFIDENT | In `check-permissions`, if `github.rest.repos.getCollaboratorPermissionLevel` fails with a network error or rate limit, the job will fail. How does the downstream `report-status` job handle the case where `needs.check-permissions.result` is `'failure'` rather than `'success'`? | The premise is not true for failures thrown by `getCollaboratorPermissionLevel`: the broad `catch` converts every such error, including a network or rate-limit error, into `allowed=false`, logs it as not being a collaborator, and leaves `check-permissions` successful. Both agent jobs are then skipped and `report-status` is also skipped because neither agent result satisfies its positive condition. If `check-permissions` itself genuinely fails outside that catch, the dependent agent jobs are skipped and the same report condition remains false, so no commit status is published. |

### Items to fix
- **F1 · CANNOT_ANSWER** — The repository changes implement the intended permission split: review commands route to a job with `contents: read`, ordinary commands retain `contents: write`, and status publication remains in a separate job. The added tests and documentation are directly related rather than unnecessary. However, the acceptance criteria explicitly require verification on a real PR, and no repository-local evidence can establish that this external verification occurred.
  _.github/workflows/claude.yml:85-96,155-168,205-246; scripts/claude_workflow_test.go:65-183; linked acceptance criterion: "Verified on a real PR before closing"_

### Important to consider
- **F3** — The stale comparison does not invalidate the token boundary, but it should be corrected so future maintainers do not rely on nonexistent precedent.
  _.github/workflows/claude.yml:148-154; .github/workflows/deliver-verify.yml:278-309; repository search for `claude-code-action\|allowed-tools` in `.github/workflows/deliver-verify.yml` returns no matches_
- **G1** — If a `pull_request` or `pull_request_review` trigger is added later, body extraction and checkout routing must be extended before enabling it.
  _.github/workflows/claude.yml:15-28,50-52_
- **G2** — A `/blis-pr-review` string on a plain issue can enter the read-only job with the default branch checked out, although the command is documented for PRs.
  _.github/workflows/claude.yml:15-21,108-122,174-184_
- **G3** — A separate workflow-schema validator would be needed to catch arbitrary GitHub Actions key typos.
  _scripts/claude_workflow_test.go:26-34,50-55; no `KnownFields`, YAML strictness, or equivalent requirement appears under docs/contributing/standards/_
- **G5** — Comments and purely stylistic YAML differences are intentionally ignored because they do not alter effective step configuration.
  _scripts/claude_workflow_test.go:125-147_
- **G6** — The test would not detect every logically unsafe expression containing the expected substring, such as an added unconditional `true \|\| ...` clause.
  _.github/workflows/claude.yml:87-89,157-159; scripts/claude_workflow_test.go:100-121_
- **G7** — A reusable workflow could also reduce duplication, but the selected design is valid and explicitly tested.
  _.github/workflows/claude.yml:69-84,85-96,155-168; scripts/claude_workflow_test.go:123-148_
- **G8** — The workflow's claim that `deliver-verify.yml` demonstrates this exact tool configuration is stale; that workflow no longer invokes Claude Code.
  _.github/workflows/claude.yml:143-166,186-203,249-301; docs/contributing/standards/agent-trust.md:63-69_
- **G10** — Runner credential hygiene remains an external operational assumption rather than a repository-enforced invariant.
  _docs/contributing/standards/agent-trust.md:71-86; .github/workflows/claude.yml:160; scripts/claude_workflow_test.go:65-98_
- **G12** — Treating transient API failures as permission denial is a pre-existing fail-closed behavior, but it can silently suppress a requested review and misstate the reason.
  _.github/workflows/claude.yml:53-67,85-89,155-159,205-243_

mtoslalibu:
Thanks — went through the qa-review pass. Summary in plain terms.

**Its one blocking item can't be fixed here.** F1 says it cannot confirm the "verified on a real PR" criterion. That's correct and expected: comment-triggered workflows always run from the default branch, so the new job can't run until this is merged. No code change closes that.

**Its one substantive finding is wrong.** F3/G8 claims `deliver-verify.yml` no longer invokes Claude, which would make my "already runs read-only in production" evidence stale. It does invoke Claude:

- `deliver-verify.yml:402` — `uses: anthropics/claude-code-action@v1`
- `:419` — `--allowed-tools Skill Agent Bash Read Edit Write` (the identical list)
- `:448` — invokes the `blis-pr-review` skill
- `:31` — `contents: read`

So the precedent holds: this exact review, with this exact tool list, has been running read-only in production. I've changed the comment to cite those line numbers, and to note the claim was disputed once, so the next reader can check it in ten seconds instead of taking either of us at our word.

**One observation was real and is fixed.** G6 correctly spotted that my routing test only checked that each gate *contains* an explicit comparison — so someone could widen a gate to `... == 'true' || true` and the test would still pass. It now also rejects `||` in either agent gate: a routing gate is a conjunction, and a disjunct can only add a way in. Verified by mutation (injecting `|| true` fails the test).

**One is pre-existing and now filed.** G12: `check-permissions` catches every error from the collaborator probe and reports it as "not a collaborator", so a rate limit or network blip silently drops a request with a log line that states a false reason. Fail-closed is right; the silence and the misattribution aren't. Filed as #1707.

**The rest are acknowledgements, not defects** — and its answers agree with the design on each: the token (not the tool list) is the control (G8), the two-job split is required because `permissions:` takes no expressions (G7), a nil permissions map fails cleanly rather than panicking (G4), `reflect.DeepEqual` does catch step drift (G5), and the workflow-level default only affects `check-permissions` (G9). I added one comment for G2: an issue body quoting the command routes to the review job with the default branch checked out — harmless, no PR to review, and the safe side of a deliberately one-sided match.

Pushed as `951748b7`. Gate re-run: `go test ./...` exit 0 (14 packages, 0 failures), `actionlint` clean, `go vet` clean, `mkdocs --strict` 0 errors.

Standing state: two pre-existing follow-ups (#1705 SHA race, #1707 silent catch), neither introduced here. The only open item on this PR is the post-merge live check — first `/blis-pr-review` after merge should show a `claude-review` job instead of `claude`, and one commit status.

scripts/claude_workflow_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, name := range []string{writeJob, reviewJob, "report-status"} {
+		if _, ok := wf.Jobs[name]; !ok {
+			t.Fatalf("job %q missing from %s — the permission split was restructured; "+
+				"re-derive these assertions rather than deleting them", name, path)
+		}
+	}
+	return wf
+}
+
+// The security property itself: the review path cannot push, the general path still can.
```
jgchn, line 74: **IMPORTANT — this pins the tokens but not the classifier.** What's pinned is that *the job named `claude-review`* can't push, not that *a review runs there*. The split has three moving parts (two gates, two permission blocks, one classifier) and the test covers two.

Two mutations I ran, both **NOT CAUGHT**, all five tests green:

| Mutation | Effect |
|---|---|
| `includes('/blis-pr-review')` → `includes('/blis-pr-reviewX')` | every review routes to `contents: write` |
| `isReview ? 'true' : 'false'` → `isReview ? 'false' : 'true'` | inverted — reviews get write, everything else read-only |

Either silently restores exactly what #1697 closes. Same class as the unenforced `KEEP THE STEPS IN SYNC` comment that motivated this file. Two assertions on `wf.Jobs["check-permissions"].Steps[0]`'s script close it: it contains `/blis-pr-review`, and it contains `isReview ? 'true' : 'false'`.

.github/workflows/claude.yml
```diff
@@ -102,8 +144,77 @@ jobs:
...
+  # agent writes in the ephemeral checkout can reach a branch, and a `git push` fails
+  # loudly with a 403 at the point of attempt instead of appearing to work.
+  #
+  # The tool list is deliberately UNCHANGED from the write path (`Edit`/`Write` still
+  # granted). `Bash` is required for a review (gh, reading the diff, the review toolkit)
+  # and `Bash` can create files anyway, so removing Edit/Write would signal intent
+  # without adding a capability boundary — while risking the review itself, since the
+  # toolkit's subagents write scratch/report files.
+  #
+  # The evidence that the token alone suffices is deliver-verify.yml, which already runs this
```
jgchn: This citation is accurate on its own terms (I checked line 31, 419, 448) but it's the wrong precedent for the risk that matters: **every** read-only `claude-code-action` invocation in this repo passes `prompt:` (agent mode). `claude.yml` is the only *tag*-mode one, and tag mode is what runs `setupBranch` — the action's docs say Contents write is "For reading repository files **and creating branches**."

I read the action source, and the conclusion holds — worth recording here so the next editor doesn't redo it:

- Open PR → `git fetch`/`git checkout`, returns **no** `claudeBranch`. Local only.
- Issue / closed-or-merged PR → only `repos.get` + `git.getRef` (both read), then *"For non-signing case, create and checkout the branch **locally only**"*. `use_commit_signing` defaults `"false"` and isn't set here, so `github-file-ops-server.ts`'s `createRef` is unreachable.
- The one `git push` (`branch-cleanup.ts:103`) is gated on `claudeBranch` being set **and existing remotely** — it returns early otherwise, and a locally-created branch never does.

So no push or ref-write is attempted on any reachable path, including the `issues:` over-match, which makes "Harmless" above load-bearing and true.

docs/contributing/standards/agent-trust.md
```diff
@@ -42,6 +42,53 @@ orchestrator using different evidence than the agent's claim.
...
+
+**If a review fails, do not restore `contents: write` to fix it.** The token scope is
+the control, and a review has no legitimate need to push. Note what the scope does and
+does not buy: `Bash` is required for a review (`gh`, reading the diff, running the
+toolkit) and `Bash` can create files, so the read-only *token* — not the tool list — is
+what makes the boundary real. Files written into the ephemeral checkout simply have no
+route to a branch, and an attempted push fails visibly with a 403 rather than appearing
+to succeed.
+
+Two limits worth stating, so nobody reads the guarantee as wider than it is:
```
jgchn: Nit: "Two limits" is followed by three bullets — the review-round commit added the append-only one without updating the lead-in. Either say "Three", or fold it into the preceding bullet, which is about the same scope.

jgchn (commented):
Reviewed for correctness. **Verdict: correct — ship it.** The security property holds; routing and the status collapse are sound; no functional break. Verification notes and one IMPORTANT coverage gap inline.

What I proved rather than took on trust:

- **All 9 reachable `(claude.result, claude-review.result)` pairs** walked through the `report-status` gate and the `!= 'skipped' && … || …` collapse: `'skipped'` matches no clause, `null != ''` is false under GitHub's numeric coercion (and is double-guarded by `result == 'success'`), and the #1675 false-green fix survives (success + empty `execution_file` → no status). Both-skipped can only arise from a gate regression, where `stateMap` correctly warns and publishes `error`.
- **Hyphenated `needs.claude-review.result` parses** — `archon.yml` and `deliver-implement.yml` already use `needs.check-permissions.outputs.*` in production.
- **The workflow-level `contents` reduction is safe for `check-permissions`**, and there is direct local evidence the PR body doesn't cite: `archon.yml` already runs the byte-identical `getCollaboratorPermissionLevel` probe in a `check-permissions` job with no job-level block, inheriting workflow-level `contents: read`. Finding #3's failure mode cannot occur.
- **The removed `conclusion` output has no consumer** — the only `conclusion` hits in `.github/`, `docs/`, `scripts/` are `gh api` JSON fields in `deliver-verify.yml`.
- **`blis-pr-review` doesn't need write**: `SKILL.md:130` already says "Do NOT run `git commit` or `git push`". The token now enforces what the skill asked for.
- **Mutation-tested the new suite**: all 7 claimed mutations fail in exactly the intended test and nothing else. `go build ./...`, `go test ./scripts/` (24.5s), `go vet`, `gofmt -l` clean.

Agreed on deferring the SHA race to #1705, and agreed it's the more consequential of the two.

Nit not worth its own thread: the steps-sync test compares steps only, so job-level drift (`runs-on: [self-hosted]` → `ubuntu-latest` on the review job, `timeout-minutes`, `outputs`) is not caught — verified. Not a security regression, but inside the test's stated purpose.

scripts/claude_workflow_test.go
mtoslalibu: You're right, and this was the actual gap — thank you. The suite pinned which token each job holds and said nothing about which job a review lands in.

Reproduced both of your mutations against the old suite (green, as you said), then fixed and re-ran:

- `includes('/blis-pr-review')` → `includes('/blis-pr-reviewX')` — now fails `TestClaudeWorkflow_ClassifierRoutesReviewsToTheReviewJob`
- `isReview ? 'true' : 'false'` → `isReview ? 'false' : 'true'` — now fails the same test
- also added: deleting the classifier entirely (`const isReview = false`) fails it, via a helper that fatals when no step sets `is_review`

Used exactly your two assertions. `787c9d4f`.

.github/workflows/claude.yml
mtoslalibu: This is the more useful half of the review and I'd have missed it. You're right that tag-vs-agent mode is the distinction that matters, not the tool list — deliver-verify.yml is agent mode (`prompt:`), so it doesn't exercise `setupBranch` at all.

I've downgraded the comment from evidence to *partial* precedent and recorded your source analysis inline so it isn't re-derived: open PR → fetch/checkout, no branch returned; issue or closed/merged PR → `repos.get` + `git.getRef` then local-only branch creation, with the remote `createRef` unreachable because `use_commit_signing` defaults false and isn't set here; the single `git push` gated on the branch existing remotely, which a local one never does.

That's also what upgrades the `issues:` over-match from 'harmless' to demonstrably harmless. `787c9d4f`.

docs/contributing/standards/agent-trust.md
mtoslalibu: Fixed — now 'Three limits'. Introduced exactly as you describe: the append-only bullet was added in the review round without updating the lead-in. `787c9d4f`.

mtoslalibu:
@jgchn thanks — this found the one thing that mattered. All three threads addressed and resolved in `787c9d4f`.

**The IMPORTANT finding was right, and I'd got the test's scope wrong.** I pinned which *token* each job holds and thought that was the property. It isn't: what matters is that a review *lands* in the read-only job, and that depends on the classifier, which nothing tested. I reproduced both your mutations against the old suite — green, exactly as you said — and either one silently hands every review a `contents: write` token. That's the same failure the file was written to prevent, one layer up.

Fixed with your two assertions, and I added a third case: deleting the classifier outright now fails too, because the helper fatals when no step sets `is_review`.

**Extended the parity test past steps,** which was your review-body nit and also inside the test's stated purpose. It now compares `runs-on`, `timeout-minutes` and `outputs` as well — moving the review onto a different runner or a shorter timeout is real drift that touches no step. Mutation-verified both. Renamed to `AgentJobsStayInSync` since it's no longer steps-only.

**Recorded your action-source analysis in the workflow rather than just accepting it.** You're right that the tool-list precedent was the wrong argument: every other read-only invocation here is agent mode, so none of them exercises `setupBranch`, and this is the repo's only tag-mode one. I downgraded the comment from "evidence" to *partial* precedent and wrote your trace inline — local-only branch creation, `use_commit_signing` defaulting false making the remote `createRef` unreachable, and the lone `git push` gated on a remote branch a local one never becomes. That's what moves the `issues:` over-match from "harmless" to demonstrably harmless, so it's worth having in the file.

Also fixed "Two limits" → "Three".

Net effect of your review: the suite went from 5 tests pinning 6 invariants to 6 tests pinning 11, and it now covers all three moving parts of the split rather than two. Gate re-run: `go test ./...` exit 0 (14 packages, 0 failures), `actionlint`, `go vet`, `gofmt`, `mkdocs --strict` all clean.

Nothing else outstanding on my side. Two pre-existing follow-ups stand (#1705 SHA race, #1707 silent catch on the collaborator probe), and the last open item remains the post-merge live check — first `/blis-pr-review` after merge should show a `claude-review` job instead of `claude`, and one commit status.

scripts/claude_workflow_test.go
```diff
@@ -0,0 +1,271 @@
...
+// ternary routes every review to the write-capable job and all the other tests stay green,
+// silently restoring what #1697 closed.
+func TestClaudeWorkflow_ClassifierRoutesReviewsToTheReviewJob(t *testing.T) {
+	wf := loadClaudeWorkflow(t)
+
+	script := classifierScript(t, wf)
+
+	// The command the routing actually keys on. A typo here sends every review to the
+	// write path.
+	if !strings.Contains(script, "includes('/blis-pr-review')") {
```
jgchn, line 153: Optional, non-blocking — two residual mutations, both strictly narrower than the ones this now catches, both still **NOT CAUGHT**:

| Mutation | Effect |
|---|---|
| `body.toLowerCase().includes(…)` → `body.includes(…)` | `/BLIS-PR-Review` routes to the write path |
| `context.payload.comment?.body ?? context.payload.issue?.body` → `context.payload.issue?.body` alone | on `issue_comment`, `issue.body` is the **PR** body, not the comment — so a `/blis-pr-review` comment on a PR whose description lacks the string routes to write |

The first is a regression of this PR's own review-round fix #2, so it's a fix that is now unpinned. The second is the more consequential: it's the common case (the command is normally in a comment, not the PR body), and it fails toward write.

Two more `strings.Contains` lines here would close both — `"body.toLowerCase()"` and `"context.payload.comment?.body ??"`. Fine to leave for later; the two highest-value literals are pinned and the split is in much better shape than when I first read it.

jgchn (commented):
Re-reviewed `787c9d4f`. **All three findings fixed and verified. Approve.**

Re-ran the mutation suite against the new tests — the two that previously slipped through are now caught, and the six earlier ones still are:

| Mutation | Before | Now |
|---|---|---|
| `includes('/blis-pr-review')` → `…ReviewX'` | not caught | `ClassifierRoutesReviewsToTheReviewJob` |
| `isReview ? 'true' : 'false'` inverted | not caught | `ClassifierRoutesReviewsToTheReviewJob` |
| classifier step deleted entirely | — | `ClassifierRoutesReviewsToTheReviewJob` (helper's `t.Fatal`) |
| review `runs-on` → `ubuntu-latest` | not caught | `AgentJobsStayInSync` |
| review `timeout-minutes` 60 → 10 | — | `AgentJobsStayInSync` |
| review `outputs` key renamed | — | `AgentJobsStayInSync` |
| M1 / M4 / M5 / M7 (tokens, default, gate, needs) | caught | still caught |

Extracting `claudeJob` was the right call — `Outputs map[string]string` pins the output *expressions*, so a job that stops exporting `execution_file` now fails the parity test rather than silently costing that path its commit status. `classifierScript`'s `t.Fatal` is a good touch: a deleted classifier is the one regression that makes both gates compare against a value nothing produces.

`gofmt -l` empty, `go vet ./scripts/` clean, `go build ./...` clean, `go test ./scripts/` ok (24.1s). No stale references to the renamed `AgentJobStepsStayInSync`. "Three limits" now matches three bullets. The `claude.yml:158-178` comment records the tag-mode/agent-mode gap accurately.

One optional follow-up inline — strictly narrower than what you just closed, non-blocking.

jgchn (approved):

PR #1719, opened 2026-09-15
hardening(invariants): anchor the declared invariants in code (Closes #1718)

### LINKED ISSUE #1718: hardening: anchor the 23 declared invariants in code (18 comments, 1 test, 1 doc line)
## Summary (read this part)

We checked all **23 declared invariants** in `docs/contributing/standards/invariants.md` against the code on `main` (`07622594`), asking one question: *for each invariant we promise, does anything in the code or tests actually reference it?*

Mostly yes — but unevenly, and the unevenness is invisible today.

| Status | Count | Meaning |
|---|---|---|
| LINKED | 19 | referenced in non-test code |
| TEST ONLY | 3 | referenced only in test files |
| UNLINKED | 1 | referenced nowhere in code at all |

The sharpest contrast: **`INV-6` (Determinism) is referenced in 69 non-test files.** It cannot be broken quietly — any PR near it lights up. **`INV-PD-2` (Pool Exclusivity) is referenced in zero.** The string `INV-PD-2` appears exactly once in the whole repo: in the document declaring it.

Same doc, same apparent status, completely different real protection.

Also: **10 of 23 have no test naming or referencing them**, including `INV-3` (Clock Monotonicity), which 12 files reference and which is about as fundamental as it gets in a discrete-event simulator.

**This is not a bug report.** We are not claiming any invariant is violated. It is an observability gap: when a PR touches code enforcing an invariant, nothing tells the reviewer an invariant is at stake unless someone wrote its ID in a nearby comment. Where that habit was followed (`INV-6`, `INV-1`, `INV-13`) it works extremely well. Where it wasn't, the invariant is protected by memory alone. It also blocks tooling — a promise linked to its evidence by name can be watched automatically ("this PR changed 4 of the 69 files claiming INV-6, and deleted the only test referencing it"); a promise linked by prose in a doc cannot.

**The proposal: ~18 one-line comments, one new test, one doc sentence.**

No renames. No behaviour changes. No refactoring. Every declared invariant ends up with at least one anchor in code and one in tests. The checklist below has exact file and line for every edit, so this is a single self-contained PR.

We considered renaming tests to carry invariant IDs (`TestINVPD5_PoolStability`). **Dropped deliberately** — a comment creates the same link, and renames would churn git blame, break local test filters, and force matching edits in `invariants.md`, which cites seven of these tests by name. Suggest we simply adopt the ID convention for *new* tests going forward and not retrofit.

Two things we are explicitly leaving out of this PR so it stays boring — both flagged at the bottom for separate discussion.

---

Declared invariants with no reference anywhere in code. An invariant that lives only in `docs/contributing/standards/invariants.md` cannot be surfaced during review and cannot be tracked by tooling. Same observability class as R1 (silent failure): nothing is wrong now, but a future regression would be invisible.

Method: for each invariant ID in the doc, search non-test Go under `sim/` and `cmd/` for the ID, search test files separately, and look for tests whose names contain it. Pure reference search, no inference. All counts are against `07622594` and reproducible with `grep`.

- [ ] Request conservation
- [x] Request lifecycle
- [x] Clock monotonicity
- [x] Signal freshness
- [ ] KV block conservation
- [ ] Causality
- [ ] Determinism
- [ ] Work-conserving
- [x] None (pure refactoring) — *no behaviour change; the boxes above mark invariants whose **anchoring** is weak, not invariants that are violated*

Also affected, not on the standard checklist: `INV-2`, `INV-11`, `INV-12`, `INV-PD-1`, `INV-PD-2`, `INV-PD-4`, `INV-PD-5`, `INV-PD-6`, `INV-PD-6b`.

**Proposed fix**

### Part 1 — anchor comments at enforcement sites (7 lines)

| # | File:line | Current code | Add reference to |
|---|---|---|---|
| 1 | `sim/cluster/pool.go:143` | `func FilterSnapshotsByPool(...)` | **`INV-PD-2`** |
| 2 | `sim/cluster/cluster.go:1270` | `func (c *ClusterSimulator) PoolMembership()` | **`INV-PD-5`** |
| 3 | `sim/cluster/cluster.go:1392` | `parent.CompletionTime = c.clock + inst.PostDecodeFixedOverhead()` | **`INV-PD-6b`** |
| 4 | `sim/batch_formation.go:456` | `preemptedRequest.State = StateQueued` | **`INV-2`** (running→queued) |
| 5 | `sim/event.go:171` | `e.Request.State = StateTimedOut` | **`INV-2`** |
| 6 | `sim/simulator.go:1139` | `req.State = StateCompleted` | **`INV-2`** |
| 7 | `sim/simulator.go:1208` | `req.State = StateCompleted` | **`INV-2`** |

Notes on the three that matter most:

- **#1 is the highest-value line in this PR.** `FilterSnapshotsByPool` is the single choke point for pool exclusivity — all six call sites funnel through it (`pd_events.go:28`, `cluster_event.go:405`, `cluster.go:2252`, `cluster.go:2317`, via `buildPoolFilteredSnapshots` at `cluster.go:1307`). One comment takes `INV-PD-2` from invisible to anchored.
- **#2** already returns a *copy* of the membership map (cited as `R8`). That copy **is** how `INV-PD-5` holds — the comment just says so.
- **#3** is literally the invariant as a line of code. Worth also noting: `cluster.go:1423` sets `parent.CompletionTime = c.clock` with no overhead. Assuming that is the non-decoded path, one clause explaining why would stop a future reader "fixing" it.

`INV-2` has **15 state-assignment sites across 7 files**; the four above are the lifecycle-meaningful edges. We are not proposing to annotate all 15.

### Part 2 — anchor comments above the verifying tests (11 lines)

The doc already names these tests in prose. A comment makes the link machine-readable. Suggested form: `// Verifies INV-PD-5 (pool membership fixed at construction).`

| # | File:line | Test | Invariant |
|---|---|---|---|
| 8 | `sim/cluster/disaggregation_test.go:141` | `TestDisaggregation_PrefillRoutedToPrefillPool` | `INV-PD-2` |
| 9 | `sim/cluster/disaggregation_test.go:164` | `TestDisaggregation_DecodeRoutedToDecodePool` | `INV-PD-2` |
| 10 | `sim/cluster/disaggregation_test.go:188` | `TestDisaggregation_RequestCompletesFullPath` | `INV-PD-1` |
| 11 | `sim/cluster/disaggregation_test.go:411` | `TestDisaggregation_PhaseCausality` | `INV-PD-4` |
| 12 | `sim/cluster/disaggregation_test.go:445` | `TestDisaggregation_PoolStability` | `INV-PD-5` |
| 13 | `sim/cluster/disaggregation_test.go:876` | `TestDisaggregation_MetricProjection_NoSubRequestKeys` | `INV-PD-6` |
| 14 | `sim/cluster/disaggregation_test.go:1648` | `TestDisaggregation_CompletionTime_IncludesNonZeroOverhead` | `INV-PD-6b` |
| 15 | `sim/batch_formation_test.go:1063` | `TestPreemption_Priority_Phase1Completeness` | `INV-12` |
| 16 | `sim/workload/session_test.go:535` | `TestSession_TimeoutCancels_NoMoreRounds` | `INV-11` |
| 17 | `sim/workload/session_test.go:903` | `TestSession_FinalRound_Completes` | `INV-11` |
| 18 | `sim/simulator_test.go:1256` | `TestWorkConserving_StepRestartsWhenWaitQNonEmpty` | `INV-8` |

### Part 3 — one new test: `INV-3` Clock Monotonicity

`INV-3` is referenced in 12 non-test files and **zero tests**. The doc says it holds "by construction" via min-heap extraction (`sim/simulator.go:314-326`, `sim/cluster/cluster.go:851-852`) — sound reasoning, but nothing checks it.

Add to `sim/simulator_test.go`:

```
TestINV3_ClockNeverDecreases — run a simulation, assert every processed
event's timestamp is >= the previous one.
```

This is the only genuine coverage gap in the list. Everything else in this PR is annotation.

### Part 4 — one sentence in the invariant doc

`INV-3` currently states, unconditionally: *"Simulation clock never decreases."*

There is one site where the cluster clock deliberately moves backwards, `sim/cluster/cluster.go:875-878`:

```go
// If ProcessNextEvent() skipped a cancelled TimeoutEvent (lazy
// cancellation — inst.Clock was not advanced), restore c.clock.
// A no-op orphaned timeout must not advance the cluster clock.
if te, ok := ev.(*sim.TimeoutEvent); ok && te.Request.State == sim.StateCompleted {
    c.clock = prevClusterClock
}
```

The clock is advanced optimistically to the instance's next event time *before* the event is known to be real. When it turns out to be an orphaned timeout, nothing happened, so the advance is undone. This reads as correct and the comment explains it well — the *statement* is what does not match the code.

Proposed: amend the INV-3 statement with a carve-out, e.g. *"…except when restoring an optimistic advance after a lazily-cancelled event, where no event was processed."* This also keeps Part 3's new test honest, since a naive assertion on `c.clock` would trip on this site.

**Not a suspected bug.** Purely a wording fix so statement and code agree.

**Files affected**

| File | Change |
|---|---|
| `sim/cluster/pool.go` | 1 comment |
| `sim/cluster/cluster.go` | 2 comments (+1 optional clarifying `:1423`) |
| `sim/batch_formation.go` | 1 comment |
| `sim/event.go` | 1 comment |
| `sim/simulator.go` | 2 comments |
| `sim/cluster/disaggregation_test.go` | 7 comments |
| `sim/batch_formation_test.go` | 1 comment |
| `sim/workload/session_test.go` | 2 comments |
| `sim/simulator_test.go` | 1 comment + 1 new test |
| `docs/contributing/standards/invariants.md` | 1 sentence (INV-3 carve-out) |

Multiple construction sites to flag: `INV-2` has 15 state-assignment sites across 7 files — we annotate 4. `INV-PD-2` is the opposite: 6 call sites all funnelling through one function, so one comment covers it.

**Design document**

Not needed. Comments, one test, one doc sentence. No architectural change.

---

### Deliberately out of scope — for discussion, not this PR

**1. Should `INV-PD-2` cover the encode pool?** The statement mentions prefill and decode only, but `FilterSnapshotsByPool` also filters `PoolRoleEncode` (`cluster.go:2317`). Either the statement widens to include encode, or it should say encode is out of scope. This changes what the invariant *means*, so it needs a team decision rather than a drive-by edit.

**2. A `validRequestTransitions` table for `INV-2`.** Request state is assigned in 15 places with no legal-transition table — even though the codebase already has exactly that pattern for instances (`validInstanceTransitions`, `sim/cluster/instance.go:416`). Adding one, plus a helper that rejects illegal transitions, would turn `INV-2` from prose into something directly testable using an established in-repo pattern. It is the only item that adds real code, so it belongs in its own issue if the team wants it.

**3. `INV-7` has no Go test at all** — the doc credits the H3 and H29 hypothesis experiments, which live on the `hypothesis-archive` branch. That may be entirely intentional (some properties are validated by experiment, not unit test), but it is worth stating explicitly in the doc so the absence reads as a decision rather than an oversight.

---

<sub>Analysis produced with archon's invariant link-map tooling. Every count is a plain reference search over `sim/` and `cmd/` at `07622594`, re-derivable with `grep`. Two measurement bugs were found and fixed while producing it — matching `INV-2` by its bare number matched every test containing "E2E", and BSD `grep -E` silently fails on negative lookahead — so if a similar search gives different numbers, those are the likely causes.</sub>

mtoslalibu:
## Summary

BLIS declares 23 invariants in `docs/contributing/standards/invariants.md`. Some are protected extremely well — `INV-6` (determinism) is named in 69 non-test files, so any PR near it lights up. Others are protected by memory alone: `INV-PD-2` (pool exclusivity) appeared exactly once in the entire repo, in the document declaring it. Same doc, same apparent status, completely different real protection.

This PR gives every declared invariant at least one named reference in code and one in tests, so a reviewer touching an enforcement site can see which promise is at stake without already knowing. Nothing here claims an invariant is violated — it is an observability fix.

- **17 anchor comments** — 9 at enforcement sites, 8 above the tests that verify them. Every added line in the six production `.go` files is a comment; there are no non-comment additions and no deletions. (Review added the 9th: an INV-2 anchor on `InstanceSimulator.EvictRequest`, whose `StateCompleted` write is a tombstone rather than a completion.)
- **One new test**, `TestINV3_ClockNeverDecreases`. `INV-3` was documented as true "by construction" via min-heap extraction. It was *not* wholly untested — `TestSimulator_ClockMonotonicity_NeverDecreases` (BC-6) predates this branch and drives the loop asserting monotonicity — but it does so on the raw `sim.Clock` **field** and never names `INV-3`. The gap this fills is that narrower one: no test named the invariant as its subject, and none used the processed-event-timestamp formulation the statement is written in, which is the only formulation that survives the cluster carve-out below.
- **One amended invariant statement** — `INV-3` said unconditionally that the clock never decreases, which its own code contradicts.

## The one subtle thing

`sim/cluster/cluster.go` deliberately moves the cluster clock **backwards**: it advances optimistically to an instance's next event time before the event is known to be real, then restores the previous value when the event turns out to be a lazily-cancelled orphaned timeout. Nothing happened, so the advance is undone. That code is correct; the *statement* was what didn't match.

So the new test asserts on **processed event timestamps, never on a raw clock field**, and skips orphans using exactly the predicate production uses. A naive `c.clock` assertion would fail at that site and look like a real bug. `INV-3`'s statement now carries the carve-out, and its two working copies (`CLAUDE.md`, `docs/concepts/core-engine.md`) are updated per the documentation single-source-of-truth rule in `principles.md`.

The test is non-vacuous by construction: it gates on >= 10 processed events, a final timestamp > 0, and at least one orphaned timeout actually occurring (so the skip branch can't silently stop being exercised if timing constants drift). A mutation check confirms the ordering assertion fires.

## Corrections to the issue

- The issue lists 11 test anchors; **3 were already anchored** on `602ae462` (`INV-PD-6`, `INV-12`, `INV-8` are already named above their tests). Adding a second anchor there is churn, so Part 2 lands 8, not 11 — 16 comments total, not 18.
- The issue says `INV-3` is referenced in "zero tests". It is referenced in four test files, and `TestSimulator_ClockMonotonicity_NeverDecreases` (BC-6) does take clock monotonicity as its subject — on the raw `sim.Clock` field, without naming `INV-3`. So the gap is real but narrower than the issue or an earlier draft of this description implied: no INV-3-named test, and no test using the processed-timestamp formulation. The count isn't right either.
- `sim/batch_formation.go:456` was cited for `preemptedRequest.State = StateQueued`; that assignment is at line 490 on current `main`. Every other cited line matched exactly.
- The issue scopes the doc edit to `invariants.md`, but `principles.md` requires updating working copies too — hence 12 files, not 10.

## Known gaps left open, on purpose

Code review surfaced three doc/code disagreements of the same species this PR fixes for `INV-3`. Each is documented in place rather than fixed, because resolving it changes what an invariant *means* and that's a team call, not a drive-by edit:

- **`INV-PD-6b`'s predicate is too broad.** It scopes on `DecodeInstanceID != ""`, which timed-out parents satisfy, so the statement reads as covering a line that correctly omits the post-decode overhead. Flagged in a comment at that line.
- **`INV-2` doesn't enumerate the edges its anchors describe.** Its statement is `queued -> running -> completed`; the code also has running -> queued (preemption) and -> timed-out. The anchors now say plainly that the statement doesn't yet name these, rather than implying it does. Widening `INV-2` is the issue's own out-of-scope item 2.
- **The cluster clock restore has no test.** The new test covers the single-instance skip; the `prevClusterClock` restore itself remains unverified.

The issue's three "deliberately out of scope" items — encode-pool scope of `INV-PD-2`, a `validRequestTransitions` table for `INV-2`, and `INV-7`'s missing Go test — are **not** implemented. They await team discussion.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` clean
- [x] `go test ./...` — full suite passes
- [x] `go test ./sim/ -run TestINV3_ClockNeverDecreases -v` passes
- [x] Mutation check: tightening the comparison to `<=` trips the assertion, so the ordering check is live rather than vacuous
- [x] No behaviour change verified mechanically — every added line in the production files is a comment (`git diff 602ae462 | grep "^+" | grep -v "^+\s*//"` is empty for all six)
- [x] INV-6 (determinism) unaffected: comments only, no output path touched
- [ ] CI green


mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

docs/contributing/standards/invariants.md
```diff
@@ -32,9 +32,11 @@ Invariants are properties that must hold at all times during and after simulatio
 
 ## INV-3: Clock Monotonicity
 
-**Statement:** Simulation clock never decreases. Every event's timestamp >= the previous event's timestamp.
+**Statement:** Simulation clock never decreases. Every **processed** event's timestamp >= the previous processed event's timestamp — except when restoring an optimistic advance after a lazily-cancelled event, where no event was processed.
 
-**Verification:** Clock is advanced in the event loop only via min-heap extraction, which guarantees non-decreasing order.
+**Verification:** Clock is advanced in the event loop only via min-heap extraction, which guarantees non-decreasing order. `sim/simulator_test.go` — `TestINV3_ClockNeverDecreases` drives the loop one event at a time and asserts the processed-timestamp sequence is non-decreasing (with a non-vacuity gate, and at least one orphaned timeout present so the carve-out below is exercised).
```
jgchn: **Over-claims the new test.** This says the test has "at least one orphaned timeout present so **the carve-out below is exercised**" — but the carve-out below is the *cluster* `prevClusterClock` restore, which the test's own comment (`simulator_test.go:1350`) says it does **not** cover.

I confirmed by mutation: deleting the orphan-skip branch entirely leaves the test passing, so it does not verify the carve-out at all. The orphans exercise the single-instance skip.

Suggest: "…and at least one orphaned timeout present so the single-instance lazy-cancellation skip is exercised (the cluster restore below is untested — see Known gaps).".

docs/contributing/standards/invariants.md
```diff
@@ -32,9 +32,11 @@ Invariants are properties that must hold at all times during and after simulatio
...
+
+**The carve-out, and why it is not a violation:** `sim/cluster/cluster.go` advances the cluster clock to an instance's next event time *before* the event is known to be real, then restores the previous value when `ProcessNextEvent()` turns out to have skipped a lazily-cancelled `TimeoutEvent` (`prevClusterClock`). A no-op orphaned timeout must not advance the cluster clock, so the advance is undone. Nothing was processed across that pair of assignments, which is why the statement is scoped to *processed* events. This also means a test asserting on the raw `c.clock` field would fail here and look like a real bug — assert on processed event timestamps instead.
```
jgchn: **One clause missing from an otherwise complete account.** `cluster.go:860` reads the *optimistically advanced* clock in the horizon guard and can `break` before the restore at `:877` — so the advance can persist unrestored.

I checked and it's benign: cluster `SimEndedTime` comes from per-instance `Finalize` (`min(Clock, Horizon)`), progress snapshots clamp at `cluster.go:1106`, and `ClusterSimulator.Clock()` has no production caller. Worth half a sentence since this paragraph is now the canonical account.

sim/cluster/cluster.go
```diff
@@ -1420,6 +1427,17 @@ func (c *ClusterSimulator) detectDecodeCompletions(inst *InstanceSimulator) {
...
 		parent := c.parentRequests[c.pendingDecodeCompletions[subReqID]]
+		// No PostDecodeFixedOverhead here, unlike the INV-PD-6b line above: a
+		// timed-out parent never finished decoding, so it never pays the
+		// post-decode overhead.
+		//
+		// Note for anyone reconciling this against the doc: INV-PD-6b's statement
+		// predicates on `DecodeInstanceID != ""`, which these timed-out parents do
+		// satisfy (detectDecodeCompletions reaches them by decode-instance match),
+		// so the statement reads as covering this line while the code correctly
+		// omits the overhead. The code is right and the statement is too broad;
```
jgchn: **"the statement is too broad" isn't accurate.** `invariants.md:262` reads: *"For all **successfully decoded** parent requests (`DecodeInstanceID != ""`)…"*. The comment quotes the parenthetical and drops the governing qualifier. A parent whose decode sub-request is `StateTimedOut` is by definition not *successfully decoded*, so the statement already excludes this line.

The risk is that this invites a future contributor to "narrow" an already-correctly-scoped invariant, with a rationale that doesn't hold.

The other two claims are correct and should stay: these parents *do* satisfy `DecodeInstanceID != ""` (guaranteed by the instance-match `continue` at `:1379`), and omitting the overhead here is right. Suggest reframing as: the parenthetical gloss is loose, the statement is not.

sim/simulator.go
```diff
@@ -1205,6 +1207,9 @@ func (sim *Simulator) processCompletions(now, currStepAdvance int64) []*Request
 				rm.LengthCapped = true
 				sim.Metrics.Requests[req.ID] = rm
 			}
+			// INV-2 (request lifecycle): the terminal running -> completed edge on
+			// the length-capped path. A force-completed request is still completed,
+			// not a distinct terminal state — the LengthCapped flag records why.
```
jgchn, line 1212: **Scope limit worth a note.** "A force-completed request is still completed, not a distinct terminal state" is true for length-capping, but a reader who generalizes it to `StateCompleted` as a whole will be wrong.

`instance.go:507` and `:523` also write `StateCompleted` on gateway eviction — `:523` from the running batch — and there it is a **tombstone, not a completion**: `EvictRequest`'s doc says it exists "to prevent dangling TimeoutEvents from double-counting", and the request is counted as `gatewayEvicted` (`cluster_event.go:315`), never via the sole `CompletedRequests++` at `simulator.go:723`.

Both anchors here are honestly path-qualified, so nothing is false — but the one *semantically anomalous* `StateCompleted` write is the one this pass left unmarked, and it's where a note pays off most.

sim/simulator_test.go
```diff
@@ -1328,6 +1328,113 @@ func TestWorkConserving_StepRestartsWhenWaitQNonEmpty(t *testing.T) {
...
+//
+// WHEN the event loop is driven one event at a time
+// THEN every event the loop actually PROCESSES has a timestamp >= its predecessor's
+// AND the run is non-vacuous: many events processed and the clock advanced past 0.
+//
+// The doc calls INV-3 true "by construction" via min-heap extraction; nothing checked
+// it. This asserts on PROCESSED EVENT TIMESTAMPS rather than on a clock field, which
+// is the formulation INV-3 is stated in and the only one that generalizes to the
+// cluster loop. Scope, stated precisely so nobody over-reads this test: within package
+// sim, sim.Clock is assigned in exactly one place (from the heap pop in
```
jgchn: Nit: "assigned in exactly one place" is literally false — there is also the constructor initializer `Clock: 0` at `simulator.go:196`. Harmless (it can't make the clock decrease), but "assigned exactly once during the run" would be accurate, and this comment's whole job is precision.

jgchn (commented):
**Verdict: correct — merge after the two doc fixes below.**

Verified locally: stdout is **byte-identical to `602ae462`** across three configs (single-instance; PD 2p/2d + `--detectors all`; flow-control + KV pressure), so all five touched production files are provably inert. `golangci-lint` clean. The new test is genuinely load-bearing: injecting a real bug (`StepEvent{time: now - 1}` in `scheduleNextStep`) makes it fail, and it is the **only** test in `./sim` that catches it. Tightening `<` to `<=` also trips it, so the assertion is live. The INV-3 carve-out is sound — I traced it and nothing can read the optimistically-advanced clock before the restore.

Everything below is comment/doc accuracy *inside the anchoring*, not behaviour. Findings 1 and 2 are worth fixing first because they are the same doc-vs-code mismatch species this PR exists to fix.

One item I can't inline (outside the diff): **`CLAUDE.md:714`** still says "**13 system invariants** (INV-1 through INV-13)". That contradicts CLAUDE.md's own list ~220 lines earlier (which includes INV-BC-DP1) and `CONTRIBUTING.md:329` (which adds "plus INV-PD-* and INV-P2-*"). Cheap fix while this file is already open.

Also noted, no action needed: **INV-7 still has 0 test anchors** (8 production files, 0 test files), so the "every invariant anchored in code *and* tests" goal lands at 22/23. Deliberate per the PR body — but `invariants.md`'s INV-7 Verification section still reads as an unexplained gap rather than a decision.

susiejojo:
### BLIS PR Review — independent verification

**Verdict: correct and low-risk — merge after the doc-accuracy fixes below.** This is a pure observability pass (16 anchor comments, one new test, one amended invariant statement). I re-ran the checks against `main` (`5bd02102`) and the PR head rather than trusting the body.

#### Archon
No `/archon-pr-review` output in prior comments. Proceeding without an archon verdict. (This PR moves no module boundary, so a FAST-TRACK is the expected shape.)

#### What I verified mechanically
- **Comments-only production change (INV-6 safe).** For all five production files (`sim/batch_formation.go`, `sim/cluster/cluster.go`, `sim/cluster/pool.go`, `sim/event.go`, `sim/simulator.go`), `git diff main..head` adds only `//` lines — zero non-comment additions, zero deletions. So the byte-identical-output claim holds by construction; cross-path parity (run/replay/observe), preemption safety, and timeout/deadline consistency are all **N/A** — no execution path is touched.
- **Build + vet clean** on the PR head; `go test ./sim/ -run TestINV3_ClockNeverDecreases -v` **passes**.
- **The new test is load-bearing for what it claims** (single-instance ordering): it asserts on `ev.Timestamp()`, not on a raw clock field, which is the only formulation that generalizes to the cluster loop. `ProcessNextEvent` assigns `sim.Clock = ev.Timestamp()` at exactly one run-time site (`simulator.go:326`), and an orphaned `TimeoutEvent` returns *before* that assignment — so skipping those events in the test faithfully mirrors production.

#### The human review by @jgchn is accurate — I independently confirmed each finding
1. **`invariants.md:37` over-claims the test.** The statement says the orphan timeout is present "so **the carve-out below is exercised**," but the carve-out is the *cluster* `prevClusterClock` restore, which the test's own comment says it does NOT cover. **Mutation-confirmed:** neutralizing the orphan-skip `continue` in the test leaves it green — the orphaned timeout's timestamp is monotonic regardless, so the skip branch does not affect the assertion. The orphans exercise the single-instance skip only. This should be reworded. *(This is the same doc-vs-code mismatch species the PR exists to fix, so it's worth fixing here.)*
2. **`sim/cluster/cluster.go:1438` comment is factually wrong.** It claims INV-PD-6b's statement "is too broad" because it predicates on `DecodeInstanceID != ""`. But `invariants.md:260` reads *"For all **successfully decoded** parent requests (`DecodeInstanceID != ""`)…"* — the governing qualifier "successfully decoded" already excludes a `StateTimedOut` parent. The comment quotes the parenthetical and drops the qualifier, then invites a future contributor to "narrow" an already-correctly-scoped invariant. The two *other* claims in that comment (these parents do satisfy `DecodeInstanceID != ""`; omitting the overhead is correct) are right and should stay. Reframe as "the parenthetical gloss is loose; the statement is not."
3. **`invariants.md:39` carve-out omits one clause.** `cluster.go:860`'s horizon guard reads the optimistically-advanced clock and can `break` before the restore at `:877`, so the advance can persist unrestored. I confirmed it's benign (`SimEndedTime` comes from per-instance `Finalize` as `min(Clock,Horizon)`; `ClusterSimulator.Clock()` has no production caller), but since this paragraph is now the canonical account, a half-sentence is warranted.
4. **`sim/simulator.go:1212` scope note.** "A force-completed request is still completed, not a distinct terminal state" is true for length-capping, but `instance.go:507`/`:523` also write `StateCompleted` as a **tombstone** on gateway eviction (counted as `gatewayEvicted`, never via the sole `CompletedRequests++`). Confirmed both writes exist. The anomalous `StateCompleted` write is the one this pass left unmarked — worth a note.
5. **`sim/simulator_test.go:1345` nit.** "assigned in exactly one place" is literally false — `simulator.go:196` has the `Clock: 0` constructor initializer. Harmless, but "assigned exactly once during the run" is accurate, and this comment's job is precision.

#### Additional items (out-of-diff, cheap while these files are open)
- **`CLAUDE.md:714`** still says "**13 system invariants** (INV-1 through INV-13)", contradicting CLAUDE.md's own INV-BC-DP1 entry and `CONTRIBUTING.md:329` ("plus INV-PD-* and INV-P2-*"). Confirmed present on `main`.
- **INV-7 lands at 22/23**, not "every invariant anchored in code and tests": 8 production references, **0 test anchors** (confirmed). The PR body flags this as deliberate, but `invariants.md`'s INV-7 Verification section still reads as an unexplained gap rather than a decision. A one-line note there would close the loop.

#### Relevance to the issue (#1718)
Faithful. The body honestly reconciles its deviations from the issue's checklist (3 anchors already existed → 16 comments not 18; `INV-3` referenced in 4 test files incidentally not "zero"; the `:456`→`:490` line drift; `principles.md` forcing the working-copy doc edits → 12 files). The three "deliberately out of scope" items are correctly left unimplemented.

**None of the findings block on correctness** — the code is provably inert. They are all comment/doc-accuracy fixes *inside the anchoring*, and #1 and #2 matter most because they are exactly the mismatch class this PR is meant to eliminate. Recommend fixing #1, #2 (and ideally the `CLAUDE.md:714` count) before merge.

mtoslalibu:
Thanks both — @jgchn and @susiejojo independently landed on the same five findings, and all five hold up. I verified each against the code before changing anything rather than taking the reports on faith, and every one was a real doc-vs-code mismatch *inside the anchoring* — the exact species this PR exists to eliminate. Fixed in `12aa102f`, plus the two adjacent items.

**1. `invariants.md:37` over-claimed the test.** Confirmed. The sentence said the orphaned timeout is present "so the carve-out below is exercised", but the carve-out below is the *cluster* `prevClusterClock` restore, and the test's own comment says it does not cover that. The orphans exercise the single-instance lazy-cancellation skip. Reworded to say exactly that, and the cluster restore is now named as untested in the same breath rather than left for a reader to discover in the PR body.

**2. `cluster.go:1438` was factually wrong.** Confirmed against `invariants.md:262`, which reads "For all **successfully decoded** parent requests (`DecodeInstanceID != ""`)". A parent whose decode sub-request is `StateTimedOut` is not successfully decoded, so the statement already excludes this line — my comment quoted the parenthetical and dropped the governing qualifier, then invited narrowing an invariant that is correctly scoped. That was the worst of the five: it would have sent a future contributor to weaken a good invariant on a rationale that doesn't hold. Reframed as you both suggested — the parenthetical gloss is loose, the statement is not — and it now says explicitly that the statement must *not* be narrowed on the strength of this line. The two correct claims are kept.

**3. `invariants.md:39` was missing the horizon-guard clause.** Confirmed at `cluster.go:862`: the guard sits between the optimistic advance and the restore and can `break`, so the advance can persist unrestored. Also confirmed benign for the reasons you both gave — `SimEndedTime` comes from per-instance `Finalize` as `min(Clock, Horizon)`, snapshots clamp, and `ClusterSimulator.Clock()` has no production caller. Since that paragraph is now the canonical account of the carve-out, it gets the half-sentence.

**4. `simulator.go:1212` scope limit.** Confirmed both writes at `sim/cluster/instance.go:507` and `:523`, and that eviction is counted as `gatewayEvicted` in `cluster_event.go:315` rather than through the sole `CompletedRequests++`. Two changes: the length-capped note is now explicitly scoped to that path, and — taking @jgchn's point that the note pays off most where the anomaly actually lives — `EvictRequest` now carries an INV-2 anchor of its own saying plainly that its `StateCompleted` write is a tombstone, not a completion, and that reading it as one would double-count. That is the anomalous write this pass had left unmarked.

**5. `simulator_test.go:1345` nit.** Correct, and worth fixing precisely because this comment's whole job is precision. `simulator.go:196` has the `Clock: 0` constructor initializer, so "exactly one place" is literally false. Now "assigned exactly once during the run", with the initializer named and why it can't make the clock decrease.

**Both adjacent items, also done:**

- **`CLAUDE.md:714`** — fixed. It said 13 invariants while the file's own INV-BC-DP1 entry and `CONTRIBUTING.md:329` say otherwise; it now mirrors CONTRIBUTING's wording.
- **INV-7's 22/23** — this was the most useful of the non-blocking notes, because it named the failure mode: a deliberate decision that *reads* as an unexplained gap will be re-litigated by the next person through. INV-7's Verification section now records that the missing Go test is a decision (out-of-scope item 3 of #1718) and why an in-process assertion would pin the snapshot plumbing rather than the tiering law, so the gap is legible as a choice.

Nothing was pushed back on — all five were right, and #1 and #2 were correctly prioritized.

**Verification:** `go build ./...` and `go vet ./...` clean; full `go test ./...` passes. All six production files remain comments-only — `git diff origin/main -- <file> | grep "^+" | grep -v "^+\s*//"` is empty for each, including the newly-touched `instance.go` — so INV-6 is untouched by construction. (Local `golangci-lint` is v1.53.3 against CI's pinned v2.9.0 and emits 50 spurious `typecheck` errors in files this PR doesn't touch; @jgchn's clean v2.9.0 run on the prior head is the real signal, and CI will confirm on this one.)

susiejojo:
### Re-review of `12aa102f` — all five findings + both adjacent items confirmed fixed

I re-checked out the PR head and verified each fix against the code rather than the changelog. All addressed correctly:

1. **`invariants.md` Verification** — now says the orphan present exercises "the single-instance lazy-cancellation skip" and adds "The *cluster* `prevClusterClock` restore described below is **not** covered by any test." Over-claim gone. ✓
2. **`cluster.go:1433` comment** — reframed exactly right: the timed-out parents *do* satisfy the parenthetical `DecodeInstanceID != ""`, but the governing qualifier "successfully decoded" excludes them, so "the parenthetical is loose; the statement is correctly scoped and must NOT be narrowed on the strength of this line." ✓
3. **Horizon-guard clause** — folded into the INV-3 carve-out paragraph: the guard between advance and restore can `break` leaving the advance unrestored, benign because `SimEndedTime = min(Clock, Horizon)` and `ClusterSimulator.Clock()` has no production caller. ✓
4. **Length-capped scope + tombstone anchor** — `simulator.go` note now says "Scoped to THIS path: do not generalize," and — better than I asked — `InstanceSimulator.EvictRequest` now carries its own INV-2 anchor calling its `StateCompleted` write "a TOMBSTONE, not a completion … counted as gatewayEvicted … never via the sole CompletedRequests++." That's the anomalous write marked where it lives. ✓
5. **`simulator_test.go:1345`** — now "assigned exactly once during the run … the only other write is the `Clock: 0` constructor initializer." Precise. ✓

Adjacent items:
- **`CLAUDE.md:714`** now mirrors CONTRIBUTING ("13 … plus INV-BC-DP1, PD disaggregation (INV-PD-*) and pool/transfer (INV-P2-*)"). ✓
- **INV-7 Verification** now records the missing Go test as a decision (out-of-scope item 3 of #1718) with the rationale that an in-process assertion would pin the snapshot plumbing rather than the tiering law. Reads as a choice, not a gap. ✓

**Independent verification on the PR head:**
- All six production `.go` files (incl. the newly-touched `sim/cluster/instance.go`) are **comments-only** vs the true merge-base `602ae462` — `git diff <merge-base> | grep "^+" | grep -v "^+\s*//"` is empty for each, so INV-6 holds by construction.
- `go build ./...` and `go vet ./sim/...` clean; `TestINV3_ClockNeverDecreases` passes.
- (Housekeeping, not a finding: the branch is behind `main` — #1713 landed after its base — so it'll want a rebase/merge before merge, but that touches no PR content.)

**All my concerns are addressed. No remaining blockers — ready to merge (pending the CI green box and the branch update).**

susiejojo:
### Side-effect / blast-radius investigation

I went past the diff to check the vectors through which even a comments-only PR can ripple across the repo, and validated each empirically (checked out the head, built both binaries, diffed output). Summary: **no adverse cross-repo consequences.** One wording nit and two pre-existing environmental red herrings worth recording so the next reviewer doesn't re-chase them.

#### Behavioral inertness — confirmed empirically, not just by diff
- **Byte-identical stdout vs the true merge-base `602ae462`** across seeds 1/42/99 (40-request run) — INV-6 holds in practice, not only by the comments-only argument.
- Run-vs-run determinism holds; `go build ./...` + `go vet ./sim/...` clean. All six production `.go` files (incl. the newly-touched `instance.go`) are comments-only vs the merge-base.

#### Positional / directive hazards — none
- No `//nolint`, `//go:build`, `//go:generate`, or `+build` directives anywhere in the diff, so no inserted comment silently detached a directive from its target by breaking adjacency.
- Every insertion lands inside a function body or is appended to an existing doc-comment — no doc-comment/func adjacency broken.

#### Documentation single-source-of-truth — propagation is complete
- Per `principles.md`, `invariants.md` is canonical. The PR updated the two working copies that carry the **statement** (`CLAUDE.md`, `core-engine.md`).
- The other listed working copies don't carry the INV-3 *statement*: `architecture.md` has none, and `.github/ISSUE_TEMPLATE/bug_report.md:24` is a terse one-line checklist item ("clock never decreases") consistent with its siblings — the processed-events/carve-out nuance doesn't belong there. No stale full-statement copy is left behind.
- The new relative link in `core-engine.md` → `../contributing/standards/invariants.md` **resolves** (docs/MkDocs CI won't break).

#### One wording nit worth a fix (same species as the doc findings already addressed)
A **pre-existing** test already takes clock monotonicity as its subject: `TestSimulator_ClockMonotonicity_NeverDecreases` (BC-6, `sim/simulator_test.go:1037`) runs the event loop and asserts monotonicity on the raw `sim.Clock` field. That softens the PR body's "nothing checking it" / "single genuine coverage gap" framing. The gap the new test fills is real but *narrower*: no test **named INV-3 as its subject**, and none used the **processed-event-timestamp** formulation the cluster carve-out requires (a raw-`c.clock` assertion would trip at the `prevClusterClock` restore). Suggest softening "zero tests" → "no test that names INV-3 or uses the processed-timestamp formulation the carve-out needs (BC-6 asserts monotonicity on the raw field, which doesn't generalize to the cluster loop)." Not a blocker.

#### Two red herrings — both pre-existing on `main`, NOT introduced here
1. **`go test ./...` shows a FAIL: `TestEveryKVCapacityCallSiteIsEPAware` (`cmd/`).** It does `filepath.WalkDir("..")` from the repo root and scans stale local git worktrees under `.claude/worktrees/` (e.g. `rename-batch-flags-1570/`) holding pre-#1656 copies of `root.go`/`replay.go`. **Reproduces identically on `origin/main`.** CI's clean checkout has no such worktrees, so it's green there — purely a local-env artifact. (Minor hardening idea for a separate issue: have that walk skip `.claude/worktrees/`, since anyone with a worktree open hits a spurious failure.)
2. **`gofmt -l` flags `simulator.go`, `pool.go`, `session_test.go`.** The wanted reformats are all in code the PR never touched (struct-field alignment `:118`, comment-indent `:108`, struct literals `:101/:423`) — **present on the base**, zero overlap with the PR's added hunks, and CI gates on `golangci-lint@v7` (formatter not enabled by default), not `gofmt -l`. The PR's own inserted comments are gofmt-clean.

**Bottom line:** genuinely inert at runtime; doc edits leave no stale siblings; the new link resolves. Only actionable item is the optional BC-6 wording tweak. The `go test`/`gofmt` results above are pre-existing environmental noise, not this PR.

mtoslalibu:
Thanks @susiejojo — the BC-6 finding is valid and I fixed it in `be6ef03b`, plus the branch update you flagged.

**The finding holds.** I confirmed `TestSimulator_ClockMonotonicity_NeverDecreases` is present on the true merge-base `602ae462` (so pre-existing, not something this branch added) and that it drives the event loop asserting monotonicity. That makes two of my claims false — and false in exactly the doc-vs-code direction this PR exists to correct, which is why I treated it as more than a body nit:

- **`sim/simulator_test.go`** said INV-3 was true "by construction" with *"nothing checked it."* Untrue — BC-6 checked it. This was the one that actually mattered, because a test comment outlives the PR description that repeats it. It now names BC-6 explicitly and states the narrower gap: BC-6 asserts on the raw `sim.Clock` **field** and never names INV-3, and a raw-field assertion doesn't generalize to the cluster loop (it trips at the `prevClusterClock` restore). So the comment now points at the sibling test instead of implying it doesn't exist.
- **PR body** — both spots corrected the same way: the summary no longer calls this "the single genuine coverage gap", and the corrections-to-the-issue bullet now says BC-6 *does* take clock monotonicity as its subject, so the gap is narrower than either the issue or my earlier draft implied. I've noted in-line that the earlier description over-claimed rather than quietly rewriting it.

While in the body I also fixed two counts that went stale when I added the `EvictRequest` anchor last round: **17 anchor comments across six production files**, not 16 across five. Worth having right given the whole point of this PR is that stated and actual protection should match.

**Branch update:** merged `origin/main` (`c4da14dc`) — #1713 had landed after the base. Clean merge, no conflicts, and my INV-3 edits to `invariants.md` survived intact alongside main's own edits to that file.

**Re-verified on the new head:**
- `go build ./...` + `go vet ./...` clean; **full `go test ./...` passes** — including `cmd/`, so `TestEveryKVCapacityCallSiteIsEPAware` is green here. That corroborates your read that it's a local-env artifact: this worktree has no nested `.claude/worktrees/` for its `WalkDir("..")` to find, which is why you saw the failure and I don't. Agreed it's worth its own hardening issue to skip that directory — happy to file it, but keeping it out of this PR.
- Both clock tests pass together (`TestINV3_ClockNeverDecreases` and BC-6).
- All **six** production `.go` files remain comments-only vs `602ae462` — `git diff 602ae462 -- <file> | grep "^+" | grep -v "^+\s*//"` is empty for each, so INV-6 still holds by construction post-merge.

No action taken on the `gofmt -l` hits, agreed as pre-existing: the wanted reformats are in untouched code (`simulator.go:118`, `pool.go:108`, `session_test.go:101/:423`), zero overlap with this PR's hunks, and CI gates on `golangci-lint` rather than `gofmt -l`.

Thanks also for the positional-hazard and doc-propagation sweep — checking for `//nolint` / `//go:build` adjacency breakage and for stale INV-3 statement copies is precisely the failure mode a comments-only PR could hide, and it's useful to have that recorded as checked rather than assumed.

susiejojo:
### Confirmed on `be6ef03b` (post-merge head) — my concern is resolved

Verified against the code, not the summary:

- **BC-6 test-comment fix is exactly right.** `sim/simulator_test.go` now names `TestSimulator_ClockMonotonicity_NeverDecreases (BC-6, above)`, states it "asserts monotonicity — but on the raw sim.Clock FIELD, and without naming INV-3," and frames the narrower gap honestly ("no test named INV-3 as its subject, and none used the formulation the invariant is actually stated in"). That was the one that mattered — the comment outlives the PR body — and it now points at the sibling test instead of implying it doesn't exist.
- **Counts corrected.** 17 anchors / six production files reconciles with the `EvictRequest` anchor added last round.
- **Post-merge inertness holds.** All six production `.go` files remain comments-only vs `602ae462`; `go build` + `go vet` clean; `TestINV3_ClockNeverDecreases` and BC-6 both pass together; and stdout is **byte-identical to the base** across seeds 3/42 — so INV-6 survives the `origin/main` merge intact.

On the `cmd/` static-scan test: I still see `TestEveryKVCapacityCallSiteIsEPAware` fail **locally**, but confirmed **every** failing path is under `.claude/worktrees/` (stale sibling worktrees its `WalkDir("..")` picks up) and **zero** come from this PR's own tree — so your green run is the real signal, and this corroborates the local-env diagnosis rather than contradicting it. Agree it deserves its own hardening issue (skip `.claude/worktrees/` in the walk); out of scope here.

No remaining blockers from me. LGTM pending CI green.

susiejojo (approved):
lgtm

Issue #1706, opened 2026-09-10
KV-offload prefill-shrink: credit the whole reloaded prefix under chunked prefill (follow-up to #1699)


susiejojo:
**What problem does this solve?**

#1699 (PR #1702) made a CPU/secondary KV-offload cache hit reduce billed prefill work (`NumNewTokens`) instead of charging a full recompute, but the credit is **capped at one chunk**. The root cause is an *ordering* difference from vLLM, not just the `ComputedTokens` clamp.

**vLLM reference** (`vllm/v1/core/sched/scheduler.py`, this fork's CPU-offload branch):

1. `num_computed_tokens = num_new_local_computed_tokens + num_external_computed_tokens` — computed **first** (`:1006-1008`).
2. `num_new_tokens = request.num_tokens - num_computed_tokens` (`:1029`).
3. **Then** the chunk cap: `if 0 < long_prefill_token_threshold < num_new_tokens: num_new_tokens = threshold` (`:1030-1032`), then `min(num_new_tokens, token_budget)` (`:1044`).
4. `allocate_slots(..., num_external_computed_tokens=...)` commits the external blocks (`:1080-1089`), and `request.num_computed_tokens = num_computed_tokens` advances progress to `local + external` **before** the chunk (`:1143`).

So vLLM's post-step progress is `(local + external) + chunk`: the external/offload prefix is folded into the computed baseline **before** the chunk is sized, and the chunk covers only the uncached remainder.

**BLIS inverts this** (`sim/batch_formation.go` Phase 2, `:338-401`): it sizes `numNewTokens` from the **GPU-only** `GetCachedBlocks` (`startIndex`), commits `[startIndex, endIndex)`, and only discovers the reload *inside* `AllocateKVBlocks` (`sim/kv/offload_chain.go` `allocateThroughChain`, `:411-448`). The post-hoc re-bill is `billedTokens = endIndex - min(newStart, endIndex)` and `ComputedTokens` stays at `endIndex`. When the reload extends the prefix past the capped chunk (`newStart > endIndex`, i.e. under chunked prefill via `--max-num-batched-tokens` / `long_prefill_token_threshold`), the remainder `[endIndex, newStart)` is **not** credited — later Phase-1 continuations bill `InputLen - ProgressIndex` and consult no cache, so it is re-billed as full recompute.

**Impact (measured, 4096-token prompts / 3072 shared prefix, billed-prefill-token counter):**

[...]

PR #1726, opened 2026-09-15
docs(kv-cache): add KV offload worked example and explanations

leipanhz:
The KV-offload section documented the config surface but never showed what offload does to a run, and left three things that repeatedly trip people up undocumented. Adds three subsections to docs/guide/kv-cache.md, aimed at a general audience (how it works, how to turn it on/off, what it buys).

### Enabling and Disabling Offload ###
  The switch is the presence of --kv-offload-config; omitting it is inert and
  byte-identical (INV-6), so the two commands are a clean A/B. Documents that
  cpu_bytes_to_use: 0 and sub-block values are rejected at startup, that a
  one-block tier is a legal offload-off control arm, and the mutual exclusion
  with the legacy --kv-cpu-blocks tier (cmd/root.go:1943).

### Sizing the CPU Tier ###
  Explains the two things the cpu_bytes_to_use YAML comment could not:

  - Block capacity. The tier is configured in bytes but used in blocks: block_capacity = floor(cpu_bytes_to_use / per_block_bytes). Adds the per_block_bytes formula, notes the division truncates, and adds a table contrasting block_capacity with --total-kv-blocks — same unit, two tiers, which is what makes block_capacity > total_kv_blocks the condition for the CPU tier to hold what the GPU evicts.
  - "Per TP rank". Both sides of that division are per-GPU, so the same cpu_bytes_to_use buys TP times more blocks while node-wide host RAM is TP * cpu_bytes_to_use. Includes a measured per_block_bytes table across --tp 1/2/4/8 for Qwen3-14B.

  Two exceptions called out: MLA models are not divided by TP (latent is
  replicated across ranks, kv_capacity.go:130-135, #1527), and the
  kv_heads < tp GQA case over-reports block capacity by tp/kv_heads. The
  latter is flagged as a known inaccuracy rather than a caveat: it is
  documented as "a known approximation (optimistic)" at kv_capacity.go:92-100
  but is silent at startup, the regime is reachable (any 8-KV-head model past
  TP=8), and it biases block capacity optimistically in both tiers.

### Example: The Measured Effect of CPU Offload ###
  A 100-tenant fixture with distinct per-member prefixes (a single shared
  prefix stays pinned on the GPU and is never evicted, so offload is a no-op
  and cannot be observed). Offload disabled vs enabled on a 1 TiB CPU tier:
  cache_hit_rate 0.0772 -> 0.3665 (4.7x) and ttft_mean_ms 52.550 -> 48.914
  (-6.9%), with throughput unchanged because the fixture is arrival-bound.
  Derives the 37.0% hit-rate ceiling so the 0.3665 is readable as ~99% of
  achievable, and notes that hit rate and TTFT are separate questions that do
  not move by the same factor.

  Also documents that prefix_length is ADDED to input_distribution rather than
  carved out of it (generator.go prepends it), so the fixture's prompts are
  2,304 tokens and not 1,280 — the reason total_input_tokens is 1382400 — and
  that --workload-spec supersedes --prefix-tokens / --rate silently.

All figures measured at this branch base (Qwen3-14B, TP=1, seed 42, single instance) and re-verified after rebasing onto origin/main, which had touched sim/cluster/kv_autocalc.go: identical to the last digit.

Docs only; no behavior change.

## Behavioral Contracts

- **BC-1:** ...
- **BC-2:** ...

## Antipattern Checklist

**Critical (correctness):**
- [ ] No silent data loss — every error path returns error, panics, or increments counter (R1)
- [ ] Struct construction sites audited for new fields (R4)
- [ ] Resource allocation loops rollback on mid-loop failure (R5)
- [ ] No `logrus.Fatalf` or `os.Exit` in `sim/` packages (R6)
- [ ] Division by runtime-derived denominators guarded (R11)
- [ ] Unbounded retry/requeue loops have circuit breakers (R19)
- [ ] No `range` over slices that can shrink during iteration (R21)

**Important (quality):**
- [ ] Map keys sorted before float accumulation or ordered output (R2)
- [ ] Every new numeric parameter validated — CLI flags AND library constructors (R3)
- [ ] Invariant tests alongside golden tests (R7)
- [ ] No exported mutable maps — use `IsValid*()` accessors (R8)
- [ ] `*float64` for YAML fields where zero is valid (R9)
- [ ] YAML strict parsing with `KnownFields(true)` (R10)
- [ ] New interfaces work for 2+ implementations (R13)
- [ ] No method spans multiple module responsibilities (R14)
- [ ] Routing scorer signals documented for freshness tier (R17)
- [ ] CLI flag values not silently overwritten by defaults.yaml (R18)
- [ ] Detectors and analyzers handle degenerate inputs (R20)
- [ ] Pre-check estimates consistent with actual operation accounting (R22)
- [ ] Parallel code paths apply equivalent transformations (R23)

**Hygiene (maintenance):**
- [ ] Golden dataset regenerated if output changed (R12)
- [ ] Stale PR references resolved (R15)
- [ ] Config params grouped by module (R16)

## Invariants

- [ ] Request conservation: injected == completed + still_queued + still_running + dropped_unservable + timed_out (+ gateway/routing/encode buckets for cluster runs) (INV-1)
- [ ] Request lifecycle: queued -> running -> completed (INV-2)
- [ ] KV block conservation: allocated + free == total (INV-4)
- [ ] Causality: arrival <= schedule <= completion (INV-5)
- [ ] Clock monotonicity (INV-3)
- [ ] Determinism: same seed produces identical output (INV-6)
- [x] N/A — this PR does not touch invariant-sensitive code

## Test Plan

```bash
go test ./...
golangci-lint run ./...
```

## Source

Fixes #

## Documentation

- [ ] CLAUDE.md updated (if new files, packages, CLI flags, or invariants added)
- [ ] README updated (if user-facing behavior changed)
- [x] CONTRIBUTING.md reviewed (if extension patterns changed)


docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+decode-offload is inert there — a degenerate offload block size (real offload block sizes track
+the GPU block size). With no `--kv-offload-config`, behavior is unchanged (INV-6).
+
+### Enabling and Disabling Offload
+
+Offload is **off by default**. There is one switch — the presence of `--kv-offload-config`:
+
+```bash
+# ENABLED — CPU staging tier
+./blis run --model qwen/qwen3-14b --workload-spec wl.yaml \
```
jgchn: **`wl.yaml` doesn't exist.** The fixture defined below is `wl_multitenant.yaml`, and the Results section says "run the enabled and disabled commands from the previous section" — so copy-paste fails on the one thing the doc most wants reproduced. Same on the disabled command below.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+# DISABLED — omit the flag
+./blis run --model qwen/qwen3-14b --workload-spec wl.yaml \
+  --total-kv-blocks 3000
+```
+
+`--kv-offload-config` is mutually exclusive with the legacy scalar `--kv-cpu-blocks` tier.
+
+### Sizing the CPU Tier
+
+Blocks reach the CPU tier only by being evicted from the GPU tier, so a CPU tier that cannot
```
jgchn: **This is the wrong mechanism, and it makes the sizing rule under-size by ~8.5×.**

CPU population is **write-through**, not eviction-driven: `MirrorToCPU` is called every step on the running batch (`sim/simulator.go:829`), with the contract "Copy newly-completed full blocks to CPU tier" (`sim/kv_store.go:17`). `sim/kv/tiered.go:203` says it outright: *"GPU prefix cache is now preserved on release; CPU tier is populated via MirrorToCPU."*

Consequence: because **all** offloadable prompt blocks are mirrored (not just the reusable prefix), the tier's real working set here is `100×64 + 600×80 = 54,400` blocks, not the 6,400 computed below. Measured sweep:

| CPU tier (blocks) | hit rate |
|---|---|
| 3,000 | 0.0772 *(= disabled)* |
| 6,400 ← the "working set" below | 0.1663 |
| 20,000 | 0.3312 |
| **54,400** | **0.3665 (saturates)** |
| 419,430 | 0.3665 |

The example's 1 TiB is fine; the *rule a reader carries to their own workload* is not. The sweep does strongly confirm the warning box below — at or below GPU capacity the run is exactly the disabled result.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+outgrow the GPU tier has nothing to hold.
+
+#### Block capacity — what the tier actually holds
+
+The tier is configured in **bytes**, but the cache uses it in **blocks**. BLIS converts once, at
+startup, by integer division:
+
+```
+per_block_bytes = 2 (K+V) × layers × kv_heads × head_dim × dtype_bytes × block_size
```
jgchn: **Formula omits `/ TP`.** The real value is divided by TP (`kv_capacity.go:182`) — as the "Per TP rank" section below states and as the TP table there reflects. At TP=8 this formula gives `2,621,440` against the true `327,680`: 8× off, in the exact expression the warning box tells readers to compute. Verified by reading back `per_block_bytes` from the trace header at `--tp 1/2/4/8`.

Also `layers` should be KV-**bearing** layers (`EffectiveKVBearingLayers()`), which matters for Kimi-K3 — named in the MLA note below — at 24 of 93.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+block_capacity  = floor(cpu_bytes_to_use / per_block_bytes)
+```
+
+`block_capacity` is the number that matters: it is what LRU/ARC evicts against, and the only
```
jgchn: **ARC isn't supported — it panics.** `eviction_policy: arc` dies with a raw Go stack trace:

```
panic: NewOffloadCache: H1 supports eviction_policy "lru" only; "arc" (e.g. arc) is a follow-up
```

`sim/kv_offload_config.go` `Validate` accepts it, so nothing catches it at the CLI. Suggest dropping "/ARC" here. (The `lru | arc` offer in the pre-existing YAML block above has the same problem, as does `blocks_per_chunk: 2` — both worth a CLI guard in a follow-up issue.)

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+
+| | Tier | How it is configured | Value for the example below |
+|---|---|---|---|
+| `--total-kv-blocks` | GPU (HBM) | directly, in blocks — or auto-derived from GPU memory when omitted | 3,000 |
+| `block_capacity` | CPU (host RAM) | *indirectly*, in bytes via `cpu_bytes_to_use`; BLIS divides | 419,430 |
+
+Both count the same thing — whole KV blocks of `block_size` tokens — so `block_capacity > total_kv_blocks`
+is the condition for the CPU tier to be able to hold what the GPU evicts. The asymmetry is only in
+the configured units: the GPU tier is set in blocks, the CPU tier in bytes (matching vLLM's knob
+for each). Both scale with TP, so the comparison must be made at the intended TP.
```
jgchn: **Only `block_capacity` scales with TP.** An explicit `--total-kv-blocks 3000` is consumed verbatim at every TP — auto-calc runs only when the flag is unset. The table row above already says this ("directly, in blocks — or auto-derived ... when omitted"), so just the "Both" needs softening.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+    Check `cpu_bytes_to_use / per_block_bytes` against `--total-kv-blocks` before concluding anything from a run.
+
+#### "Per TP rank" — both sides of that division are per-GPU
+
+Under tensor parallelism the KV cache is **sharded across ranks**: each GPU holds the same
+*logical* blocks (the same token ranges) but only its own slice of the KV heads. So each GPU
+needs only its own slice of host memory to stage those blocks. BLIS therefore treats both sides
+of the division as per-rank — `cpu_bytes_to_use` is the host budget for **one** GPU, and
+`per_block_bytes` is one block's size on **one** rank (`KVBytesPerToken(model, tp) × block_size`,
+divided by TP at [`sim/latency/kv_capacity.go:182`](https://github.com/inference-sim/inference-sim/blob/main/sim/latency/kv_capacity.go#L182)).
```
jgchn: **Line-anchored links pinned to `main` will rot.** These three `#L182` / `#L130-135` / `#L92-100` anchors are the first line-anchored links in `docs/` (the tree has 3 `blob/main/` links, none with `#L`). All three currently match their claims verbatim — worth pinning to a commit SHA to keep it that way.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+
+    When `kv_heads >= tp` the division is exact — and a `kv_heads` not evenly divisible by `tp`
+    is rejected outright rather than approximated.
+
+### Example: The Measured Effect of CPU Offload
+
+#### The workload needs prefixes the GPU will evict
+
+Offload only helps if blocks are **evicted** from the GPU and later requested again. A single
+shared prefix (`--prefix-tokens N`) stays pinned hot on the GPU and is never evicted, so the
```
jgchn: **This justification is false — measured.** A shared prefix is evictable like anything else. With `--prefix-tokens 1024 --prompt-tokens 1280 --prompt-tokens-stdev 0 --output-tokens 16 --output-tokens-stdev 0 --rate 40 --num-requests 600 --seed 42`:

| `--total-kv-blocks` | offload off | offload on |
|---|---|---|
| 3,000 | 0.4406 | 0.4406 *(byte-identical)* |
| 500 | 0.4406 | **0.4312** |
| 200 | 0.5961 | **0.5535** |

Byte-identity holds only when the GPU isn't under pressure — and a reader shrinking `--total-kv-blocks` to force eviction (this doc's own logic) lands in the diverging regime. Note the enabled arm goes *down*; see the comment on the results table for why.

The 100-distinct-prefix fixture is still a good choice — it isolates the effect cleanly. It's the stated reason that needs rewriting.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+    prefix_sharing: per_member     # 100 DISTINCT prefixes => GPU must evict
+    prefix_length: 1024            # ADDITIVE on input_distribution
+    rate_fraction: 1.0
+    arrival: {process: poisson}
+    input_distribution:  {type: constant, params: {value: 1280}}
+    output_distribution: {type: constant, params: {value: 16}}
+```
+
+```yaml
+# offload_cpu.yaml — one CPU tier, large enough to hold the whole working set
```
jgchn: **"the whole working set" is ~54,400 blocks, not the 6,400 derived above.** Sizing at 6,400 yields 0.1663 — about 45% of the achievable 0.3665. The 1 TiB here is comfortably above the real requirement, so the example is unaffected; the comment just overstates what suffices. (See the write-through comment above for why.)

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+
+#### Results
+
+Run the enabled and disabled commands from the previous section. `Cache Hit Rate` is printed to
+stdout under `=== KV Cache Metrics ===`; add `--metrics-path m.json` for the full-precision
+`cache_hit_rate` field.
+
+| | `cache_hit_rate` | `ttft_mean_ms` | `e2e_mean_ms` | `responses_per_sec` |
+|---|---|---|---|---|
+| Offload **disabled** | 0.0772 | 52.550 | 250.303 | 4.1640 |
```
jgchn, line 248: **`cache_hit_rate` is not defined identically across these two arms.**

`OffloadCache.CacheHitRate()` (`sim/kv/offload_chain.go:277`) is `hits / (hits + gpu.CacheMisses + offloadMissCount)`, while single-tier (`cache.go:453`) is just `hits/(hits+misses)`. `offloadMissCount` increments when there's no CPU-resident continuation **and `gpu.AllocateKVBlocks` fails** (`offload_chain.go:456`) — i.e. it counts GPU allocation failures, and it lands in the denominator only.

So enabling offload adds a denominator-only term that grows with GPU pressure. Here the 4.7× hit gain swamps it and **the conclusion is sound**, but under pressure the enabled arm can report a *lower* rate for an accounting reason rather than a caching one (that's the shared-prefix table above). Worth one sentence saying the two arms' hit rates aren't strictly comparable and a drop doesn't mean offload hurt — and probably worth its own BLIS issue.

All four figures in this table reproduce exactly, incidentally. One caveat for the methodology line: they require `891facee` (#1702) — on an older base the enabled arm shows no TTFT change at all.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+| Offload **enabled** (1 TiB CPU tier) | **0.3665** | **48.914** | **243.994** | 4.1641 |
+
+Cache hit rate rises **4.7×** and mean TTFT falls **6.9%** (−3.6 ms). Throughput is unchanged
+because this workload is arrival-bound — 4 req/s offered against an unsaturated instance. Offload
+buys latency here; it buys *throughput* only under saturation, where spending fewer prefill tokens
+per request lets more requests into each step.
+
+Why the hit rate lands near 0.37: only the 1,024-token prefix is shareable and each tenant's first
+request must miss, so the ceiling is `1024 × 500 / 1382400` = **37.0%**. The enabled run reaches
```
jgchn: **This isn't the metric's definition.** `cache_hit_rate` is a *block* ratio — `hits/(hits+misses)`, incremented per block (`cache.go:277/350/415`) — and its denominator includes decode-allocated blocks. The token formula lands within 0.1pp here only because block size is uniform and output is just 16 tokens; a reader generalizing it to a 512-token-output workload would be materially off. Worth noting it's an approximation.

Also: 4.7× is the most seed-sensitive figure here — 3.8–4.8× across seeds 42/7/1234, while the TTFT improvement is stable at 6.7–7.1%.

docs/guide/kv-cache.md
```diff
@@ -143,10 +143,209 @@ tier). Reuse is single-instance (offload tiers are per-instance and invisible to
...
+0.3665 — essentially every reuse the workload contains.
+
+!!! tip "Large cohorts need `--lazy-generation`"
+    Eager workload generation costs roughly `population × num_requests × 44 KB` in resident
```
jgchn: **Constant is ~1.6× high; the scaling law is right.** Measured peak RSS: `pop×nr` 30,000 → 0.87/0.82 GB (both product-equal configs agree, confirming the product law), 60,000 → 1.66 GB, 120,000 → 3.22 GB ⇒ **~26–28 KiB per unit, not 44 KB**.

Ran this exact 300×3,000 config without `--lazy-generation`: **peak RSS 23.6 GB**, so "~38 GB" should read **~24 GB**. "is OOM-killed" is machine-dependent — true on a 16–32 GB laptop, but it completes on a 69 GB box, so consider phrasing it as a threshold. Also worth noting the doc's own example (100×600 = 60,000) sits above the stated "few tens of thousands" while correctly omitting the flag.

jgchn (commented):
## Review: correctness pass

**Verdict: approve with changes.** Docs-only and nothing here is unsafe to merge — but four statements are wrong in ways a reader would act on.

**The measurements are excellent.** I rebuilt at this PR's base (`5bd02102`) and re-ran everything: all 8 figures in the results table, the entire TP `per_block_bytes` table, and every derived number (419,430 blocks · 6,400 · 44% · 37.0% · 4.7× · 6.9% · 99%-of-ceiling) reproduce **to the digit**, deterministically, and the TTFT conclusion holds across three seeds. All three `kv_capacity.go` citations, `cmd/root.go:1943`, and the `generator.go` quote are accurate; every YAML key exists; `--rate`/`--prefix-tokens` really are inert alongside `--workload-spec`. That's better empirical hygiene than most docs PRs get.

**Every defect is in the explanatory model wrapped around those correct numbers** — which is worth fixing precisely because this doc's job is to teach rules readers will apply to their own workloads:

1. **CPU population is write-through, not eviction-driven** (line 168) — and correcting it fixes the sizing rule, which currently under-sizes by ~8.5× (6,400 stated vs 54,400 measured to saturate)
2. **`per_block_bytes` formula omits `/ TP`** (line 177) — 8× off at TP=8
3. **The shared-prefix "byte-identical" justification is false** (line 277) — diverges under GPU pressure
4. **`cache_hit_rate` isn't comparable across the offload on/off boundary** (line 333) — the enabled arm carries an extra denominator term and can report a *lower* rate under pressure

Then ~7 quick edits inline: ARC isn't supported (it panics), "both scale with TP", the memory constant (~24 GB measured, not ~38 GB), the `wl.yaml` filename, the ceiling-formula caveat, and SHA-pinning the three code links.

Two things outside the diff:
- **PR body is still the unfilled template** — `**BC-1:** ...` placeholders, no checkboxes ticked, `Fixes #` with no number. The commit message is genuinely good and most of it belongs in the body.
- **"Worth raising upstream; not yet filed"** — per CLAUDE.md's issue-filing guidance, worth filing the GQA `kv_heads < tp` issue and linking it rather than shipping a doc pointing at an untracked defect. Same for the `arc`/`blocks_per_chunk` panics and the hit-rate comparability.
- Minor: `docs/reference/configuration.md` has no row for `--kv-offload-config`, yet this doc's Further Reading links there promising "all KV cache flags". Pre-existing, but this is the natural PR to close it in.

Reproduction commands for anything above are available on request.

susiejojo:
## `/blis-pr-review` — docs(kv-cache): CPU-offload worked example

**Verdict: Merge after fixing one incorrect claim (ARC).** Docs-only, no behavior change. I reproduced the worked example end-to-end on this branch (`4f8ba2ef`, Qwen3-14B, TP=1, seed 42, single instance) and **every measured figure matches to the last digit.** I then did a second pass hunting for over-arching or incorrect claims — one is wrong (below), the rest hold up.

### Reproduction (all exact)

| Metric | Doc — disabled / enabled | My run — disabled / enabled |
|---|---|---|
| `cache_hit_rate` | 0.0772 / 0.3665 | 0.07724 / 0.36651 ✓ |
| `ttft_mean_ms` | 52.550 / 48.914 | 52.55036 / 48.91381 ✓ |
| `e2e_mean_ms` | 250.303 / 243.994 | 250.30256 / 243.99358 ✓ |
| `responses_per_sec` | 4.1640 / 4.1641 | 4.16398 / 4.16411 ✓ |
| `total_input_tokens` | 1,382,400 | 1,382,400 ✓ |

I also reproduced the **full TP block-capacity table** by reading `per_block_bytes` back out of the trace header at each TP — not just TP=1:

| `--tp` | doc `per_block_bytes` / blocks | trace-header readback |
|---|---|---|
| 1 | 2,621,440 / 6,553 | 2,621,440 → 6,553 ✓ |
| 2 | 1,310,720 / 13,107 | 1,310,720 → 13,107 ✓ |
| 4 | 655,360 / 26,214 | 655,360 → 26,214 ✓ |
| 8 | 327,680 / 52,428 | 327,680 → 52,428 ✓ |

Arithmetic verified independently: `per_block_bytes` = 2×40×8×128×2×16 = 2,621,440; 1 TiB → 419,430 blocks; truncation `5000000//2621440 = 1`; working set `100×1024/16 = 6400`; hit ceiling `1024×500/1382400 = 37.0%` (500 = the `(6−1)×100` reuse requests); input `1280+1024 = 2304`, `600×2304 = 1,382,400`; shareable frac `1024/2304 = 44%`. All ✓.

Code/behavior citations confirmed by running them: `cpu_bytes_to_use: 0` → fatal at startup (`sim/kv_offload_config.go:166-167`); `--kv-offload-config` + `--kv-cpu-blocks` → fatal, mutually exclusive (`cmd/root.go:1988`); `prefix_length` prepend snippet present at `generator.go:273`/`:611`; MLA / `/tp` / GQA citations land on the cited code.

### Findings

1. **INCORRECT CLAIM — ARC eviction is not supported and panics at runtime.** Line 181 states `block_capacity` "is what **LRU/ARC** evicts against," presenting ARC as a working eviction policy alongside LRU. It is not. The config *validator* accepts `eviction_policy: arc` (`sim/kv_offload_config.go:182`), but the H1 offload mechanism hard-panics on anything but `lru`:

   ```
   panic: NewOffloadCache: H1 supports eviction_policy "lru" only; "arc" (e.g. arc) is a follow-up
   ```
   (`sim/kv/offload_chain.go:138-139` — I confirmed this by running `eviction_policy: arc`.)

   A reader who follows this doc and picks ARC gets a panic, not an eviction-policy choice. Drop the "/ARC" — say LRU (ARC is a documented follow-up), or note explicitly that ARC is accepted by the parser but not yet implemented.

2. **Line-pinned permalinks to `main` will rot.** Three `blob/main/...#L182` / `#L130-L135` / `#L92-L100` links pin exact line ranges on a moving branch and will drift to unrelated code as `main` evolves. The inline `cmd/root.go:1943` citation is already slightly off (the actual fatal is at `:1988`). Pin to a commit SHA or cite the function name instead.

3. **"Same unit" comparison silently assumes matching block sizes.** The `block_capacity` vs `--total-kv-blocks` comparison (lines 185-198) is "the same unit" only because the offload `block_size: 16` equals the default GPU `--block-size-in-tokens 16`. That holds for the example, but the claim is stated generally; a reader who sets a different offload `block_size` breaks the direct block-count comparison. Worth one clause noting the comparison assumes equal block sizes.

4. **PR metadata unfilled.** `Fixes #` has no issue number and the checklist is blank. Link the motivating issue if one exists.

### Cross-path parity
N/A — docs only. Offload is `run`/`replay`-scoped by design (observe is a black-box dispatcher, per CLAUDE.md); the doc frames it correctly.

The A/B fixture design (100 distinct per-tenant prefixes to force eviction, with the explicit note that a single shared prefix stays GPU-pinned and makes offload a no-op) is exactly the part most people get wrong, and it's correct here. Once the ARC claim is corrected, this is a strong, well-evidenced doc.

docs/guide/kv-cache.md
leipanhz: Updated with wl_multitenant.yaml

docs/guide/kv-cache.md
leipanhz: The writing in the doc caused confusion, updated to reflect that claim that the blocks from CPU tier are only retrieved if they are evicted from GPU

docs/guide/kv-cache.md
leipanhz: Updated the writing, cc @susiejojo regarding similar comments

susiejojo:
Hi @leipanhz — thanks for this, it's a really well-put-together worked example. 🙏

One heads-up: [#1725](https://github.com/inference-sim/inference-sim/pull/1725) ("credit the WHOLE reloaded prefix under chunked prefill", Closes #1706) merged to `main` earlier today (`5497174e`) and touches exactly the offload-billing path this doc measures. Since your example prompt is **2,304 tokens** — above the default `--max-num-batched-tokens 2048`, so it's in the chunked-prefill regime #1725 changes — the enabled-arm numbers do shift. **Could you please rebase onto latest `main` and refresh the figures when you get a chance?**

I re-ran both arms on `origin/main` (which now includes #1725), same fixtures/seed, to show what moves:

| metric | doc (enabled) | main + #1725 | change |
|---|---|---|---|
| `cache_hit_rate` | 0.3665 | **0.3678** | +0.4% |
| `ttft_mean_ms` | 48.914 | **40.605** | **−17%** |
| `e2e_mean_ms` | 243.994 | **235.619** | −3.4% |
| `responses_per_sec` | 4.1641 | 4.1645 | ~0 |

The disabled arm is unchanged (0.0772 / 52.550 / 250.303 / 4.1640). Good news is the story gets *stronger*, not weaker:

- **The headline TTFT claim grows.** "mean TTFT falls **6.9%** (−3.6 ms)" becomes roughly **−22.7% (−11.9 ms)** (52.550 → 40.605). That's the whole point of #1725 — the full reloaded prefix is now credited instead of just one chunk.
- **`cache_hit_rate` barely moves** (0.3665 → 0.3678), still ~99% of the 37.0% ceiling, so that derivation and the "4.7×" both still hold.

Two spots that go stale beyond the results table:

1. **The "Chunked-prefill limitation (#1706)" note (lines ~137–144)** describes the one-chunk cap as a *current* limitation — but #1725 *closes* #1706, so that whole admonition is now describing fixed behavior. It probably wants to be removed or reworded to past tense.
2. The `~6.9% / −3.6 ms` figure in the Results prose (line ~336) needs the same refresh as the table.

Everything else I checked reproduces exactly to the last digit against the pre-#1725 base, so it's purely these offload-latency numbers plus that one note. Thanks again — happy to re-verify once it's rebased!

docs/guide/kv-cache.md
leipanhz: The block capacity calculation is rewritten, as I feel it adds a lot of details about model calculation. Make this doc focus on kv offloading related discussion.

docs/guide/kv-cache.md
leipanhz: Fixed in the updated version

docs/guide/kv-cache.md
leipanhz: 54400 blocks are correct, fixed in the doc. Thank you

leipanhz (commented):
Thanks @jgchn and @susiejojo for reviewing the PR. I made changes based on the feedback and added replies. Please let me know if you have any questions.

docs/guide/kv-cache.md
leipanhz: Doc is updated.

docs/guide/kv-cache.md
leipanhz: see the comments above.

docs/guide/kv-cache.md
leipanhz: updated doc

leipanhz:
> Hi @leipanhz — thanks for this, it's a really well-put-together worked example. 🙏
> 
> One heads-up: [#1725](https://github.com/inference-sim/inference-sim/pull/1725) ("credit the WHOLE reloaded prefix under chunked prefill", Closes #1706) merged to `main` earlier today (`5497174e`) and touches exactly the offload-billing path this doc measures. Since your example prompt is **2,304 tokens** — above the default `--max-num-batched-tokens 2048`, so it's in the chunked-prefill regime #1725 changes — the enabled-arm numbers do shift. **Could you please rebase onto latest `main` and refresh the figures when you get a chance?**
> 
> I re-ran both arms on `origin/main` (which now includes #1725), same fixtures/seed, to show what moves:
> 
> metric	doc (enabled)	main + #1725	change
> `cache_hit_rate`	0.3665	**0.3678**	+0.4%
> `ttft_mean_ms`	48.914	**40.605**	**−17%**
> `e2e_mean_ms`	243.994	**235.619**	−3.4%
> `responses_per_sec`	4.1641	4.1645	~0
> The disabled arm is unchanged (0.0772 / 52.550 / 250.303 / 4.1640). Good news is the story gets _stronger_, not weaker:
> 
> * **The headline TTFT claim grows.** "mean TTFT falls **6.9%** (−3.6 ms)" becomes roughly **−22.7% (−11.9 ms)** (52.550 → 40.605). That's the whole point of [fix(kv-offload): credit the WHOLE reloaded prefix under chunked prefill (Closes #1706) #1725](https://github.com/inference-sim/inference-sim/pull/1725) — the full reloaded prefix is now credited instead of just one chunk.
> * **`cache_hit_rate` barely moves** (0.3665 → 0.3678), still ~99% of the 37.0% ceiling, so that derivation and the "4.7×" both still hold.
> 
> Two spots that go stale beyond the results table:
> 
> 1. **The "Chunked-prefill limitation ([KV-offload prefill-shrink: credit the whole reloaded prefix under chunked prefill (follow-up to #1699) #1706](https://github.com/inference-sim/inference-sim/issues/1706))" note (lines ~137–144)** describes the one-chunk cap as a _current_ limitation — but [fix(kv-offload): credit the WHOLE reloaded prefix under chunked prefill (Closes #1706) #1725](https://github.com/inference-sim/inference-sim/pull/1725) _closes_ [KV-offload prefill-shrink: credit the whole reloaded prefix under chunked prefill (follow-up to #1699) #1706](https://github.com/inference-sim/inference-sim/issues/1706), so that whole admonition is now describing fixed behavior. It probably wants to be removed or reworded to past tense.
> 2. The `~6.9% / −3.6 ms` figure in the Results prose (line ~336) needs the same refresh as the table.
> 
> Everything else I checked reproduces exactly to the last digit against the pre-#1725 base, so it's purely these offload-latency numbers plus that one note. Thanks again — happy to re-verify once it's rebased!

Hi @susiejojo the latest change updated numbers in the table. I will remove "Chunked-prefill limitation" section per suggestion, thank you.

leipanhz (commented):
Removed "Chunked-prefill limitation" section per suggestion, please review and let me know if you have any questions, thank you!

susiejojo:
Thanks for the quick rebase, @leipanhz! 🙏 I rebuilt the branch and the refreshed table reproduces exactly (0.3678 / 40.605 / 235.619 / 4.1645), and the ARC wording, permalinks, and sizing rewrite all look great.

Two small things before this is good to merge:

**1. The "Remove stale section" commit (`7ad59ca`) removed the wrong block.** It deleted the unrelated `## Chunked Prefill` section (HOL blocking / `--long-prefill-token-threshold`), while the note you meant to drop — **"Chunked-prefill limitation (#1706)" at lines 137–144** — is still there:

> The reload credit is **currently** capped at **one prefill chunk** … the realized TTFT benefit is roughly `chunk_size / prompt_length` — **small** …

That's now describing behavior that no longer exists: the branch already contains #1725, whose own CLAUDE.md entry states *"#1706 resolves the former one-chunk cap — vLLM's pre-cap ordering now credits the whole reloaded prefix."* So the note contradicts both the code and the doc's own new −22.7% result just below it. Could you remove (or past-tense) that admonition? The fix is just deleting lines 137–144.

**2. The PR now shows conflicts.** `main` has moved on since the last rebase (it's currently `CONFLICTING` / `DIRTY`), and the conflict is in `docs/guide/kv-cache.md` itself — around the Chunked Prefill area, which overlaps the section above. **Could you rebase onto latest `main` once more to resolve it?** Handy to do in the same pass as fix 1, since they touch the same neighborhood.

Sorry for the extra round-trip! Everything else is verified and reproduces to the last digit. Thanks again 🙏

susiejojo:
The rebase resolved the conflict (branch is `MERGEABLE` again) and the refreshed numbers still reproduce exactly — thank you! 🙏 The one red check, `test (workload & cmd)`, is worth a quick note because the cause isn't obvious:

**What's failing:** `TestDocExamplesPassDeploymentFlags` (`cmd/docs_examples_test.go`), reproduced locally:

```
../docs/guide/kv-cache.md:156: documented example omits --hardware and --tp — both are
  required flags since #1733 (NS-6), so this command aborts as written
../docs/guide/kv-cache.md:160: documented example omits --hardware and --tp — ...
```

**Why:** #1754 (NS-6 / #1733) landed on `main` and made `--hardware` and `--tp` **required** on `blis run`/`blis replay`, plus a doc-lint test that asserts every documented example passes them. Your rebase pulled that test in, and it flags the two `blis run` commands in the new "Enabling and Disabling Offload" section — they predate the rule. (Only those two: the file's other examples at lines 7/32/63/290 already carry `--hardware H100 --tp 1`, since #1754 updated the pre-existing ones but couldn't know about this open PR's additions.)

**Fix:** add `--hardware H100 --tp 1` to both commands (lines 156 & 160), matching every other example in the file — that's also the config the numbers were measured on. Natural to fold into the same pass as the #1706 note below.

---

**Status of the earlier review — everything else is addressed** ✅

- ARC "evicts against" claim → fixed
- Line-pinned `blob/main` permalinks → removed
- "Same unit" sizing framing → reworked cleanly (new saturation calc verified)
- Rebase + number refresh → done, reproduces to the last digit
- Merge conflict → resolved

**Two small items remain, both one-line doc edits in the same area:**

1. **The stale "Chunked-prefill limitation (#1706)" note (lines 137–144)** — still present; #1725 closed #1706, so it now describes fixed behavior. (The `7ad59ca` "Remove stale section" commit removed the adjacent `## Chunked Prefill` section by mistake.)
2. **`--hardware H100 --tp 1` on lines 156 & 160** — the CI fix above.

Once those two land (and CI goes green), this is **ready to approve** from my side — it's a strong, well-evidenced doc. Thanks for bearing with the rounds of review! 🙏

leipanhz:
> The rebase resolved the conflict (branch is `MERGEABLE` again) and the refreshed numbers still reproduce exactly — thank you! 🙏 The one red check, `test (workload & cmd)`, is worth a quick note because the cause isn't obvious:
> 
> **What's failing:** `TestDocExamplesPassDeploymentFlags` (`cmd/docs_examples_test.go`), reproduced locally:
> 
> ```
> ../docs/guide/kv-cache.md:156: documented example omits --hardware and --tp — both are
>   required flags since #1733 (NS-6), so this command aborts as written
> ../docs/guide/kv-cache.md:160: documented example omits --hardware and --tp — ...
> ```
> 
> **Why:** #1754 (NS-6 / #1733) landed on `main` and made `--hardware` and `--tp` **required** on `blis run`/`blis replay`, plus a doc-lint test that asserts every documented example passes them. Your rebase pulled that test in, and it flags the two `blis run` commands in the new "Enabling and Disabling Offload" section — they predate the rule. (Only those two: the file's other examples at lines 7/32/63/290 already carry `--hardware H100 --tp 1`, since #1754 updated the pre-existing ones but couldn't know about this open PR's additions.)
> 
> **Fix:** add `--hardware H100 --tp 1` to both commands (lines 156 & 160), matching every other example in the file — that's also the config the numbers were measured on. Natural to fold into the same pass as the #1706 note below.
> 
> **Status of the earlier review — everything else is addressed** ✅
> 
> * ARC "evicts against" claim → fixed
> * Line-pinned `blob/main` permalinks → removed
> * "Same unit" sizing framing → reworked cleanly (new saturation calc verified)
> * Rebase + number refresh → done, reproduces to the last digit
> * Merge conflict → resolved
> 
> **Two small items remain, both one-line doc edits in the same area:**
> 
> 1. **The stale "Chunked-prefill limitation ([KV-offload prefill-shrink: credit the whole reloaded prefix under chunked prefill (follow-up to #1699) #1706](https://github.com/inference-sim/inference-sim/issues/1706))" note (lines 137–144)** — still present; [fix(kv-offload): credit the WHOLE reloaded prefix under chunked prefill (Closes #1706) #1725](https://github.com/inference-sim/inference-sim/pull/1725) closed [KV-offload prefill-shrink: credit the whole reloaded prefix under chunked prefill (follow-up to #1699) #1706](https://github.com/inference-sim/inference-sim/issues/1706), so it now describes fixed behavior. (The `7ad59ca` "Remove stale section" commit removed the adjacent `## Chunked Prefill` section by mistake.)
> 2. **`--hardware H100 --tp 1` on lines 156 & 160** — the CI fix above.
> 
> Once those two land (and CI goes green), this is **ready to approve** from my side — it's a strong, well-evidenced doc. Thanks for bearing with the rounds of review! 🙏

Hi @susiejojo , thanks for the review. The additional changes are added to the review. Hopefully this resolves all issues!

susiejojo (approved):
LGTM, please edit the PR description into the repo template before merging.

Issue #1715, opened 2026-09-15
feat(ci): add QA_VERDICT gate signal and wire qa-review into deliver-verify (round 0)


namasl:
## What problem does this solve?

The L1 delivery loop (#1654, hardened by #1686) gates a PR on CI, the archon plan signal, and one Anthropic reviewer (`blis-pr-review`, via the `DELIVER-VERDICT` marker → `AGENT_VERDICT`). It has **no cross-vendor review**. RFC #1603's whole thesis is that a decorrelated second opinion — a questioner and answerer from a different model family than the Opus code-writer and the Sonnet reviewer — catches what a single-vendor review misses. This PR makes the qa-review two-agent pass a **first-class, blocking gate signal** in the loop.

Depends on the qa-review tooling landing first (#1714).

## Proposed solution

Add a new gate signal **`QA_VERDICT ∈ {PASS, BLOCK, MISSING}`**, parallel to the existing `AGENT_VERDICT`, and produce it from a new verify-phase step that runs the committed `scripts/qa-review/` questioner + answerer against the PR head.

This touches exactly the three artifacts that `AGENT_VERDICT` already touches — the gate, its test, and the verify workflow — which is the whole reason a new parallel signal is cleaner than folding qa-review into the existing verdict.

### 1. `scripts/deliver-gate.sh` — the tested decision point

- Add `QA_VERDICT` to the required-env loop (fail-closed: unset/empty → `usage`/exit 2). The loop currently guards **six** vars (`CI_STATUS PLAN_GATE AGENT_VERDICT DISMISSALS ROUND MAX_ROUNDS`); `QA_VERDICT` becomes the seventh, guarded exactly like the other six.
- Add a domain `case`: `PASS | BLOCK | MISSING` accepted; anything else → `emit needs-human` with a message naming the offending value (catch-all, like the other domains).
- `QA_VERDICT=MISSING` → `needs-human` ("the verify phase posted no QA-VERDICT marker, so the qa-review verdict could not be read"), mirroring the existing `AGENT_VERDICT=MISSING` row.
- **`ready` additionally requires `QA_VERDICT=PASS`.** The only path to `ready` (CI success, plan `pass`/`absent`, `AGENT_VERDICT=GREEN`, `DISMISSALS=none`) gains this conjunct.
[...]

PR #1736, opened 2026-09-16
feat(ci): vendor the qa-review two-agent tooling to scripts/qa-review/

### LINKED ISSUE #1714: feat(ci): commit the qa-review two-agent tooling to scripts/qa-review/

The **qa-review** cross-vendor two-agent review exists today only as a user-level `/qa-review` skill (uncommitted, exercised on live PRs — the RFC #1603 AI-review prototype). To make it a standard, blocking gate in the L1 automated delivery loop (#1654, hardened by #1686), CI must be able to run it — and a GitHub Actions job cannot invoke a user-level Claude skill. The code has to live in the repository.

This is the first PR of the three-PR stack under epic **#1717**. It **only vendors the tooling and tests it** — it wires nothing into CI. That keeps it small, independently reviewable, and a clean prerequisite for the gate wiring (#1715) and the adjudicate-only re-verify (#1716).

## Proposed solution

Add the four qa-review scripts, a README, and tests under a new `scripts/qa-review/` directory (beside the existing delivery tooling `deliver-gate.sh`, `archon-*.sh`).

**Canonical source:** the working `/qa-review` prototype. Vendor those four scripts **verbatim** — do **not** reimplement the cross-vendor prompting, the JSON-repair regex, or the verdict taxonomies from scratch; they are tuned against the LiteLLM proxy and reinventing them will silently degrade the review. The contracts below are the acceptance surface, not a spec to re-derive.

### The two — and only two — deviations from byte-verbatim

Both are **strictly additive and behavior-preserving**: with the new surface unused, each script is byte-identical to the prototype. Neither touches a prompt, the JSON-repair regex, or a verdict taxonomy. They exist because the two downstream PRs cannot use the prototype as-is:

1. **Extract questioner.py's inline JSON-repair into a module-level `repair_json()`.** Today the repair (`re.sub(...)` → `json.loads`) lives inline in `main()` (questioner.py lines ~122–136, right after the POST). Move that exact block, unchanged, into a module-level `repair_json(payload) -> str` (or `parse_generated`) that `main()` then calls. This is what makes the repair path unit-testable (AC #2) without a live model. No regex change.
2. **Add an off-by-default `--no-exec` flag to `answerer.py` and `adjudicator.py` that drops the `go` build/test tool.** Both scripts register a `go` tool in **both** their `TOOLS_IMPL` and `TOOLS_SCHEMA` (`answerer.py`; `adjudicator.py`) that runs `go build ./...` / `go test` against the worktree — i.e. it *executes the code under review*. `--no-exec` removes `go` from **both** the implementation map and the advertised tool schema (leaving `read_file`/`grep`/`list_dir`, and the adjudicator's read-only `gh_issue`/`pr_diff`, intact). **Flag absent ⇒ the `go` tool is present ⇒ verbatim prototype behavior.** This seam is what lets #1715/#1716 run the answerer/adjudicator on the self-hosted runner while preserving `deliver-verify.yml`'s invariant that the PR's code is never compiled or executed there (see #1715). Implement it as a pure `tools_for(no_exec) -> (impl, schema)` selector so the drop is deterministically testable (AC #2).

All four scripts are stdlib-only Python 3 (`urllib`, `json`, `re`, `argparse`, `subprocess`) — **no new dependency**, no `requirements.txt`. They talk to the LiteLLM proxy over its OpenAI-compatible surface.

### Scripts and their contracts

1. **`questioner.py`** — cross-vendor probing-question generator.
   - Env: `OPENAI_BASE_URL`, `OPENAI_API_KEY` (falls back to `LITELLM_KEY`), `QA_QUESTIONER_MODEL` (default `gcp/gemini-3.6-flash`).
   - Emits 3 FIXED policy questions (`F1..F3`) verbatim plus model-generated topic-seeded questions (`G1..`) across 8 topics (correctness, invariants, rules, parity, tests, rationale, scope, errors).
   - One `POST /chat/completions`. Includes JSON-repair for invalid backslash escapes (cross-vendor models emit regex/shell like `\s` inside strings) — extracted to `repair_json()` per carve-out (1) above.
   - stdout: `{"model", "questions":[{id,topic,question}]}`.

2. **`answerer.py`** — cross-vendor answerer (agentic, read-only).
   - Env: `QA_ANSWERER_MODEL` (default `azure/gpt-5.6-sol`), same base/key vars.
   - Runs an OpenAI-compatible function-calling loop with **read-only tools sandboxed to `--worktree`** (`read_file`, `grep`, `list_dir`, and — unless `--no-exec` — a guarded `go build`/`go test`). Investigates the actual PR-head code to answer each question.
   - Per-question contract `status ∈ {CONFIDENT, CANNOT_ANSWER, FLAW_FOUND}` plus `answer`, `evidence` (file:line / repro), and an optional non-blocking `note`.
   - `--no-exec` (carve-out (2)) drops the `go` tool; all other flags/behavior unchanged.
   - stdout: `[{id,status,answer,evidence,note}]`.

3. **`render_report.py`** — report renderer (+ optional PR posting).
   - Verdict `BLOCK` iff any answer status ∈ `{FLAW_FOUND, CANNOT_ANSWER}`, else `PASS`.
   - Output shape (and nothing else): a one-line verdict **header** (`## qa-review — PR #N: ✅ PASS` / `⛔ BLOCK`), the full untruncated `ID | Topic | Result | Question | Answer` table, an **Items to fix** section (blocking findings + evidence), and an **Important to consider** section (non-blocking `note` fields). Note the verdict is a header emoji, **not** a trailing machine marker — deriving the `QA-VERDICT: PASS|BLOCK` gate marker from this output is #1715's job, not this PR's.
   - Flags: `--questions`, `--answers`, `--pr`, `--title`, `--qmodel`, `--amodel`, `--out`, `--post-to-pr`, `--repo`, `--banner`. `default_banner(qmodel, amodel)` is built from the actual models so it can't drift. Vendored verbatim (no seam).

4. **`adjudicator.py`** — author-defence adjudicator (used by #1716).
   - Env: `QA_ADJUDICATOR_MODEL` (default `azure/gpt-5.6-sol`), same base/key vars.
   - Reads the most recent qa-review comment's **Items to fix** (`parse_items_to_fix`) and every later comment (the author's responses), then per prior blocking finding verifies with read-only worktree tools + `gh_issue` + `pr_diff` and returns `RESOLVED` / `WAIVED_JUSTIFICATION` / `WAIVED_DEFERRED` / `STILL_OPEN`.
   - **Skeptical default-block** (defaults `STILL_OPEN`) and an **ambiguity-pushback rule** (on a genuine acceptance-criterion interpretation fork it returns `STILL_OPEN` and asks the author to pin the interpretation rather than guessing). Verdict `BLOCK` iff **any finding is `STILL_OPEN` or left un-adjudicated** (an id with no verdict blocks — see `render()`); the aggregate verdict is emitted on **stderr** as `[adjudication verdict: PASS|BLOCK]` (not a PR marker — #1716 derives the marker).
   - `--no-exec` (carve-out (2)) drops the `go` tool, keeping `gh_issue`/`pr_diff`. `--post-to-pr`.

5. **`README.md`** — how the pieces fit and the env surface. May adapt the prototype's `SKILL.md` orchestration notes (resolve PR → materialize the PR head into a throwaway read-only worktree → questioner → answerer → render/post → cleanup; `--adjudicate` mode), and must document the `--no-exec` seam and why it exists (the self-hosted-runner no-execution invariant #1715 relies on). Env knobs: `QA_REPO` (default `inference-sim/inference-sim`), `QA_REPO_DIR`, and the model/base/key vars above.

### Which components are affected?
- [x] New package/directory needed (`scripts/qa-review/`)
- [ ] No Go production code (`sim/`, `cmd/`) changes
- [ ] No CI workflow changes (deferred to #1715 / #1716)

## Acceptance criteria

1. `scripts/qa-review/{questioner.py,answerer.py,render_report.py,adjudicator.py,README.md}` are added, vendored from the working prototype, executable (`chmod +x`) with `#!/usr/bin/env python3` shebangs. The **only** deviations from byte-verbatim are the two additive, behavior-preserving seams above — (a) questioner's `repair_json()` extraction and (b) the off-by-default `--no-exec` flag on `answerer.py` and `adjudicator.py` — each a strict no-op when unused. No prompt, JSON-repair regex, or verdict taxonomy is regenerated.
2. **Tests over the deterministic surface, run entirely under `go test`.** The repo has an established shell-tool test precedent — `scripts/deliver_gate_test.go` (`package scripts_test`) drives `deliver-gate.sh` via `exec.Command` — and **no** Python test convention (no pytest/tox/pyproject/`*_test.py`). So follow the Go-subprocess precedent: add `scripts/qa_review_test.go` that shells out to `python3` (mirroring how `deliver_gate_test.go` shells out to `bash`), so every qa-review test runs under `go test ./scripts/...` with **no** new test framework and **no** CI wiring. It must cover the model-free surface:
   - **`render_report.py`** — invoke it with fixture `--questions`/`--answers` JSON and assert (i) `BLOCK` iff any status ∈ `{FLAW_FOUND, CANNOT_ANSWER}` else `PASS`, and (ii) the exact output shape (verdict header + full table + `Items to fix` + `Important to consider`).
   - **`questioner.repair_json()`** — a payload with invalid `\s`-style escapes parses after repair; an already-valid `\\` pair is not double-escaped (drive via `python3 -c` importing the module).
   - **`adjudicator.parse_items_to_fix()`** — extracts finding ids/text from a rendered qa-review comment; plus the `BLOCK` iff any `STILL_OPEN` **or un-adjudicated** verdict rule.
   - **`--no-exec` seam** — assert `tools_for(no_exec=True)` excludes `go` while `tools_for(no_exec=False)` includes it, for both `answerer.py` and `adjudicator.py` (deterministic, model-free).
   The model-calling paths (`questioner` generation, `answerer`/`adjudicator` tool loops) are **not** unit-tested here — they need the proxy — but every pure function around them is.
3. No new third-party dependency; scripts run under the repo's Python 3.
4. **Nothing is wired into CI in this PR** — no `.github/workflows/*.yml` changes. The scripts are inert until #1715.
5. `go build ./...`, `go test ./...`, and `golangci-lint run ./...` are unaffected (green) — including the new `scripts/qa_review_test.go`.

## Alternatives considered
- **Keep qa-review as a user-level skill and shell out to it from CI.** Impossible — a GitHub Actions runner has no access to a user's `~/.claude/skills/`. The code must be in-repo.
- **Reimplement from the contracts instead of vendoring.** Rejected: the cross-vendor prompts and JSON-repair are tuned; regenerating them is how the review quietly stops finding real flaws.
- **A `*_test.py` / pytest suite instead of Go-subprocess tests.** Rejected: the repo has zero Python test infrastructure; adding pytest to CI is out of scope for a vendoring PR, and the Go-subprocess precedent already runs shell-tool tests under `go test ./...` with no extra wiring.
- **Ship the scripts byte-verbatim with no seams and solve execution/testability downstream.** Rejected: the `go` tool cannot be disabled at the call site without a flag (so #1715 could not honor the no-execution invariant), and the inline repair cannot be tested without a live model. Both seams are additive no-ops, so "verbatim" is preserved for every existing behavior.

- Design: RFC #1603 (the two-agent evidence-gated AI-review + adjudication).
- Prerequisite for #1715 (QA_VERDICT gate signal + verify wiring) and #1716 (adjudicate-only re-verify).
- Part of the qa-review delivery-gate epic **#1717**.

**Depends on:** nothing.

mtoslalibu:
Closes #1714


namasl:

## qa-review — PR #1736: ⛔ BLOCK
_feat(ci): vendor the qa-review two-agent tooling to scripts/qa-review/ · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ❌ FLAW_FOUND | Does this PR fully implement the issue it closes, without unnecessary additions? | The scripts and README are present, the deterministic tests pass, and no CI workflow is wired, but AC #2 is not fully met: the renderer test does not assert the exact output shape. It only checks selected substrings, so additions, omissions, reordering, duplicated sections, or unexpected trailing output can pass. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The new README documents the four components, orchestration, model and proxy environment variables, QA_REPO/QA_REPO_DIR, the no-exec security seam, verdict behavior, and test strategy required by the issue. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No materially stale code comments were found. Comments describing the repair extraction, no-exec behavior, output contracts, and subprocess-test setup match the implementation. |
| G1 | correctness | ✅ CONFIDENT | In `fetch_comments()` in `adjudicator.py`, if a GitHub comment's `body` field is `None` (e.g. `{"body": null}` in the API response), `c.get("body", "")` evaluates to `None`, causing `"## qa-review — PR #" in c.get("body", "")` to raise a `TypeError`. How is `None` handled for comment bodies? | None is not explicitly handled: both the marker search and later response join assume comment bodies are strings. For the actual `gh pr view --json comments` producer, comment bodies are expected as strings, so this is defensive handling of a hand-crafted or schema-invalid response rather than a demonstrated production defect. |
| G2 | correctness | ✅ CONFIDENT | In `urllib.request.urlopen()` across `questioner.py`, `answerer.py`, and `adjudicator.py`, no HTTP timeout parameter is set. What prevents a hanging LiteLLM proxy connection from stalling a script or CI step indefinitely? | Nothing in these scripts imposes an HTTP timeout; `urlopen` uses its default timeout behavior. This matches the vendored prototype and the issue did not permit unrelated behavioral changes or wire these scripts into CI yet. |
| G3 | invariants | ✅ CONFIDENT | In `render_report.py`, `verdict_for(answers)` returns `PASS` when `answers` is an empty list (`[]`). Does this risk passing a PR gate if question answering returns an empty payload or fails to parse answers? | Yes, an empty answer list produces PASS, but that follows the explicit acceptance contract: BLOCK iff at least one answer has FLAW_FOUND or CANNOT_ANSWER. Parsing failures do not become empty lists—they raise before rendering. Completeness validation, if desired for the future gate, is outside this renderer contract. |
| G4 | invariants | ✅ CONFIDENT | How do `tool_list_dir` and `tool_grep` in `answerer.py` and `adjudicator.py` guarantee deterministic, environment-independent tool output ordering across different OS environments (e.g. locale-dependent `listdir` sorting or grep file traversal order)? | `tool_list_dir` explicitly sorts names using Python's deterministic Unicode string ordering. `tool_grep` does not normalize traversal order and relies on the project's actual Unix `grep` behavior; the repository's byte-identical simulation-output invariant does not apply to transient model tool observations. |
| G5 | rules | ✅ CONFIDENT | In `answerer.py` and `adjudicator.py`, when `--no-exec` is false, `tool_go` executes `go build ./...`, `go test ./...`, or `go vet ./...` via `subprocess.run` without checking exit codes, silently swallowing non-zero exit statuses and returning truncated stdout/stderr strings (`[-8000:]`). Does this comply with repository error-checking standards? | The subprocess exit code is not returned explicitly, but stdout and stderr—including normal Go build/test diagnostics—are returned to the model rather than discarded. This agent tool is outside the `sim/` library error-handling boundary, and its behavior is part of the prototype that this issue requires to remain unchanged except for the two named seams. |
| G6 | rules | ✅ CONFIDENT | In `adjudicator.py`, `_ITEM_RE` uses a middle dot character (`·`) to match finding headers (`- **F2 · FLAW_FOUND**`). Will this regex fail if comment bodies use standard dashes, spaces, or alternate UTF-8 representations generated by different LLM formatting outputs? | The parser intentionally matches the canonical format emitted by this repository's renderer, not arbitrary LLM formatting. The renderer always emits the same middle dot and em dash, so standard dashes or spaces are not produced on the shipped path. |
| G7 | parity | ✅ CONFIDENT | How is parity maintained between `TOOLS_SCHEMA` and `TOOLS_IMPL` when `--no-exec` is toggled in `answerer.py` versus `adjudicator.py`, ensuring the schema advertised to the model never references tools missing from the implementation map? | Each module's `tools_for` derives both outputs from its paired constants and removes exactly the `go` entry from both when no-exec is true. The tests verify presence in both surfaces when false, absence from both when true, preservation of all read-only tools, and a one-entry size difference. |
| G8 | parity | ✅ CONFIDENT | In `questioner.py`, fixed policy questions `F1..F3` are prepended to the generated questions. Does the prototype's ID numbering scheme expect model-generated questions to start at `G1` or continue sequentially after `F3`? | Generated questions intentionally use a separate G namespace and start at G1; they do not continue numerically after F3. This is documented and implemented consistently. |
| G9 | tests | ✅ CONFIDENT | In `scripts/qa_review_test.go`, `runPython` executes `python3` with `cmd.Dir = qaScript(t, ".")`. Will this helper properly locate `scripts/qa-review` when tests are run from subdirectories via `go test ./...` versus running from the repository root? | Yes. Go runs package tests with the package source directory as the test process's working directory, so for package `scripts`, `qa-review` resolves to `scripts/qa-review`. The helper converts that to an absolute path before assigning `cmd.Dir`. |
| G10 | tests | ✅ CONFIDENT | In `scripts/qa_review_test.go`, `TestRepairJSON` tests `repair_json()` by passing inline Python code strings to `python3 -c`. Why pass raw strings via `-c` instead of invoking `questioner.py` directly or testing via JSON fixture inputs? | The `-c` probe imports and calls the newly extracted pure function directly, which is the required model-free seam. Invoking questioner.py would enter `main`, require proxy credentials, and perform the live POST before exercising parsing; fixture files would not avoid that. |
| G11 | rationale | ✅ CONFIDENT | Why was `repair_json` implemented using custom regex substitution in `questioner.py` rather than using standard JSON parser fallback strategies or standard library string unescaping? | The regex is a targeted repair for invalid JSON backslash escapes emitted in regex or shell text. It preserves every valid JSON escape and doubles only otherwise-invalid backslashes. More general string unescaping could alter valid JSON semantics. Most importantly, the issue explicitly requires preserving the prototype's tuned repair regex verbatim. |
| G12 | rationale | ✅ CONFIDENT | What was the rationale for adding `--no-exec` as a flag on `answerer.py` and `adjudicator.py` rather than controlling tool availability via an environment variable (e.g. `QA_NO_EXEC`) matching the other configuration knobs? | A command-line flag makes the security-sensitive tool set explicit and deterministic at each invocation, defaults to prototype behavior when absent, and exposes a pure `tools_for(no_exec)` selector for tests. An ambient environment variable would be less visible and was not the seam specified by the acceptance criteria. |
| G13 | scope | ✅ CONFIDENT | Does `adjudicator.py`'s invocation of `gh issue view` and `gh pr diff` via `subprocess.run` introduce any implicit dependencies on external GitHub CLI authentication that are outside the scope of vendor-only tooling in PR #1736? | The adjudicator does depend on an installed and authenticated `gh` CLI for fetching comments, issues, diffs, and optional posting. That is intentional and within this issue's stated adjudicator contract, not an undeclared Python dependency; the scripts themselves remain stdlib-only. |
| G14 | scope | ✅ CONFIDENT | Are all environment variables documented in `scripts/qa-review/README.md` (such as `QA_REPO_DIR`) actively consumed by the vendored Python scripts in this PR? | No. `QA_REPO_DIR` is documented as the local clone from which orchestration creates the throwaway worktree, but none of the four Python scripts consumes it directly. They receive the materialized worktree through `--worktree`. The issue expressly required README documentation of this orchestration-level knob. |
| G15 | errors | ✅ CONFIDENT | In `parse_verdicts()` in `adjudicator.py`, if the model returns non-JSON text surrounding markdown code blocks, `json.loads(text)` will raise an unhandled `JSONDecodeError`. Is there a JSON repair or extraction fallback like `parse_generated()` in `questioner.py`? | No. `parse_verdicts` only strips a code fence when it begins the entire response, then calls json.loads. Surrounding prose or malformed JSON raises JSONDecodeError. This matches the prototype path, and the issue explicitly excludes model-calling paths from deterministic unit coverage and permits no additional parser rewrite. |
| G16 | errors | ✅ CONFIDENT | In `answerer.py`, if `json.loads()` fails on `tool_calls` function arguments, `arguments` defaults to `{}` and calls `fn(worktree, **{})`. If the tool has required arguments (e.g. `path` in `read_file`), how is the resulting `TypeError` handled and communicated back to the model? | The malformed argument JSON is replaced with `{}`; calling a tool that requires arguments then raises TypeError inside `run_tool`. `run_tool` catches that exception and returns a `tool error (<name>): <message>` string, which is appended as the tool-call response so the model can correct its request. |

### Items to fix
- **F1 · FLAW_FOUND** — The scripts and README are present, the deterministic tests pass, and no CI workflow is wired, but AC #2 is not fully met: the renderer test does not assert the exact output shape. It only checks selected substrings, so additions, omissions, reordering, duplicated sections, or unexpected trailing output can pass.
  _scripts/qa_review_test.go:133-171 builds a mustContain list and uses strings.Contains rather than comparing the complete expected report; it also does not assert the rendered evidence line despite AC #2 requiring exact output shape. Repro: `go test ./scripts/...` passes, but inserting arbitrary extra output into render_report.py would not necessarily fail TestRenderOutputShape._

### Important to consider
- **G1** — Using `(c.get("body") or "")` would cheaply harden both sites.
  _scripts/qa-review/adjudicator.py:425-431 obtains comments from `gh pr view`; scripts/qa-review/adjudicator.py:433-439 directly uses `c.get("body", "")` in membership and join operations. A synthetic `{body: null}` would raise TypeError._
- **G2** — The downstream CI integration should apply a job/process timeout, or a later change should add explicit network timeouts.
  _scripts/qa-review/questioner.py:125-133, scripts/qa-review/answerer.py:210-220, and scripts/qa-review/adjudicator.py:278-288 call `urllib.request.urlopen(req)` without a timeout; scripts/qa-review/README.md:18-21 states this PR does not wire the tooling into CI._
- **G3** — The downstream gate may want to reject an empty answer set or require one answer per question.
  _scripts/qa-review/render_report.py:51-56 returns PASS after iterating an empty list; scripts/qa-review/render_report.py:133-159 uses json.loads, so malformed answers raise rather than silently becoming []; README.md:54-66 documents the same iff rule._
- **G4** — Sorting grep result lines would make prompts more reproducible across filesystem layouts, but no required output contract depends on their order.
  _scripts/qa-review/answerer.py:88-102 and scripts/qa-review/adjudicator.py:125-135; docs/contributing/standards/invariants.md:61-70 scopes INV-6 to simulation stdout and identifies Go map ordering concerns._
- **G5** — Including `proc.returncode` would make a silent nonzero exit unambiguous, but that would be a prototype behavior change outside this PR's permitted deviations.
  _scripts/qa-review/answerer.py:105-122 and scripts/qa-review/adjudicator.py:138-150 capture and return combined stdout/stderr; docs/contributing/standards/principles.md:36-46 defines error strategies primarily for CLI/library/runtime layers._
- **G6** — Hand-edited comments using alternate separators would be ignored; accepting them would be optional defensive flexibility.
  _scripts/qa-review/render_report.py:97-105 emits `- **<id> · <status>** — <answer>`; scripts/qa-review/adjudicator.py:77-99 parses that exact canonical shape; scripts/qa_review_test.go:230-285 tests parsing a rendered-format comment._
- **G13** — The eventual runner setup must provide authenticated `gh`; this PR intentionally does not add CI wiring.
  _scripts/qa-review/adjudicator.py:153-172 defines gh_issue/pr_diff; scripts/qa-review/adjudicator.py:420-430 fetches PR comments through gh; scripts/qa-review/README.md:14-16 states the Python dependency boundary and README.md:68-79 documents the GitHub-backed adjudication flow._
- **G15** — A future hardening change could convert parse failures into an explicit BLOCK rather than terminating the adjudicator.
  _scripts/qa-review/adjudicator.py:333-339; the system prompt at adjudicator.py:68-70 instructs the model to return only JSON without fences; scripts/qa-review/README.md:121-127 says model-calling paths are not unit-tested here._

namasl:

## qa-review adjudication — PR #1736: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The change fixes the missing evidence-line assertion, but it does not resolve the original exact-output-shape finding. TestRenderOutputShape still uses independent strings.Contains checks, so unexpected output, duplicated sections, reordered table content, or extra trailing content can still pass. The section slice only enforces that Items to fix precedes Important to consider. To resolve this, compare the complete rendered report with an exact expected string, or add equivalently strict assertions covering the entire output, ordering, uniqueness, and absence of extra content. |

### Still blocking
- **F1** — The change fixes the missing evidence-line assertion, but it does not resolve the original exact-output-shape finding. TestRenderOutputShape still uses independent strings.Contains checks, so unexpected output, duplicated sections, reordered table content, or extra trailing content can still pass. The section slice only enforces that Items to fix precedes Important to consider. To resolve this, compare the complete rendered report with an exact expected string, or add equivalently strict assertions covering the entire output, ordering, uniqueness, and absence of extra content.
  _scripts/qa_review_test.go:145-161 checks expected fragments with strings.Contains rather than full-output equality. scripts/qa_review_test.go:168-179 bounds and checks only the Items to fix section and its evidence. The renderer's complete deterministic output is returned at scripts/qa-review/render_report.py:123, but the test does not compare against that complete expected output._

### Cleared
- _None._

namasl (changes requested):
See https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5703589710

namasl:

## qa-review adjudication — PR #1736: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The evidence-line gap was fixed, but the original broader finding remains: the test still uses substring checks rather than asserting the complete exact report. Arbitrary extra output, duplicated sections, and some reorderings could still pass. Replace these checks with full expected-output equality or an equivalently strict assertion that enforces all lines, ordering, section cardinality, and absence of unexpected output. |

### Still blocking
- **F1** — The evidence-line gap was fixed, but the original broader finding remains: the test still uses substring checks rather than asserting the complete exact report. Arbitrary extra output, duplicated sections, and some reorderings could still pass. Replace these checks with full expected-output equality or an equivalently strict assertion that enforces all lines, ordering, section cardinality, and absence of unexpected output.
  _scripts/qa_review_test.go:145-161 defines selected mustContain strings and validates them with strings.Contains. The section check at scripts/qa_review_test.go:168-179 only bounds and checks the Items-to-fix content; it does not reject extra or duplicated output or enforce the complete report shape. The renderer's complete output contract is documented at scripts/qa-review/render_report.py:9-17._

### Cleared
- _None._

namasl:

## qa-review adjudication — PR #1736: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The evidence-line gap was fixed, but the original finding also requires assertion of the exact complete output shape. The test still uses independent strings.Contains checks rather than comparing the rendered report with a complete expected value, so unexpected additions, duplicated content, and some reorderings or trailing output can still pass. Replace or supplement these checks with a full-output equality assertion, or equivalent strict assertions that reject all unexpected output and enforce the complete ordering. |

### Still blocking
- **F1** — The evidence-line gap was fixed, but the original finding also requires assertion of the exact complete output shape. The test still uses independent strings.Contains checks rather than comparing the rendered report with a complete expected value, so unexpected additions, duplicated content, and some reorderings or trailing output can still pass. Replace or supplement these checks with a full-output equality assertion, or equivalent strict assertions that reject all unexpected output and enforce the complete ordering.
  _scripts/qa_review_test.go:145-161 defines selected mustContain substrings and checks them with strings.Contains. Lines 168-179 only bound and inspect the Items-to-fix section. Lines 152 and 177-178 confirm the evidence line is now covered, but no assertion compares the entire output or rejects arbitrary extra content. `go test ./scripts/...` passes._

### Cleared
- _None._

namasl (changes requested):
See https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5705211430

namasl:

## qa-review adjudication — PR #1736: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The evidence-line gap was fixed, but the original exact-output-shape finding remains. The test still uses independent strings.Contains checks rather than comparing the complete rendered report, so unexpected additions, duplicated content, most reordering, and trailing output can still pass. There is also an unresolved acceptance-criterion ambiguity over whether “exact output shape” requires byte-for-byte expected output or only selected structural assertions. The author must either implement a complete expected-output comparison or cite where issue #1714 explicitly pins the weaker interpretation. |

### Still blocking
- **F1** — The evidence-line gap was fixed, but the original exact-output-shape finding remains. The test still uses independent strings.Contains checks rather than comparing the complete rendered report, so unexpected additions, duplicated content, most reordering, and trailing output can still pass. There is also an unresolved acceptance-criterion ambiguity over whether “exact output shape” requires byte-for-byte expected output or only selected structural assertions. The author must either implement a complete expected-output comparison or cite where issue #1714 explicitly pins the weaker interpretation.
  _scripts/qa_review_test.go:145-161 checks selected substrings with strings.Contains. Lines 168-179 enforce only the relative bounds and contents of the Items-to-fix section; they do not validate the complete report. The newly added evidence assertions are present at lines 152, 155, and 177-179. `go test ./scripts/...` passes, but that does not cover arbitrary extra or duplicated output._

### Cleared
- _None._

namasl (changes requested):
**Requesting changes** — one item to address before this is ready.

### `TestRenderOutputShape` does not assert the exact output shape (issue #1714, AC #2)

`scripts/qa_review_test.go` verifies `render_report.py`'s output with a list of `strings.Contains` fragment checks plus a bounded scan of the "Items to fix" section. That confirms certain lines are *present*, but it does not verify the **complete** rendered report. As written, the test still passes if the renderer emits an extra line, duplicates a section, reorders the table rows, or appends trailing output — none of which the current assertions catch. AC #2 calls for asserting the exact output shape, and substring checks don't meet that bar.

**What to change:** assert the full rendered report by exact equality instead of substrings. `render(...)` is fully deterministic for fixed inputs — `renderFixture` pins the questions fixture, `--pr 42`, `--title "My PR"`, `--amodel azure/gpt-5.6-sol`, and passes no banner — so the entire report is reproducible. Build the complete expected string and compare the whole thing:

```go
want := strings.Join([]string{
    "## qa-review — PR #42: ⛔ BLOCK",
    "_My PR · 2 questions · questioner `<default-qmodel>` · answerer `azure/gpt-5.6-sol`_",
    "",
    "| ID | Topic | Result | Question | Answer |",
    "|----|-------|--------|----------|--------|",
    "| F1 | fixed | ✅ CONFIDENT | Implements the issue? | Fully implemented. |",
    "| G1 | tests | ❌ FLAW_FOUND | Covered by tests? | A test is missing. |",
    "",
    "### Items to fix",
    "- **G1 · FLAW_FOUND** — A test is missing.",
    "  _y_test.go:9_",
    "",
    "### Important to consider",
    "- **F1** — a minor doc nit",
    "  _x.go:1_",
}, "\n") + "\n"

if out != want {
    t.Errorf("rendered report != expected\n--- got ---\n%s\n--- want ---\n%s", out, want)
}
```

Fill `<default-qmodel>` and the exact question text from the `rrQuestions` fixture. A whole-output equality check subsumes every fragment assertion you have now and additionally fails on any added, missing, reordered, duplicated, or trailing line — which is what "exact output shape" means.

If you deliberately intend the weaker structural check rather than byte-for-byte equality, then pin that intent explicitly — state it in the test's doc comment and in issue #1714 — and still add assertions that enforce line ordering, exactly-one-of each section, and the absence of any unexpected content. But full-output equality is the simpler and stronger option, and I'd prefer it here.

Everything else looks good — the four scripts, the module-level `repair_json()` extraction, the off-by-default `--no-exec` seam on `answerer.py`/`adjudicator.py`, and the README all match the issue. This test is the only blocker.

namasl:

## qa-review adjudication — PR #1736: ⛔ BLOCK

|----|----|----|----|
| F1 | FLAW_FOUND | ❌ STILL-OPEN | The fix adds evidence-line assertions and bounds the Items-to-fix section, but it still does not assert the complete rendered output. The test continues to use independent strings.Contains checks, so unexpected output, duplicated sections or rows, and some reorderings can still pass. To resolve the original finding, compare the complete output with an expected report or add equivalent strict assertions covering exact order, uniqueness, and absence of extra/trailing content. |

### Still blocking
- **F1** — The fix adds evidence-line assertions and bounds the Items-to-fix section, but it still does not assert the complete rendered output. The test continues to use independent strings.Contains checks, so unexpected output, duplicated sections or rows, and some reorderings can still pass. To resolve the original finding, compare the complete output with an expected report or add equivalent strict assertions covering exact order, uniqueness, and absence of extra/trailing content.
  _scripts/qa_review_test.go:145-161 checks a mustContain list using strings.Contains. The section-specific checks at scripts/qa_review_test.go:168-179 verify only the presence and relative placement of the two section headers plus the blocking text/evidence; they do not reject duplicate or unexpected content. The renderer's complete output contract is documented at scripts/qa-review/render_report.py:9-17._

### Cleared
- _None._

namasl (changes requested):
**Requesting changes — this is a demonstrated defect, not a theoretical one.**

`TestRenderOutputShape` in `scripts/qa_review_test.go` does not assert the exact output shape that issue #1714 AC #2 requires. It checks a `mustContain` list with `strings.Contains`, plus a bounded scan of the Items-to-fix section. Because every assertion is containment-only, malformed reports pass. I verified this on the current head (`d5565180`): each mutation below is a one-line edit to `scripts/qa-review/render_report.py` that produces a **wrong report**, yet `go test ./scripts/... -run TestRenderOutputShape` **still passes**.

1. **Stray trailing line.** Add `lines.append("STRAY DEBUG LINE — renderer is broken")` just before `render()`'s `return`. → **PASS.** The report now carries garbage after `### Important to consider`; nothing is asserted past that header.
2. **Doubled table.** Emit the `| ID | Topic | Result | Question | Answer |` header + `|----|...|` separator twice. → **PASS.** A report containing two tables is accepted.
3. **Duplicated section content.** Inject an extra `- **DUP · FLAW_FOUND** — ...` bullet into `### Items to fix`. → **PASS.** The bounded-section check only asserts the real finding is *present*, not that nothing else is.

Repro (any of them): apply the one-line edit, run `go test ./scripts/... -run TestRenderOutputShape`, observe it stays green. That is precisely the class of regression "exact output shape" exists to catch, and the current test does not.

On interpretation: "the exact output shape (verdict header + full table + Items to fix + Important to consider)" cannot be satisfied by a test that accepts a doubled table, a duplicated section, or trailing garbage — under **any** reading, structural included. Mutations 1–3 are wrong renderers that ship green. "AC #2 doesn't require full-string equality" does not answer them.

### Fix to implement

Assert the complete rendered output by equality, not substrings. `render(...)` is fully deterministic for the fixture inputs (fixed questions, `--pr 42 --title "My PR" --amodel azure/gpt-5.6-sol`, no banner), so build the whole expected report and compare `out == want`:

```go
// Pin the subtitle by passing an explicit --qmodel in renderFixture so `want` is fully determined.
want := strings.Join([]string{
	"## qa-review — PR #42: ⛔ BLOCK",
	subtitle, // "_My PR · 2 questions · questioner `<qmodel>` · answerer `azure/gpt-5.6-sol`_"
	"",
	"| ID | Topic | Result | Question | Answer |",
	"|----|-------|--------|----------|--------|",
	"| F1 | fixed | ✅ CONFIDENT | Implements the issue? | Fully implemented. |",
	"| G1 | tests | ❌ FLAW_FOUND | Covered by tests? | A test is missing. |",
	"",
	"### Items to fix",
	"- **G1 · FLAW_FOUND** — A test is missing.",
	"  _y_test.go:9_",
	"",
	"### Important to consider",
	"- **F1** — a minor doc nit",
	"  _x.go:1_",
}, "\n") + "\n"
if out != want {
	t.Errorf("rendered report != expected\n--- got ---\n%s\n--- want ---\n%s", out, want)
}
```

Exact-equality subsumes every current assertion and fails all three mutations above. (If you prefer to keep structural assertions, they must additionally reject extra, duplicated, and trailing lines and enforce section uniqueness — but equality is simpler and complete.)

### If you believe no change is warranted

That's a legitimate outcome — but the justification must engage these three mutations directly: for each, explain why a renderer that emits that malformed output while the test stays green is acceptable under AC #2. A general restatement that full-string equality isn't required does not address mutations 1–3, which are wrong under the structural reading too.

Everything else in the PR is fine — the four scripts, the module-level `repair_json()`, the off-by-default `--no-exec` seam, and the README all match the issue. This test is the only blocker.

namasl (commented):
The requested change is now on this branch — pushed as `a1ad1651`.

`TestRenderOutputShape` now asserts the **complete** rendered report byte-for-byte (a `want := strings.Join(...)` equality check against `render_report.py`'s full output), alongside the existing substring/section checks. This closes the gap the earlier review raised: the three malformed-output cases that previously passed the test — a stray trailing line, a doubled results table, and a duplicated `Items to fix` bullet — now each **fail** it (verified locally by re-applying them).

Change is test-only (`scripts/qa_review_test.go`, +28 lines); `gofmt -l` is clean and `go vet ./scripts/...` + `go test ./scripts/...` are green. Please re-verify on the current head.

namasl (approved):

mtoslalibu:
## Review notes

Ran code-review, silent-failure, and test-coverage analysis on the vendored scripts. Architecture is clean — no overengineering, the `--no-exec` seam is well-designed, and the byte-exact output-shape test is excellent. A few things worth addressing before merge:

---

### Bugs

**1. Tool-loop exhaustion crashes instead of producing a result**
`answer_loop` and `adjudicate_loop` — when `MAX_TOOL_TURNS` (24) is hit, the fallback `messages[-1].get("content", "")` grabs the last *tool-result* message (a raw string like file contents), not the last assistant reply. `parse_answers`/`parse_verdicts` then calls `json.loads` on that and crashes with an unhelpful `JSONDecodeError`.
- `answerer.py` ~L167, `adjudicator.py` ~L206
- Fix: track the last assistant message separately, or search backwards for `role == "assistant"`.

**2. `_load()` crashes with confusing error when neither arg is provided**
`render_report.py`'s `_load("", "")` passes empty string to `open()`, raising `FileNotFoundError: [Errno 2] No such file or directory: ''`. Easy to hit if `--questions` and `--questions-inline` are both omitted.
- Fix: validate in argparse or check before `open()`.

---

### Silent-failure concerns

**3. Malformed tool arguments silently cascade through three fallback layers**
When a model emits broken JSON in `arguments`:
1. `json.loads` fails → falls back to `{}`
2. Tool is called with no args → raises `TypeError`
3. `run_tool`'s bare `except Exception` converts it to a string the model sees as normal output

The operator never sees that the model is producing garbage. All three layers are individually reasonable but compound into invisible failure. A `sys.stderr.write` on the `JSONDecodeError` in step 1 would break the silence.
- `answerer.py` ~L155, `adjudicator.py` ~L198

**4. `run_tool` bare except — consider logging to stderr**
The `except Exception` catch-all is a valid design choice (surface errors to the model, don't crash the loop), but operators get zero visibility into tool failures. A one-line `sys.stderr.write` before returning the error string would help CI debugging without changing behavior.

---

### Test gaps

**5. `_safe_path` has zero test coverage**
This is the sandbox boundary that prevents the model from reading outside the worktree. No test verifies that `../../etc/passwd` is rejected. Worth a small test given it's a security-relevant function.

**6. `repair_json` tests are narrow**
Only covers `\s` (invalid) and `\\` (valid). Doesn't test that valid escapes like `\n`, `\t`, `\uXXXX` are left intact, or that adjacent valid+invalid patterns (`\\\s`) work. These are the exact patterns the docstring says cross-vendor models emit.

---

### Minor / optional

- `tool_grep` returns `""` for both no-match (exit 1) and error (exit 2) — indistinguishable.
- `questioner.py` silently emits entries with `"question": ""` if the model returns empty questions; could waste an answerer turn and produce a spurious BLOCK.
- `adjudicator.render()` output shape is untested (unlike `render_report.py` which has byte-exact coverage).
- HTTP errors from `post_chat_completion` produce raw tracebacks — fine for now since nothing is wired into CI yet, but worth a `try/except HTTPError` before #1715 lands.

namasl (changes requested):
See https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5715126286https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5715126286https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5715126286https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5715126286https://github.com/inference-sim/inference-sim/pull/1736#issuecomment-5715126286

@mtoslalibu in the future, submit a review with changes requested so the workflow triggers off it

namasl (approved):

jgchn (approved):

PR #1749, opened 2026-09-16
test(latency): catalog↔fixture ModelConfig equivalence gate (Closes #1748)

### LINKED ISSUE #1748: test(latency): catalog↔fixture ModelConfig equivalence gate for R1 fresh-fetch configs (blis-catalog#1 merge dep)
**Describe the task**

Merge-dependency gate for blis-catalog PR #1 (inference-sim/blis-catalog#1), R1 option-(b) decision: the catalog ships *fresh, verbatim-from-vendor* `config.json` files rather than the older trimmed/stale `model_configs/` fixtures. For three models the upstream config is *fuller* than the fixture it replaces (`glm-5.2-fp8`, `llama-2-7b-hf`, `llama-3.1-8b-instruct`). R1 acceptance test #3 requires byte-identical stdout, which holds only if BLIS parses equivalent values from the fuller configs. This issue enshrines that as an automated regression (an ad-hoc `reflect.DeepEqual` run is not a gate).

**What to add**

A table test (suggested home: `sim/latency/config_test.go`) that, for each of the three models, parses **both** the catalog `config.json` and the retained `model_configs/<name>/config.json` fixture through `latency.GetModelConfig` and asserts the resulting `*sim.ModelConfig` are equal:

```go
for _, name := range []string{"glm-5.2-fp8", "llama-2-7b-hf", "llama-3.1-8b-instruct"} {
    catalog, err := latency.GetModelConfig(catalogPath(name))   // blis-catalog models/<name>/config.json
    require.NoError(t, err)
    fixture, err := latency.GetModelConfig(fixturePath(name))   // model_configs/<name>/config.json
    require.NoError(t, err)
    require.Equal(t, fixture, catalog, "parsed ModelConfig must match for %s", name)
}
```

Additionally cover the one **read** key that differs between fixture and upstream — `llama-3.1-8b-instruct`'s `rope_scaling` (added upstream, type `llama3`). Assert the effective-max-length path is unchanged:

- `applyRopeScaling(maxPosEmb, "", {"rope_type":"llama3","factor":8.0,...})` returns `(maxPosEmb, false)` — i.e. `llama3` is blacklisted and does **not** scale `max_position_embeddings`. (This is why the fuller config is equivalent; the test pins it so a future change to `applyRopeScaling`'s blacklist can't silently break byte-identity for Llama 3.1.)

**Verified baseline (why this is expected to pass)**

Running the three catalog configs and their fixtures through `latency.GetModelConfig` today yields byte-identical `ModelConfig` (`reflect.DeepEqual` true) for all three; the extra upstream keys (`eos_token_id`, `pad_token_id`, `transformers_version`, `modules_to_not_convert`, DSA `index_*`/`indexer_*`, …) are not bound by `HFConfig`/`ModelConfig`, and `rope_scaling`=`llama3` is a no-op in `applyRopeScaling`. The test locks that in.

**Scope notes**

- `kimi-k3` is intentionally **excluded** from this equivalence test: it carries corrected real-model values (`kv_lora_rank: 512`, `moe_intermediate_size: 3072`, 24 full-attn layers) that legitimately differ from the older stale copy; it is not an equivalence case.
- The concrete `catalogPath`/`fixturePath` wiring depends on how the catalog loader (`--catalog`/`BLIS_CATALOG`, S4/S6) is implemented; this test can start by pointing at a vendored copy of the three catalog configs under `testdata/` and migrate to the real loader when it lands.

**Merge dependency**

blis-catalog PR #1 should not merge to the point of being consumed by the loader until this regression exists. Linking both directions.

susiejojo:
## R1 migration gate: catalog↔fixture ModelConfig equivalence

Closes #1748. Merge dependency for **blis-catalog PR #1** (inference-sim/blis-catalog#1).

### Why

blis-catalog PR #1 ships **fresh, verbatim-from-vendor** `config.json` files rather than the older trimmed `model_configs/` fixtures. For three models the upstream config is *fuller* than the fixture it replaces (extra tokenizer/runtime/quantization keys the vendor ships that BLIS doesn't read):

- `glm-5.2-fp8` — 51-line hand-authored stub → 776-line upstream config
- `llama-3.1-8b-instruct` — adds `rope_scaling`, array `eos_token_id`, `pretraining_tp`
- `llama-2-7b-hf` — adds `_name_or_path`, drops `pad_token_id`

R1 **acceptance test #3** requires byte-identical stdout after the migration. That holds only if BLIS parses an *equivalent model shape* from the fuller config. This PR makes that an automated gate instead of an ad-hoc check.

### What

A table test (`sim/latency/catalog_equivalence_test.go`) that, for each of the three models, parses **both** the vendored catalog `config.json` (`testdata/catalog_configs/`) and the retained `model_configs/` fixture through the real loader `latency.GetModelConfig`, and asserts the resulting `*sim.ModelConfig` are `reflect.DeepEqual`.

- **Verified non-vacuous:** perturbing a read field (e.g. `num_hidden_layers` → 999) makes the test fail; restoring it passes.
- **Passes today** for all three — the fuller configs parse to identical shape.

### Scope / lifecycle

Intentionally a **migration gate**, not a permanent regression. It compares against the `model_configs/` fixture that **S6 will remove**; its job is to prove the fresh-fetch migration is behaviour-preserving *before* the old baseline is deleted. Once S6 lands, it is replaced by a golden `*sim.ModelConfig` keyed directly off the catalog config (no fixture needed).

The one read-relevant key that differs — `llama-3.1`'s `rope_scaling` (type `llama3`) — is **not** read by `GetModelConfig`; it's consumed only by `cmd.applyRopeScaling`, where `llama3` is blacklisted (a no-op) and already covered permanently by `cmd/root_test.go:TestApplyRopeScaling` (`"llama3 excluded" → wantApplied:false`). So `ModelConfig` equality is the complete equivalence claim at this layer.

`kimi-k3` is intentionally **not** included: it carries corrected real-model values that legitimately differ from an older stale copy; it is not an equivalence case (and was not a C1 fixture).

### Testdata provenance

`testdata/catalog_configs/{glm-5.2-fp8,llama-2-7b-hf,llama-3.1-8b-instruct}.json` are byte-identical copies of the corresponding `models/<name>/config.json` in blis-catalog @ `f047360`, themselves byte-identical to the recorded upstream HF revisions.


namasl (approved):
Reviewed as the blis-catalog#1 migration dependency. The test exercises the real `latency.GetModelConfig` path for all three refreshed fixtures, the three vendored inputs are byte-identical to blis-catalog head `ca81210`, the existing command test covers the separate Llama 3.1 `applyRopeScaling` no-op, and the full CI matrix is green.

susiejojo:
@claude /blis-pr-review

Issue #1746, opened 2026-09-16
hardening: cluster-only conservation counters are absent from stdout, so no CLI test can assert INV-1


mtoslalibu:
## Summary

INV-1's canonical form is a twelve-term cluster equation. Seven of those twelve terms cannot be observed from `blis run` stdout, so no CLI-level test can assert the invariant — the one in `cmd/` is structurally stuck at the five-term single-instance form even when it is checking a cluster.

## Where the gap is

The seven cluster-only counters exist on `cluster.RawMetrics` (`sim/cluster/metrics.go`, the block commented `INV-1 extended`):

- `GatewayQueueDepth`, `GatewayQueueShed`, `GatewayQueueRejected`, `GatewayEvicted`, `GatewayExpired`
- `RoutingRejections`, `EncodeRoutingRejections`

The JSON emitted per instance and for the cluster aggregate is `sim.MetricsOutput` (`sim/metrics_utils.go`). It carries `completed_requests`, `still_queued`, `still_running`, `injected_requests`, `dropped_unservable` and `timed_out_requests` — none of the seven above.

`RawMetrics` is built by `CollectRawMetrics`, called only from `cmd/root.go` and `cmd/replay.go`, and is consumed for fitness evaluation rather than serialised into the metrics objects.

## Consequence

`clusterConservationHolds` (`cmd/dp_placement_test.go`, 12 call sites across `cmd/dp_placement_test.go` and `cmd/dp_replay_parity_test.go`) parses the `instance_id: "cluster"` object and asserts the five-term form. It passes today only because its fixtures use default admission with no flow control, so all seven missing buckets are zero. A CLI test with `--flow-control-enabled` would either assert nothing meaningful or fail with no way to see which bucket the requests went to.

It also means an operator reading `blis run` output cannot reconcile the numbers: with the gateway queue active, `injected_requests` will not equal the sum of the buckets they can see, and nothing in the output explains where the difference went.

## How to verify the gap

```
go run . run --num-instances 2 --flow-control-enabled --flow-control-detector concurrency \
  --flow-control-max-concurrency 1 --num-requests 20 2>/dev/null | grep '"instance_id": *"cluster"'
```

The emitted object has no `gateway_*` or `*_rejections` field.

## Suggested fix

[...]

Issue #1745, opened 2026-09-16
hardening: INV-1's exclusivity clause is asserted by nothing, and no sum can assert it


mtoslalibu:
## Summary

INV-1 makes two promises. Conservation — every injected request is accounted for — is now asserted everywhere by a shared twelve-term helper (#1720). The second promise is not asserted anywhere, and cannot be by any conservation sum.

From `docs/contributing/standards/invariants.md` (the cluster-only term definitions paragraph):

> Each terminal bucket is mutually exclusive — a request lands in exactly one.

## Why a sum cannot check this

The helper asserts `injected == sum(twelve buckets)`. Consider a run where one request is counted in **both** `still_running` and `gateway_evicted`, and a different request is lost entirely. The sum is +1 from the double-count and −1 from the loss, so it balances and the assertion passes. Two real bugs, cancelling.

This is not hypothetical in shape: the buckets are incremented from different code paths (instance finalisation, the gateway queue, the router, the encode pool), and a request that transitions through more than one of them relies on each site removing it from the previous container. There is no single place that records "this request's terminal bucket is X".

## How to check it

Asserting exclusivity needs per-request terminal-bucket information, which does not exist today. Two shapes:

1. **A terminal-bucket field on `Request`**, written exactly once at each terminal transition, with a guard that panics (or counts) on a second write. Then `sum(buckets) == accounted` can be cross-checked against the tally of per-request tags, and a double-write is caught at the moment it happens rather than inferred from a mismatch.
2. **A terminal-transition ledger** on the simulator — an append-only list of `(requestID, bucket, tick)` — asserted in tests to contain each request exactly once. Heavier at runtime but needs no `Request` field, and it also gives INV-2 the transition record it currently lacks.

Option 1 is closer to the existing pattern: `sim/cluster/instance.go` already has `validInstanceTransitions` plus a `TransitionTo()` that panics on an illegal edge, for *instances*. Requests have no equivalent.

Either way this is a production change, not a test change, which is why #1720 (test-only) left it out.

## Context

Found while reviewing the plan for #1720. That issue split each invariant statement into clauses and asked which clause each test actually asserts — this is a clause neither the issue nor the first draft of the fix had noticed, so it is a direct instance of the pattern #1720 exists to find.

Recorded in the registry under INV-1 as a known-unasserted clause, so the twelve-term equality is not read as a completeness proof.

[...]

PR #1734, opened 2026-09-16
hardening(invariants): make every code-cited INV-* ID resolvable, fix INV-1's contradiction (Closes #1721)

mtoslalibu:
## Summary

The invariant registry (`docs/contributing/standards/invariants.md`) had two problems. **First**, INV-1's statement contradicted itself: it said a five-term request-conservation equation held "at simulation end (all levels)", then the next paragraph gave a *twelve*-term cluster equation as an "extension". A developer writing a cluster test and asserting five terms was following the document correctly — which is why #1720 found 18 cluster tests doing exactly that. **Second**, seven invariant IDs that the code cites by ID were documented nowhere, including `INV-A`, which has a runtime verifier and prints `"INV-A violation on node %s"` at 11 call sites. Someone reading that error had nowhere to look it up.

This PR makes the twelve-term form canonical with the five-term form as an explicitly single-instance specialisation, documents every ID the code cites, and regroups the registry so it stays readable as it grows. Documentation, plus two comment-only edits in Go files.

**Merge readiness: ready.** No behaviour change — the only Go edits are comments. `go build ./...`, `go vet ./sim/...`, `go test ./sim/... ./cmd/...` (13 packages) and `mkdocs build --strict` all pass.

## What changed, in plain terms

- **INV-1 now states one equation, not two contradictory ones.** Twelve terms is the canonical cluster form; five terms is labelled the single-instance specialisation; and the entry says explicitly which to assert where. It also fixes a miscount — the doc said "all six terms are always zero" for single-instance, but there are seven cluster-only terms.
- **Three new full entries**: `INV-A` (per-node `allocated_gpus + free_gpus == total_gpus`), `INV-A2` (a failed `PlaceInstance` must be logged, never silently dropped), `INV-W3` (cohort expansion is pure).
- **The LoRA family (`INV-L1`–`INV-L7`) gets pointer lines.** The design plan stays authoritative; the registry just makes lookup work, per the scoping decision in #1721's comment.
- **`INV-L2` gets a code anchor.** It was declared in the plan and cited nowhere — a capacity bound with nothing pointing at it. Now anchored at the load-start slot reservation the plan identifies.
- **Grouping**: run-level and subsystem (graded by scope) plus a cross-cutting code-boundary group (graded by kind), with an ID-ordered index. This is the "group by tier" ask from #1721's comment.
[...]

PR #1752, opened 2026-09-16
hardening(invariants): assert the invariant clauses no test was checking (Closes #1720)

### LINKED ISSUE #1720: hardening: invariant clauses that no test asserts (conservation missing GatewayExpired, INV-5 chain, INV-9 whitelist)
## Summary (read this part)

Follow-up to #1718. That issue asked "is each invariant *mentioned* in the code?" This one asks the harder question: **is every clause of each invariant statement actually asserted by a test?**

We split invariant statements into their individual clauses and compared each against what the tests really check. Four findings, one of which is a latent defect.

**Finding 1 — the conservation equation is 12 terms in the doc, and 3–6 terms in almost every test.**

INV-1 at cluster level names twelve buckets. There is no shared helper, so every test hand-rolls its own sum, and they disagree about how many terms conservation has. Across **29 conservation-style assertions** in `sim/` and `cmd/`:

| Term | Present in |
|---|---|
| `CompletedRequests` | 29 / 29 |
| `StillQueued`, `StillRunning` | 27 / 29 |
| `DroppedUnservable` | 22 / 29 |
| `TimedOutRequests` | 13 / 29 |
| `RoutingRejections` | 6 / 29 |
| `GatewayQueueDepth`, `GatewayQueueShed`, `GatewayQueueRejected`, `GatewayEvicted`, `EncodeRoutingRejections` | **1 / 29** |
| `GatewayExpired` | **0 / 29** |

Most sites assert 3–6 terms. One asserts 11.

**And the 11-term one is missing a bucket.** `sim/cluster/epd_test.go:186` is the most complete INV-1 assertion in the codebase, and it omits `GatewayExpired`:

```go
accounted := m.CompletedRequests + m.StillQueued + m.StillRunning + m.DroppedUnservable +
    m.TimedOutRequests + cs.RoutingRejections() + cs.EncodeRoutingRejections() +
    cs.GatewayQueueDepth() + cs.GatewayQueueShed() + cs.GatewayQueueRejected() +
    cs.GatewayEvicted()
// GatewayExpired is absent
```

`GatewayExpired` is a live metric — incremented in `cluster.go`, exposed at `cluster.go:1610`, and its own count is tested in `gateway_queue_ttl_test.go:34`. It is simply in no conservation sum anywhere.

This is **latent, not currently breaking**: gateway TTL defaults off, so `GatewayExpired` is always 0 in that test today. But the first person to write a conservation test with `--gateway-ttl` enabled gets a false violation, and a genuine leak into the TTL bucket is invisible right now.

There is also a *second, different* conservation equation in `progress_hook_test.go:185` (`sum(ShedByTier) == RejectedRequests + GatewayQueueShed + GatewayEvicted + GatewayExpired`) which does include it. So the suite contains two conservation equations and neither matches the doc's twelve terms.

**Finding 2 — INV-5 states a 4-term chain; every test asserts a 2-term derived form.**

The statement is `arrival <= enqueue <= schedule <= completion`. Every test citing INV-5 checks the same two things instead — `TTFT >= 0` and `E2E >= TTFT`. Both are durations measured *from arrival*, so that only says the last token follows the first. **The enqueue and schedule links are never asserted.** The timestamps exist (`GatewayEnqueueTime`, `RequestSchedulingDelays`); nothing compares them.

Notably the code is stricter than the tests here — `cluster.go:2006` carries a defensive comment about "completion before arrival", and `infra_lifecycle_event.go:273` reasons about the full chain explicitly.

**Finding 3 — INV-9's test is a hardcoded whitelist and three routing files are not on it.**

`TestINV9_OracleKnowledgeBoundary_NoOutputTokensInControlPlane` is, for our money, the best-designed invariant test in the repo — it scans control-plane *source* for `OutputTokens` rather than testing behaviour. But the file list is hardcoded, and `sim/router_state.go`, `sim/routing_nohit_lru_scorer.go`, and `sim/routing_precise_prefix_scorer.go` are absent from it. **We checked all three: clean, no violation.** Two are scorers, though — exactly the class already covered — so the next scorer added will also escape silently.

**Finding 4 — INV-7 has no Go test, and that is probably correct.**

INV-7 isn't a true/false predicate; it's a table describing which routing signals are synchronous vs periodic. It was validated by the H3/H29 hypothesis experiments instead. Fine — but the doc doesn't say so, so the absence reads as an oversight rather than a decision.

**What we propose here:** one shared conservation helper (replacing 29 hand-rolled sums), one new INV-5 test, a whitelist→glob change in the INV-9 test, and one doc sentence for INV-7.

**Deliberately not in this issue,** both discussed at the bottom: INV-2's missing transition guard (needs a design decision first) and migrating 8 of 9 INV-13 parity tests to byte-identical stdout (a sequence of PRs).

---

An invariant statement is usually several promises in one sentence. A test that references the invariant may assert only one of them. `INV-2` is the clearest case — *"requests transition queued → running → completed. **No invalid transitions.**"* — where the second clause has no enforcement and no test, while the first is well covered. A name-level search sees "INV-2 appears in 5 test files" and reports it covered.

The result is invariants that look protected and are not. Finding 1 is the concrete instance: 29 assertions that all claim to check conservation, none of which checks the conservation the doc defines.

Method: split each statement into clauses, locate what asserts each clause, read it, and judge. All counts below are against `07622594`.

- [x] Request conservation
- [x] Request lifecycle
- [x] Causality
- [x] Signal freshness
- [ ] KV block conservation
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Work-conserving

**Proposed fix**

### Part 1 — one shared conservation helper, used everywhere

Add a single helper that implements the equation once, taken from the field list INV-1 names:

```
AssertRequestConservation(t, metrics)          // single-instance, 5 terms
AssertClusterRequestConservation(t, cs)        // cluster, all 12 terms incl. GatewayExpired
```

Suggested home: alongside the existing shared test infrastructure in `sim/internal/testutil` (which already hosts golden-dataset helpers), or next to `Metrics` if an import cycle makes that awkward.

Then replace the 29 hand-rolled sums with calls. Highest-value single edit inside that: **add `GatewayExpired` to `sim/cluster/epd_test.go:186`**, or better, replace that block with the helper.

Why this shape: adding a bucket to INV-1 then means editing one function, and every existing test inherits it. When the cluster buckets were added (#882, #1190, #1228, #1193, #1264), the ~27 existing sums did not gain them — they still assert the pre-gateway version of the invariant. A helper makes that class of drift impossible.

**Related suggestion, optional:** INV-4 has a runtime checker, `verifyBlockConservation()`, which walks the free list independently for debug assertions. INV-1 — the invariant the doc says caused a months-long silent bug (#183) — has no equivalent. A `VerifyRequestConservation()` on the same model would let non-test code assert it too. Happy to leave this out if it is scope creep.

### Part 2 — `TestINV5_Causality_FullChain`

Assert the actual statement, per completed request:

```
ArrivalTime <= GatewayEnqueueTime
            <= GatewayEnqueueTime + SchedulingDelay
            <= CompletionTime
```

One caveat for whoever writes it: on the single-instance path there is no gateway, so `GatewayEnqueueTime` may be unset. The test should skip or substitute rather than assert on a zero field — otherwise it fails for the wrong reason.

The existing `TTFT >= 0` / `E2E >= TTFT` checks are worth keeping. They are a valid weaker property, just not this invariant's statement.

### Part 3 — INV-9 test: whitelist → glob

In `sim/simulator_test.go`, replace the two hardcoded filename lists with a pattern scan, so new files are covered by default:

```
scan: sim/routing*.go, sim/admission*.go, sim/scheduler*.go, sim/slo*.go, sim/router_state.go
plus an explicit, commented opt-out list for files that legitimately aggregate
TotalOutputTokens (the existing line-level exclusion already handles this case)
```

The point is to invert the default from opt-in to opt-out. A new scorer then has to *deliberately* exempt itself, which is a reviewable act rather than an omission.

### Part 4 — one sentence for INV-7

State in `invariants.md` that INV-7 is validated by the H3/H29 hypothesis experiments rather than a Go test, and why (it describes a freshness hierarchy, not a single predicate). Makes an intentional absence read as intentional.

**Files affected**

| Part | Change | Files |
|---|---|---|
| 1 | new conservation helper | `sim/internal/testutil/` (new func) |
| 1 | replace hand-rolled sums | ~15 test files across `sim/`, `sim/cluster/`, `cmd/` — **29 call sites** |
| 1 | the specific omission | `sim/cluster/epd_test.go:186` |
| 2 | new causality test | `sim/cluster/cluster_test.go` or a new `causality_test.go` |
| 3 | whitelist → glob | `sim/simulator_test.go` |
| 4 | doc sentence | `docs/contributing/standards/invariants.md` |

Multiple construction sites to flag: **the 29 conservation sums are the whole point of Part 1** — they are the duplication being removed. Part 1 is mechanical but wide; it may be worth landing the helper plus the `epd_test.go` fix first, and migrating the other 28 call sites in a follow-up so the diff stays reviewable.

**Design document**

Not needed. One test helper, one new test, one test refactor, one doc sentence.

---

### Deliberately out of scope — for discussion

**1. `INV-2` has no transition guard and no invalid-transition test.** Request state is assigned by direct field write in **15 places across 7 files**. There is no legal-transition table and no test that attempts an illegal transition and expects rejection — so the statement's "no invalid transitions" clause is entirely unprotected.

The codebase already has the right pattern, applied to instances: `validInstanceTransitions` (`sim/cluster/instance.go:416`) plus `TransitionTo()` (`:428`) which panics on an illegal edge, tested by a legal/illegal edge table at `sim/cluster/instance_lifecycle_test.go:252`.

Blocking question first, though: **`sim/request.go` declares five states** — `StateQueued`, `StateRunning`, `StateCompleted`, `StateTimedOut`, `StateWaitingForRemoteKVs` — and INV-2's statement names three. Is `running → waiting_for_remote_kvs → running` legal? Nobody can write the table, let alone the test, until the state set and legal edges are written down. Suggest that decision lands first, in its own issue.

**2. 8 of 9 INV-13 parity tests use the weaker form the statement says to move away from.** The statement already says byte-identical stdout "is what new parity tests should assert" and names `TestINV13_RunReplayParity_MoEDPPlacement` as the model. Classifying all nine: that one compares stdout; the other eight (`_PD`, `_PD_CLI`, `_SpecDecode`, `_FP8KV`, `_Offload`, `_Offload_DecodeOffload`, `_Offload_Jitter`, `_CacheHitRate`) compare `RequestTTFTs`/`RequestE2Es` maps.

Not proposing a mechanical migration, because the statement itself raises the reason not to: byte-identity "is satisfied by two identically-wrong runs," so each migration needs the non-vacuity gate and conservation companion the model test carries. That makes it a sequence of small independent PRs.

**3. Still unchecked.** We audited a subset at clause level, not all 23. `INV-PD-4`'s seven-link phase-causality chain is the obvious next one — given Finding 2, we no longer assume a chain is fully asserted just because the doc says a test checks it. `INV-4`, `INV-8`, `INV-10`, `INV-12`, `INV-BC-DP1`, `INV-PD-1`, `INV-PD-6`, `INV-PD-6b`, `INV-P2-1`, `INV-P2-2` were not examined at this depth and should not be read as cleared.

---

<sub>Produced with archon's invariant tooling plus manual reading. Counts are re-derivable at `07622594`: the 29 sites are conservation-style sums in `*_test.go` under `sim/` and `cmd/` (an assignment whose right side adds to `CompletedRequests`), with terms counted in a 6-line window. Clause splitting is a judgement call — someone may read INV-2 as two clauses rather than three — but that changes the accounting, not the findings.</sub>

mtoslalibu:
## What this is, in plain English

An invariant statement is usually several promises in one sentence, and a test that cites the invariant may only check one of them. That makes coverage look real when it is not. Issue #1720 went through the invariant registry clause by clause and found four cases. This PR fixes them, and adds tests whose job is to stop the same gap reopening.

The concrete defect: BLIS's request-conservation invariant (INV-1) has **twelve** buckets a request can end in. Almost every test hand-rolled its own sum, and they disagreed about how many terms conservation has — anywhere from three to eleven. Three cluster-level assertions left out the `gateway_expired` bucket entirely. That was invisible because gateway TTL is off by default, so the bucket was always zero: no conservation test had ever enabled TTL. The first person to write one would have got a false violation, and a genuine leak into that bucket was undetectable.

**What a reviewer should look at:** the shared helpers in `sim/cluster/inv1_conservation_test.go`, and whether the migrated call sites assert something stronger than what they replaced. Everything else follows from those.

**Merge readiness:** ready. Build, all 15 packages, `-race -count=2`, and lint are green, with zero new lint or `gofmt` findings against `main`.

## What changed

1. **One conservation helper per observation layer**, replacing 30 hand-rolled sums. Adding a bucket to INV-1 is now one edit instead of a manual search.
2. **The `gateway_expired` bucket** added to the three assertions that omitted it (`sim/cluster/epd_test.go`, `cluster_test.go`, `cluster_tier_test.go`), plus the first conservation test that actually drives a TTL expiry.
3. **`TestINV5_Causality_FullChain`** — INV-5 states `arrival <= enqueue <= schedule <= completion`, but every test citing it asserted `TTFT >= 0 && E2E >= TTFT`. Both of those are measured from arrival, so together they only say the last token follows the first; the enqueue and schedule links were asserted by nothing.
4. **INV-9's source scan** changed from a hardcoded six-file list to globs with an explicit opt-out list, and it now checks the second pattern the registry already claimed was grep-verified.
5. **Guard tests** so this does not silently regress: a registry-drift test, an AST scan rejecting new hand-rolled sums, and a check pinning the ledger's field-to-metric mapping.

Part 4 of the issue (a doc sentence for INV-7) is **not** here: it already landed in #1734, after the issue was written.

## Behavioral contracts

**BC-1: a request that ends in the TTL-expiry bucket is accounted for.**
GIVEN a cluster run whose gateway queue expires at least one request, WHEN conservation is asserted at simulation end, THEN the run is reported as conserving, and a run with an unaccounted request is reported as violating.

Evidence, and how to check it: `TestINV1_Conservation_GatewayTTLExpiry` fails against the old eleven-term sum with `injected=2 != accounted=1`. That is how the defect was confirmed real rather than theoretical — the test was written with the old arithmetic inline first, observed to fail, and only then switched to the helper.

**BC-2: the cluster conservation equation is stated in one place per observation layer.**
GIVEN a contributor adds a thirteenth bucket, WHEN they add it to the shared helper, THEN every cluster-level assertion checks it with no further edits, and a newly hand-rolled sum anywhere in `sim/` or `sim/cluster/` is reported with its location.

Evidence: `TestINV1_NoInlineClusterConservationSums` and its `sim/` counterpart, both over `internal/invariantscan`. Verified against the pre-fix tree: the scanner flags `cluster_tier_test.go:207`, one of the three defective sites, which a line-based grep misses because that sum accumulates through local variables.

**BC-3: the helpers cannot drift from the registry.**
GIVEN someone edits the conservation equation in `docs/contributing/standards/invariants.md` without touching the tests, WHEN the suite runs, THEN the mismatch is reported naming the terms present in one and absent from the other. The registry asks for exactly this ("any shared conservation helper should be generated from the twelve-term list above") and nothing enforced it.

**BC-4: every link of the causality chain is asserted independently.**
GIVEN a cluster run with the gateway queue active, WHEN causality is asserted per completed request, THEN each of arrival→enqueue, enqueue→dispatch and dispatch→schedule→completion holds on its own, a request missing a timestamp is skipped rather than compared against a zero, and the run fails if too few requests exercised the gateway links for the assertion to mean anything.

**BC-5 / BC-6: a new control-plane file is scanned without being registered, and the oracle boundary is checked through its alias.**
GIVEN a contributor adds a routing scorer, WHEN the INV-9 test runs, THEN it is scanned by default; exempting it requires a named entry, and a stale entry naming a deleted file fails. The previous list omitted `router_state.go`, `routing_nohit_lru_scorer.go` and `routing_precise_prefix_scorer.go` — all three clean, but two of them scorers, so the next scorer would have escaped too.

## Testing

```
go build ./...                      # exit 0
go test ./... -count=1              # 15/15 packages ok
go test ./internal/... ./sim/ ./sim/cluster/ -count=2 -race   # ok
golangci-lint run ./...             # zero new findings vs main
gofmt -l sim cmd internal           # zero new findings vs main
```

Two claims worth verifying rather than trusting, both reproducible:

- **The defect was real.** Revert `sim/cluster/gateway_queue_ttl_test.go`'s helper call to the eleven-term sum from `epd_test.go` and it fails `injected=2 != accounted=1`.
- **The guards actually fire.** Every detector has a fixture test proving it rejects the shape it targets, including the shapes it must *not* flag. The ledger mapping check was mutation-tested by swapping `gatewayEvicted` and `gatewayExpired` — two buckets that are zero together in every fixture, so no assertion over totals catches it — and it reports both fields.

## Review history

Two review rounds, both of which found defects in this PR that were worth the round:

- Round 1 found that the helper's doc comment and the registry both claimed it asserts INV-1's full-pipeline clause. It does not — that is an arithmetic identity given how `injected` is derived — and the migration had deleted the two sites that genuinely checked it against `len(Metrics.Requests)`. Restored, and the claim corrected. It also found that passing `cs.RejectedRequests()` at sites whose original equation had no rejected term silently excused a spurious rejection at 21 of them; those now pass a `noRejections` constant.
- Round 2 found that the AST guard could be evaded with `+=` accumulation, while the guard's own comment asserted that form was rejected; that the `sim/` guard was line-based and so missed wrapped sums; and that a mutation test I had cited as proof only exercised the easy case. Fixing the first two produced `internal/invariantscan`, and pointing it at `sim/` found four more hand-rolled sums the original grep had missed.

## Discovered issues

Filed rather than fixed here, both with reproduction steps:

- **#1745 — INV-1's exclusivity clause is asserted by nothing, and no sum can assert it.** The registry says each bucket is mutually exclusive. A twelve-term equality passes when one request is double-counted in one bucket and another is lost, because the errors cancel. Catching it needs per-request terminal-bucket tagging, which is a production change. This is the most valuable thing this PR found and did not fix, and it means the PR closes #1720's findings, not INV-1's full clause set.
- **#1746 — the twelve-term equation is invisible from stdout.** The seven cluster-only counters live on `cluster.RawMetrics` and are never serialised into the JSON output, so no CLI-level test can assert the canonical equation. `clusterConservationHolds` in `cmd/` is structurally stuck at five terms for that reason.

## Out of scope, per the issue

INV-2's transition guard (blocked on deciding the legal request state set), and migrating the eight INV-13 parity tests to byte-identical stdout (a sequence of PRs, each needing its own non-vacuity gate).

Closes #1720


mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
## Correction on round 5's "production changes" section

Round 5 approved the PR, and its round-4 follow-up verification is accurate. But its P1 / P7 / P8 analysis — the "secondary production changes" to `sim/batch_formation.go`, `sim/kv/offload_chain.go`, `sim/kv/offload_deferral.go`, `sim/kv/tiered.go`, `sim/kv_store.go` and `CLAUDE.md`, described as a "design reversal" of the #1706 billing fix — is not this PR. Those files are not in its diff, and the "Scope disclosure" observation that the test-only label is inaccurate should be withdrawn.

**What happened.** The review appears to have diffed two-dot (`git diff main HEAD`) rather than three-dot (`git diff main...HEAD`). At review time the branch's merge-base was `67f202c2`, and `main` had since advanced by two commits — one of them `5497174e fix(kv-offload): credit the WHOLE reloaded prefix under chunked prefill (Closes #1706) (#1725)`. A two-dot diff renders those newer commits as *inverted* changes on the branch side, which is why #1725's own pre/post design reads as a reversal, with the "old design (pre-PR)" and "new design (this PR)" labels the wrong way round. The PR was never reversing anything; it simply did not contain #1725 yet.

**How to check it.** Against the merge-base as it stood:

```
git diff --name-only 67f202c2...HEAD | grep -E 'batch_formation\.go|kv/|kv_store\.go|CLAUDE\.md'
# → only sim/batch_formation_test.go (a test file); no production file, no CLAUDE.md

git log --oneline 67f202c2..origin/main
# → 4fa5ca1e test(latency): ... (#1749)
# → 5497174e fix(kv-offload): credit the WHOLE reloaded prefix ... (#1725)
```

And against current `origin/main`, the branch's whole diff is test files, docs and one new test-only package:

```
git diff --name-only origin/main...HEAD | grep -vE '_test\.go$|^docs/|^\.github/|^internal/invariantscan/'
# → no output
```

**Acted on rather than just disputed.** The finding was wrong about authorship but pointed at something real: the branch was behind `main`, and #1725 added new test files (`sim/batch_formation_reload_test.go`, `sim/kv/offload_*_test.go`, `sim/cluster/offload_deferral_e2e_test.go`) that this PR's new guards scan. If any of them hand-rolled a conservation sum, the merged tree would have failed `TestINV1_NoInlineConservationSums`. `origin/main` is now merged in (`579322f9`) and verified against fresh baselines taken from the new `main`:

- `go build ./...` — clean
- `go test ./... -count=1` — all packages pass
- `gofmt` — no new findings vs `main` (67 pre-existing, unchanged)
- `golangci-lint run ./...` — no new findings vs `main`

Zero file overlap between #1725/#1749 and this branch, so the merge was textual-conflict-free, and the new offload test files do not trip the guards.

Nothing else in round 5 is affected: BC-1 through BC-6, the ten perspectives on the invariant work, and the round-4 fix verification all stand.

mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
@claude /blis-pr-review

susiejojo:
## BLIS PR Review — #1752 `hardening(invariants): assert the invariant clauses no test was checking`

**Verdict: Ready to merge.** Independent review (6th round). No blockers found. This is a test-only, invariant-hardening PR closing #1720; it adds no production code, so cross-path parity (run/replay/observe) and preemption safety are N/A by construction.

### Archon
No `/archon-pr-review` output in prior comments. Proceeding without an archon verdict. (Test-only, no boundary movement — a FAST-TRACK-equivalent posture.)

---

### What I verified independently (not just re-read)

Both reproducibility claims in the PR body check out against the actual tree:

1. **The defect was real.** I reconstructed the old 11-term `epd_test.go` sum (the most complete pre-#1720 assertion, which omits `GatewayExpired`) and ran it against the TTL-expiry fixture: `injected=2, accounted=1, gatewayExpired=1`. The old sum would have **falsely passed a genuine leak** into the TTL bucket, and now fails. The new `assertClusterINV1Conservation` closes it. ✅
2. **The AST scanner catches what a grep misses.** Ran `invariantscan.FindConservationSums` over `main`'s `sim/cluster/cluster_tier_test.go`: it flags line 207 — a sum laundered through local variables (`completed`, `queued`, …) that a line-based grep cannot see. ✅
3. **The guard actively fires.** Introduced a fresh inline 5-term cluster sum into `sim/cluster/` → `TestINV1_NoInlineClusterConservationSums` failed naming `zz_evasion_test.go:7`. It is not vacuous. ✅
4. **No false positives on the new floors.** `causality_test.go`'s non-vacuity floors (`len(requests) - shed - rejected - depth - expired`) read four cluster counters but no instance-side bucket, so `isLedger` correctly declines to classify them as ledgers. The full-tree scan passes. ✅
5. **Build + affected packages green.** `go build ./...` = 0; `internal/invariantscan`, `sim`, `sim/cluster` all pass, including the 15 new INV-1/INV-5 tests. (My local `cmd` run flagged `TestEveryKVCapacityCallSiteIsEPAware`, but that is a **local environment artifact** — a stray `.claude/worktrees/` copy that the test's `filepath.WalkDir("..")` doesn't skip; the PR touches none of the files it names. Not a PR defect.)

---

### Contract verification (BC-1 … BC-6)

| Contract | Evidence | Status |
|---|---|---|
| BC-1 TTL-expiry request accounted | `TestINV1_Conservation_GatewayTTLExpiry` (non-vacuity gate + fixture-margin guard on r1 completion) | ✅ verified reproducible |
| BC-2 one ledger per layer; new inline sum reported | `TestINV1_NoInlineClusterConservationSums` (+ `sim/` twin), `internal/invariantscan` | ✅ fires on probe |
| BC-3 helpers cannot drift from registry | `TestINV1_HelperMatchesRegistry` (parses 2 `injected_requests ==` equations: 12 + 5 terms) | ✅ term-set diff both directions |
| BC-4 every causality link asserted independently | `TestINV5_Causality_FullChain` (per-link counters + per-link floors), `..._SingleInstancePath` | ✅ |
| BC-5 new control-plane file scanned by default | glob `simControlPlaneGlobs` + required-file non-vacuity list | ✅ |
| BC-6 oracle boundary checked through the derived accessor | `TestINV9_OracleReadDetectorFires` (`completionProgressIndex` fixture + comment-not-a-violation case) | ✅ |

The triple-lock against swapped-bucket drift (`ledgerFieldSources` AST table ↔ `terms()` ↔ `clusterLedgerTerms`, checked by `TestNewClusterLedger_FieldSources` / `TestClusterLedger_TermsMatchNames` / `TestClusterLedger_Accounted` with power-of-two witnesses) is the strongest part of the PR: it catches a `gatewayEvicted`↔`gatewayExpired` swap that no aggregate equality can, because both are zero-together in every fixture.

---

### Findings

**None blocking.** Three informational notes, all either already-tracked or deliberate design:

**F1 (informational) — `isLedger` heuristic gap.** A hand-rolled cluster sum that reads exactly **one** instance-side bucket other than `CompletedRequests` plus cluster buckets (e.g. `StillQueued + GatewayQueueDepth() + GatewayEvicted() + GatewayExpired()`) escapes detection, because `isLedger` requires `CompletedRequests` OR ≥2 instance-side buckets. I confirmed this escapes the scanner. This is a **deliberate, documented trade-off** (scan.go:159-188) — widening it would false-positive on the INV-5 floors and the shed-accounting identity, earning the guard an exemption list. The shape is unnatural for a real ledger (a conservation sum that omits `CompletedRequests` and all-but-one terminal bucket is not a conservation sum). Acceptable as-is; worth a one-line mention in the `isLedger` doc that single-instance-bucket ledgers are out of scope.

**F2 (informational, already tracked) — the two most valuable gaps are filed, not fixed.** #1745 (INV-1 exclusivity — no aggregate sum can detect a double-count that cancels a loss) and #1746 (the seven cluster-only counters never reach stdout JSON, so `clusterConservationHolds` in `cmd/` is stuck at 5 terms). Both are honestly disclosed in the PR body and the registry, and both require production changes out of this PR's test-only scope. The registry's new "Known-unasserted clause" paragraph correctly prevents the twelve-term equality from being mistaken for a completeness proof.

**F3 (informational, already documented as a known gap) — cluster control-plane INV-9 files stay a hand-maintained list.** The `sim/` scan is glob-based (new scorers covered by default), but the four `sim/cluster/*.go` control-plane files remain an explicit list because they carry the legitimate `TotalOutputTokens` aggregate and no glob shape separates them from the rest of the package. A new *cluster* control-plane file must still be registered by hand. This is the inverse of the `sim/` default and is explicitly called out in both the test comment and the registry — a known, accepted limitation, not an oversight.

---

### Doc / registry accuracy
The `invariants.md` INV-1/INV-5/INV-9 rewrites are accurate and match the code: the registry now states the twelve-term equation as the source of truth, documents the `noRejections` vs `RejectedRequests()` distinction, the `Metrics.Requests`-map caveats (drop guards/drain/PD-collapse undercount; timeouts overcount), and the AST-guard rationale. `TestINV1_HelperMatchesRegistry` mechanically enforces the registry↔helper term-set match, so the doc cannot silently drift. The package doc in `internal/invariantscan/scan.go` (corrected in round 3) correctly states `cmd` is not a consumer.

### Review-thread hygiene
All 5 prior review rounds' findings are addressed (verified in the round-4/round-5 status tables): the round-1 helper tautology, round-2 `+=` evasion (→ `internal/invariantscan`), round-3 package-doc/`disaggregation.go`-glob/arg-builder fixes, round-4 non-vacuity floors, and the post-round-5 mechanical merge-conflict fix (`clusterConservationHolds` caller `expected=20`). No unaddressed reviewer comments remain.

---

### Bottom line
A disciplined, high-quality hardening PR that closes a real latent defect and installs guards (registry-drift, AST anti-sum, per-link causality, oracle-alias) that stop the same gap reopening. The two things it found and could not fix (#1745, #1746) are correctly scoped out and filed. **Ready to merge.**

*Review generated with Claude Code (Opus 4.8). Read-only review — no code modified.*

susiejojo (approved):
lgtm

PR #1759, opened 2026-09-17
hardening(ci): retry dispatch with backoff on all delivery phase hand-offs

mtoslalibu:
## Summary

A single un-retried `gh workflow run` dispatch terminally stalls the delivery loop on a transient API failure — observed live on PR #1736 (verify run `35180721918`, step 24 "Hand off to correct"). The stall-sweep backstop only flags; it does not re-dispatch.

- Add 3-attempt retry with exponential backoff (5s, 10s) to **all three** phase hand-off dispatch sites: verify→correct, correct→verify, implement→verify
- On exhaustion the step fails and the existing failure reporter fires (no silent loss)
- First-attempt success adds zero delay

Closes #1757

## Files changed

- `.github/workflows/deliver-verify.yml` — "Hand off to correct" step (~line 983)
- `.github/workflows/deliver-correct.yml` — "Hand back to verify" step (~line 326)
- `.github/workflows/deliver-implement.yml` — "Hand off to verify" step (~line 684)

## Test plan

- [ ] Verify retry loop structure matches across all 3 sites (identical pattern)
- [ ] Confirm `exit 1` on exhaustion triggers the existing `always() && !success()` failure reporter
- [ ] Confirm first-attempt success exits immediately with no sleep
- [ ] Backoff total (5+10 = 15s) is well within step timeout budget

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1723, opened 2026-09-15
hardening(ci): make deliver-implement crash-survivable and resumable

mtoslalibu:
## What this changes, in plain English

The implement phase of the L1 delivery loop used to do all its work and then, as its very last
action, push a branch and open a PR. If anything interrupted it before that moment, everything was
lost — no branch, no PR, nothing for a re-run to pick up. That happened twice. On #1706 the second
attempt had already finished the implementation with passing tests when its runner was evicted 75
seconds before it would have pushed; all 52 minutes were discarded.

It was also invisible. `deliver-stall-sweep.yml` is the backstop for a dead runner, but it finds
deliveries *from the PR side*, so a run that died before opening a PR could not be seen by the one
detector built to catch it.

**Now the branch is pushed by the workflow before the agent starts, and the agent's first action —
before it writes any code — is to open a draft PR on it.** The agent pushes as it goes and marks the
PR ready when it's done. So an interrupted run leaves recoverable commits, the delivery becomes
visible almost immediately, and re-issuing `/approve-issue-for-pr-delivery` continues that branch
instead of re-deriving the design.

Two secondary fixes come from the same incident. The prompt no longer tells the agent to run the
full test suite and the linter — `ci.yml` already owns those, verify dispatches it, and having the
agent build `golangci-lint` from source on the 4 GiB self-hosted runner was itself a cause of the
eviction. And the prompt now says the phase is unattended, because on #1706's *first* attempt the
agent invoked `superpowers:brainstorming`, posted a design write-up, and ended its turn waiting for
a human reply that cannot arrive — 2m18s, $1.83, zero commits, and the phase reported success.

**Merge readiness:** ready for review. Build, vet, `go test ./...` and CI `lint` are all green.

## Why the workflow does not open the PR itself

An earlier revision of this PR had the workflow open the draft PR with `gh pr create`. That is the
stronger guarantee, and I moved away from it deliberately.

`gh pr create` from a workflow step uses `GITHUB_TOKEN`, which GitHub gates behind
**"Allow GitHub Actions to create and approve pull requests"**. That is a **single** toggle
(`can_approve_pull_request_reviews`) granting creation **and approval** to *every* workflow in the
repository — there is no way to take only the half this loop needs. This loop must never approve
anything: it labels, and a human merges. Acquiring an approval capability repo-wide, on a public
repository that deliberately runs agent workflows, to save the agent one command is the wrong trade.

The agent already opens PRs with an App installation token obtained via OIDC, which that setting does
[...]

PR #1760, opened 2026-09-17
hardening(ci): dispatch deliver-correct on the delivery branch, not github.ref_name

### LINKED ISSUE #1729: hardening(latency): derive layer count from the block-type array when num_hidden_layers is absent (NS-4) [R1/S2]
*Part of the R1 tracking issue #1727 (task **S2**). Independent — no blocker.*

## What this fixes

Some model configs carry the layer count as a top-level scalar `num_hidden_layers`; others (Nemotron-3-Ultra-550B) omit that scalar and instead express the count as the **length of a block-type array** (`layers_block_type`). BLIS reads the scalar, gets 0, and **aborts before any simulation**:

```
$ blis run --model nvidia/Nemotron-3-Ultra-550B
fatal: NumLayers must be > 0, got 0
```

The 30B sibling (`Nemotron-3.5-Lightning-30B`) has the **same architecture**, declares `num_hidden_layers: 52`, and runs. BLIS didn't lack physics for the 550B — it asked the wrong question. `num_hidden_layers` has no answer in that config; *what is the layer list?* always does, because every config describes its layers somehow. This is **NS-4**: derive the count from the list when the scalar is absent.

## Where it breaks today

- Field: [`ModelConfig.NumLayers` (`json:"num_hidden_layers"`)](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/model_hardware_config.go#L12).
- Parse site: [`config.go#L482`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L482) — `NumLayers: getInt("num_hidden_layers")`, inside [`GetModelConfigFromHF`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L337). The [`getInt` closure](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L338-L343) returns **0** when the key is absent — the fall-through that yields `NumLayers == 0`.
- Two validators emit the fatal, one per latency backend:
  - roofline: [`config.go#L577-L578`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L577-L578), made fatal at [`latency.go#L195-L197`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/latency.go#L195-L197);
  - trained-physics: [`trained_physics_model.go#L1078-L1079`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/trained_physics_model.go#L1078-L1079).
- `layers_block_type` is referenced **nowhere** in the Go code today (no `layers_block_type` / `block_type` / `BlockType` match) — this task introduces the handling.

## The change

In [`GetModelConfigFromHF`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L337), when `num_hidden_layers` is absent (or 0), derive `NumLayers` from `len(layers_block_type)`.

There is an existing precedent to mirror exactly — reading a layer count from a nested config array: [`HFConfig.LinearAttnFullLayerCount`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L152-L168) reads `linear_attn_config.full_attn_layers` and returns `len(...)`, with callers falling back to `NumLayers` (call site [`config.go#L459`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L459)). A `layers_block_type` reader would follow the same `len(list)` shape. Nested keys are reachable at the top level because [`ParseHFConfig` pivots `text_config` onto the top-level map](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L251-L268).

**Scope guard:** this task only *counts* the array to get a total layer count. It does **not** introduce layer groups or per-type tallies (that is R4a). The presence-detection helper for `num_hidden_layers` at [`cmd/hfconfig.go#L224-L244`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/hfconfig.go#L224-L244) may also need to recognise the block-type-array case so a fetched/located config is considered valid.

## Acceptance criteria (the contract)

1. `blis run --model nvidia/Nemotron-3-Ultra-550B …` (with a config that has `layers_block_type` and no `num_hidden_layers`) **reaches completion** instead of the `NumLayers must be > 0` fatal.
2. A model that **declares** `num_hidden_layers` (the 30B sibling, and every currently-catalogued model) is **byte-identical** to before (INV-6): the scalar still wins when present; the array is only a fallback.
3. Both latency backends (roofline and trained-physics) accept the derived count — the fatal at [`config.go#L577-L578`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go#L577-L578) and [`trained_physics_model.go#L1078-L1079`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/trained_physics_model.go#L1078-L1079) no longer fires for a config whose only layer-count evidence is the array.
4. A config with **neither** `num_hidden_layers` **nor** a usable `layers_block_type` still fails loudly (no silent 0).

## Invariants at risk

- **INV-6 (determinism)** — the scalar-present path must be byte-identical; the array path is a pure fallback.

## Files affected

- [`sim/latency/config.go`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/sim/latency/config.go) (`GetModelConfigFromHF`, a new `layers_block_type` reader), possibly [`cmd/hfconfig.go`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/hfconfig.go#L224-L244) (presence detection) + tests.

## Origin

R1 task **S2** / **NS-4** of the operator-basis design note (reconciling Discussion #1700). No architecture change.

### LINKED ISSUE #1730: docs: correct two false model-onboarding claims (models.md, CONTRIBUTING.md step 3) [R1/S3]
*Part of the R1 tracking issue #1727 (task **S3**). Independent — no blocker. Docs-only: no code, no output change.*

## What this fixes

Two documentation claims contradict the code. R1 corrects them so a reader following the docs isn't misled.

### (a) `docs/reference/models.md` — "no per-model setup or calibration"

[`docs/reference/models.md#L3`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md#L3) claims:

> BLIS supports **any transformer model with a HuggingFace `config.json`** — no per-model setup or calibration required. Both latency backends (roofline and trained-physics) generalize across architectures.

The same document then contradicts it:
- a **"Validated Architectures"** list at [`#L9-L18`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md#L9-L18) (L18 hedges "hasn't been formally validated");
- an **MFU calibration** note at [`#L23`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md#L23) (directly contradicting "no calibration required");
- **per-architecture KV/shape handling** at [`#L26-L34`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md#L26-L34);
- **ten "known approximations"** at [`#L36-L45`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md#L36-L45).

Reword L3 so it matches the rest of the page: BLIS *runs* many models day-zero with standard CLI configuration, but fidelity is architecture-dependent and some shapes carry known approximations (which the page already lists). Don't overclaim "no per-model setup or calibration."

### (b) `CONTRIBUTING.md` — a `models:` list that does not exist

The **"Adding a New Model to defaults.yaml"** section ([`CONTRIBUTING.md#L261`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/CONTRIBUTING.md#L261)) step 3 says:

> 3. If trained coefficients exist, add a corresponding entry to the `models:` list — [`CONTRIBUTING.md#L267`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/CONTRIBUTING.md#L267)

**No `models:` section exists in `defaults.yaml`.** Its top-level sections are `defaults:` ([L1](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L1)), `workloads:` ([L122](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L122)), `trained_physics_coefficients:` ([L163](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L163), a **global** coefficient block, not a per-model list), `lora:` ([L192](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L192)), `kv_offload_devices:` ([L221](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L221)), `version:` ([L227](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/defaults.yaml#L227)). The parse target is a fixed struct ([`cmd/default_config.go#L25`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/default_config.go#L25)), so a `models:` key would map to nothing — and strict YAML parsing would reject it outright.

Remove/replace step 3. **Steps 1 and 2 are correct and stay** ([`#L265`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/CONTRIBUTING.md#L265): add a `defaults:` entry with `GPU` + `tensor_parallelism`; [`#L266`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/CONTRIBUTING.md#L266): add `hf_repo`).

## Acceptance criteria (the contract)

1. `docs/reference/models.md` L3 no longer claims "no per-model setup or calibration"; it is consistent with the validated-architectures list, the MFU-calibration note, and the ten known approximations already on the page.
2. `CONTRIBUTING.md` "Adding a New Model" no longer references a `models:` list in `defaults.yaml`. Steps 1 and 2 are unchanged. A reviewer following the section succeeds against the real `defaults.yaml` schema.
3. No code and no simulator output changes (docs-only).

## Scope note

This corrects only the two false statements. It does **not** rewrite the onboarding flow for the later catalog/`--catalog` world (that follows S4/S6 and R4a); step 1's `defaults:`-entry instruction is still true on `main` today and stays.

## Invariants at risk

- None (documentation only).

## Files affected

- [`docs/reference/models.md`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/docs/reference/models.md), [`CONTRIBUTING.md`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/CONTRIBUTING.md).

## Origin

R1 task **S3** of the operator-basis design note (reconciling Discussion #1700).

### LINKED ISSUE #1751: hardening(ci): deliver-verify "Hand off to correct" uses github.ref_name → HTTP 422 on review-triggered runs, dead-ending post-convergence findings at needs-human

`deliver-verify.yml`'s hand-off to the correction phase dispatches `deliver-correct.yml` with the wrong git ref on **review-triggered** runs, so a finding that arrives via a PR review after the loop has converged can never reach the correction phase. The delivery dead-ends at `needs-human` on an infrastructure error — not on any real verdict, and not on round exhaustion.

**Symptom** (verified on `main` @ `32808364`)

A `pull_request_review`-triggered *Deliver — Verify* run that decides `final_decision: correct` fails at the **Hand off to correct** step:

```
gh workflow run deliver-correct.yml --repo … --ref "1742/merge" -f pr_number=1742 -f issue_number=1730
  final_decision: correct
  MAX_ROUNDS: 3
could not create workflow dispatch event: HTTP 422: No ref found for: 1742/merge
##[error]Process completed with exit code 1.
```

The `verify` job then fails and the fail-safe reporter [`Report a failed verify`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/.github/workflows/deliver-verify.yml#L1006-L1029) posts *"Delivery stopped — the verify phase failed or timed out"* and applies `needs-human`. Note this is **not** round exhaustion: `MAX_ROUNDS` was 3 and the gate's decision was `correct`.

**Root cause**

[`.github/workflows/deliver-verify.yml#L983-L988`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/.github/workflows/deliver-verify.yml#L983-L988) dispatches with `--ref "${{ github.ref_name }}"`:

```yaml
- name: Hand off to correct
  if: steps.paused.outputs.paused == 'false' && env.final_decision == 'correct'
  run: |
    gh workflow run deliver-correct.yml --repo "$REPO" --ref "${{ github.ref_name }}" \
      -f pr_number="$PR" -f issue_number="$ISSUE_NUMBER"
```

- On `workflow_dispatch` / `push` triggers, `github.ref_name` is the delivery branch `deliver/issue-<N>` → a valid dispatch ref (this is why implement-driven and push-driven correction rounds hand off fine).
- On `pull_request_review` / `pull_request_review_comment` triggers, `github.ref_name` is the PR merge ref `<pr>/merge` (e.g. `1742/merge`). `gh workflow run` (workflow_dispatch) requires a branch or tag, so GitHub returns **HTTP 422: No ref found**.

**Impact**

The review triggers exist precisely so that "a finding that arrives after the loop converged should be acted on, not just able to halt it" ([deliver-verify.yml#L62-L67](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/.github/workflows/deliver-verify.yml#L62-L67), and the changes-requested/commented handling at [L217-L227](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/.github/workflows/deliver-verify.yml#L217-L227)). That path is broken: a post-convergence request-changes/comment review correctly flips the gate to `correct`, then can never hand off — it always dead-ends at `needs-human`. Observed on two delivery PRs, both left at `needs-human`:

- run `35132122044` — PR #1742 (closes #1730)
- run `35132976778` — PR #1743 (closes #1729)

**Proposed fix**

Resolve the delivery branch explicitly for the hand-off dispatch rather than trusting `github.ref_name`. The issue number is already in scope, so `--ref "deliver/issue-$ISSUE_NUMBER"` is deterministic; alternatively derive it from the PR head ref (`gh pr view "$PR" --json headRefName -q .headRefName`). Audit any other `gh workflow run --ref "${{ github.ref_name }}"` in the deliver-* workflows that can be reached from a review event for the same defect.

- [x] None (CI workflow correctness) — but the bug silently converts a real, fixable finding into a `needs-human` stop, so the correction loop never runs.

**Files affected**
- `.github/workflows/deliver-verify.yml` — the `Hand off to correct` step ([L983-L988](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/.github/workflows/deliver-verify.yml#L983-L988)).

**Design document**
None needed — one-line ref fix.

Sibling to the delivery-robustness hardening (#1722). Surfaced while re-driving R1 sub-issue #1730 (tracker #1727): a request-changes review carrying qa-review findings triggered verify, which decided `correct` but could not hand off.

mtoslalibu:
## What this fixes

When a review comes in **after** an L1 delivery has already converged, the loop is supposed to route that finding back to the correction phase. Today it can't: the hand-off always dies with `HTTP 422: No ref found` and the delivery dead-ends at `needs-human` on an infrastructure error — not a real verdict, and not round exhaustion. Two deliveries hit this: PR #1742 (closes #1730) and PR #1743 (closes #1729).

## Why

`deliver-verify.yml`'s `Hand off to correct` step dispatched `deliver-correct.yml` with `--ref "${{ github.ref_name }}"`:

- On `workflow_dispatch` / `push` runs, `github.ref_name` is the delivery branch `deliver/issue-<N>` — a valid `workflow_dispatch` ref. This is why implement- and push-driven correction rounds hand off fine.
- On `pull_request_review` / `pull_request_review_comment` runs, `github.ref_name` is the PR **merge** ref `<pr>/merge`. `gh workflow run` (workflow_dispatch) requires a branch or tag, so GitHub returns `HTTP 422: No ref found`. The verify job fails, the fail-safe reporter posts "verify failed or timed out" and applies `needs-human`.

## The fix

Dispatch on `steps.target.outputs.branch` — the resolved `deliver/issue-<N>` that:
- the CI dispatch already uses (`deliver-verify.yml` L441),
- is validated in `Validate the target PR` to equal `deliver/issue-<N>`,
- is populated on **every** trigger, so it is a valid dispatch ref regardless of event.

This also makes the review path consistent with the push path, which already dispatched correct on the delivery branch.

## Sibling audit (per the issue's request)

The issue asked to audit other `gh workflow run --ref "${{ github.ref_name }}"` in the deliver-* workflows reachable from a review event:

| Location | Trigger(s) | Reachable from a review event? | Action |
|---|---|---|---|
| `deliver-verify.yml:986` (Hand off to correct) | workflow_dispatch, push, **pull_request_review[_comment]** | **yes** | **fixed** |
| `deliver-implement.yml:688` (dispatch verify) | `issue_comment` only | no | left as-is |
| `deliver-correct.yml:329` (dispatch verify) | `workflow_dispatch` only | no | left as-is |

Only the verify hand-off is reachable from a review event, so only it carries the 422 defect. The other two are left untouched to avoid scope creep; both are dispatched with a branch ref today and work.

## Testing

- **TDD:** added `TestDeliverVerifyHandsOffOnTheDeliveryBranchNotTheEventRef` in `scripts/deliver_guards_test.go` — red against the old `github.ref_name`, green after the fix. It asserts the hand-off dispatches on the delivery branch and never on `github.ref_name`, matching the existing guard-test pattern (GitHub Actions expressions can't be unit-tested, so the phases are guarded against reintroducing the defect).
- `go test ./scripts/ -count=1` — pass.
- `actionlint` on the deliver-* workflows — clean.

## Behavioral contract

- **GIVEN** a delivery PR that has converged, **WHEN** a `commented` / `changes-requested` review arrives and flips the gate to `correct`, **THEN** verify dispatches `deliver-correct.yml` on `deliver/issue-<N>` and the correction round runs, instead of failing with HTTP 422 and applying `needs-human`.

Closes #1751


mtoslalibu:
@claude /blis-pr-review

mtoslalibu:
## Review findings addressed + merge conflict resolved

### Merge with `main` (#1759)
`main` advanced onto these same two files via #1759, which routed all delivery hand-offs through `scripts/dispatch-with-retry.sh` (exponential-backoff retry) — but kept `--ref "${{ github.ref_name }}"`, so #1751 was still unfixed, just wrapped in retry. Resolved by keeping #1759's retry helper **and** #1759's `TestDeliveryHandoffDispatchesUseRetryScript`, and dispatching on `${{ steps.target.outputs.branch }}`. The repo-wide guard was broadened to match **both** dispatch spellings (`gh workflow run` and `dispatch-with-retry.sh`) so a `github.ref_name` regression can't hide behind the retry helper.

### Findings from the first review pass
1. **Comment stated a wrong `workflow_dispatch` premise (Important).** The comment claimed `github.ref_name` is `deliver/issue-<N>` on `workflow_dispatch`; it is actually `main` (implement dispatches verify on its `issue_comment` ref). Round-1 correction previously ran `main`'s copy of the workflow; the fix makes every trigger consistent on the delivery branch. **Fixed** — the comment now states this honestly instead of hiding the one real behavior change.
2. **Guard tested the consumer but not the producer (test-hardening).** If `Validate the target PR` stops writing `branch=` to `$GITHUB_OUTPUT`, the output goes empty and `--ref ""` silently falls back to the default branch. **Fixed** — added a producer assertion and anchored the consumer match to `--ref`.
3. **Guard pinned one named step in one file; the defect is a class (test-hardening).** **Fixed** — added `TestNoReviewReachableWorkflowDispatchesOnTheEventRef`: globbed, asserts no workflow with a `pull_request`/`pull_request_review*` trigger dispatches on `github.ref_name`, mirroring the #1723 break-tested repo-wide guard pattern.

### Verification
- `go test ./scripts/` — pass (full suite). Both new guards **break-tested**: reverting the ref to `github.ref_name` trips the consumer + repo-wide guards; deleting the producer's `echo "branch=$branch"` trips the producer half alone; adding a review trigger to `deliver-correct.yml` trips the class guard.
- `actionlint` on the deliver-* workflows — clean.
- `gofmt` — clean.

### Verdict
Ready to merge. The fix is correct on every trigger (`steps.target.outputs.branch` is provably populated whenever the hand-off runs — same `proceed` gate as the producer step), the guards fail for the right reasons, and the merge preserves #1759's retry work.

Closes #1751

mtoslalibu:
## Full staged review (code / tests / comments / errors) — findings addressed

Ran the complete review methodology locally (four specialized perspectives), not a single pass. Results:

- **Silent-failure:** sound — traced the `if:` gate chain and confirmed `--ref ""` is structurally unreachable (the hand-off can only fire after `Validate the target PR` has written a non-empty `branch=`). No new silent failures. `dispatch-with-retry.sh` fails loudly on a bad ref.
- **Comments:** every factual claim verified accurate. **One fix:** the hardcoded `L441` reference is a rot hazard → now names the step (`Run the repository CI on the delivery branch`).
- **Code:** solid, CLAUDE.md-compliant. **Minor:** producer assertion was coupled to the exact shell var name → loosened to `branch=<non-empty>` (still trips if the line is dropped; no longer false-alarms on a rename).
- **Tests (Medium):** the repo-wide class guard only saw `github.ref_name` in a step's `run:`, so a ref staged via step `env:` (`env: REF: ${{ github.ref_name }}` + `--ref "$REF"`), or a raw-REST dispatch (`gh api -X POST .../dispatches`), slipped past. **Fixed:** the guard now matches all three dispatch spellings and scans step-level `env:` too. Break-tested both new paths (each trips the guard; green after removal). Job-/workflow-level env is a documented residual gap — the consumer assertion locks the actual hand-off regardless.

### Verification
- `go test ./scripts/` full suite — pass. Guards break-tested: revert ref → both trip; drop producer output → producer half trips alone; env-staged ref and raw-REST dispatch on a review-reachable workflow → class guard trips.
- `actionlint` — clean. `gofmt` — clean.

### Verdict
Ready to merge. Fix is correct on every trigger; guards are env/REST-aware and fail for the right reasons; merged cleanly with #1759's retry helper.

Closes #1751

namasl:

## qa-review — PR #1760: ✅ PASS
_hardening(ci): dispatch deliver-correct on the delivery branch, not github.ref_name · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The hand-off now dispatches `deliver-correct.yml` on the validated delivery branch, directly fixing the review-triggered 422. The added tests cover both the specific producer/consumer contract and the requested audit for the same defect in review-reachable workflows. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | Yes. No external user-facing documentation or standards change is required for this internal CI correction. The workflow's inline operational documentation and regression-test comments were updated to explain the ref contract. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale code comment was found. The new comments match the current implement-driven path and the corrected hand-off. |
| G1 | correctness | ✅ CONFIDENT | If `Validate the target PR` is skipped or fails to set its output, `steps.target.outputs.branch` will be empty. What ref does `dispatch-with-retry.sh` receive when `--ref ""` is passed, and does `gh workflow run` reject it or silently fall back to the default branch? | `dispatch-with-retry.sh` forwards the literal `--ref ""` unchanged. In GitHub CLI, an empty ref is treated like an unspecified ref and resolves to the repository default branch rather than producing the review merge-ref 422. The new producer assertion exists specifically to prevent that silent fallback. |
| G2 | correctness | ✅ CONFIDENT | If a review trigger fires on a PR after its delivery branch has been deleted or renamed on the remote, how does `gh workflow run` handle dispatching on `steps.target.outputs.branch` versus the merge ref? | A branch already deleted or renamed before validation is rejected before the hand-off: the PR must still be same-repository, open, and exactly on `deliver/issue-$ISSUE_NUMBER`. If deletion occurs after validation but before dispatch, dispatching the saved branch name fails because the named ref no longer exists. Dispatching the review merge ref is not a useful fallback: workflow_dispatch rejects `<pr>/merge` as described by the reported 422. |
| G3 | invariants | ✅ CONFIDENT | How does switching the dispatch ref from `github.ref_name` to `steps.target.outputs.branch` impact causality (INV-5) if a `pull_request_review` event arrives while a new `push` commit is being pushed to the delivery branch? | This change does not affect simulator causality INV-5, which concerns request lifecycle timestamps. Operationally, all delivery phases share a branch-keyed concurrency group, so the push-triggered verify cannot execute concurrently with the current verify/correction chain. A branch can still move during a running verify; CI explicitly checks that its run used the initially captured SHA, while the correction dispatch intentionally resolves the delivery branch at dispatch time. |
| G4 | invariants | ✅ CONFIDENT | Does dispatching `deliver-correct.yml` on `steps.target.outputs.branch` preserve deterministic state transitions (INV-6) across review-triggered runs compared to implement-triggered runs? | Yes for the delivery state machine: review-triggered and implement-triggered verifies now hand off using the same validated `deliver/issue-N` branch, so both select the branch's workflow copy instead of one using an invalid merge ref and the other using the default branch. Simulator INV-6 itself is not implicated because it concerns byte-identical simulator output for a fixed seed. |
| G5 | rules | ✅ CONFIDENT | In `triggerNames` (`deliver_guards_test.go`), when `n.Kind` is `yaml.MappingNode`, the loop steps `i += 2` over `n.Content`. Does this correctly handle event definitions with configuration maps (such as `pull_request: { types: [opened] }`), or could mapping values be incorrectly parsed as event names? | Yes. A yaml.v3 mapping node stores alternating key and value nodes in `Content`. Incrementing by two and reading `Content[i]` therefore returns only top-level event keys; a configuration mapping such as the value of `pull_request` is skipped, not mistaken for another event. |
| G6 | rules | ✅ CONFIDENT | The test comment notes that job-level and workflow-level `env:` indirection is not checked by `TestNoReviewReachableWorkflowDispatchesOnTheEventRef`. Does this violate strict static analysis rules by leaving open unmonitored pathways for `github.ref_name` dispatches? | No. The project standards do not require this test to be a complete general-purpose static analyzer. It is explicitly defense in depth, while the exact changed hand-off is independently pinned to `steps.target.outputs.branch`. The known job/workflow-level env gap therefore does not leave the reported regression unguarded. |
| G7 | parity | ✅ CONFIDENT | When `deliver-verify.yml` is executed via `workflow_dispatch` versus `pull_request_review`, does using `steps.target.outputs.branch` guarantee that `deliver-correct.yml` executes the exact same commit and workflow version across both entry points? | It guarantees that both entry points dispatch from the same branch name and therefore use the delivery branch's workflow version at dispatch time. It does not guarantee the same immutable commit across separate runs: a branch can advance between the two dispatches, and `--ref` is a branch ref rather than the previously captured SHA. |
| G8 | parity | ✅ CONFIDENT | If `deliver-correct.yml` or `deliver-implement.yml` dispatch back to `deliver-verify.yml` using `github.ref_name`, is there an asymmetry in how ref resolution is handled across different state transitions in the delivery loop? | There is a syntactic asymmetry, but it is safe under the current trigger topology. `deliver-correct.yml` is workflow-dispatched on the delivery branch, so its `github.ref_name` is that branch when it hands back. `deliver-implement.yml` runs only on `issue_comment`, normally on the default branch, so its initial hand-off starts verify from the default branch; verify then resolves and validates the actual PR head. Only verify is reachable from review events where `github.ref_name` can be `<pr>/merge`. |
| G9 | tests | ✅ CONFIDENT | `TestDeliverVerifyHandsOffOnTheDeliveryBranchNotTheEventRef` looks up steps by hardcoded string names (`"Hand off to correct"` and `"Validate the target PR"`). If a developer renames these step titles in `deliver-verify.yml`, will the test fail cleanly or produce false results? | It fails cleanly. Renaming either step causes an explicit fatal test message saying the named step is missing; it does not silently inspect a different step or pass. |
| G10 | tests | ✅ CONFIDENT | In `TestNoReviewReachableWorkflowDispatchesOnTheEventRef`, dispatch commands are identified by checking strings like `gh workflow run` or `dispatch-with-retry.sh`. Would this test detect a workflow step that dispatches via a custom GitHub Action or external script instead of raw CLI commands? | No. The generic test only recognizes the three encoded command patterns: direct `gh workflow run`, the retry wrapper, or a matching `gh api` REST dispatch. A `uses:` custom action or an opaque external script with another name would not be detected by this generic scan. The specific verify-to-correct regression remains protected because its dedicated test requires the known retry wrapper and exact branch ref. |
| G11 | rationale | ✅ CONFIDENT | Why rely on `steps.target.outputs.branch` output from an earlier step rather than constructing `deliver/issue-$ISSUE_NUMBER` directly in the shell script or querying `gh pr view "$PR" --json headRefName`? | The output is preferable because it reuses the branch value obtained from the authoritative PR API and already validated against the expected delivery convention. Constructing the string again would duplicate that contract, while another `gh pr view` would add a redundant API call and a second failure/race point. |
| G12 | rationale | ✅ CONFIDENT | Why was the ref fix implemented specifically at the `deliver-verify.yml` hand-off step rather than standardizing branch resolution inside `dispatch-with-retry.sh` itself? | Branch resolution belongs at the workflow call site because `dispatch-with-retry.sh` is a generic retry/forwarding utility and has no PR number, issue semantics, or knowledge of delivery branch conventions. Making it rewrite refs would couple the generic wrapper to one transition and could alter legitimate callers. The verify job already has the validated branch available. |
| G13 | scope | ✅ CONFIDENT | This PR adds `TestNoReviewReachableWorkflowDispatchesOnTheEventRef` which scans all workflow YAML files in `.github/workflows/`. Could this generic guard break non-delivery utility workflows that legitimately dispatch on `github.ref_name` upon PR events? | It could reject a future non-delivery workflow if that workflow is review-reachable and has a recognized dispatch command plus any `github.ref_name` occurrence in the same run script or step env, even if the occurrence is not actually used as the dispatch ref. It does not currently break such a workflow: the present workflow set passes. |
| G14 | scope | ✅ CONFIDENT | Does asserting the producer step `Validate the target PR` inside `TestDeliverVerifyHandsOffOnTheDeliveryBranchNotTheEventRef` expand test scope beyond verifying the reported dispatch regression? | It broadens the assertion from the consumer line to its direct producer, but that is necessary to verify the effective behavior. Checking only that the hand-off references `steps.target.outputs.branch` would pass even if the output were removed, in which case the empty ref could fall back to the default branch. |
| G15 | errors | ✅ CONFIDENT | If `steps.target.outputs.branch` is empty or unset, does `dispatch-with-retry.sh` fail fast with a non-zero exit code or allow `gh workflow run` to fall back to default branch behavior? | `dispatch-with-retry.sh` does not fail fast on an empty ref. It forwards `--ref ""`; if `gh workflow run` accepts that as an unspecified ref and dispatches the default branch, the wrapper exits successfully on the first attempt. It only returns non-zero when `gh` itself fails on all three attempts. |
| G16 | errors | ✅ CONFIDENT | In `TestNoReviewReachableWorkflowDispatchesOnTheEventRef`, if `yaml.Unmarshal` fails on an invalid workflow file, `t.Fatalf` stops execution immediately—does the failure message provide sufficient context regarding which YAML structure or line failed? | Yes. The failure includes the exact workflow path and yaml.v3's parse error, which normally carries line information for syntax errors. That is sufficient to locate the invalid file; identifying a higher-level expected structure is unnecessary because unmarshalling failed before structural inspection. |

### Items to fix

### Important to consider
- **F1** — The repository-wide guard is broader than the one-line runtime fix, but it directly supports the issue's request to audit other reachable dispatches.
  _.github/workflows/deliver-verify.yml:316-346 validates the PR head and emits `branch`.; .github/workflows/deliver-verify.yml:985-1006 passes `steps.target.outputs.branch` to the correction dispatch.; scripts/deliver_guards_test.go:584-652 guards the specific regression.; scripts/deliver_guards_test.go:673-768 audits review-reachable workflows.; `go test ./scripts -count=1` passes._
- **F3** — The shorthand `workflow_dispatch: main` describes the current implement-driven dispatch. A manually initiated workflow_dispatch can select another valid ref, but that does not affect the fix.
  _.github/workflows/deliver-implement.yml:683-692 dispatches verify using the implement run's event ref.; .github/workflows/deliver-verify.yml:988-1003 explains why that ref differs from the resolved delivery branch and why the latter is now used.; scripts/deliver_guards_test.go:566-583 states the same dependency enforced by the test._
- **G2** — There remains an unavoidable remote-ref deletion race between validation and dispatch; it fails visibly rather than dispatching the wrong workflow copy.
  _.github/workflows/deliver-verify.yml:330-339 reads and validates the live PR head repository, branch, and state.; .github/workflows/deliver-verify.yml:343-346 only emits the branch after those checks pass.; .github/workflows/deliver-verify.yml:996-999 documents that the PR merge ref is rejected with HTTP 422.; scripts/dispatch-with-retry.sh:10-23 retries a failed dispatch three times and then exits non-zero._
- **G3** — The head-movement recheck only gates `ready`, not `correct`; therefore a correction may start on a newer branch head than the one whose findings were collected. That behavior predates this ref correction and is serialized by the delivery concurrency group.
  _docs/contributing/standards/invariants.md:115-119 defines INV-5 as request timestamp ordering.; .github/workflows/deliver-verify.yml:69-100 defines shared branch-keyed, non-cancelling concurrency.; .github/workflows/deliver-verify.yml:462-468 rejects CI running on a SHA other than the validated SHA.; .github/workflows/deliver-verify.yml:892-927 rechecks head movement before a `ready` outcome.; .github/workflows/deliver-verify.yml:1004-1006 dispatches correction on the live branch ref._
- **G6** — The generic guard can be extended if job- or workflow-level ref indirection is later introduced.
  _scripts/deliver_guards_test.go:669-672 explicitly scopes the residual env-indirection gap and identifies the consumer assertion as the primary guard.; scripts/deliver_guards_test.go:616-636 directly verifies the actual verify-to-correct command and its `--ref` expression.; docs/contributing/standards/rules.md:1-20 lists the applicable rule tiers; no rule mandates exhaustive static analysis of every possible dispatch indirection._
- **G7** — This is intended current-branch behavior, not immutable-run parity; exact commit pinning would require a different dispatch contract.
  _.github/workflows/deliver-verify.yml:330-346 captures both the current head SHA and branch.; .github/workflows/deliver-verify.yml:1004-1006 passes the branch, not `steps.target.outputs.sha`, to workflow dispatch.; .github/workflows/deliver-verify.yml:462-468 demonstrates that the workflow already accounts for branches moving between operations._
- **G8** — Using explicit resolved branches in every transition could improve uniformity, but the remaining uses are not exposed to the review merge-ref defect.
  _.github/workflows/deliver-correct.yml:27-35 has only `workflow_dispatch` and .github/workflows/deliver-correct.yml:328-333 uses its event ref.; .github/workflows/deliver-implement.yml:42-45 has only `issue_comment` and .github/workflows/deliver-implement.yml:683-692 uses its event ref.; .github/workflows/deliver-verify.yml:37-67 includes review triggers.; .github/workflows/deliver-verify.yml:260-346 resolves and validates the PR branch before later work._
- **G9** — A title-only rename requires updating the test, which is expected for this structural workflow contract.
  _scripts/deliver_guards_test.go:605-612 returns `false` when an exact step name is absent.; scripts/deliver_guards_test.go:616-621 fatals if `Hand off to correct` is absent or no longer dispatches the expected workflow.; scripts/deliver_guards_test.go:639-642 fatals if `Validate the target PR` is absent._
- **G10** — If the repository adopts a custom dispatch action, the static guard should be extended alongside that change.
  _scripts/deliver_guards_test.go:732-745 contains the complete dispatch-recognition logic.; scripts/deliver_guards_test.go:590-596 parses only step name and `run` for the dedicated test, not `uses`.; scripts/deliver_guards_test.go:616-636 requires the actual hand-off to invoke `dispatch-with-retry.sh deliver-correct.yml` with the target branch._
- **G13** — `pull_request_target` normally uses the base branch ref rather than a PR merge ref, so including it makes this future-facing guard stricter than the defect's exact trigger class.
  _scripts/deliver_guards_test.go:673-682 globs every workflow YAML file, not only `deliver-*` files.; scripts/deliver_guards_test.go:687-692 treats four PR/review trigger types as review-reachable.; scripts/deliver_guards_test.go:732-758 flags based on co-occurrence in the step rather than tracing the value specifically into `--ref`.; `go test ./scripts -count=1` passes against the current workflows._
- **G15** — Adding generic non-empty validation for `--ref` in the wrapper would be additional hardening, but the new producer test prevents the changed path from losing its output unnoticed.
  _scripts/dispatch-with-retry.sh:8-15 performs no argument validation and exits zero on any successful `gh workflow run`.; scripts/dispatch-with-retry.sh:16-23 retries failures and exits one only after exhaustion.; scripts/deliver_guards_test.go:581-583 and 648-650 explicitly document the empty-ref default-branch fallback that the producer check prevents._

namasl (approved):

PR #1763, opened 2026-09-17
hardening(ci): auto-resolve merge conflicts in the L1 delivery loop

### LINKED ISSUE #1758: hardening(ci): L1 delivery loop stalls silently when a delivery PR conflicts with / falls behind main — auto-update the branch

The L1 delivery loop has **no handling for a `deliver/issue-N` PR that conflicts with (or falls behind) `main`**. When such a PR goes `mergeStateStatus: DIRTY`, the loop stalls **silently** and a human must merge `main` and resolve conflicts by hand before anything else can proceed.

**Why it stalls (mechanism)**

A conflicting PR has **no `refs/pull/N/merge` ref** (GitHub can't compute the merge commit). So:

1. A `pull_request_review` requesting changes can't be processed against the merge ref, and the hand-off to correct fails anyway (see #1751). The requested changes **never route to the correction agent** — no round label, no "Delivery stopped" comment.
2. The PR is left at a **stale `ready-for-merge`** label with `reviewDecision: CHANGES_REQUESTED` and `mergeStateStatus: DIRTY` — unmergeable, and nothing retrying.

**Evidence (verified 2026-09-17)**

R1 sub-issues **S4 (PR #1755, closes #1731)** and **S6 (PR #1754, closes #1733)** both went `DIRTY` — a change to `cmd/hfconfig.go` landed on `main` (`7c727681`) after both branches forked (`4fa5ca1e`), and both PRs rewrite `resolveModelConfig` in that file. Both carried a `CHANGES_REQUESTED` review whose findings were **never picked up**, and both sat at stale `ready-for-merge`. Each required a **manual `main` merge + conflict resolution + push** to make any progress.

**Today's behavior (verified)**

`grep -nEi "mergeable|conflict|DIRTY|rebase|merge .*main" .github/workflows/deliver-*.yml` finds **no git-conflict handling**. The only `conflicts` logic (`deliver-verify.yml`, the `verdict == CONFLICTS` gate) is the **archon-plan dist** gate — plan distance, not git. A git-conflicting branch is entirely unhandled.

**Proposed automation** (any of, roughly in order of value)

1. **Auto-update the delivery branch.** Before verify/correct proceeds, check `mergeStateStatus`; if `DIRTY`/`BEHIND`, merge (or rebase) `origin/main` into `deliver/issue-N` and push (a `GITHUB_TOKEN` push re-triggers push-verify). Attempt auto-resolution; on unresolvable conflicts, hand to the correction agent with an explicit *"resolve merge conflicts against main"* instruction.
2. **Keep branches current proactively** — merge `main` at the start of each correction round, so drift never accumulates into a hard conflict.
3. **At minimum, fail loudly.** Detect `DIRTY` and stamp `needs-human` with an explicit *"merge conflict — resolve against main"* message. Today it stalls with **no comment at all**, which is the worst failure mode (indistinguishable from "still working").

**Acceptance**

A delivery PR that conflicts with `main` is either (a) auto-updated and re-verified to a real verdict, or (b) explicitly flagged `needs-human` naming the conflict — **never a silent stall at stale `ready-for-merge`**.

- [x] None (CI workflow correctness) — but the silent stall makes a stuck delivery indistinguishable from an in-progress one.

**Files affected**
- `.github/workflows/deliver-verify.yml` and/or `deliver-correct.yml` (add a mergeable/base-sync step), possibly a new small `deliver-*` step or the stall-sweep.

- **#1751** — the review-hand-off ref bug; compounds this (even with conflicts cleared, review-driven fixes need #1751 fixed or a push-triggered re-verify). Fixing both is what makes post-convergence, drifted-branch corrections work end to end.
- #1654 (the L1 loop), #1722 (closed, runner death), #1737 (correct-phase efficiency). Surfaced re-driving R1 tracker #1727.

mtoslalibu:
## What this fixes

Fixes #1758.

A `deliver/issue-N` PR that **conflicts with main** stalled the L1 delivery loop **silently**. The verify phase reads three signals — CI, the archon plan, and the review verdict — and **none of them see a git conflict**. So a conflicting PR could be marked `ready-for-merge` (a stale label a human cannot act on, since GitHub cannot compute its merge ref), and the loop sat there with nothing retrying. Evidenced in #1758 by R1 sub-issues S4 (#1755) and S6 (#1754), which both went `DIRTY` after a `cmd/hfconfig.go` change landed on main and each needed a manual merge.

## Approach

Detect the conflict, route it to the correction agent to resolve, and let the normal review loop catch a bad resolution — exactly the design from the issue discussion. **Scope: conflicts only** (`DIRTY`); a merely-behind-but-clean branch stays mergeable and does not spend a correction round.

- **`deliver-verify.yml`** — a new `mergestate` step reads the PR's `mergeable_state` over REST (polled, because GitHub computes mergeability asynchronously) and feeds `MERGE_STATE` to the gate. A signal-table row surfaces it in the decision comment.
- **`deliver-gate.sh`** (the tested decision brain) — a conflicting branch can **never** be marked `ready`; it routes to a correction round. This is deliberately *not* treated as the GREEN-vs-blocking disagreement (the reviewer approved the code, not the mergeability), so it corrects rather than stopping at `needs-human`. It is bounded by the existing round cap. An undeterminable state withholds `ready` loudly.
- **`deliver-correct.yml`** — the agent's first task (STEP 1) is `git merge origin/main` and resolve conflicts using the PR's intent; findings become STEP 2 (conditional, since a merge-only round has no NOT-GREEN comment). An unresolvable conflict stops for a human rather than forcing a guess.

The resolution then re-runs build/test/lint + blis-pr-review, so a bad merge is caught like any other agent code. Every path ends at a real verdict or an explicit `needs-human` — **never a silent stall**.

## Behavioral contracts

- **BC-1** GIVEN a delivery PR whose review is GREEN, CI green, plan clean WHEN the branch conflicts with main THEN the gate returns `correct` (not `ready`), and the correction agent is dispatched to merge main.
- **BC-2** GIVEN a conflicting branch at the correction round cap THEN the gate returns `needs-human` naming the conflict and the cap (never `ready`).
- **BC-3** GIVEN the branch's mergeability cannot be determined WHEN every other signal is green THEN the gate withholds `ready` (`needs-human`), rather than trusting an unverified merge state.
- **BC-4** GIVEN a mergeable branch THEN behavior is unchanged (regression guard).
- **BC-5** GIVEN a GREEN review but a *failing objective signal* (CI) on a conflicting branch THEN the existing disagreement guardrail still wins (`needs-human`).

## Testing

- `scripts/deliver_gate_test.go` — `MERGE_STATE` added to the wiring/domain/cross-product coverage plus a dedicated conflict-routing suite (conflict→correct, conflict-at-cap→needs-human, unknown→needs-human, mergeable→ready, disagreement precedence).
- `scripts/deliver_guards_test.go` — asserts verify **derives** `mergeable_state` and **feeds** `MERGE_STATE` to the gate step *before* it decides (ordering + env wiring), and that the correct prompt tells the agent to merge `origin/main`, resolve conflicts, and stop for a human when unresolvable.
- `go test ./scripts/...`, `go build ./...`, `go vet ./scripts/...`, `gofmt` — all green.

Live end-to-end (a real conflicting delivery PR auto-resolved on the self-hosted runner) is an operator step; the loop's decision logic is fully exercised by the gate cross-product here.


mtoslalibu:
## Self-review (local `pr-review-toolkit:code-reviewer`, independent brief)

Ran an independent code review over `git diff origin/main...HEAD`, plus the full `go test ./scripts/...` suite and **mutation testing** of the two load-bearing spots (the gate conflict branch and the verify→gate wiring). Both mutants were caught by the tests.

**Verdict: GREEN — no CRITICAL, IMPORTANT, or MINOR findings.**

Confirmed:
- **A conflicting branch can never reach `ready`.** `ready` is emitted from one line, reachable only with `blocking=""` AND `GREEN` AND `DISMISSALS=none` AND `MERGE_STATE=mergeable`. `conflicting` falls through to the round cap → `correct` (or `needs-human` at cap); `unknown` emits `needs-human` directly. The conflict path does **not** bypass the GREEN-vs-blocking disagreement guardrail (evaluated first) or the outstanding-dismissal block (same nesting level). `reason` is set on every cap-reaching path, so `set -u` never trips.
- **The mergeability derivation fails closed.** `dirty→conflicting`, `""/null/unknown→unknown` (after polling), everything else→`mergeable`. A `gh api` failure or an all-unknown poll yields `unknown` → gate withholds `ready`. Robust under `set -uo pipefail`; `REPO`/`PR`/`GH_TOKEN` are job-level env.
- **The correct-phase prompt has no hole.** STEP 1 (`git merge origin/main` + resolve) always runs, so a merge-only round is never a no-op; the `DELIVER-DISMISSALS: <n>` contract is intact (0 on a merge-only round); the reworded constraint permits only the `origin/main`→branch update, not a push to main or a merge of the PR itself; an unresolvable conflict routes to `needs-human`.
- **Tests pin behavior (mutation-confirmed), not structure**, and the `AlwaysDecides` cross-product now includes the merge-state dimension. No R1-R23 violations — every unknown/unreadable path is loud.

No fixes were needed. Ready for a human to merge.

mtoslalibu:
## blis-pr-review (run locally, full methodology)

Ran the `blis-pr-review` skill's methodology locally — `pr-review-toolkit:review-pr` across the applicable code-review perspectives plus the silent-failure, test-quality, and comment-accuracy specialists. Cross-path parity, preemption, and timeout/deadline perspectives are **N/A** (no `sim/` command paths, per-request metrics, or timeout code touched — CI-only change).

### Findings and how they were addressed

**IMPORTANT (fixed in 6e043c9e) — the mergeable_state mapping was untested.** The `dirty→conflicting / ""→unknown / *→mergeable` mapping lived inline in the verify workflow step, proven only by its inverse (the gate rejecting raw values). A typo there (`dirty)`→`dirtry)`) would have mapped a conflicting branch to `mergeable`→`ready` — the exact #1758 bug — with every test green. Extracted the pure mapping into `scripts/map-merge-state.sh` (same script-not-YAML rationale as `deliver-gate.sh`) and added `scripts/map_merge_state_test.go` cross-producting every GitHub `mergeable_state` value. The guards test now asserts the step maps through that script.

**IMPORTANT (fixed) — gate/derivation run-condition drift.** Added a guards assertion that the `mergestate` and `gate` steps share the same `if:` condition, so the gate can never run with an empty `MERGE_STATE` (which would exit 2). Fails closed already, now pinned.

**MINOR (fixed) — precedence untested.** Added gate cases: an open dismissal and a plan-regression disagreement both beat the conflict path (`needs-human`), while an honest NOT-GREEN on a conflicting branch still corrects. Hedged the "behind is mergeable" comment to note it relies on the correction phase merging main each round.

**No CRITICAL findings.** Verified (mutation-tested) that a conflicting branch can never reach `ready`; every mergeability failure mode (API error, rate limit, null, persistent unknown) fails **closed** to `unknown`→`needs-human`, never open to `mergeable`; the correct-phase agent cannot launder a still-conflicting branch past the gate because mergeability is re-measured from GitHub each round. Comments verified accurate against the control flow. BC-1..BC-5 each have a matching behavioral test that fails on revert.

### Verify
`go test ./scripts/...`, `go build ./...`, `go vet ./scripts/...`, `gofmt` — all green.

**Verdict: ready to merge.** No open findings.

namasl:

## qa-review — PR #1763: ⛔ BLOCK
_hardening(ci): auto-resolve merge conflicts in the L1 delivery loop · 15 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. A `dirty` PR is mapped to `conflicting`, prevented from reaching `ready`, and routed to correction while rounds remain; the correction prompt explicitly merges `origin/main`, resolves conflicts, and pushes for re-verification. At the round cap it becomes `needs-human`, satisfying both accepted outcomes. Proactively merging main on other correction rounds is also one of the issue's proposed approaches rather than unrelated scope. |
| F2 | fixed | ✅ CONFIDENT | Has all documentation been updated to reflect the changes in this PR? | The affected workflow and script contracts are documented inline: the new `MERGE_STATE` domain, raw-state mapping, asynchronous polling, correction behavior, and conflict escalation are all described. There is no existing delivery-loop documentation under `docs/` referencing mergeability that required a corresponding update. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No correctness-affecting stale comments were found. Some introductory wording is now imprecise: the correction workflow says it fixes findings from a verdict, and the agent prompt initially says there is an open verdict with numbered findings, although a conflict-only correction can now have no findings. The later STEP 2 text correctly handles that case. |
| G1 | correctness | ❌ FLAW_FOUND | In `deliver-verify.yml`, `gh api "/repos/$REPO/pulls/$PR"` polls `mergeable_state` 5 times with a 3-second sleep. If GitHub's async mergeability computation takes longer than 15 seconds during API load, `MERGE_STATE` resolves to `unknown` and `deliver-gate.sh` emits `needs-human`. Could this cause transient false-positive `needs-human` escalations on clean PRs? | Yes. The five reads occur at approximately 0, 3, 6, 9, and 12 seconds. If all five return null/`unknown`—or transiently fail—`state` remains empty, maps to `unknown`, and an otherwise all-green PR is immediately given the terminal `needs-human` decision. The final sleep does not lead to a sixth read. This is a realistic false-positive escalation during delayed GitHub computation or temporary API trouble. |
| G2 | correctness | ✅ CONFIDENT | In `deliver-correct.yml` STEP 1, the prompt instructs the agent to run `git fetch origin main` and `git merge --no-edit origin/main`. If the agent's workspace contains untracked build artifacts or modified files from a previous step, `git merge` will fail. How does the workflow guarantee a clean git working tree before executing STEP 1? | Before the agent runs, the workflow uses `actions/checkout@v5` with a full-history checkout. The checkout action's default cleaning behavior resets and cleans the repository, and the only intervening workflow step merely records a timestamp. The agent is then told to check out the PR branch before STEP 1. Thus artifacts from an earlier job on the persistent runner are cleaned before the merge; files created by the agent before obeying its first task instruction cannot be mechanically prevented. |
| G3 | rules | ✅ CONFIDENT | BLIS rule R1 forbids silent error suppression. In `deliver-verify.yml` line 894, `gh api "/repos/$REPO/pulls/$PR" --jq '.mergeable_state // ""' 2>/dev/null \|\| echo ""` suppresses `stderr` and masks API call failures. Why are API failures silenced rather than logged or treated as explicit workflow errors? | The API failure itself is not propagated, but its result is not silently trusted or dropped: it is converted to an empty read, each empty attempt is logged, and after retries the state becomes `unknown`, which fails closed as `needs-human`. Therefore this does not silently lose a delivery or permit a bad merge under R1. It does, however, conceal the API's actual error message and makes an API outage indistinguishable from pending computation. |
| G4 | rules | ✅ CONFIDENT | In `scripts/map-merge-state.sh`, the wildcard case `*) echo mergeable ;;` maps any unrecognised GitHub `mergeable_state` to `mergeable`. Does defaulting unhandled state values to `mergeable` risk violating strict state handling if GitHub introduces new non-mergeable states? | For GitHub's currently handled REST state surface, only `dirty` denotes a merge conflict; states such as `behind`, `blocked`, `unstable`, `has_hooks`, and `draft` can still have a merge ref and are intentionally considered mergeable for this conflicts-only gate. The wildcard behavior is explicitly tested and therefore is not accidental. A hypothetical future state with different semantics could be misclassified, but that is not a defect against the current API behavior. |
| G5 | tests | ✅ CONFIDENT | In `scripts/deliver_guards_test.go`, `TestDeliverVerifyFeedsMergeStateToTheGate` inspects step IDs and environment keys using YAML parsing. If step IDs or internal workflow step structures in `deliver-verify.yml` are refactored, will this structural test break even if the workflow runtime behavior remains unchanged? | Yes, renaming the `gate` or `mergestate` IDs, moving the logic into a composite action, or changing the environment wiring would require updating this test even if equivalent runtime behavior were retained. That coupling is intentional: those IDs and outputs are the workflow's actual expression-level wiring, and the test is specifically guarding order, co-firing conditions, and data flow rather than merely searching for text. |
| G6 | tests | ✅ CONFIDENT | `scripts/map_merge_state_test.go` unit-tests `map-merge-state.sh` in isolation. Are there end-to-end integration tests that verify how `deliver-verify.yml` executes when `gh api` outputs `dirty` versus `unknown`? | No live end-to-end test executes the GitHub Actions workflow with mocked `gh api` responses. Coverage is compositional: one test executes the mapping script for `dirty` and `unknown`, gate tests verify their resulting decisions, and a YAML structural test verifies that the workflow reads `mergeable_state`, invokes the mapping script, and passes its output to the gate. The scripts package test suite passes. |
| G7 | rationale | ✅ CONFIDENT | Why was `map-merge-state.sh` created as a standalone shell script rather than adding the mapping logic directly into `deliver-gate.sh` alongside its existing input validation? | The repository's stated rationale is separation of concerns and direct testing. The workflow owns the asynchronous REST polling, the mapper is a pure normalization layer with no network or state, and `deliver-gate.sh` consumes only its small validated domain. Making the mapping standalone lets tests exercise every known raw GitHub state and catch a `dirty` mapping typo independently of the decision cross-product. |
| G8 | rationale | ⚠️ CANNOT_ANSWER | In `deliver-correct.yml`, STEP 1 instructs the agent to merge `origin/main` into the PR branch using `git merge --no-edit`. What alternatives (such as `git rebase` or GitHub's update-branch REST API) were evaluated, and why was local git merging in the runner selected? | The committed code establishes why merge is operationally suitable, but it does not record which alternatives were actually evaluated. Local merging preserves the delivery branch's existing commits, avoids a force-push required by rebase, works for both clean updates and agent-assisted conflict resolution, and does not depend on GitHub's update-branch endpoint being able to resolve conflicts. I cannot verify from the repository that rebase or the update-branch API were formally evaluated before this choice. |
| G9 | scope | ✅ CONFIDENT | Issue #1758 targets handling PRs in `dirty` (conflicting) state. However, `deliver-correct.yml` instructs the correction agent to run `git merge origin/main` on *every* correction round, even when `mergeable_state` was clean. Is merging `main` on non-conflicting correction rounds intentional, and could it introduce unnecessary merge commits? | Yes, it is explicitly intentional and matches the issue's proactive keep-current option. If the branch is already current, `git merge origin/main` creates no commit. If it is behind but non-conflicting, it creates a merge commit, trading some history noise for preventing drift from accumulating before later correction rounds. |
| G10 | scope | ✅ CONFIDENT | The acceptance criteria in issue #1758 allow flagging `needs-human` on unresolvable conflicts. How does the PR ensure that an agent attempting to auto-resolve complex merge conflicts does not silently introduce incorrect code changes or semantic regressions during conflict resolution? | The safeguard is layered rather than an absolute proof: the prompt requires intent-based resolution and forbids blindly choosing either side; uncertain conflicts must be aborted and handed to a human; resolved code must pass build, test, and lint before push; and the subsequent verify phase reruns repository CI and an independent review before `ready` is possible. The correction round cap bounds repeated failed attempts. As with any agent-authored code, semantic defects not detected by tests or review cannot be categorically excluded. |
| G11 | errors | ✅ CONFIDENT | In `scripts/map-merge-state.sh`, how does the script handle invocation with no arguments versus an explicit empty string `""`, and could an unbound parameter cause execution failure under `set -uo pipefail`? | No argument and an explicitly empty first argument behave identically: `${1-}` expands to an empty string in either case, which matches the empty/`unknown` arm and prints `unknown`. The `-` default expansion prevents an unbound-parameter error under `set -u`. |
| G12 | errors | ✅ CONFIDENT | In `deliver-correct.yml`, if `git merge --abort` fails (for instance, if files were partially staged or modified during conflict resolution), how does the agent recover and ensure the workflow terminates cleanly with a `needs-human` status? | There is no special scripted recovery for a failed `git merge --abort`; conflict handling is delegated to the agent. The safe outcomes are nevertheless bounded: the prompt forbids forcing a guessed resolution and asks the agent to report and stop; if the action fails, the always-running correction failure reporter comments, removes `ready-for-merge`, applies `needs-human`, and does not hand back for verification. If the agent returns success without pushing, the unchanged remote branch remains conflicting and cannot be marked ready; subsequent verification/correction is still bounded by the round cap. |

### Items to fix
- **G1 · FLAW_FOUND** — Yes. The five reads occur at approximately 0, 3, 6, 9, and 12 seconds. If all five return null/`unknown`—or transiently fail—`state` remains empty, maps to `unknown`, and an otherwise all-green PR is immediately given the terminal `needs-human` decision. The final sleep does not lead to a sixth read. This is a realistic false-positive escalation during delayed GitHub computation or temporary API trouble.
  _.github/workflows/deliver-verify.yml:889-911 performs five reads and leaves `state` empty; scripts/map-merge-state.sh:27-30 maps empty to `unknown`; scripts/deliver-gate.sh:138-143 emits `needs-human` for unknown on the ready path; .github/workflows/deliver-verify.yml:971-995 applies that terminal decision._
- **G8 · CANNOT_ANSWER** — The committed code establishes why merge is operationally suitable, but it does not record which alternatives were actually evaluated. Local merging preserves the delivery branch's existing commits, avoids a force-push required by rebase, works for both clean updates and agent-assisted conflict resolution, and does not depend on GitHub's update-branch endpoint being able to resolve conflicts. I cannot verify from the repository that rebase or the update-branch API were formally evaluated before this choice.
  _.github/workflows/deliver-correct.yml:192-208 specifies local merge and conflict resolution; .github/workflows/deliver-correct.yml:246-248 explicitly limits the permitted merge to `origin/main` into the delivery branch._

### Important to consider
- **F1** — The short polling window can falsely escalate a clean PR when GitHub remains asynchronous; see G1.
  _.github/workflows/deliver-verify.yml:877-927; scripts/map-merge-state.sh:27-30; scripts/deliver-gate.sh:129-159; .github/workflows/deliver-correct.yml:192-208,246-264; .github/workflows/deliver-verify.yml:971-1019_
- **F3** — Consider updating the introductory wording to mention branch synchronization and conflict-only rounds.
  _.github/workflows/deliver-correct.yml:3-8,168-170,210-215_
- **G2** — The workflow relies on the checkout action's default `clean: true`; setting it explicitly would make this guarantee more visible.
  _.github/workflows/deliver-correct.yml:122-133 performs checkout before the agent; .github/workflows/deliver-correct.yml:135-190 makes branch checkout the agent's first repository operation; .github/workflows/deliver-correct.yml:192-200 then performs the fetch and merge._
- **G3** — Capturing stderr and emitting a warning after each failed API call would materially improve diagnosis and would help distinguish this from GitHub's legitimate asynchronous `unknown` state.
  _docs/contributing/standards/rules.md:24-32 defines R1; .github/workflows/deliver-verify.yml:892-910 logs empty attempts and maps the result; scripts/deliver-gate.sh:138-143 fails closed on unknown._
- **G4** — Failing unknown future values closed would be stricter and more consistent with the gate's general distrust of unmapped external values.
  _scripts/map-merge-state.sh:19-23,27-30; scripts/map_merge_state_test.go:35-58,76-86_
- **G5** — This is normal contract-test maintenance rather than a runtime defect.
  _scripts/deliver_guards_test.go:837-846 documents the structural intent; scripts/deliver_guards_test.go:872-925 asserts IDs, ordering, conditions, mapping invocation, and output wiring._
- **G6** — The polling and API-failure behavior itself is not exercised, which leaves the transient escalation in G1 outside test coverage.
  _scripts/map_merge_state_test.go:18-74; scripts/deliver_gate_test.go:316-391; scripts/deliver_guards_test.go:847-927; reproduced with `go test ./scripts -count=1` (pass)._
- **G7** — The mapping could also have been implemented and unit-tested inside `deliver-gate.sh`; the standalone choice is architectural, not required for correctness.
  _scripts/map-merge-state.sh:10-17; scripts/deliver-gate.sh:31-33,82-85; scripts/map_merge_state_test.go:11-16,35-58_
- **G9** — Repositories preferring linear history could use rebase, but that would require force-push handling and is a design tradeoff rather than a defect.
  _.github/workflows/deliver-correct.yml:192-202 explicitly says every round merges main to prevent harder conflicts; scripts/map-merge-state.sh:19-23 treats `behind` as mergeable while noting correction always merges main._
- **G12** — The prompt could explicitly say what to do if `git merge --abort` itself fails, such as stop immediately, avoid pushing, and include `git status` in the report.
  _.github/workflows/deliver-correct.yml:202-208; .github/workflows/deliver-correct.yml:350-357 only hands back after successful preceding steps; .github/workflows/deliver-correct.yml:359-392 reports failure and applies `needs-human`; scripts/deliver-gate.sh:139-157 prevents a conflicting branch from becoming ready and enforces the cap._

mtoslalibu:
## Response to qa-review (⛔ BLOCK) — addressed in c071344c

Thanks — the two flagged items were both real. Reasoned through every item; fixes and rationale below.

### G1 · FLAW_FOUND (the block) — FIXED
Correct: the poll did 5 reads at ~0/3/6/9/12s and the final `sleep` led to no sixth read, so a slow or troubled GitHub API could escalate an all-green PR to a terminal `needs-human`. Fixed in `deliver-verify.yml`: the loop now **reads on every attempt including the last** (sleeps only *between* attempts) and widens to **8 reads over ~21s**. It still fails closed, but the spurious-escalation window is much smaller and no read is wasted. (Note: this step runs *after* CI + review, so the head has sat on GitHub for minutes and mergeability is normally long since computed — the poll now only bridges genuine API trouble.)

### G3 · fail-closed but opaque — FIXED
`2>/dev/null` hid whether an `unknown` was an API outage or GitHub's legitimate async state. Now the gh-api stderr is captured and surfaced as a `::warning::` on failure (still fails closed to `unknown`) — the R1 posture is now *visible and safe*, not silent.

### G4 · wildcard mapped future states to `mergeable` (fail-open) — FIXED
This was inconsistent with `deliver-gate.sh`'s own rule that an out-of-domain value is not trusted. `map-merge-state.sh` now **enumerates** the known mergeable states (`clean|behind|blocked|unstable|has_hooks|draft`) and the wildcard **fails closed to `unknown`**. A genuinely-new GitHub state now stops for a human (recoverable next run) until classified deliberately, rather than being assumed safe to merge. All eight documented `mergeable_state` values are handled explicitly; `map_merge_state_test.go` pins each — including that `blocked` (a protected PR's normal state) stays `mergeable`, so protection rules don't escalate every PR.

### G8 · CANNOT_ANSWER (merge vs rebase/update-branch) — DOCUMENTED
Added the rationale to `deliver-correct.yml`'s header: a local `git merge origin/main` preserves the branch's commits (no force-push, which the push-triggered re-verify and the agent's checkout would fight) and is the only form an agent can carry through when the update **conflicts** — the update-branch API only fast-forwards a clean branch and cannot resolve conflicts, which is the whole point of #1758.

### F3 · imprecise intro wording — FIXED
Reworded the correct-phase header and prompt intro to state a round may update the branch, fix findings, or both — a conflict-only round legitimately has no findings.

### Considered, not changed (with reasoning)
- **G2** (explicit `clean: true` on checkout): the `actions/checkout@v5` default already cleans, and the agent's first op is `gh pr checkout`; the guarantee holds. Left as-is to keep the diff focused.
- **G12** (`git merge --abort` itself failing): already bounded — the `always() && !success()` failure reporter comments, removes `ready-for-merge`, applies `needs-human`, and does not hand back; a no-push agent leaves the branch conflicting, which cannot be marked ready and is bounded by the round cap.
- **G5/G6/G7/G9/G10/G11**: all marked CONFIDENT (non-flaws) and I concur — the structural coupling is intentional, coverage is compositional, proactive main-merge is the issue's own option, and conflict resolution is layered (intent-based + abort-on-uncertain + build/test/lint + independent re-review + round cap).

### Verify
`go test ./scripts/...`, `go build ./...`, `go vet ./scripts/...`, `gofmt` — all green. Re-reviewed the changed error-handling with the silent-failure specialist: fails closed on every path, no stale `raw`, `blocked` regression avoided, full state coverage.

**Every blocking/actionable item is addressed. Ready for a human to merge.**

namasl:

## qa-review adjudication — PR #1763: ⛔ BLOCK

|----|----|----|----|
| G1 | FLAW_FOUND | ❌ STILL-OPEN | The change increases polling from five reads to eight and removes the wasted final sleep, but it does not eliminate the reported false-positive terminal escalation. If all eight reads return empty/unknown or transiently fail, `state` remains empty, maps to `unknown`, and an otherwise green PR is still immediately stopped as `needs-human`. An implementation fix must make temporary mergeability uncertainty nonterminal/retryable, or the author must provide a concrete, independently checkable justification that this terminal behavior is required and the remaining window cannot realistically cause the reported false positive. |
| G8 | CANNOT_ANSWER | ✅ RESOLVED | The current workflow now records the requested comparison and rationale: local merge preserves existing commits without a force-push, while rebase would conflict with the push-triggered verification/agent checkout model, and the update-branch API cannot carry conflict resolution through. This directly documents why local merge was selected over both alternatives. |

### Still blocking
- **G1** — The change increases polling from five reads to eight and removes the wasted final sleep, but it does not eliminate the reported false-positive terminal escalation. If all eight reads return empty/unknown or transiently fail, `state` remains empty, maps to `unknown`, and an otherwise green PR is still immediately stopped as `needs-human`. An implementation fix must make temporary mergeability uncertainty nonterminal/retryable, or the author must provide a concrete, independently checkable justification that this terminal behavior is required and the remaining window cannot realistically cause the reported false positive.
  _.github/workflows/deliver-verify.yml:896-914 leaves `state` empty after eight unresolved reads; scripts/map-merge-state.sh:33-36 maps empty and unknown values to `unknown`; scripts/deliver-gate.sh:138-143 emits terminal `needs-human` for that state on the all-green path; .github/workflows/deliver-verify.yml:984-1008 labels and announces that the delivery loop has stopped._

### Cleared
- **G8** (✅ RESOLVED) — The current workflow now records the requested comparison and rationale: local merge preserves existing commits without a force-push, while rebase would conflict with the push-triggered verification/agent checkout model, and the update-branch API cannot carry conflict resolution through. This directly documents why local merge was selected over both alternatives.
  _.github/workflows/deliver-correct.yml:12-16 explicitly compares local `git merge origin/main` with rebase and GitHub's update-branch API and states the operational reasons for the choice._

mtoslalibu:
## Response to qa-review adjudication — G1 now fixed in code (16673ba4)

Agreed — widening the poll was mitigation, not a fix, and the adjudicator is right that the terminal escalation was the real problem. I took the implementation route (make temporary uncertainty non-terminal) rather than defend terminal behavior, because terminal behavior here is **not** required.

### The fix
The gate now has a fourth, **non-terminal** decision: `recheck`. It is returned from exactly one cell — every signal green (`CI success`, plan clean, `GREEN`, no open dismissal) **but** the branch's mergeability could not be read (`MERGE_STATE=unknown`).

`verify`'s `Apply the decision` handles `recheck` by:
- applying **no terminal label** (not `ready`, not `needs-human`),
- posting a comment that explicitly says the delivery is **NOT stopped** and will re-verify on the next event,
- **not** handing off, and exiting **success** so the `always() && !success()` failure reporter does not fire (that reporter is what would otherwise stamp `needs-human`).

So an undeterminable mergeability is now **re-checkable, not terminal**:
- **Transient blip (the common case):** the PR keeps no terminal label and stays in the stall sweep's in-flight set; the next push/review/dispatch re-runs verify, which re-reads mergeability (by then long since computed) and proceeds. No false "delivery stopped — a human is needed."
- **Persistent unreadable mergeability (a real, sustained outage):** the stall sweep — the loop's existing bounded backstop — escalates it after the quiet window. A human is still reached, but only when the problem is real, not on a single blip.

This is the same "don't emit a false verdict when you can't read an input; rely on the stall sweep as backstop" idiom the verify phase already uses for its `Resolve the delivery target` API-failure path.

### Why `recheck` and not the alternatives
- Not `ready`: an unverified mergeability must never merge (the whole #1758 safety property).
- Not `needs-human`: that is terminal ("stopped, won't resume") — wrong for a re-checkable blip.
- Not auto-redispatching verify: that would re-run the full CI + review cycle just to re-read one field, on a path that is rare by construction. The next natural event + the stall-sweep backstop cover it without the waste.

### Coverage
- `deliver_gate_test.go`: the unknown-would-be-ready case now asserts `recheck` and explicitly `!= ready` **and** `!= needs-human`; `AlwaysDecides` includes `recheck` in the valid decision set.
- `deliver_guards_test.go`: a new test pins that verify keeps a `recheck)` arm with an empty label, so a future edit can't silently drop it and send `recheck` down the `*)` → exit 1 → `needs-human` path.

`go test ./scripts/...`, `go build ./...`, `go vet`, `gofmt` — all green.

G8 was already cleared. With G1 resolved in code, both prior blocking findings are addressed.

namasl:

## qa-review adjudication — PR #1763: ✅ PASS

|----|----|----|----|
| G1 | FLAW_FOUND | ✅ RESOLVED | The unresolved mergeability path is no longer converted into the terminal `needs-human` decision. On otherwise-green inputs, the gate emits the non-terminal `recheck`; the workflow handles it with no terminal label, clears any stale terminal labels, records the decision successfully, and does not dispatch correction. The existing stall sweep later escalates genuinely persistent inactivity. |
| G8 | CANNOT_ANSWER | ✅ RESOLVED | The current workflow explicitly records the comparison that was previously unverifiable: local merge preserves existing commits without a force-push, while rebase would conflict with the push-triggered verification and agent-checkout model, and the update-branch API cannot carry conflict resolution through. |

### Still blocking

### Cleared
- **G1** (✅ RESOLVED) — The unresolved mergeability path is no longer converted into the terminal `needs-human` decision. On otherwise-green inputs, the gate emits the non-terminal `recheck`; the workflow handles it with no terminal label, clears any stale terminal labels, records the decision successfully, and does not dispatch correction. The existing stall sweep later escalates genuinely persistent inactivity.
  _scripts/deliver-gate.sh:145-150 emits `recheck` when MERGE_STATE is unknown. .github/workflows/deliver-verify.yml:984-994 handles `recheck` with an empty label; lines 1023-1038 clear both terminal labels and record the successful final decision. .github/workflows/deliver-stall-sweep.yml:143-167 selects in-flight deliveries, and lines 217-250 escalate only after the quiet cutoff. `go test ./scripts/...` passes._
- **G8** (✅ RESOLVED) — The current workflow explicitly records the comparison that was previously unverifiable: local merge preserves existing commits without a force-push, while rebase would conflict with the push-triggered verification and agent-checkout model, and the update-branch API cannot carry conflict resolution through.
  _.github/workflows/deliver-correct.yml:12-16 explicitly compares local `git merge origin/main` against both rebase and GitHub's update-branch API and documents why local merge was chosen._

namasl (approved):
Approving: qa-review adjudication is ✅ PASS (both prior blocking findings resolved — G1 mergeability uncertainty is now the non-terminal `recheck` decision, G8 rationale documented). See adjudication comment https://github.com/inference-sim/inference-sim/pull/1763#issuecomment-5720795164. Will merge once the pending heavy-cluster CI check completes.

mtoslalibu:
## Self-review of the `recheck` fix — GREEN, ready to merge

Ran the review specialist over the `recheck` commit (16673ba4) — the non-terminal fix for qa-review G1. **No CRITICAL or IMPORTANT findings.** Verified point by point:

- **Reachability:** `recheck` is emitted from exactly one gate cell — no blocking signal (CI success, plan clean), `GREEN`, `DISMISSALS=none`, `MERGE_STATE=unknown`. A failing-CI or conflicting branch can never reach it (they take the blocking / conflict arms first).
- **Truly non-terminal:** `recheck` applies neither `ready-for-merge` nor `needs-human` (empty label), does not hand off (hand-off is gated on `final_decision=='correct'`), and the step exits success so the `always() && !success()` failure reporter never fires. There is no path where `recheck` ends up stamping `needs-human`.
- **No loop/spin:** no auto-hand-off; it comments and exits, resuming on the next event with the stall-sweep as the bounded backstop.
- **No bad interaction:** the moved-head and ci-touched downgrades only fire for `decision==ready`, so they never touch `recheck`.
- **Tests hold:** gate asserts `unknown→recheck` (and `!=ready`, `!=needs-human`); `AlwaysDecides` includes `recheck` so exhaustiveness holds; the new guards test fails if the verify `recheck)` arm is dropped or wired to a terminal label.

One MINOR (a stale error-message string in the exhaustiveness test) fixed in 1a6090a0.

`go test ./scripts/...`, `go build ./...`, `go vet`, `gofmt` — all green.

**Verdict: ready to merge.** Both prior blocking findings (G1, G8) resolved; no open issues across the blis-pr-review and qa-review rounds.

Issue #1774, opened 2026-09-17
hardening(cmd): settle the --catalog location convention (clone-root vs entries-root) [R1 cleanup]


namasl:
Part of **#1767** (R1 cleanup). **This gates #1769, #1770, #1771, and R2** — settle and land the catalog-location contract before wiring any catalog reader or deleting the bundled fixtures.

**What antipattern or issue does this address?**

R1 (S4, #1731) made the model catalog an explicitly-located directory, but the simulator and the authoritative [`blis-catalog`](https://github.com/inference-sim/blis-catalog) repo disagree on **what `--catalog` points at**:

- BLIS implements **entries-root** semantics — [`catalogModelDir`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/cmd/hfconfig.go#L265-L285) resolves `<catalog>/<short-name>/config.json` and does **not** append `models/`. [`TestCatalogModelDir`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/cmd/hfconfig_test.go#L326) pins that behavior.
- `blis-catalog` stores configs at `<clone-root>/models/<name>/config.json`, with `workloads/`, `devices/`, `hardware/`, and `networks/` as **sibling namespaces** under the same clone root ([tree @ 77de4553](https://github.com/inference-sim/blis-catalog/tree/77de4553792f5877ff9cd3a8caf2b75a14e3f4ad)).

So an operator must today pass `--catalog <clone-root>/models`, while #1767 and the catalog README describe the clone root. Because #1769 (workloads) and #1770 (devices) read *sibling* namespaces, and R2 relocates coefficients into the same catalog, every downstream reader inherits this one choice. It must be settled first.

**Decision (recommended in [#1767 comment §1](https://github.com/inference-sim/inference-sim/issues/1767#issuecomment-5720445880)):** make `--catalog <clone-root>` the stable contract and have model resolution append `models/`. The sibling namespaces only compose cleanly off a single clone root.

[...]

Issue #1767, opened 2026-09-17
Tracking: R1 cleanup — retire dead defaults.yaml fields & resolve catalog duplication


susiejojo:
## Purpose

Tracking issue for **R1 cleanup**: residual `defaults.yaml` ↔ catalog debt that R1 (#1727) deliberately left behind. R1 established NS-6 ("a model runs iff it is in the catalog") but scoped itself to the **model config** only — it pointed `resolveModelConfig` at `<catalog>/<name>/config.json` and stopped there. It explicitly deferred everything else:

> "C6 hardware/workloads/devices dirs are NOT S-task prerequisites, so they live in blis-catalog and are not shown here." — #1727 DAG legend
> "coefficient sections … stay put until R2" — #1727, *Not in R1*

Meanwhile the catalog-side C6 work **already staged copies** of the workloads and storage-device data into [`blis-catalog`](https://github.com/inference-sim/blis-catalog) (`workloads/*.yaml`, `devices/storage.yaml` — the latter's header literally reads *"Split verbatim from the kv_offload_devices: section of the simulator's defaults.yaml"*). But **no `inference-sim` reader was ever pointed at them.** The result is a transitional state with three distinct problems:

| Block in `defaults.yaml` | In catalog? | Read from catalog? | Read from `defaults.yaml`? | Status |
|---|---|---|---|---|
| `defaults:` (GPU/TP) | ❌ (only `hf_repo` identity lives in `model.yaml`) | ❌ | ❌ (dead since NS-6) | **dead — remove (#1768)** |
| `workloads:` | ✅ (staged, C6) | ❌ not wired | ✅ `cmd/convert.go` | **duplicated — drift risk (#1769)** |
| `kv_offload_devices:` | ✅ (staged, C6) | ❌ not wired | ✅ `cmd/kv_offload.go` | **duplicated — drift risk (#1770)** |
| `model_configs/` (bundled tree) | it IS a stale copy | docs/tests wrongly point here | fixtures + doc sites | **NOT a catalog — delete (#1771)** |

None of this was filed anywhere at first — #1727 is closed, and #1602 (config-unification) is a different axis (CLI/flag surface, not source-of-truth relocation). This tracker collects the follow-ups. **#1750** (the C6 strict catalog-load CI gate) is *not* model-only — its scope covers loading **models, hardware, workloads, and devices** through the loader, so reconcile its ownership with #1769/#1770 and R2 rather than treating it as unrelated.

## Provenance

[...]

Issue #1733, opened 2026-09-16
hardening(cmd): remove run-time HF fetch and require --hardware/--tp (NS-6) [R1/S6]


namasl:
*Part of the R1 tracking issue #1727 (task **S6**). This is the deliberate NS-6 behaviour change.*
*⟂ Cross-repo prerequisites (human-coordinated, NOT machine-enforced): blis-catalog **C3** (every model has `models/<name>/{config.json,model.yaml}`) and **C4** (the 25 defaults-only entries decided). Refusing an uncatalogued model is only safe once every intended model is catalogued. Land C3 + C4 first.*
*Soft ordering: recommended to land after **S4** (#1731) so the "not in the catalog" refusal can name the `--catalog` path. Kept a soft note, not a hard `Depends on:`, per the design's stated S6 deps (= C3, C4).*

## What this fixes

Two run-time behaviours violate **NS-6 (a model runs iff it is in the catalog)**:

1. **Run time fetches and writes the catalog.** When a model config is absent locally, the run path downloads `config.json` from HuggingFace and **writes it into `model_configs/`** — so running an unknown model *adds a catalog entry as a side effect*. That is exactly how the catalog drifts.
2. **The deployment is inferred.** When `--hardware`/`--tp` are absent, BLIS looks them up per-model in `defaults.yaml` and **warns-and-continues on a deployment nobody chose**:

```
$ blis run --model qwen/qwen3-14b --num-requests 3
warning: Finding default values of TP for model=qwen/qwen3-14b
warning: Using default tp=1
warning: Finding default values of hardware for model=qwen/qwen3-14b
  ... completes, emits metrics    # on a deployment nobody chose
```

## Where it happens today

[...]

Issue #1731, opened 2026-09-16
hardening(cmd): locate the catalog via --catalog/BLIS_CATALOG and retire --model-config-folder [R1/S4]


namasl:
*Part of the R1 tracking issue #1727 (task **S4**).*
*⟂ Cross-repo prerequisite (human-coordinated, NOT machine-enforced): blis-catalog **C1** — the catalog repository exists and holds `catalog/models/`. Land C1 before merging this.*

## What this fixes

There must be **exactly one way** to tell BLIS where a model config lives, and it must not have a working-directory-dependent default. Today there are two overlapping mechanisms and a silent default:

- `--model-config-folder` — registered once in the shared helper [`registerSimConfigFlags` at `cmd/root.go#L1521`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/root.go#L1521) (var decl [`cmd/root.go#L65`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/root.go#L65)), consumed by `run` and `replay` (helper called at [`cmd/root.go#L3208`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/root.go#L3208) and [`cmd/replay.go#L1085`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/replay.go#L1085)).
- The bundled `model_configs/` fallback + `--defaults-filepath` (default `"defaults.yaml"`, [`cmd/root.go#L1520`](https://github.com/inference-sim/inference-sim/blob/328083645e9c81f3f5e27046d64a86d764d7a039/cmd/root.go#L1520)) resolved against the working directory — so `blis run` silently only works from the repo root. A relative default inherits that hazard, and it contradicts NS-6 (nothing about the model is inferred).

[...]

PR #1797, opened 2026-09-19
hardening: delete the stale model_configs/ tree [R1 cleanup]

### LINKED ISSUE #1771: hardening: delete the stale model_configs/ tree — blis-catalog is the sole catalog; migrate fixtures, docs, and the doc-guard test [R1 cleanup]
Part of **#1767** (R1 cleanup).

Depends on: #1774

> **The authoritative model catalog is [`inference-sim/blis-catalog`](https://github.com/inference-sim/blis-catalog) — and ONLY that.** The in-repo `model_configs/` tree is a **stale duplicate that must be deleted**, not a "valid catalog," not a "bundled subset," not a "day-zero fallback." Any doc or test that presents it as a catalog is wrong and is in scope here.

**Why this depends on #1774:** the docs' `BLIS_CATALOG` instruction must point at whatever layout #1774 settles. Under #1774's recommended clone-root contract, resolution appends `models/`, so the docs instruct `export BLIS_CATALOG=<blis-catalog-clone>` (the **clone root**, not `<clone>/models`). #1774 also removes the transition fallback here, once the bundled tree is gone.

**What this addresses**

R1 removed every reason for `model_configs/` to be the catalog (since S4/#1731 the location is `--catalog`/`BLIS_CATALOG` with no default, and nothing in the binary privileges `model_configs/`), but left three problems:

1. **The directory still exists** (13 stale configs), able to drift from the ~23-model authoritative catalog.
2. **The docs systematically teach it as the catalog** — they instruct `export BLIS_CATALOG=$PWD/model_configs` and `--catalog model_configs` throughout, pointing users at the wrong source of truth.
3. **The doc-example inventory is under-counted, and the guard-test claim was wrong** (both corrected below).

No production code depends on `model_configs/` — grep of `cmd/`+`sim/` finds it only in comments. It survives solely as test fixtures and the thing the docs point at.

**Corrections from [#1767 comment §3](https://github.com/inference-sim/inference-sim/issues/1767#issuecomment-5720445880):**

- **The guard test does NOT lock in `$PWD/model_configs`.** [`TestDocExamplesDocumentTheCatalogEnvVar`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/cmd/docs_examples_test.go#L172) requires that the docs document **both** `--catalog` and `BLIS_CATALOG` **and** carry an `export BLIS_CATALOG=` instruction in quickstart — it does not pin the exact value. So repointing the docs to a `blis-catalog` clone likely passes it **as-is**; touch the test only if the export-instruction *shape* changes. (The earlier "a guard test locks the mistake in" framing was wrong.)
- **Corrected doc-example counts** (same multiline matcher as `cmd/docs_examples_test.go`):

  | Scope | `run`/`replay` examples | Omit `--catalog` | Omit `--hardware`/`--tp` |
  |---|---:|---:|---:|
  | All live documentation roots | 127 | 122 | 0 |
  | `docs/guide` + `docs/reference` | 44 | 40 | 0 |

  The remaining usability question is **catalog context, not deployment flags** (0 examples omit `--hardware`/`--tp`).

**§3 — this issue owns the catalog-locator documentation prerequisite policy.** Beyond replacing literal `model_configs` strings, decide and apply ONE policy for the 122 examples that omit `--catalog`: (a) a per-page prerequisite block, (b) make examples self-contained (`--catalog` on each), or (c) explicitly accept an entry-point-only export policy (quickstart/index set `BLIS_CATALOG` once; inner pages assume it). Document the choice.

- [x] Determinism (INV-6) — the equivalence gate ([`sim/latency/catalog_equivalence_test.go`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/sim/latency/catalog_equivalence_test.go#L34)) keeps stdout byte-identical across the fixture→catalog swap; its replacement (catalog-keyed golden `*ModelConfig`) must preserve that.
- [x] None otherwise (deleting a stale duplicate + migrating fixtures/docs).

**Proposed fix** (one coordinated change — do NOT delete the dir before the fixtures/docs move)

*Code / tests:*
1. Rewrite [`TestCatalogConfigEquivalence`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/sim/latency/catalog_equivalence_test.go#L34) to a golden `*ModelConfig` keyed directly off the catalog config — its own comment already prescribes this ("once S6 lands and the fixtures are gone, this test is replaced by a golden `*ModelConfig` keyed directly off the catalog config").
2. Repoint the fixture-consuming test files off `../model_configs` — `cmd/catalog_cli_test.go`, `cmd/dp_placement_test.go`, `cmd/dp_replay_parity_test.go`, `cmd/ep_mode_cli_test.go`, `cmd/lora_noop_test.go`, `cmd/saturation_cli_parity_test.go`, `cmd/saturation_stdout_parity_test.go`, `sim/latency/dp_ep_config_test.go`, `sim/latency/glm_model_shape_test.go` — to committed `testdata/` mini-catalogs or a checked-out `blis-catalog` in CI (decide in the design note).
3. **Expanded inventory (from [comment §3](https://github.com/inference-sim/inference-sim/issues/1767#issuecomment-5720445880)) — the indirect / non-obvious dependents:**
   - **Golden `ModelConfigDir` deps** (indirect — the golden dataset JSON names the dir): [`sim/cluster/roofline_golden_test.go#L123`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/sim/cluster/roofline_golden_test.go#L123) and [`sim/cluster/trained_physics_golden_test.go#L118-L126`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/sim/cluster/trained_physics_golden_test.go#L118-L126).
   - [`scripts/find-saturation.sh#L24`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/scripts/find-saturation.sh#L24) (`CATALOG="${CATALOG:-model_configs}"`), [`scripts/README.md#L127-L159`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/scripts/README.md#L127-L159), [`.gitignore#L48`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/.gitignore#L48), [`CONTRIBUTING.md#L265`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/CONTRIBUTING.md#L265), [`docs/reference/project-structure.md#L102`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/docs/reference/project-structure.md#L102), and any `testdata` references.
   - **Not a bundled-tree consumer:** [`cmd/catalog_provenance_test.go#L145`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/cmd/catalog_provenance_test.go#L145) merely creates a **temp directory named** `model_configs`; it does not read the committed tree. Leave it (or rename for clarity) — it is not a deletion blocker.
4. Delete `model_configs/`.

*Docs — repoint EVERY `model_configs` reference to `blis-catalog` (clone + `BLIS_CATALOG=<clone-root>` per #1774; the whole roofline.md fetch narrative included):*
- `docs/index.md`, `docs/getting-started/quickstart.md`, `docs/getting-started/installation.md`, `docs/getting-started/tutorial.md`, `docs/guide/latency-models.md`, `docs/guide/observe-replay-calibrate.md`, [`docs/concepts/roofline.md#L3`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/docs/concepts/roofline.md#L3) **and** the stale "previously fetched" narrative at [`#L70-L79`](https://github.com/inference-sim/inference-sim/blob/72876bff34903351a5fdf8965c19f9f02841ca83/docs/concepts/roofline.md#L70-L79), `docs/reference/models.md`, `docs/reference/configuration.md`
- `README.md`
- `CLAUDE.md` (the `export BLIS_CATALOG=$PWD/model_configs` decision + the `model_configs/` mentions)
- Confirm `TestDocExamplesDocumentTheCatalogEnvVar` still passes after repointing (rewrite only if the export-instruction shape changes).

**Files affected**
- `model_configs/` (delete)
- `sim/latency/catalog_equivalence_test.go` (rewrite to catalog-keyed golden), the fixture-consuming test files, the two golden datasets' `ModelConfigDir`, `scripts/find-saturation.sh`, `scripts/README.md`, `.gitignore`, `CONTRIBUTING.md`, `docs/reference/project-structure.md`
- all doc/README/CLAUDE.md sites listed above

**Notes / coordination**
- `docs/reference/configuration.md` and `docs/reference/models.md` are also edited by #1775 (content falsehoods, different lines) — land #1775 first or rebase (`⇄` #1775).
- `docs/concepts/roofline.md` is owned **entirely here** (both the `model_configs/` path and the fetch narrative) — #1775 explicitly excludes it.

**Design document**
Note needed: (a) test-fixture strategy — committed `testdata/` mini-catalog vs. clone `blis-catalog` in CI; (b) how getting-started obtains the catalog (git clone step, submodule, or CI checkout); (c) the §3 prerequisite policy; (d) confirm docs land in the SAME PR as the deletion so the quickstart never references a deleted path.

namasl:
Closes #1771. Part of #1767 (R1 cleanup). Depends on #1774 (merged).

## What

The in-repo `model_configs/` tree was a **stale duplicate** of the authoritative
[`blis-catalog`](https://github.com/inference-sim/blis-catalog) repository — documented as
"the catalog," able to drift, and the thing every doc pointed users at. NS-6/S4
(#1733/#1731) already made the catalog an explicitly-located input with nothing in the
binary privileging `model_configs/`. This PR:

1. **Deletes `model_configs/`** and the #1774 flat-layout **transition fallback** that kept
   it resolving. `cmd.catalogModelDirs` now returns a single canonical candidate, so there is
   exactly ONE layout: `<catalog>/models/<short-name>/config.json`.
2. **Migrates the test fixtures** the tree provided into a committed, clone-root-shaped
   fixture `testdata/catalog/` (`models/<name>/config.json` + `workloads/*.yaml`).
3. **Repoints every doc/README/CONTRIBUTING/scripts/CLAUDE.md reference** to a `blis-catalog`
   clone.

This is a **deletion + fixture/doc migration only — no simulation number changes**
(INV-6/INV-13).

## Migration decisions (issue §3 + design note)

- **Test-fixture strategy (a):** committed `testdata/catalog/` mini-catalog (clone-root
  shaped, mirroring blis-catalog) rather than a CI checkout of blis-catalog, so
  `go test ./...` resolves models and presets **offline** with no network dependency. The
  config.json files are **byte-identical git renames** of the deleted tree (git shows pure
  `R`), so parsing — and every downstream number — is unchanged. `codellama-34b-instruct-hf`
  was dropped (no test read its config; it appears only as a string in a path-derivation
  unit test). `.gitignore` whitelists `testdata/catalog/models/**/config.json` in place of
  the retired `model_configs/*/config.json`.
- **Doc-locator policy (c), unchanged:** entry points state the catalog location **once**
  and inner-page examples keep omitting `--catalog`. The one-time instruction now
  `git clone`s blis-catalog and `export BLIS_CATALOG=$PWD/blis-catalog` instead of pointing
  at the deleted `model_configs/`. `TestDocExamplesDocumentTheCatalogEnvVar` (which pins that
  entry points name both forms + quickstart shows the export, not the exact value) and
  `TestDocExamplesPassDeploymentFlags` both still pass.
- **`docs/concepts/roofline.md`** (owned entirely here) rewritten — both the `model_configs/`
  path and the stale "previously fetched into `model_configs/`" narrative.
- **Equivalence gate rewrite:** `sim/latency/catalog_equivalence_test.go` becomes a golden
  `*sim.ModelConfig` keyed off the vendor-verbatim catalog config in `testdata/catalog_configs/`
  (exactly what its own predecessor comment prescribed once the fixtures were gone). The
  golden values ARE what the retired `model_configs/` fixtures parsed to.
- **Flat-fallback tests:** `cmd/catalog_layout_test.go`'s `_ByteIdenticalToFlat` tests (the
  flat-vs-clone equivalence law is moot with one layout) become `_ResolvesAndRuns`
  determinism tests; `cmd/hfconfig_test.go` flat-fallback unit tests are removed or rewritten
  (`_NotBypassedByFlatFallback` → `_Refused`, `CanonicalLayoutIsFirst` →
  `OnlyCanonicalCandidate`). Shared temp-catalog helpers and the two golden datasets'
  `model_config_dir` fields point at the `models/` layout.

## Behavior change on the record

The pre-#1774 invocation `--catalog <clone-root>/models` no longer resolves (the fallback
that masked it is gone) — pass the clone root. An uncatalogued model's refusal now names a
single canonical path instead of both a `models/` candidate and a flat fallback.

## INV-6 / INV-13 evidence

- Relocated `config.json` files are **byte-identical git renames** (no content diff).
- `TestNoOpByteIdentity_AdapterBlindRunMatchesBaseline` still matches the pre-#1733 golden
  `specs/007-lora-control-plane/testdata/baseline_noop.json` (untouched).
- `TestCatalogConfigEquivalence`, the sim/cluster roofline + trained-physics golden suites,
  and the run/replay clone-root determinism tests all pass.
- `go build ./...` and `go test ./...` are green.


namasl:
## blis-pr-review — verdict: GREEN (ready to merge)

Reviewed as a **deletion + fixture/doc migration** (issue #1771, R1 cleanup). The core invariant is INV-6/INV-13: no simulation number may change. Verified across the 10 code-review perspectives, cross-path parity, and a fresh-eyes independent audit of the diff.

**Findings**

1. **[FIXED] Stale comment** — `cmd/catalog_workloads.go:76` still claimed `catalogModelDirs` carries a flat fallback (this PR removed it). Reworded to reflect the single clone-root layout (commit `0128f969`). Comment-only, no runtime effect.

All other items clean:

- **INV-6 (byte-identity):** the 12 relocated `testdata/catalog/models/*/config.json` are pure git renames (`R100`) of the deleted tree — byte-identical, so parsing and every downstream number are unchanged. `TestNoOpByteIdentity_AdapterBlindRunMatchesBaseline` still matches the pre-#1733 golden (untouched). `codellama` (dropped) is read by no active code/golden.
- **Single-layout resolution:** `catalogModelDirs` returns one canonical candidate; `readCatalogEntry`'s loop + errno classification are preserved and correct. Uncatalogued refusal names the one canonical path.
- **Cross-path parity (run/replay/observe):** N/A — catalog resolution is shared (`resolveModelConfig`), no per-path behavior added. `observe` resolves no model config.
- **Preemption / timeout / DES:** untouched by this PR.
- **Tests:** `go build ./...` + `go test ./...` green. Doc guards `TestDocExamplesPassDeploymentFlags` and `TestDocExamplesDocumentTheCatalogEnvVar` both pass. Flat-fallback unit tests correctly removed/rewritten; `catalog_equivalence_test.go` rewritten to a catalog-keyed golden `*sim.ModelConfig` (golden values verified consistent with the vendor-verbatim configs).
- **`.gitignore`:** relocated configs whitelisted (`testdata/catalog/models/**/config.json`) and tracked.
- **Golden datasets:** `model_config_dir` fields repointed to `testdata/catalog/models/<name>`; all referenced dirs exist; sim/cluster golden suites pass.
- **Docs:** entry points repointed to a `blis-catalog` clone (policy c preserved); `roofline.md` fetch narrative rewritten. Remaining `model_configs` mentions are append-only change-history blocks (the repo's documented history pattern), not active guidance.

**Behavior change on the record (intended, per issue):** `--catalog <clone-root>/models` no longer resolves — pass the clone root.

No blocking issues remain.

namasl:

## qa-review — PR #1797: ⛔ BLOCK
_hardening: delete the stale model_configs/ tree [R1 cleanup] · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ❌ FLAW_FOUND | Does this PR fully implement the issue it closes, without unnecessary additions? | The production/test migration is implemented and the relevant Go test suites pass, but the issue is not fully closed because CLAUDE.md still contains an obsolete paragraph that presents the deleted bundled `model_configs/` tree as a valid catalog and instructs users to export it. The later #1771 paragraph contradicts rather than removes that stale guidance. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. Most user-facing documentation was repointed to the external clone-root layout, but CLAUDE.md retains actionable stale documentation for the deleted tree. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No materially stale production-code comment was found. References to `model_configs/` in `cmd/hfconfig.go`, `cmd/catalog_layout_test.go`, and `sim/latency/catalog_equivalence_test.go` describe retired behavior historically, while current behavior is described as the sole `models/` layout. |
| G1 | correctness | ✅ CONFIDENT | In `cmd/hfconfig.go`, `readCatalogEntry` now checks only the canonical candidate path `<catalog>/models/<short-name>/config.json`. If `<catalog>/models` exists as a file instead of a directory, does `os.Stat` or `os.ReadFile` return an error that is correctly reported, or could it be misclassified as missing entry absence? | If `<catalog>/models` is a regular file, `os.Stat(<catalog>/models/<name>/config.json)` returns ENOTDIR. The implementation deliberately classifies ENOTDIR as absence and returns the normal uncatalogued-model refusal, including the exact canonical path. It therefore does not report the `models` node itself as a filesystem error. |
| G2 | correctness | ✅ CONFIDENT | In `cmd/catalog_provenance_cli_test.go`, `dirtyCatalogEntry` was updated to use `testCatalogConfigPath(catalogRoot, shortName)`. If a catalog root path contains trailing slashes or relative directory components, does `captureCatalogProvenance`'s `git status --porcelain -- .` correctly filter git status for files nested inside `models/`? | Yes. Git is invoked with `git -C <catalogRoot>`, so the pathspec `.` is evaluated from that catalog directory and includes nested `models/` files. Git and the OS accept relative roots, `..` components, and trailing separators without changing that scope. |
| G3 | invariants | ✅ CONFIDENT | With `sim/latency/catalog_equivalence_test.go` (`TestCatalogConfigEquivalence`) rewritten to validate against hardcoded golden `*sim.ModelConfig` structs instead of directly comparing two physical catalog locations, how do we enforce that changes in `blis-catalog` will not quietly diverge from these hardcoded test goldens (INV-6)? | The test enforces INV-6 against committed snapshots of the relevant vendor catalog files: each snapshot is parsed by the real loader and compared with the pre-deletion `ModelConfig` golden. It does not automatically monitor future commits in the separate `blis-catalog` repository; updating those snapshots is a human-coordinated cross-repository step, and any changed parsed shape then fails the golden comparison. |
| G4 | invariants | ✅ CONFIDENT | In `cmd/catalog_layout_test.go`, `TestReplayCmd_CatalogCloneRootLayout_ResolvesAndRuns` verifies that replay succeeds using the new clone root layout. Does this test verify that replay output remains byte-identical to trace runs generated before `model_configs/` was removed (INV-13)? | No, that specific test does not compare against a trace/run produced before deletion. It exports a trace using the new clone-root fixture and compares two replay executions using that same fixture. Pre-change numerical compatibility is covered separately by retained goldens and the model-config equivalence gate. |
| G5 | rules | ✅ CONFIDENT | Rule R1 states that BLIS must not silently swallow errors or continue on broken inputs. In `cmd/hfconfig.go`'s `readCatalogEntry`, if `config.json` is present in `<catalog>/models/<short-name>/` but fails due to permission denied (EACCES), does the error bubble up with the exact path rather than getting swallowed into an uncatalogued model refusal? | Yes. EACCES is neither `IsNotExist` nor ENOTDIR, so stat failures immediately return `cannot be checked` and read failures return `exists but cannot be read`; both include the exact `config.json` path. |
| G6 | rules | ✅ CONFIDENT | Rule R23 requires canonical helpers for repeated logic. Now that the flat transition fallback is deleted, are there any remaining places in `cmd/` or `sim/` where model config paths are manually constructed with `filepath.Join` instead of calling `resolveModelConfig` or `catalogModelDirs`? | No duplicate production model-catalog path constructor was found. `catalogModelDirs` is the production constructor, and `readCatalogEntry` appends `config.json`. Other joins in `cmd/` are test fixture creation/assertions; the cluster golden tests intentionally consume explicit `ModelConfigDir` dataset paths rather than performing CLI catalog resolution. |
| G7 | parity | ✅ CONFIDENT | Does `registerCatalogFlag` in `cmd/hfconfig.go` register identical flag definitions and help text across `blis run`, `blis replay`, `blis observe`, and `blis convert preset`, ensuring consistent catalog CLI flags across all subcommands? | Yes. All four commands use the same `registerCatalogFlag` helper, so the name, binding, default, and help text are identical. |
| G8 | parity | ⚠️ CANNOT_ANSWER | For tests like `sim/cluster/roofline_golden_test.go` that reference golden dataset JSON files containing `ModelConfigDir`, do `blis run` and `blis replay` observe identical model shape parameters when pointed at `testdata/catalog` versus an external `blis-catalog` clone? | Run and replay share the same resolver and therefore observe identical shape parameters when given identical `config.json` bytes. However, this worktree contains no external `blis-catalog` checkout or revision pin covering every golden-dataset model, so equality between all committed mini-catalog entries and the current external catalog cannot be verified here. |
| G9 | tests | ✅ CONFIDENT | In `cmd/catalog_layout_test.go`, `TestRunCmd_CatalogCloneRootLayout_UncataloguedModelStillRefused` checks that missing models are refused. Is there an invariant unit test verifying that a malformed `config.json` inside `<catalog>/models/<short-name>/` fails with a parse error naming the file rather than returning an 'uncatalogued model' error? | Yes. A unit test writes a malformed canonical entry, asserts resolution fails, and asserts the error names that file. The production error identifies it as an invalid HuggingFace `config.json`, not as an uncatalogued model. |
| G10 | tests | ✅ CONFIDENT | The PR migrates test fixtures from `model_configs/` to `testdata/catalog/`. If `testdata/catalog/` is missing a expected model config during a test run, does the test fail fast with an explicit catalog assertion message or an opaque simulation failure? | Yes. CLI-driven tests receive the resolver's explicit uncatalogued-model error naming the expected canonical path, while tests that read fixtures directly fail at setup with the attempted path. Golden dataset tests similarly report `GetModelConfig(<path>)`. |
| G11 | rationale | ✅ CONFIDENT | What were the design trade-offs considered between using a committed `testdata/catalog/` mini-catalog fixture vs shallow-cloning `blis-catalog` in CI, and why was the committed `testdata/` fixture chosen? | The committed mini-catalog was chosen to keep `go test ./...` deterministic, offline, and independent of network availability or movement in another repository. A CI clone would provide automatic freshness against `blis-catalog`, but would introduce network/cross-repository availability and revision-control concerns. The fixture preserves the real clone-root shape while limiting data to models and workloads actually consumed by tests. |
| G12 | rationale | ✅ CONFIDENT | Regarding prerequisite policy choice for the 122 doc examples that omit `--catalog`, why was option (c) — setting `BLIS_CATALOG` once in quickstart/CLAUDE.md/README — chosen over explicit `--catalog` flags on every single example command? | Option (c) was chosen because `BLIS_CATALOG` is explicitly the reusable environment form of the locator, whereas required deployment inputs have no environment equivalent. Setting it once keeps the large body of examples readable while still making them runnable from documented entry points, and a guard test prevents that prerequisite from disappearing. |
| G13 | scope | ✅ CONFIDENT | The PR diff updates comments and logic in `cmd/catalog_workloads.go` and `cmd/catalog_provenance.go`. Does this PR alter any workload preset or provenance emission behavior beyond updating catalog paths and removing flat-layout fallbacks? | No workload preset or provenance emission behavior is changed by these edits. Workload resolution remains `<catalog>/workloads/<name>.yaml`; only comments and fixture locations changed. Provenance still runs Git from the supplied catalog root, scopes dirty status to `.`, degrades to unknown on error, and is emitted through the same shared run/replay option helper. |
| G14 | scope | ✅ CONFIDENT | During the migration from `model_configs/` to `testdata/catalog/`, `codellama-34b-instruct-hf` was omitted. Were any other model configurations, parameters, or test workloads modified or dropped as part of this PR? | The resulting mini-catalog contains the other 12 model entries expected after dropping the unused CodeLlama entry, and the four workload presets remain present. No simulation parameters are intentionally changed: three fuller vendor configurations are guarded by parsed `ModelConfig` goldens, and preset values are compared field-for-field with the retired defaults. |
| G15 | errors | ✅ CONFIDENT | When `resolveCatalogRoot` in `cmd/hfconfig.go` fails because neither `--catalog` nor `BLIS_CATALOG` is provided, does it fail via `logrus.Fatalf` in the CLI layer while leaving core library packages free of process-exiting logic? | Yes. `resolveCatalogRoot` and `resolveModelConfig` return ordinary errors. The shared CLI resolution layer converts those errors to `logrus.Fatalf`; the simulator/library packages are not involved in process termination. |
| G16 | errors | ✅ CONFIDENT | In `cmd/catalog_provenance.go`, if `git` is not installed or the provided catalog root is not a git repository, does `captureCatalogProvenance` log a warning and return `sim.UnknownCatalogRevision` without failing the overall execution of `blis run` or `blis replay`? | Yes. Failure to execute Git or a non-repository error is returned by `runCatalogGit`, caught by `captureCatalogProvenance`, logged as a warning, and represented with `sim.UnknownCatalogRevision` and `Dirty=false`; no error is returned to abort run or replay. |

### Items to fix
- **F1 · FLAW_FOUND** — The production/test migration is implemented and the relevant Go test suites pass, but the issue is not fully closed because CLAUDE.md still contains an obsolete paragraph that presents the deleted bundled `model_configs/` tree as a valid catalog and instructs users to export it. The later #1771 paragraph contradicts rather than removes that stale guidance.
  _CLAUDE.md:738 says `The bundled model_configs/ tree is a valid catalog` and recommends `export BLIS_CATALOG=$PWD/model_configs`; CLAUDE.md:742 later says that tree is gone. `go test ./cmd`, `go test ./sim/latency`, and `go test ./sim/cluster` all pass._
- **F2 · FLAW_FOUND** — No. Most user-facing documentation was repointed to the external clone-root layout, but CLAUDE.md retains actionable stale documentation for the deleted tree.
  _CLAUDE.md:738 still says `--catalog` names a root with one directory per model, gives the migration example `--catalog model_configs`, calls the bundled tree valid, and recommends `export BLIS_CATALOG=$PWD/model_configs`. This conflicts with the sole `<catalog>/models/<short-name>/config.json` layout documented at CLAUDE.md:742 and cmd/hfconfig.go:383-433._
- **G8 · CANNOT_ANSWER** — Run and replay share the same resolver and therefore observe identical shape parameters when given identical `config.json` bytes. However, this worktree contains no external `blis-catalog` checkout or revision pin covering every golden-dataset model, so equality between all committed mini-catalog entries and the current external catalog cannot be verified here.
  _cmd/root.go:942-978 calls `resolveModelConfig` for both analytical backends through the shared `resolveLatencyConfig`; testdata/roofline_goldendataset.json and testdata/trained_physics_iter29.json point to `testdata/catalog/models/...`. sim/latency/catalog_equivalence_test.go:42-74 compares only three committed catalog snapshots with hardcoded shapes._

### Important to consider
- **F3** — CLAUDE.md:738 is stale documentation, but it is not a code comment.
  _cmd/hfconfig.go:178-205 and cmd/hfconfig.go:383-433 describe the one-candidate layout; cmd/catalog_layout_test.go:140-168 describes the post-removal tests; sim/latency/catalog_equivalence_test.go:24-31 explicitly describes the deleted fixture historically._
- **G1** — A dedicated “models is not a directory” diagnostic would be clearer, but this is defensive handling of a malformed hand-created catalog and still fails rather than continuing.
  _cmd/hfconfig.go:214-228 treats `os.IsNotExist` or `syscall.ENOTDIR` as absence; cmd/hfconfig.go:254-262 returns the uncatalogued refusal with the canonical path. cmd/hfconfig_test.go:257-275 explicitly tests a file named `models` and expects that disposition._
- **G3** — There is no CI checkout or revision pin that automatically detects upstream `blis-catalog` drift.
  _sim/latency/catalog_equivalence_test.go:12-31 documents the snapshot/golden gate; sim/latency/catalog_equivalence_test.go:42-74 loads `sim/latency/testdata/catalog_configs/<name>.json` and performs `reflect.DeepEqual`. The committed snapshots are present under sim/latency/testdata/catalog_configs/._
- **G4** — A direct old-trace/new-catalog combination would add coverage, but the shared replay resolver and existing migration goldens cover the core behavior.
  _cmd/catalog_layout_test.go:166-190 creates the catalog, exports the trace through it, then compares two replay runs through the same catalog. sim/latency/catalog_equivalence_test.go:24-31 preserves parsed pre-deletion model shapes, and cmd/lora_noop_test.go:57-73 compares a migrated-catalog run with a pre-feature stdout golden._
- **G11** — Freshness against the external catalog remains human-coordinated.
  _CLAUDE.md:742 records option (a), stating that `testdata/catalog/` lets tests resolve models and presets offline with no network or catalog checkout. testdata/catalog contains `models/` and `workloads/`, matching cmd/hfconfig.go:383-433 and cmd/catalog_workloads.go's clone-root contracts._

namasl:
## Author response to qa-review BLOCK (head `457a4cf8`)

**F1 / F2 (FLAW_FOUND — stale CLAUDE.md guidance) — FIXED in `457a4cf8`.**
The finding is correct: the `#1731` change-history paragraph (CLAUDE.md:738) still read
`The bundled model_configs/ tree is a valid catalog` and showed
`export BLIS_CATALOG=$PWD/model_configs` as if it were current guidance, contradicting the
`#1771` entry below it. I annotated both claims inline — the tree "was a valid catalog at the
time (**#1771 has since DELETED it** …)" and the export instruction now carries
"(**#1771 repointed that instruction to a `blis-catalog` clone: … `export BLIS_CATALOG=$PWD/blis-catalog`**)".
This keeps the append-only history record intact (the repo's convention — e.g. `#1733` annotating
`#435`'s removed auto-fetch) while removing the actionable-but-wrong reading. Verified: no
un-annotated `export BLIS_CATALOG=$PWD/model_configs` remains, and
`TestDocExamplesDocumentTheCatalogEnvVar` + `TestDocExamplesPassDeploymentFlags` still pass.

**G8 (CANNOT_ANSWER — can't verify the committed mini-catalog equals the *external* blis-catalog) — out of scope for this PR; covered where it belongs.**
G8 asks to verify equality between `testdata/catalog/` and the *current external* `blis-catalog`
repo, which is inherently unverifiable in this worktree (no checkout, by design). It is not a flaw
in this PR:
- This PR's INV-6 obligation is byte-identity with the tree it **replaces**, not with the external
  repo. Guaranteed structurally: `git diff --find-renames` shows all 12 relocated `config.json`
  as pure `R100` renames (zero content change) — so parsing and every downstream number are
  identical to pre-PR, and `TestNoOpByteIdentity_AdapterBlindRunMatchesBaseline` still matches the
  pre-#1733 golden.
- Shape-equivalence between the vendor-verbatim configs and what BLIS parses is exactly what
  `sim/latency/catalog_equivalence_test.go` (`TestCatalogConfigEquivalence`) guards, for the three
  shape-critical models; a divergent parse fails that golden.
- Keeping the committed fixture in step with the external catalog's full model set is cross-repo
  prerequisite **C3/C4** under tracker #1767 — explicitly human-coordinated, not something a
  deletion PR verifies. The answerer's own G3/G11 notes acknowledge this is a human-coordinated
  cross-repo step; it was labeled `CANNOT_ANSWER`, not `FLAW_FOUND`.

No code change is warranted for G8. Requesting adjudication.

namasl:

## qa-review adjudication — PR #1797: ✅ PASS

|----|----|----|----|
| F1 | FLAW_FOUND | ✅ RESOLVED | The obsolete guidance is now explicitly historical and immediately states that #1771 deleted the bundled tree and that users must use an external blis-catalog clone. It no longer presents the deleted path as current actionable guidance. |
| F2 | FLAW_FOUND | ✅ RESOLVED | CLAUDE.md now distinguishes the historical S4 behavior from the current catalog location and layout, removing the actionable contradiction identified by the finding. |
| G8 | CANNOT_ANSWER | 🟡 WAIVED (justification) | The requested comparison against the current contents of a separate external checkout is not the migration invariant assigned to this deletion PR. The independently checked tracker defines #1771 as deleting the stale bundled tree and migrating its fixtures/equivalence gate, while its binding INV-6 constraint is byte identity against the pre-cutover simulator behavior. It does not require this worktree to validate every fixture against an unpinned current external repository. |

### Still blocking

### Cleared
- **F1** (✅ RESOLVED) — The obsolete guidance is now explicitly historical and immediately states that #1771 deleted the bundled tree and that users must use an external blis-catalog clone. It no longer presents the deleted path as current actionable guidance.
  _CLAUDE.md:738 says the bundled model_configs/ tree "was a valid catalog at the time," then explicitly says #1771 deleted it and directs users to point --catalog/BLIS_CATALOG at a blis-catalog clone. The current tree contains no unannotated `export BLIS_CATALOG=$PWD/model_configs` or `bundled model_configs/ tree is a valid catalog` match._
- **F2** (✅ RESOLVED) — CLAUDE.md now distinguishes the historical S4 behavior from the current catalog location and layout, removing the actionable contradiction identified by the finding.
  _CLAUDE.md:738 marks the old model_configs guidance as historical, states that #1771 deleted that tree, and directs users to the external clone root. CLAUDE.md:742 documents the current sole `<catalog>/models/<short-name>/config.json` layout. The document header also shows cloning blis-catalog and exporting `BLIS_CATALOG=$PWD/blis-catalog`._
- **G8** (🟡 WAIVED (justification)) — The requested comparison against the current contents of a separate external checkout is not the migration invariant assigned to this deletion PR. The independently checked tracker defines #1771 as deleting the stale bundled tree and migrating its fixtures/equivalence gate, while its binding INV-6 constraint is byte identity against the pre-cutover simulator behavior. It does not require this worktree to validate every fixture against an unpinned current external repository.
  _GitHub issue #1767 identifies blis-catalog as the authoritative source, scopes #1771 to deleting model_configs and migrating the equivalence gate, fixtures, and docs, and states that each cutover must be byte-identity testable against main. sim/latency/catalog_equivalence_test.go:24-31 explicitly implements the post-deletion gate as parsed ModelConfig goldens captured from the retired fixtures; lines 42-74 verify the three shape-critical vendor configs. testdata/roofline_goldendataset.json and testdata/trained_physics_iter29.json now consistently reference testdata/catalog/models entries, and `go test ./sim/latency` passes._

namasl:

## qa-review — PR #1797: ⛔ BLOCK
_hardening: delete the stale model_configs/ tree [R1 cleanup] · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | The substantive issue is implemented: the in-repo model_configs directory is gone, the flat-layout fallback is removed, fixtures use a clone-root-shaped test catalog, affected scripts/tests/golden datasets were migrated, and the documented policy is entry-point-only BLIS_CATALOG export. The production change is limited to removing the transitional candidate; the remaining additions are tests, fixtures, and documentation needed for that migration. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. Most live documentation was migrated, but CLAUDE.md still describes the current resolver as reading `<catalog>/<short-name>/config.json`. The implemented and otherwise documented sole layout is `<catalog>/models/<short-name>/config.json`, so this is a concrete misleading instruction in maintained project documentation. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | The changed production comments generally describe the new single-layout behavior correctly. One minor stale plural remains: resolveModelConfigInCatalog says an entry can be “Absent from every layout” even though catalogModelDirs now returns exactly one layout. Other model_configs mentions in code comments are explicitly historical explanations, and the temp directory named model_configs in the provenance test is intentional. |
| G1 | correctness | ✅ CONFIDENT | In `cmd/catalog_layout_test.go`, `TestRunCmd_CatalogCloneRootLayout_UncataloguedModelStillRefused` checks refusal for `test-org/not-catalogued`. How does `resolveModelConfigInCatalog` behave if `--catalog` points to a valid directory that lacks a `models/` subdirectory entirely, versus when `models/` exists but the specific model entry is missing? | Both cases have the same disposition. The resolver derives `<catalog>/models/<short-name>/config.json`; if models/ is absent, os.Stat returns ENOENT, and if models/ exists but that model/config is absent, os.Stat also returns ENOENT. Both are treated as absence and produce the same not-in-catalog refusal naming the canonical path. Neither case creates files. A regular file named models yields ENOTDIR and is handled as absence too. |
| G2 | correctness | ✅ CONFIDENT | The PR notes that `codellama-34b-instruct-hf` was dropped during fixture migration to `testdata/catalog/models/`. Are all other 12 models from `model_configs/` preserved in `testdata/catalog/models/`, and are any test assertions broken if a test attempts to reference a model not included in `testdata/catalog/`? | Yes. testdata/catalog/models contains the other 12 migrated model directories, with only codellama-34b-instruct-hf omitted. No test attempts to resolve CodeLlama through that catalog: its remaining hfconfig_test occurrence only checks pure path derivation, while the old general golden dataset names CodeLlama but constructs its simulator configuration independently rather than loading this catalog. Any CLI test that really requested an absent fixture model would now fail explicitly as uncatalogued rather than silently falling back. |
| G3 | invariants | ✅ CONFIDENT | INV-6 requires determinism and stdout byte-identity. In `sim/latency/catalog_equivalence_test.go`, `TestCatalogConfigEquivalence` was rewritten to test against a golden `*sim.ModelConfig` instead of comparing `model_configs/` against `blis-catalog`. How is the golden `*sim.ModelConfig` verified to maintain byte-identical latency outputs against historical baselines when model parameters are loaded? | The rewritten equivalence test runs the real latency.GetModelConfig parser over three vendor-verbatim catalog snapshots and DeepEqual-compares every parsed ModelConfig field with values captured from the retired fixtures. Since the latency models consume that parsed shape, exact structural equality prevents a fixture move from changing their numeric inputs. Separate migrated CLI and cluster golden tests then exercise actual simulations against unchanged historical baselines. |
| G4 | invariants | ✅ CONFIDENT | INV-13 requires `run` and `replay` parity. With the flat transition fallback removed from `cmd.catalogModelDirs`, both commands resolve models strictly via `<catalog>/models/<short-name>`. If a user previously passed `--catalog path/to/models` (pointing directly at the `models` folder), how do `run` and `replay` fail, and do both commands output identical refusal diagnostics? | Both commands treat path/to/models as the clone root and therefore look for `path/to/models/models/<short-name>/config.json`. When that path is absent, both abort with the same resolver-generated not-in-catalog message naming that doubled models path and instructing the user to pass the clone root. Run and replay register the same catalog flag and call the same resolveLatencyConfig/resolveModelConfig path, so the catalog-refusal text is identical for equivalent inputs. |
| G5 | rules | ✅ CONFIDENT | Rule R1 requires explicit refusals and no silent fallbacks. In `cmd/catalog_layout_test.go` and `cmd/hfconfig.go`, now that `catalogModelDirs` returns only the single canonical candidate `<catalog>/models/<short-name>`, how does `readCatalogEntry` distinguish between a missing `config.json` file and a malformed/unreadable `config.json` inside the canonical path? | readCatalogEntry stats the canonical config path first. ENOENT, ENOTDIR, or a directory at config.json is classified as absence and ends in the actionable uncatalogued-model refusal. Other stat errors are returned as “cannot be checked,” and non-absence read errors are returned as “exists but cannot be read,” both naming the path and preserving the wrapped errno. Readable malformed content reaches resolveModelConfigInCatalog, which rejects it as not a HuggingFace config and names the file rather than treating it as missing. |
| G6 | rules | ✅ CONFIDENT | Rule R23 requires single-source-of-truth configuration paths. The committed test catalog `testdata/catalog/` mirrors the clone root layout with `models/`, `workloads/`, and `devices/`. Are there any places in `cmd/` or `sim/` where test helpers bypass this clone root and read directly from subdirectories, violating canonical resolution? | Production catalog consumers do not bypass the clone root: model resolution, preset loading, and device-class loading all derive their namespace from the same resolved root. Some tests intentionally read fixture files directly—for example, copying a model config into a temporary clone-root catalog or loading a golden dataset's explicit ModelConfigDir—but those are fixture setup or lower-level parser tests, not competing production configuration paths. |
| G7 | parity | ✅ CONFIDENT | The issue notes that 122 documentation examples across `docs/guide` and `docs/reference` omit `--catalog` and rely on `BLIS_CATALOG`. Do `blis run`, `blis replay`, `blis observe`, and `blis convert preset` handle relative versus absolute paths in `BLIS_CATALOG` identically across all four CLI entry points? | Yes, whenever those commands consume the catalog. All four register the same --catalog binding, and production reads BLIS_CATALOG only through resolveCatalogRoot. That function preserves the supplied path verbatim; relative paths are consequently resolved by filesystem operations against the process working directory, while absolute paths remain absolute. Run/replay use it for model resolution, and run/observe/convert preset use it through the shared preset loader; replay does not load presets. |
| G8 | parity | ✅ CONFIDENT | In `CLAUDE.md`, `README.md`, and `docs/getting-started/quickstart.md`, the documentation instructs cloning `blis-catalog` and setting `export BLIS_CATALOG=$PWD/blis-catalog`. Does `cmd/docs_examples_test.go` (`TestDocExamplesDocumentTheCatalogEnvVar`) verify that every documentation page featuring `run`/`replay` examples includes or references this export instruction? | No. The guard intentionally enforces policy (c), not a per-page rule. It checks six designated entry-point documents for mentions of both BLIS_CATALOG and --catalog, then checks only quickstart.md for an actual `export BLIS_CATALOG=` instruction. It does not walk every page containing run/replay examples or verify that each such page links back to the instruction, and it does not pin the export value to blis-catalog. |
| G9 | tests | ❌ FLAW_FOUND | In `cmd/catalog_layout_test.go`, the tests previously named `_ByteIdenticalToFlat` were renamed to `_ResolvesAndRuns`. Without the flat-vs-clone-root comparison, do these tests assert non-vacuous simulation results (e.g. non-zero metrics and specific request completion counts) rather than just asserting zero exit code? | No. The renamed tests verify successful exit, require stdout merely to contain the string `completed_requests`, and compare two runs byte-for-byte. They do not parse the metrics, require completed_requests to be non-zero, or assert the expected count of 20. Therefore output such as `{"completed_requests":0,...}` would satisfy their stated non-vacuity check. |
| G10 | tests | ✅ CONFIDENT | The golden datasets in `sim/cluster/roofline_golden_test.go` and `trained_physics_golden_test.go` had their `ModelConfigDir` fields repointed to the new fixture paths. Are there companion invariant tests verifying that parsing configs from `testdata/catalog/models/` yields exact byte-identical outputs against `testdata/roofline_goldendataset.json` and `testdata/trained_physics_iter29.json`? | Yes, the two companion golden tests load config.json through latency.GetModelConfig from each dataset's repointed `testdata/catalog/models/...` path, run the real roofline or trained-physics simulation, and compare against the unchanged dataset metrics. Request/token counts and TTFT sums are exact; floating metrics use a 1e-9 relative tolerance rather than serialized byte comparison. Both also assert conservation and causality independently of the golden values. |
| G11 | rationale | ✅ CONFIDENT | Why was a committed `testdata/catalog/` fixture directory selected over checking out `blis-catalog` as a git submodule or CI step? What trade-offs between repository size, offline testing, and catalog drift were considered? | The recorded choice favors hermetic offline tests: a small committed clone-root-shaped fixture lets `go test ./...` run without network access, credentials, submodule initialization, or a separate CI checkout. It also keeps test inputs stable instead of allowing upstream catalog movement to change test outcomes. Repository size is limited by retaining only the 12 model configs tests need plus workloads/devices; CodeLlama was omitted because no test consumed it. Drift is accepted for these test fixtures and controlled through historical output goldens and the catalog-config-to-ModelConfig equivalence test rather than by treating testdata/catalog as an authoritative catalog. |
| G12 | rationale | ✅ CONFIDENT | The issue §3 presented three policy options for the 122 doc examples that omit `--catalog`: (a) per-page prerequisite block, (b) self-contained `--catalog` on each example, or (c) entry-point-only export policy. Why was policy (c) chosen over making each example self-contained with explicit `--catalog` flags? | Policy (c) was chosen because BLIS_CATALOG exists to state the shared catalog location once, while adding --catalog to roughly 122 examples would duplicate the same prerequisite throughout the docs. Entry documents clone and export the catalog once, and inner examples assume that environment. This differs from --hardware/--tp, which have no environment-variable form and therefore must appear on every command. |
| G13 | scope | ✅ CONFIDENT | Beyond deleting `model_configs/` and repointing tests and documentation, does this PR introduce any changes to how model parameters (such as `head_dim` or `kv_lora_rank`) are parsed or applied in `sim/latency/`? | No. The production parsing and application of head_dim, kv_lora_rank, qk_rope_head_dim, and related model-shape fields remain in the existing sim/latency loader. This PR changes catalog location/fixtures and rewrites a latency test; it does not add or alter those parser branches or their downstream semantics. |
| G14 | scope | ✅ CONFIDENT | In `cmd/catalog_provenance.go` and `cmd/catalog_provenance_cli_test.go`, catalog provenance dirty-checking paths were updated. Did this change alter the scope of git status checks when evaluating whether a catalog repo has uncommitted modifications? | No. Production provenance still runs `git status --porcelain -- .` with the catalog root as git's working directory, so dirty detection remains scoped to the catalog subtree rather than the whole containing repository. The CLI test change only updated how it locates a config inside the new models/ fixture layout; the separate temporary subtree test still verifies that unrelated repository edits do not mark the catalog dirty. |
| G15 | errors | ✅ CONFIDENT | When `--catalog` points to a non-existent directory or a file, is the error returned as a structured Go `error` from `resolveCatalogRoot` and `resolveModelConfig`, or does it call `logrus.Fatalf` directly inside the resolution helper? | The resolution helpers return structured Go errors. resolveCatalogRoot distinguishes nonexistent, unstatable, and non-directory roots; resolveModelConfig propagates that error. The CLI boundary in resolveLatencyConfig then calls logrus.Fatalf on the returned error. There is no Fatalf inside resolveCatalogRoot or resolveModelConfig itself. |
| G16 | errors | ✅ CONFIDENT | If a config file at `<catalog>/models/<short-name>/config.json` contains malformed JSON or invalid data types, does `resolveModelConfigInCatalog` surface the exact underlying JSON unmarshaling error and file path, or does it collapse the failure into a generic refusal message? | resolveModelConfigInCatalog does not expose the exact JSON unmarshalling error. It first uses json.Valid and a shape check; malformed JSON or readable JSON lacking recognized HuggingFace shape is collapsed into a contextual “not a HuggingFace config.json” error that includes the exact file path. If JSON passes that gate, the analytical configuration path later calls latency.ParseHFConfig/GetModelConfigFromHF; ParseHFConfig wraps exact JSON decoding errors, while wrong field types may become zero/default values and be rejected by later model validation rather than by resolveModelConfigInCatalog. |

### Items to fix
- **F2 · FLAW_FOUND** — No. Most live documentation was migrated, but CLAUDE.md still describes the current resolver as reading `<catalog>/<short-name>/config.json`. The implemented and otherwise documented sole layout is `<catalog>/models/<short-name>/config.json`, so this is a concrete misleading instruction in maintained project documentation.
  _CLAUDE.md:602 claims `cmd.resolveModelConfig` is a single read of `<catalog>/<short-name>/config.json`; cmd/hfconfig.go:25-32 and cmd/hfconfig.go:383-436 implement only `<catalog>/models/<short-name>`; docs/reference/configuration.md:166 documents the correct sole layout_
- **G9 · FLAW_FOUND** — No. The renamed tests verify successful exit, require stdout merely to contain the string `completed_requests`, and compare two runs byte-for-byte. They do not parse the metrics, require completed_requests to be non-zero, or assert the expected count of 20. Therefore output such as `{"completed_requests":0,...}` would satisfy their stated non-vacuity check.
  _cmd/catalog_layout_test.go:104-127 supplies `--num-requests 20`; cmd/catalog_layout_test.go:145-163 and 169-191 only use strings.Contains and string equality_

### Important to consider
- **F1** — There is one remaining incorrect path claim in CLAUDE.md, detailed under F2.
  _cmd/hfconfig.go:383-436; testdata/catalog/models/; scripts/find-saturation.sh:21-42; CLAUDE.md:745; cmd/catalog_layout_test.go:140-205; go test ./cmd, ./sim/latency, and ./sim/cluster all pass_
- **F3** — The stale plural is harmless comment imprecision, but changing it to “absent from the catalog” would avoid confusion.
  _cmd/hfconfig.go:158-161 contains the stale plural; cmd/hfconfig.go:178-205 and cmd/hfconfig.go:383-436 correctly describe the one-candidate behavior; cmd/catalog_provenance_test.go:138-180 uses a deliberately named temporary subtree_
- **G3** — The ModelConfig equivalence test is structural rather than a direct stdout comparison for each of its three models; downstream output coverage is supplied by separate golden tests.
  _sim/latency/catalog_equivalence_test.go:12-37 and 38-76; cmd/lora_noop_test.go:57-74; sim/cluster/roofline_golden_test.go:58-78 and 119-140; sim/cluster/trained_physics_golden_test.go:60-77 and 118-136_
- **G4** — There is no dedicated end-to-end test comparing run and replay stderr for this obsolete invocation, but the shared call path makes the diagnostic common by construction.
  _cmd/hfconfig.go:383-436; cmd/hfconfig_test.go:541-548; cmd/root.go:928-940 and 942-978; cmd/root.go:1571-1580_
- **G7** — A command that does not actually require catalog content, such as observe without a named preset, need not resolve BLIS_CATALOG merely because the flag is registered.
  _cmd/hfconfig.go:35-65 and 89-119; cmd/root.go:1571-1580; cmd/observe_cmd.go:137-144; cmd/convert.go:170-177; cmd/catalog_workloads.go:174-186; cmd/hfconfig_test.go:295-335_
- **G8** — This is consistent with the explicitly chosen entry-point-only policy, but it would not detect an isolated inner page that becomes an alternative entry point without catalog context.
  _cmd/docs_examples_test.go:145-156 and 158-198_
- **G10** — The comments call this byte-for-byte identity, but floating-point assertions are tolerance-based, not literal output-byte comparisons.
  _testdata/roofline_goldendataset.json:8-9; testdata/trained_physics_iter29.json:23-26; sim/cluster/roofline_golden_test.go:119-140 and 199-285; sim/cluster/trained_physics_golden_test.go:118-140 and 202-292; go test ./sim/cluster passes_
- **G11** — The fixture can still drift from external blis-catalog coverage; it is deliberately a test fixture, not a catalog mirror.
  _CLAUDE.md:745 records option (a), offline operation, the clone-root fixture shape, and dropping unused CodeLlama; testdata/catalog/ contains only devices/, models/, and workloads/; sim/latency/catalog_equivalence_test.go:12-31_
- **G16** — The generic malformed-entry refusal is pre-existing defensive behavior and still provides the actionable path, but it intentionally does not preserve JSON syntax details at the initial resolver gate.
  _cmd/hfconfig.go:152-175 and 286-327; sim/latency/config.go:455-483; cmd/root.go:1013-1023_

namasl:
## Author response to qa-review round-2 BLOCK (head `edef7a57`)

Both findings fixed.

**F2 (FLAW_FOUND — CLAUDE.md still describes the resolver reading the flat `<catalog>/<short-name>/config.json`) — FIXED.**
Corrected/annotated every remaining flat-layout claim so none misdescribes the sole
`models/` layout:
- **CLAUDE.md:556** (NS-6 *invariant* entry — describes current behavior): rewritten to state
  `<catalog>/models/<short-name>/config.json` is **the sole clone-root layout** and that #1771
  deleted the bundled tree + the flat transition fallback (dropped the "transition fallback to
  the flat …" clause and "absent from BOTH layouts").
- **CLAUDE.md:602** (#1733 *history* entry): the flat-path claim now carries an inline note that
  #1774 moved the entry under `models/` and #1771 made it the sole layout (append-only history
  convention, matching the existing `#1733`-annotates-`#435` style).
- **docs/getting-started/quickstart.md** and **docs/contributing/standards/invariants.md**: the
  active `<catalog>/<model>/config.json` / `<catalog>/<short-name>/config.json` guidance now reads
  `<catalog>/models/<...>/config.json`.
Verified: no un-annotated flat-layout current claim remains; `TestDocExamplesDocumentTheCatalogEnvVar`
and `TestDocExamplesPassDeploymentFlags` still pass.

**G9 (FLAW_FOUND — `_ResolvesAndRuns` non-vacuity gate is a bare `strings.Contains("completed_requests")`, which `"completed_requests": 0` would satisfy) — FIXED.**
Added `requireCompletedRequests(t, label, stdout)` (regex-extracts `completed_requests` and
asserts **> 0**) and use it in both `TestRunCmd_CatalogCloneRootLayout_ResolvesAndRuns` and
`TestReplayCmd_CatalogCloneRootLayout_ResolvesAndRuns` in place of the substring check. A leg
that resolved the model but completed nothing now fails the gate. Both tests pass with the
stronger assertion.

`go build ./...` and `go test ./...` remain green. Requesting adjudication.

namasl:

## qa-review adjudication — PR #1797: ✅ PASS

|----|----|----|----|
| F2 | FLAW_FOUND | ✅ RESOLVED | The misleading current-layout statement is corrected. The active invariant now identifies `<catalog>/models/<short-name>/config.json` as the sole layout, while the former flat path at the historical entry is explicitly labeled as superseded by #1774 and #1771. |
| G9 | FLAW_FOUND | ✅ RESOLVED | The tests no longer accept `completed_requests: 0`. They extract the numeric metric and fail unless it is greater than zero, and both run and replay tests invoke this strengthened non-vacuity check. The cmd test package also passes. |

### Still blocking

### Cleared
- **F2** (✅ RESOLVED) — The misleading current-layout statement is corrected. The active invariant now identifies `<catalog>/models/<short-name>/config.json` as the sole layout, while the former flat path at the historical entry is explicitly labeled as superseded by #1774 and #1771.
  _CLAUDE.md:556 states that `cmd.resolveModelConfig` reads `<catalog>/models/<short-name>/config.json` as the sole clone-root layout. CLAUDE.md:602 immediately annotates the historical `<catalog>/<short-name>/config.json` path with the later move to `models/` and identifies `<catalog>/models/<short-name>/config.json` as the sole current layout._
- **G9** (✅ RESOLVED) — The tests no longer accept `completed_requests: 0`. They extract the numeric metric and fail unless it is greater than zero, and both run and replay tests invoke this strengthened non-vacuity check. The cmd test package also passes.
  _cmd/catalog_layout_test.go:14-30 defines `requireCompletedRequests`, parses `completed_requests`, and rejects missing, unparsable, or non-positive values. cmd/catalog_layout_test.go:175-178 applies it to the run output; cmd/catalog_layout_test.go:201-204 applies it to replay output. `go test ./cmd` completed successfully._

susiejojo (approved):
LGTM

Issue #1806, opened 2026-09-22
hardening(ci): AI flows must filter untrusted comment text by author write-permission before reading it


mtoslalibu:
**What issue does this address?**

The AI flows read comment text this repository does not control. *Who* may trigger them is already gated to `admin`/`maintain`/`write` (`check-permissions` / `authorize`). *What* they **read** is not gated at all — and this repo is public, so anyone can comment on any PR or issue. Comment text is therefore a prompt-injection surface into flows that run on a persistent self-hosted runner with credentials in the environment (and, in the correction phase, `contents: write`).

**Scope: filter external-authored comment text out of EVERY AI flow whose inputs we control.** Not just the delivery-loop agents — the qa-review reviewer too.

The flows that read issue/PR comments:

1. **`deliver-verify.yml`** — the review agent (agent mode; we own the prompt).
2. **`deliver-correct.yml`** — the correction agent (agent mode; we own the prompt).
3. **`scripts/qa-review/answerer.py` + `adjudicator.py`** — the cross-vendor reviewer reads issue/PR comments through its own `gh` tools and renders them into an LLM prompt.

All three MUST read comment text only through the trusted-comment filter.

**Mechanism**

Comment text reaches these flows only through `scripts/deliver-trusted-comments.sh`, which keeps authors holding `admin`/`maintain`/`write` plus the automation's own comments (an explicit allowlist: `github-actions[bot]`, `claude[bot]`) and drops everyone else. Trust is the author's real repository **permission**, never `author_association`.

- Flows 1 & 2 (agent mode): a step assembles the filtered digest before the agent, and the prompt reads that file instead of running `gh pr view --comments` / raw `gh api .../comments`.
- Flow 3 (Python): the answerer/adjudicator's comment-reading tools route through the same filter (or are handed a pre-filtered digest) so no unfiltered comment enters their prompt.

**Explicitly out of scope (documented, tracked elsewhere)**

- **`claude.yml` (`@claude`, `/blis-pr-review`)** — runs `claude-code-action` in **tag mode**: the action assembles the PR/comment context itself, with no prompt seam to insert a filter. Filtering it would require converting it to agent mode (a separate, larger change to the interactive command). Its external-**authored** exposure is covered by #1813 (refuse AI flows on external-authored PRs/issues); the residual (external *comments* on a trusted-authored PR reaching tag mode) is a documented limitation, not this issue's scope.
[...]

PR #1820, opened 2026-09-22
hardening(ci): gate AI flows on the PR/issue AUTHOR's write access (#1813)

### LINKED ISSUE #1813: hardening(ci): gate AI flows on the PR/issue AUTHOR's write-access — refuse outside-authored PRs and issues
**What issue does this address?**

The AI flows (bot delivery: `implement`/`verify`/`correct`; interactive `@claude`; `/blis-pr-review`; `/approve-issue-for-pr-delivery`) can be pointed at a PR or issue that was **authored by someone without write access**. On such an item the untrusted content is not just the comments (which #1806 now filters for the delivery loop) — it is the **PR/issue body and, on a PR, the diff itself**. No comment filter helps there: reading the outside-authored body/diff is the whole task, and it flows straight into an agent running on a persistent self-hosted runner with credentials (and, in the correction phase, `contents: write`).

A maintainer triggering the command is **not** a safeguard: triggering does not sanitize the body/diff, and expecting the maintainer to read the whole thing for injection defeats the point of the automation.

**The rule we want: gate on the AUTHOR of the PR/issue, not on fork status.** If the PR or issue was authored by an account that does not hold `admin`/`maintain`/`write` on this repo, the AI flows must not run on it. Author is what matters — a same-repo branch opened by an outsider is as untrusted as a fork; a fork opened by a maintainer is fine.

This is the container-level complement to #1806 (which filters comment *text*). #1806 handles the trusted-container case (a delivery PR, bot-authored) where only stray comments are untrusted; this handles the case where the container itself is untrusted.

**Proposed fix**

As an early gate in each AI-flow workflow, before any agent step:

1. Resolve the PR/issue **author's** repository permission (`repos/{repo}/collaborators/{author}/permission`), the same check already used for the *triggerer* (`check-permissions` / `authorize`).
2. If the author is not `admin`/`maintain`/`write`: do **not** run the agent. Skip with a clear, logged reason (and, where a comment channel exists, a one-line note so the requester knows why).
3. Gate on `author`, independent of `head.repo.full_name` (fork-ness). Fork status is neither necessary nor sufficient.

Applies to: `deliver-implement.yml`, `deliver-verify.yml`, `deliver-correct.yml`, `claude.yml`. Delivery PRs are bot-authored (`github-actions[bot]`/`claude[bot]`), which the check must treat as trusted automation (same bot-recognition the trigger gates use), so normal deliveries are unaffected.

**Relationship to other work**
- Complements #1806 (comment-text filter for the delivery loop). Distinct surface: this gates the whole untrusted *container*, #1806 filters trusted-container *comments*.
- Supersedes the "skip fork PRs" framing floated in the earlier #1703 discussion: the correct signal is the author's permission, not fork status.

- [x] None (CI/workflow gating; no simulator behaviour)

**Design document**

Not needed — workflow gating, no architectural impact.

mtoslalibu:
## Summary

Every AI flow — the bot delivery loop (`implement`/`verify`/`correct`), interactive `@claude`, and `/blis-pr-review` — could be pointed at a PR or issue **authored** by an account without write access. There the untrusted content is the PR/issue **body** and, on a PR, the **diff** itself: reading it *is* the task, so #1806's comment-text filter does not help, and a maintainer triggering the run does not sanitise it. Every existing gate (`check-permissions` / `authorize`) checks only the **triggerer** (`context.actor`), never the author.

This adds the container-level complement: gate on the **author's** write access, independent of fork-ness (a same-repo branch opened by an outsider is as untrusted as a fork; a fork opened by a maintainer is fine).

- **`scripts/deliver-author-gate.sh`** — the single tested decision, reused by all four workflows. Same `admin`/`write`/`maintain` law as the triggerer gate (and as `deliver-issue-refinements.sh`), with two single-author differences: `github-actions[bot]`/`claude[bot]` are trusted **without** a probe (the loop's own PRs/issues are bot-authored and the permission API 404s for a bot login), and a probe failure fails closed **and loud** (exit 3), distinct from a clean-404 denial (exit 1). Reuses the sibling's bounded-`gh` watchdog so a stalled probe cannot hang a job.
- **`claude.yml` / `deliver-implement.yml`** — gate in the ubuntu `check-permissions` job, so a refused author never wakes the self-hosted runner; ANDed into the agent job(s).
- **`deliver-verify.yml` / `deliver-correct.yml`** — gate before the agent, on the **sub-issue** author (the delivery PR is bot-authored + same-repo by existing guards, so the issue — read via `gh issue view` — is the untrusted container). The generic failure reporter and the `always()`-gated cleanup/dismissal steps are excluded on a handled block, so a refusal draws exactly one specific comment + `needs-human`.

## Behavioral contracts

- **BC-1 Trusted author runs** — `admin`/`write`/`maintain` author ⇒ gate passes, agent runs as before.
- **BC-2 Untrusted author refused** — `read`/`none` (fork **or** same-repo) ⇒ agent does not run; a one-line reason is posted; independent of `head.repo.full_name`.
- **BC-3 Bot author trusted** — `github-actions[bot]`/`claude[bot]` ⇒ pass **without** a network probe; normal deliveries unaffected.
- **BC-4 Probe failure fails closed, loudly** — 5xx/403/network/deadline ⇒ no agent, reported as a probe failure (exit 3), never a silent drop (R1); a clean 404 is a denial (exit 1).
- **BC-5 Complement, not replacement** — ANDed with the existing triggerer gate; both must pass.
- **BC-6 One decision, four callers** — all four reach the same decision via one script; only author-login resolution (event-shape-specific) stays per-workflow.

## Testing

- `go test ./scripts/` — decision (`deliver_author_gate_test.go`: write/admin/maintain allow, read/none/404 block, bot allowlist without a probe, non-allowlisted bot blocked without a probe, 5xx/403/network/deadline read-failed, empty/malformed login refused) and wiring (`deliver_author_gate_wiring_test.go`: each flow invokes the gate, verify/correct reporters exclude the block, both `claude` agent jobs gate on `author_allowed`, ubuntu gate jobs check out the trusted tree).
- `go build ./...`, `go vet ./scripts/...` clean. YAML validated for all four workflows.
- No `sim/` or `cmd/` change — INV-6 untouched (this is CI/workflow gating only).

## Discovered issues

None filed — no pre-existing bugs surfaced in the touched surface.

Fixes #1813


mtoslalibu:
## Self-review — part 1: analysis & design reasoning

Posting my reasoning as I review, per the delivery workflow. This is a **CI/workflow-only** PR (no `sim/`/`cmd/` change), so the simulator-specific review lenses (DES, vLLM, preemption safety, timeout/deadline consistency, cross-path run/replay/observe parity) are **not applicable** — recorded here so their absence is a decision, not an omission. The applicable lenses are Substance, Code-Quality/Error-Handling, Test-Quality, Automated-Reviewer-Sim, and Security/Robustness.

### Why the issue is valid (verified against the code)
Every gate today probes `context.actor` — the **triggerer** — never the author:
- `claude.yml` `getCollaboratorPermissionLevel({username: context.actor})`
- `deliver-verify.yml` `authorize` and the `deliver-implement`/`deliver-correct` permission jobs — same `context.actor`.

So a maintainer commenting `@claude` / `/approve-issue-for-pr-delivery` on an **outside-authored** PR or issue passes the gate, and the agent then reads the untrusted **body/diff** (the read *is* the task, so #1806's comment filter cannot help) on a self-hosted runner with credentials. `claude.yml` additionally checks out `refs/pull/N/head` with no fork guard.

### Design decisions (and the trade-offs)
1. **Gate on the author, not fork-ness** — as the issue argues: a same-repo branch opened by an outsider is as untrusted as a fork; a maintainer's fork is fine. Fork status is neither necessary nor sufficient.
2. **Which author, per flow** — the delivery loop's untrusted spec is the **sub-issue** (verify feeds the agent `gh issue view $ISSUE`; implement builds the body). The delivery PR is bot-authored **and** same-repo by existing guards, so its author is already trusted — therefore all three delivery flows gate the **issue** author; `claude.yml` gates the PR author (PR triggers) or issue author (issue triggers).
3. **One tested decision, four callers** (`scripts/deliver-author-gate.sh`) rather than four copies — reuses the exact `admin/write/maintain` + 404-vs-probe-failure law already trusted in `deliver-issue-refinements.sh`, plus a bot-allowlist short-circuit (the permission API 404s for a bot login, so probing would deny the loop's own PRs).
4. **Placement** — `claude.yml`/`deliver-implement.yml` gate in the **ubuntu** `check-permissions` job, so a refused author never even wakes the self-hosted runner. `deliver-verify`/`deliver-correct` gate before the agent (their permission job can't cheaply resolve the branch→issue for push/review triggers), and I excluded the generic failure reporter + `always()`-gated cleanup/dismissal/conflict-check steps on a handled block so a refusal draws exactly one specific comment + `needs-human`.
5. **Bot allowlist is a login list, not a `[bot]`-suffix test** — a third-party App (e.g. `dependabot[bot]`) authoring a fork PR is blocked, not waved through; only `github-actions[bot]`/`claude[bot]` are trusted.

### Fail-closed semantics
`allowed` (exit 0) → run. `blocked` (exit 1, clean 404 / `read` / `none`) → refuse with a definitive reason. `read-failed` (exit 3, 5xx/403/network/deadline) → refuse **and** report a probe failure distinct from a denial (R1), never a silent drop. Bounded-`gh` watchdog reused so a stalled probe can't hang a job.

Cross-vendor QA + Q/A findings + verdict to follow.

mtoslalibu:
## Self-review — part 2: cross-vendor QA (GPT-5.4 + Gemini-3.1-Pro)

Ran an independent critical review of the gate script + all four wirings through two non-Claude vendors via litellm, focused on bypass/injection, shell-decision correctness, `always()`-step wiring, and silent/double-comment failures. I triaged every finding against the code (trust but verify).

### Acted on (1 genuine improvement — pushed)
- **Silent refusal if the comment API fails (verify/correct).** The refusal comment was `gh ... || true` and then the generic reporter was suppressed via `blocked=true` — so a comment-API outage left a red run with no explanation. **Fixed:** `blocked=true` is now set *only* when the specific comment posts; otherwise `blocked` is left unset so `Report a failed verify/correction` fires as a fallback. A refusal can no longer be silent (R1). (commit `hardening(ci): author-gate refusal falls back to the generic reporter…`)

### Verified NON-issues (reported by a model, checked, and dismissed with evidence)
- **"`run_bounded` may exec a PATH-resolved `gh`" (GPT, IMPORTANT).** `command -v gh` returns an **absolute** path for an external binary (a bare name only for builtins/aliases/functions), and the idiom is copied verbatim from the production `scripts/deliver-issue-refinements.sh`. No PATH hijack. 
- **"`author_allowed` unset → silent skip" (GPT, IMPORTANT).** `author_allowed` is only empty when `steps.check.outputs.allowed != 'true'` — i.e. the **triggerer** was already denied and handled by the existing upstream gate. When `allowed == 'true'`, the gate step always runs and sets the output. No new hole.
- **"`timeout --kill-after` / orphaned `sleep` are non-portable" (Gemini, MINOR).** Same watchdog as the sibling `deliver-issue-refinements.sh` that already runs in production on these self-hosted runners (GNU coreutils present); orphaned sleep is swept by job cleanup. No behavior change vs. the trusted sibling.
- **Injection / bypass / checkout safety.** Both vendors independently confirmed: login is regex-gated (`^[A-Za-z0-9-]+$`) before any URL/shell use and passed via env/quoted arg; the bot allowlist is exact-match with a bot-shaped catch-all denial; and for `issue_comment`/`pull_request_review_comment`/`issues` the checkout is the trusted default branch (no `ref:` override), so only BLIS's own script runs.

### Noted, won't-fix (rationale)
- **Exit code `2` (usage) classified as probe-failure by callers (GPT, IMPORTANT→downgraded).** Unreachable: every caller invokes the script with exactly one argument, so the script returns only `{0,1,3}`. Adding branch logic for an unreachable code is the speculative robustness the guidelines say to avoid; the reachable `{0,1,3}` are handled correctly.
- **`--permission` seam treats garbage as denial; author-expr brittle to future events (MINOR).** Test-only seam; the `on:` block has exactly three events and the expression is correct for all three. Not current defects.

Gemini's overall read: *"highly robust, well-reasoned hardening change… no bypasses or injection vectors found."* GPT found no CRITICAL. My own Q/A + verdict next.

mtoslalibu:
## Self-review — part 3: Q/A self-audit + verdict

### Q/A (probing questions, answered with evidence)
- **Fully implements #1813?** Yes — all four flows gate on the author: `claude.yml` (both agent jobs AND `author_allowed`), `deliver-implement.yml` (fails `check-permissions` → `deliver` skipped), `deliver-verify.yml` / `deliver-correct.yml` (gate before the agent). CONFIRMED.
- **Any residual path where an outside author reaches an agent?** No. claude: both jobs require `author_allowed == 'true'`. implement: gate exit 1 fails the ubuntu job, so the self-hosted `deliver` never starts. verify/correct: gate exit 1 → the agent step (plain `if: paused=='false'`, implicitly `success()`-gated) skips. CONFIRMED.
- **Normal (bot-authored) delivery unaffected?** Sub-issue authors are maintainers (write) or `github-actions[bot]` (allowlisted, no probe); both pass. An outside-authored sub-issue is exactly what #1813 says to refuse. CONFIRMED.
- **`always()` steps on the block path?** verify: only a worktree-cleanup (safe no-op) + the now-excluded reporter. correct: `Record the dismissal state` and `Did the conflict actually get resolved?` (which could `git fetch`/comment) are both now excluded via `blocked != 'true'`. CONFIRMED.
- **report-status on an author-block (claude)?** Both agent jobs `skipped` ⇒ no `failure`/`cancelled`/`success-with-execution_file` clause matches ⇒ report-status skips, exactly as for a triggerer denial. No false status. CONFIRMED.
- **Determinism / INV-6?** No `sim/` or `cmd/` change; simulator output untouched. CONFIRMED.
- **Known trade-off (documented, not a defect):** in `deliver-correct` the round counter advances before the gate, so a blocked correction consumes one round — acceptable defense-in-depth, since verify's identical gate blocks first in normal flow and the loop halts on `needs-human` regardless.

### Verification gate
- `go build ./...` — OK
- `go test ./scripts/ -count=1` — ok (full package, 136s; all pre-existing workflow structural tests still pass)
- `go vet ./scripts/...` — OK; `gofmt -l` — clean
- YAML parse (4 workflows) — OK; `bash -n` on all four embedded gate blocks — OK
- `golangci-lint`: not run locally (my local binary predates Go 1.24 and errors on `unicode`/`yaml` imports in *pre-existing* files); CI's pinned v2.9.0 will run it. New files are vet/gofmt-clean.

### Verdict: **READY TO MERGE**
Closes the container-level hole #1813 describes: every AI flow now refuses an outside-authored PR/issue, gating on the **author** (not fork-ness), with `github-actions[bot]`/`claude[bot]` trusted, one tested decision reused by four callers, fail-closed-and-loud on a probe outage, and — after the cross-vendor-surfaced fix — no silent refusal even if the comment API is down. Two independent vendors found no CRITICAL and no bypass. It complements #1806 (comment-text filter) without overlapping it.

Human review welcome on: (a) gating the **issue** author (vs also the PR author) in verify/correct — justified because the delivery PR is bot-authored + same-repo by existing guards, so the issue is the only outside-authorable content source; (b) adding `actions/checkout` to the two ubuntu permission jobs to run the shared script (trusted default-branch tree only).

.github/workflows/claude.yml
```diff
@@ -148,6 +153,59 @@ jobs:
               }
             }
 
+      # Trusted tree, for scripts/deliver-author-gate.sh only. For every trigger here — an
+      # issue_comment, a pull_request_review_comment, or an issue — `github.ref` is the DEFAULT
+      # branch (GitHub does not associate these events with a PR head), so this checks out BLIS's
+      # own main-branch script, never PR content, and nothing runs but that script.
+      - uses: actions/checkout@v5
```
namasl: [P0] Pin this checkout to the trusted default branch. GitHub documents `pull_request_review_comment` with `GITHUB_REF=refs/pull/PULL_REQUEST_NUMBER/merge`, so the default checkout here is the attacker-controlled PR merge tree, not `main`. Line 191 then executes the PR copy of `deliver-author-gate.sh` before the author decision; a fork PR can replace that script with an allow result and a maintainer inline `@claude` comment will admit the self-hosted agent. Please set an explicit trusted ref and strengthen `TestUbuntuGateJobsCheckoutTrustedTree` to inspect that ref rather than merely finding an `actions/checkout` string. Event semantics: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_review_comment

.github/workflows/deliver-verify.yml
```diff
@@ -407,6 +407,57 @@ jobs:
...
+      # maintainer triggered the run — reading that outside-authored body IS the task, so no comment
+      # filter (#1806) helps. deliver-author-gate.sh applies the same admin/write/maintain law the
+      # triggerer gate uses (bots trusted without a probe; a probe failure fails closed AND loud,
+      # distinct from a denial). The delivery PR itself is bot-authored and same-repo (Validate the
+      # target PR), so the issue — not the PR — is the untrusted container this phase must gate.
+      #
+      # On a refusal this posts its own specific comment and moves the label, then exits non-zero;
+      # the downstream steps skip (their `if:` is implicitly `success()`-gated) and `Report a failed
+      # verify` is excluded via `author_gate.outputs.blocked`, so there is exactly one comment.
+      - name: Gate on the sub-issue author's write access
```
namasl: [P0] This gate is also sourced from the content it is meant to distrust. The checkout immediately above defaults to the PR merge ref for `pull_request_review` / `pull_request_review_comment` and to the delivery ref for a push. An existing outside-authored delivery can therefore replace `scripts/deliver-author-gate.sh` and self-authorize before any agent step. Explicitly check out the default branch for the gate; only fetch/read PR objects after it passes. The nearby `Trusted tree` comment is currently false for these event types.

.github/workflows/deliver-correct.yml
```diff
@@ -136,6 +136,52 @@ jobs:
...
+      # deliver-author-gate.sh applies the same admin/write/maintain law as the triggerer gate (bots
+      # trusted without a probe; a probe failure fails closed AND loud). In normal operation the
+      # verify phase's identical gate blocks first, so this is defence-in-depth for a correction
+      # dispatched by any other route; the delivery PR is bot-authored and same-repo (Validate the
+      # target PR), so the issue is the untrusted container.
+      #
+      # On a refusal it posts a specific comment and moves the label, then exits non-zero: the
+      # downstream steps skip (implicitly `success()`-gated) and `Report a failed correction` is
+      # excluded via `author_gate.outputs.blocked`, so there is exactly one comment.
+      - name: Gate on the sub-issue author's write access
```
namasl: [P0] `deliver-correct.yml` is dispatched with `--ref deliver/issue-N`, and the checkout at line 134 has no explicit ref, so the script invoked here is the delivery-branch copy. That makes the defense-in-depth gate bypassable by changing the gate script on the content branch. Please run the gate from an explicit trusted default-branch checkout, then check out the delivery branch only after the gate passes.

docs/contributing/standards/agent-trust.md
```diff
@@ -106,6 +106,23 @@ reason — a status, or a category when there is no status — never the caught
...
+and there the untrusted input is the PR/issue **body** and, on a PR, the **diff** — reading it *is*
+the task, so #1806's comment-text filter does not help, and a maintainer triggering the run does not
+sanitise it. `scripts/deliver-author-gate.sh` is the one place that decision lives, applying the same
+`admin`/`write`/`maintain` law as the triggerer gate with two differences for a single-author gate:
+`github-actions[bot]`/`claude[bot]` are trusted **without** a probe (the delivery loop's own PRs and
+issues are bot-authored, and the permission endpoint 404s for a bot login), and a probe failure fails
+closed **and** loud (exit 3), distinct from a clean-404 denial (exit 1). The signal is the **author**,
+never fork-ness: a same-repo branch or issue opened by an outsider is as untrusted as a fork, and a
+fork opened by a maintainer is fine. `claude.yml` and `deliver-implement.yml` run it in their
+ubuntu `check-permissions` job (so a refused author never wakes the self-hosted runner);
```
namasl: [P2] Please update the user-facing and script documentation too. `docs/contributing/automated-delivery.md` still says the command is merely restricted to collaborators and that any single deliverable issue works, without the new issue-author requirement; its Security section also says the root checkout always comes from `main`, which is currently false for review events and correction dispatches. `docs/contributing/pr-workflow.md` repeats only the triggerer rule, and `scripts/README.md` does not document the new shared gate. Updating only this trust standard leaves the documented behavior and security model inconsistent.

.github/workflows/claude.yml
```diff
@@ -148,6 +153,59 @@ jobs:
...
+            echo "author_allowed=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "author_allowed=false" >> "$GITHUB_OUTPUT"
+          if [[ "$rc" -eq 1 ]]; then
+            body="\`@claude\` will not run here: this was authored by \`$AUTHOR\`, who does not hold write access to this repository. The agent reads the body and diff, so it is not run on outside-authored content (#1813). A maintainer can reproduce the request on an item they authored."
+          else
+            body="\`@claude\` was not run: could not verify whether \`$AUTHOR\` has write access. This is a probe failure, not a permission decision — retrying should work."
+          fi
+          gh issue comment "$TARGET" --repo "$GITHUB_REPOSITORY" --body "$body" || true
```
namasl: [P1] Do not swallow this comment failure for a permission-probe failure. When the gate returns exit 3 and this API call also fails, `author_allowed=false` skips both agent jobs and `report-status`, while this job exits 0. The workflow is green and the requester gets no explanation: the same silent-refusal bug the follow-up commit fixed in verify/correct remains here. At minimum, make the check job fail when an exit-3 explanation cannot be posted, and cover that fallback.

namasl (changes requested):
Requesting changes. The central author-permission decision is sensible and the script tests pass, but the current wiring does not yet establish the claimed trust boundary: `claude`, `deliver-verify`, and `deliver-correct` can execute the gate script from PR/delivery-controlled refs; the Claude probe-error path can still fail silently if its explanatory comment fails; and the user-facing delivery/security docs remain inconsistent with the new rule. I found no unrelated functional additions, and `/archon-pr-review` is deterministic static analysis rather than another LLM agent path, so I do not consider its omission from this issue a gap. Full `go test ./scripts/ -count=1` and `git diff --check` pass locally; CI is green.

.github/workflows/deliver-correct.yml
```diff
@@ -131,11 +131,73 @@ jobs:
...
+      # `.gate-trusted` path — never the delivery-branch checkout above. Running the gate from the
+      # delivery branch would let an outside-authored delivery replace deliver-author-gate.sh with
+      # an "allow" result and self-authorize; pinning to `default_branch` guarantees BLIS's own
+      # script runs.
+      - name: Check out the trusted gate script
+        if: steps.paused.outputs.paused == 'false'
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          path: .gate-trusted
```
namasl: [P0] This path is not isolated from the preceding delivery-branch checkout. That ref can track `.gate-trusted` as a symlink (including to an absolute directory). `actions/checkout@v5` validates the input lexically with `path.resolve`, but its existence check uses `fs.statSync` and its existing-directory preparation follows the symlink; it can therefore clean/init/write the symlink target on this persistent runner. `deliver-verify` has the same ordering. Please check out the default branch at workspace root and run the gate *before* checking out the event/delivery ref (then switch the root checkout only after the gate passes), and add an ordering assertion. A nested path beneath an untrusted tree is not a trust boundary.

.github/workflows/deliver-verify.yml
```diff
@@ -401,12 +401,81 @@ jobs:
...
+          fetch-depth: 1
+
+      # Container-level author gate (#1813), before the CI dispatch and the agent. The TRIGGERER is
+      # already write-checked (the authorize job), but that is not the untrusted axis here: this
+      # phase feeds the agent `gh issue view $ISSUE_NUMBER` for the contracts (see the review step's
+      # prompt) plus the diff, and the SUB-ISSUE's author may hold no write access even when a
+      # maintainer triggered the run — reading that outside-authored body IS the task, so no comment
+      # filter (#1806) helps. deliver-author-gate.sh applies the same admin/write/maintain law the
+      # triggerer gate uses (bots trusted without a probe; a probe failure fails closed AND loud,
+      # distinct from a denial). The delivery PR itself is bot-authored and same-repo (Validate the
```
namasl: [P0] `Validate the target PR` does not validate this assertion: it checks only `.head.repo.full_name`, `.head.ref`, and `.state`, never `.user.login`. Verify/correct therefore gate only the sub-issue author while leaving the PR author unguarded, even though this job reads the PR body/diff and #1813 explicitly covers PR-authored content. Please enforce the allowlisted bot author (or gate the PR author too) in both phases and cover that structurally; same-repository plus branch shape does not prove bot authorship.

docs/contributing/automated-delivery.md
```diff
@@ -305,7 +311,7 @@ Not yet automated, each its own follow-up: sequencing sub-issues `0..N` and open
...
+**The workflow's own steps run trusted code — but the root checkout ref is not universally `main`, so trust must be pinned where it matters.** For `workflow_dispatch` (the normal delivery path) the root checkout is the dispatched ref; for a `push` it is the delivery branch; and for a `pull_request_review`/`pull_request_review_comment` GitHub sets the ref to the PR *merge* tree. So the older claim that `scripts/`/`.archon-version` "always come from `main`" was not true for the review and correction paths. The **author gate** (below) therefore does **not** run its script from the root checkout: it uses a dedicated `actions/checkout` pinned to `github.event.repository.default_branch` in a separate `.gate-trusted` path, so the decision that admits the agent can never be sourced from content the author controls (#1813). `deliver-gate.sh` runs post-verdict from the root tree; building archon from a possibly-untrusted root ref on the review/correction paths is a separate, pre-existing hardening item tracked outside #1813.
```
namasl: [P1] This correction leaves adjacent security claims false. The next paragraph still says PR code never runs on the self-hosted runner, while review events put the PR merge tree at the root and `Build archon` executes that tree's `scripts/archon-build.sh`. The `deliver-verify.yml` header still says its only workspace checkout is the default branch, and line 479 still says there is no PR checkout/execution in the job. Please restore the documented invariant or update every stale doc/code comment to state the actual behavior, and link the concrete follow-up for the item described only as ‘tracked outside #1813’.

namasl (changes requested):
Re-review of `94b79a1`: the original five findings are substantially addressed, and the focused local checks pass, but I found two remaining trust-boundary gaps plus stale security documentation. The nested trusted checkout can be redirected by a tracked symlink after the untrusted checkout, and verify/correct do not actually validate their assertion that the delivery PR is bot-authored. The PR also currently conflicts with `main`, and GitHub reports no checks on this head. I cannot approve until these are fixed, the contradictory security claims are made accurate, the branch is rebased, and the resulting diff/checks are re-reviewed.

Local verification on this head: `git diff --check`, `bash -n scripts/deliver-author-gate.sh`, `go test ./scripts/ -count=1`, and `go vet ./scripts/...` all pass.

.github/workflows/deliver-correct.yml
namasl: Checked `cee4f21`: setting `working-directory: .gate-trusted` does not fix this. The second checkout resolves/follows `.gate-trusted` before the gate step starts, and the runner also follows the same symlink when entering that working directory. The checkout ordering therefore remains the blocking issue.

.github/workflows/claude.yml
mtoslalibu: Fixed in 04d8d898. The gate no longer relies on the implicit ref: `check-permissions` now checks out `ref: ${{ github.event.repository.default_branch }}` at the workspace root (this job does no other checkout) and runs the trusted script from there, so the `pull_request_review_comment` PR-merge-tree ref can no longer supply the gate script. `TestGateRunsFromTrustedDefaultBranchCheckout` now asserts the checkout *immediately preceding* the gate is pinned to the default branch (not merely that an `actions/checkout` exists).

.github/workflows/deliver-verify.yml
mtoslalibu: Fixed in 04d8d898. `deliver-verify`'s single workspace checkout is now pinned to `ref: default_branch`, so scripts/.archon-version (and the gate) always come from trusted code for every trigger including the review events — which also closes the archon-from-untrusted concern and makes the header's "trusted tree" claim true. The gate runs before the CI dispatch and agent.

.github/workflows/deliver-correct.yml
mtoslalibu: Fixed in 04d8d898. `deliver-correct` now checks out the default branch at root FIRST, runs the gate, and checks out the delivery branch it edits only AFTER the gate passes. `TestCorrectChecksOutDeliveryBranchOnlyAfterGate` pins that ordering.

docs/contributing/standards/agent-trust.md
mtoslalibu: Fixed. Updated `automated-delivery.md` (issue-author rule at the top + a dedicated author-gate paragraph in Security, and the "always from main" claim is now accurate after pinning), `pr-workflow.md` (author rule beside the triggerer rule), `scripts/README.md` (documents the shared gate), and `agent-trust.md`.

.github/workflows/claude.yml
mtoslalibu: Fixed in 94b79a14. On the refusal path claude.yml now conditions on whether the comment posted: if it posts, exit 0 (clean skip); if it fails, `::error` + `exit 1` so the run goes RED rather than green-with-no-explanation. `TestClaudeGateFailsLoudWhenRefusalCommentCannotPost` pins it.

.github/workflows/deliver-correct.yml
mtoslalibu: Fixed in 04d8d898. Agreed — a nested path under an untrusted checkout is not a boundary. Replaced `.gate-trusted` with a default-branch checkout at the workspace ROOT placed BEFORE the delivery-branch checkout; the gate runs there, and the delivery branch is checked out only after it passes. No nested path remains.

.github/workflows/deliver-verify.yml
mtoslalibu: Fixed in 04d8d898. Both phases now gate the PR author too. The canonical PR-author login is resolved via REST `.user.login` in `Validate the target PR` (a GET — `gh pr view --json author` drops the `[bot]` suffix and would break the allowlist) and passed to the gate as `PR_AUTHOR`; keeping the `gh api .../pulls` read out of the gate step also avoids tripping the repo-wide no-PR-creation guard beside the label POST. `TestVerifyAndCorrectGateBothAuthors` pins both authors are resolved and checked.

docs/contributing/automated-delivery.md
mtoslalibu: Fixed in 04d8d898. Rather than caveat the claim, I made it true: pinning verify's root checkout to `default_branch` means scripts/.archon and the gate always come from main, so "PR code never runs on the self-hosted runner" and the header hold for review events too. The "tracked outside #1813" line is removed (archon-from-untrusted is resolved by the pin). correct edits the delivery branch, so it checks that branch out — but only after the gate.

mtoslalibu:
## Review round 2 — addressed @namasl's CHANGES_REQUESTED

Thanks — the P0 was a real bypass I missed, and my own cross-vendor QA got it wrong (both models asserted the checkout was the default branch for these events; it isn't for `pull_request_review_comment`). All findings are fixed. Commits `94b79a14` (round 1) and `04d8d898` (round 2).

### What changed and how

**P0 — the gate ran untrusted code (all 3 workflow comments + the symlink follow-up).**
The gate is now executed from a checkout **pinned to `github.event.repository.default_branch` at the workspace ROOT, before any event/delivery checkout** — not a nested `.gate-trusted` path (which, as you noted, isn't a boundary: a delivery/PR ref could carry a symlink there that `actions/checkout` would follow).
- `claude.yml` / `deliver-implement.yml`: their `check-permissions` job does no other checkout, so it checks out `default_branch` at root and runs the gate from it.
- `deliver-verify.yml`: its single workspace checkout is now pinned to `default_branch`, so `scripts/`, `.archon-version`, `deliver-gate.sh` and the gate all come from trusted code for **every** trigger (this also closes the archon-from-untrusted concern and makes the job header true).
- `deliver-correct.yml`: checks out `default_branch` at root → gate → then the delivery branch it edits. Ordering pinned by `TestCorrectChecksOutDeliveryBranchOnlyAfterGate`.

**P0 — PR author was unguarded (verify:435).**
Correct — `Validate the target PR` proves same-repo + `deliver/issue-N` but not bot authorship. Both phases now gate **both** the PR author and the sub-issue author. The canonical PR login is resolved via REST `.user.login` in `Validate` (a GET; `gh pr view --json author` drops the `[bot]` suffix and would break the allowlist) and passed to the gate as `PR_AUTHOR` — which also keeps the `gh api …/pulls` read out of the gate step, so it doesn't trip the repo-wide no-PR-creation guard beside its label POST. Pinned by `TestVerifyAndCorrectGateBothAuthors`.

**P1 — claude silent refusal on comment-post failure (claude:203).**
claude.yml now fails loudly (`::error` + `exit 1`) if the refusal comment can't post, instead of exiting green with no explanation. Pinned by `TestClaudeGateFailsLoudWhenRefusalCommentCannotPost`.

**P2 — docs.** `automated-delivery.md` (issue-author rule + author-gate Security paragraph; the "always from main" claim is now accurate after pinning, and the "tracked outside #1813" caveat is removed since the pin resolves it), `pr-workflow.md`, `scripts/README.md`, `agent-trust.md` all updated.

### Self-review + cross-vendor QA of the fix
Re-ran the two-vendor QA (GPT-5.4 + Gemini-3.1-Pro) on the rework — **no CRITICAL, no security defects**. Two notes I triaged:
- *"verify now runs later scripts from `default_branch`"* (GPT, IMPORTANT): not a regression — the normal `workflow_dispatch` path already checked out `default_branch` (`--ref github.ref_name`), and verify's tooling is meant to be trusted/main, not the branch under review. The pin only makes review/push events consistent with the path that already ran that way. qa-review reads the PR head from an ephemeral fetched worktree, not the root checkout.
- Fail-closed confirmed by both vendors: an empty/missing `PR_AUTHOR` (env, so empty not unset under `set -u`) or a failed `/issues` read → probe failure → refuse.

### Verification
`go build ./...`, full `go test ./scripts/ -count=1` (incl. the previously-tripped `TestDeliverImplementDoesNotCreatePRsWithTheWorkflowToken` — now green), `go vet`, `gofmt`, YAML parse ×4, and `bash -n` on every gate/validate block all pass.

### Verdict: **ready for re-review.**
All five inline findings and the summary are addressed with tests that pin each property (trusted-checkout ordering, correct-phase delivery-checkout-after-gate, both-author gating, claude loud-fail). The `/archon-pr-review` omission you flagged as acceptable is unchanged.

mtoslalibu:
## Final pre-merge review — pr-review-toolkit subagents + cross-vendor Q/A

Ran the full `pr-review-toolkit` (4 specialized subagents, in parallel) plus an independent cross-vendor Q/A (GPT-5.4 + Gemini-3.1-Pro) over the complete PR diff vs `main`.

### Verdict: **NO BLOCKERS.**

| Reviewer | Result |
|---|---|
| **code-reviewer** | No CRITICAL/IMPORTANT. Fail-closed exit-code split, bot-allowlist-before-probe, login regex validation, env-passed logins, bounded-`gh` watchdog, trusted-default-branch checkout before the untrusted ref — all sound. |
| **silent-failure-hunter** | "No blockers." Every refusal path fails closed **and** loud; no path ends green-with-no-signal. |
| **comment-analyzer** | One stale doc reference (fixed below). |
| **pr-test-analyzer** | "Adequate for merge." Tests are behavioral (survive a refactor); the order-sensitive checkout assertion is what actually prevents the P0 bypass. |
| **GPT-5.4** | One IMPORTANT (implement refusal loudness — addressed below); no bypass, ordering/both-author/shell all clean. |
| **Gemini-3.1-Pro** | "NO BLOCKERS" across all six checks (bypass, fail-open/closed, checkout ordering, both-author + canonical bot login, shell correctness, silent failure). |

### Three cheap items surfaced and fixed (commit `hardening(ci): toolkit + cross-vendor review nits`)
1. **Stale doc** — `automated-delivery.md` still referenced a "trusted `.gate-trusted` checkout" from the pre-rework design; corrected to the default-branch checkout at the workspace root.
2. **implement loudness parity** — `deliver-implement`'s refusal now `::error::`s the reason before the best-effort comment, so a comment-API outage still records *why* on the run (the job already failed loudly via `exit 1`; this matches `claude.yml`).
3. **test rigor** — the wiring test now asserts the two `check_author` calls target **distinct** authors (`"delivery PR"` and `"sub-issue"`), not a count a copy-paste could satisfy.

Non-blocking follow-up noted by pr-test-analyzer (not fixed, not required for merge): no test asserts the *workflow-level* routing of exit 3 (probe failure) into a distinct comment body — the script-level tri-state (0/1/3) is fully proven, so this is a test-strength gap, not a defect.

### State
All CI checks green (build, every test group, lint, claude-md-size); no merge conflict (merged `main`, resolved the `scripts/README.md` clash). The only remaining gate is this review's `CHANGES_REQUESTED` status — @namasl, whenever you have a moment to re-review. Thanks for the catches; the P0 checkout-trust and PR-author findings were exactly right.

.github/workflows/deliver-verify.yml
```diff
@@ -401,12 +407,83 @@ jobs:
...
+          # (blocked=true) ONLY when the SPECIFIC comment posts; if the comment API fails, leave
+          # `blocked` unset so that reporter fires as a fallback rather than the requester getting
+          # a red run with no explanation.
+          if gh pr comment "$PR" --repo "$REPO" --body "$body"; then
+            echo "blocked=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "::warning::could not post the author-gate refusal on #$PR; the generic failure reporter will report instead"
+          fi
+          gh api -X DELETE "/repos/$REPO/issues/$PR/labels/ready-for-merge" >/dev/null 2>&1 || true
+          gh api -X POST "/repos/$REPO/issues/$PR/labels" -f "labels[]=needs-human" >/dev/null 2>&1 || true
```
jgchn: Minor (non-blocking): a probe failure (exit 3) also strips `ready-for-merge` and adds `needs-human`, halting the loop — but the message above says "retry should work," and claude/implement don't halt on the same condition. Fail-closed is fine; a one-line note that the halt is intentional would reconcile the two. (Same in deliver-correct.)

namasl (approved):
Re-reviewed through `c6e58c4`. All of my blocking concerns are addressed: the gate runs from a default-branch checkout at the workspace root before any untrusted checkout; verify/correct gate both the delivery PR author and sub-issue author; the Claude refusal path fails loudly; and the documentation/code comments now match the implemented trust boundary. The branch is mergeable, all reported CI checks pass, and focused local validation (`git diff --check`, shell syntax, `go test ./scripts/ -count=1`, `go vet ./scripts/...`, and gofmt) passes. I found no unnecessary unrelated additions or remaining security blocker.

jgchn (approved):

Issue #1824, opened 2026-09-23
Shrink CLAUDE.md to operating-context minimum and stop per-PR changelog growth


namasl:
**What antipattern or issue does this address?**

`CLAUDE.md` has grown to **161,357 chars / 795 lines** — it now exceeds the 150k harness limit, so it is both truncation-prone and expensive to load on *every* session. The root cause is a single process antipattern: **CLAUDE.md is being used as a per-PR changelog** rather than as durable agent operating context. This is one root cause with one remediation (shrink + stop the mandate that regrows it); the tasks below are facets of that fix, not independent problems — hence one issue, not a tracker.

Measured breakdown (reproduce with `wc -c CLAUDE.md` + section math on current `main`):

| Section | Chars | % of file |
|---|---|---|
| `### Latency Estimation` | 69,743 | **43%** |
| `## Build and Run Commands` | 29,793 | 18% |
| `### Current Implementation Focus` | 23,655 | 15% |
| `## Post-Hoc Saturation Detection` | 12,555 | 8% |
| `### Key Invariants to Maintain` | 10,754 | 7% |

Those five sections are **91% of the file**, and they are the changelog-shaped ones. Concentration is extreme: **one paragraph (the #1530/#1694 inter-node-network writeup, line ~720) is 12,901 chars — 8% of the entire file by itself.** There are **261 issue/PR `#refs` across 103 distinct PRs**; nearly every merged PR appended a dense implementation-narrative paragraph ("why the three land as one PR", "behavior change on the record", "the trap avoided", byte-identity golden evidence, INV citations). That is commit-message / design-doc material — written once to justify a change, never needed again by an agent doing daily work.

**Why this is a standards violation, not a preference:** `docs/contributing/standards/principles.md` (lines 92-102) defines a **canonical-source registry** — every durable fact has exactly one home (`invariants.md`, `rules.md`, `principles.md`, `project-structure.md`, …) and CLAUDE.md is listed as holding only a **"summary" or "pointer"**. CLAUDE.md today does the opposite: it inlines large partial copies of docs that total 121 KB on disk (`invariants.md` 88 KB, `rules.md` 25 KB, `principles.md` 8 KB) — copies that carry the disclaimer *"If this section diverges, X is authoritative"*, i.e. they concede they are stale-prone duplicates.

**Root cause of the growth (verified on HEAD — must be fixed or it recurs):** the process actively instructs it.
[...]

Issue #1801, opened 2026-09-21
hardening(ci): verify does not read inline PR review comments, so inline reviewer findings are dropped


mtoslalibu:
## What issue does this address?

The delivery loop's **verify** phase does not reliably read **inline (line-level) PR review comments**, so a reviewer's inline point may never reach the reviewer/verdict — and therefore never becomes a finding a correction round can fix.

## Root cause

Inline review comments live at a different API than conversation comments:

- conversation comments: `GET /repos/{owner}/{repo}/issues/{n}/comments` — what `gh pr view --comments` returns (verify reads these)
- inline review comments: `GET /repos/{owner}/{repo}/pulls/{n}/comments` — **not** returned by `gh pr view --comments`

The verify prompt (`.github/workflows/deliver-verify.yml`) never fetches the second endpoint, so unless the agent happens to call it, inline comments are invisible to the review/verdict step. Conversation-level reviews are seen; inline ones are missed.

## Effect

A collaborator who leaves a precise inline fix-request on a delivery PR can have it silently dropped: verify never sees it → the bot reviewer never re-raises it as a NOT-GREEN finding → the correction round (whose work list is only the bot findings) never fixes it. Observed on PR #1680, where two valid inline comments from a reviewer were not incorporated.

## Suggested fix (keep it small)

In the verify phase, fetch inline review comments and include them in the review agent's inputs, e.g.:

```sh
gh api "/repos/$REPO/pulls/$PR/comments?per_page=100" --paginate
```

Feed the result into the prompt alongside the diff and conversation comments. This closes the "never seen" gap. No change to the trust model is required — inline comments remain **evidence to assess on the merits**, exactly like conversation comments today (a human comment still can't command a verdict; the SECURITY note already covers that).

## Scope / non-goals

- Verify phase only. The correction phase's work list stays the two bot findings (blis-pr-review NOT-GREEN + qa-review BLOCK) — unchanged.
- No new trust for human comments beyond what conversation comments already get.
- Permission-gating inline comments by reviewer write-access is a possible later refinement, not part of this fix.

## Acceptance criteria

- [ ] The verify phase fetches `pulls/{n}/comments` and passes them to the review agent.
- [ ] An inline review comment on a delivery PR is visible to the review/verdict reasoning (a valid inline point can produce a NOT-GREEN verdict just as a conversation point can).
- [ ] No change to the correction phase's findings source.

## Invariants at risk

None (CI-tooling change; no simulation code touched).

PR #1822, opened 2026-09-22
docs(claude): trim CLAUDE.md from 814 to 361 lines

mtoslalibu:
## Summary

`CLAUDE.md` had grown to **814 lines / 176KB** and is loaded into context on every session and subagent. Most of the bulk was per-PR changelog narrative (107 lines carried a `#nnnn` reference) and content duplicated from the canonical standards docs. This trims it to **361 lines (55.7% reduction; PR refs 107 → 4)** with **zero loss of durable guidance**.

Governing rule applied throughout: **keep the current rule, delete the PR story, never delete a fact that lives only in CLAUDE.md — relocate it first.** No PR narrative was bulk-copied into the guides (that would just move the bloat); PR-history prose is deleted (git holds it), and the one piece of reference found to live only in CLAUDE.md was relocated.

## Changes

- **Deleted** "Current Implementation Focus" (pure status/history).
- **Compressed** Build/Run Commands (456 → 119 lines; all command families kept with one-line comments), Latency Estimation, and Post-Hoc Saturation Detection to durable summaries + guide pointers.
- **Summarized** Key Invariants (one line per INV-1…19 + a families pointer) and BDD/TDD, matching the source-of-truth map's `(summary)` designation for those working copies — so **no map edit was needed**.
- **Relocated** the saturation flag migration table to its canonical home, `docs/contributing/saturation-analyzer-extension.md` (the only reference found to live solely in CLAUDE.md).
- **Removed stale refs**: `$PWD/model_configs`, `--catalog model_configs`, `catalog/models/`, present-tense `model_configs/` prose (that tree was deleted in #1771; the top-of-file `blis-catalog` clone instruction is retained).

Every trimmed fact has a verified home: latency backends/quant/KV-dtype/EP/inter-node → `docs/guide/latency-models.md`; MLA/hybrid → `docs/reference/models.md`; saturation flags + migration table → `docs/contributing/saturation-analyzer-extension.md`.

## Verification

- **KEEP sections byte-identical vs `main`** (Project Overview, Testing, Design Principles, PR Workflow, Code Review Standards, Antipattern Prevention, Extension Recipes, Code Style, CI/CD, Agent Behavioral Instructions, File Organization, Key Data Flow, Governance, Active Technologies, Change History) — verified programmatically.
[...]

PR #1825, opened 2026-09-23
docs(claude): shrink CLAUDE.md to operating-context minimum + guard regrowth

### LINKED ISSUE #732: docs: stale Recent work paragraph in CLAUDE.md
The 'Recent work' line in CLAUDE.md (around line 148) lists #450, #435, #458, #460 but significant work has been completed since (observe/replay/calibrate pipeline, tiered cache, client behavior model, etc.).

### LINKED ISSUE #1818: Trim CLAUDE.md: it is 814 lines / 176KB of accreted PR changelog loaded into every session

`CLAUDE.md` has grown to **814 lines / 176KB** and is loaded into context on every Claude Code session (and every subagent). The bulk is not durable guidance — it is per-PR changelog narrative that reads like squashed commit messages and duplicates content the file itself declares canonical elsewhere.

Quantified: **107 of 814 lines carry a `#nnnn` PR reference.** Several single "blocks" are 30–50 lines of wrapped prose describing one PR (e.g. the `#1548` EP block, the `#1530` inter-node network block, the catalog `#1768`/`#1769`/`#1770`/`#1771`/`#1774`/`#1776` series). The file even states at the bottom that `git log --oneline main` is "the definitive commit history" — which contradicts the changelog bloat above it.

This is a pure documentation/hardening cleanup. No code changes, no invariant risk.

- [ ] Request conservation
- [ ] Request lifecycle
- [ ] KV block conservation
- [ ] Causality
- [ ] Clock monotonicity
- [ ] Determinism
- [ ] Signal freshness
- [ ] Work-conserving
- [x] None (pure refactoring)

**Proposed fix**

Three categories of trim, each independently safe:

**1. Replace self-declared-authoritative sections with one-line pointers.** Three sections explicitly say *"If this section diverges, X.md is authoritative"*:
- **Key Invariants to Maintain** (~32 lines) restates INV-1…INV-19 + NS-6 verbatim from `docs/contributing/standards/invariants.md`.
- **BDD/TDD Development** (~13 lines) → `docs/contributing/standards/principles.md`.
- **Engineering Principles** (~18 lines) → `docs/contributing/standards/principles.md`.

Each should shrink to its pointer line. The registry/rules docs are the canonical source; keeping a second copy in CLAUDE.md guarantees drift.

**2. Delete / compress the per-PR narrative blocks (the main bloat).** These are commit messages `git log` already holds:
- **Current Implementation Focus** (~26 lines) — "PR17/PR18 completed", `#883`, `#1469`… pure history.
- The **Latency Estimation** region and the wall of `**Feature (#nnnn)**:` blocks (`#1530`, `#1522`, `#1656`, `#1548`, `#1527`, `#1591`, `#1699`…) — keep the durable 2-mode summary (trained-physics default vs roofline) and move detail to `docs/guide/latency-models.md`.
- The interleaved **catalog series** (`#1768`/`#1769`/`#1770`/`#1771`/`#1774`/`#1776`/`#1732`/`#1729`/`#1728`) — each is a verbatim PR description ("R1 cleanup of tracker #1767", test names, "Behavior change on the record"). Collapse each to a single durable sentence stating the *current* rule; the rationale lives in the PR.
- **Post-Hoc Saturation Detection** (~54 lines) — condense to a pointer + the migration table.

**3. Trim the Build/Run command comments.** Build and Run Commands is ~455 lines because nearly every example carries a paragraph-length `#nnnn` rationale comment. Keep ~15 canonical command forms; move the rationale into `docs/guide/*`.

**Also fix stale content found during the audit** (low-effort, do it in the same pass):
- Historical blocks still show `export BLIS_CATALOG=\$PWD/model_configs` and `--catalog model_configs` as worked examples, but `model_configs/` was **deleted** by #1771 (the top-of-file header is already correct: clone `blis-catalog`). Test fixtures now live in `testdata/catalog/`.
- The `#1769`/`#1770` blocks describe the bundled `model_configs/` presets/`devices/storage.yaml` in present tense, then append "(#1771 removes that tree)".
- Cross-repo note says `catalog/models/`; settled layout (#1774) is `<root>/models/` with no `catalog/` level.

Expected outcome: **~60–70% reduction** (well under ~300 lines) with zero loss of actual guidance — rules, invariants, command forms, and agent-behavioral instructions all preserved via pointers to the canonical docs.

**Files affected**
- `CLAUDE.md` (the trim)
- Possibly `docs/guide/latency-models.md`, `docs/guide/*` for saturation and command rationale, if any detail worth keeping isn't already there (verify before deleting — most already is).

No code, no construction sites, no invariant surface.

**Design document**
None needed — mechanical doc cleanup against already-canonical standards docs.

namasl:
## Summary

Shrinks `CLAUDE.md` from **818 lines / ~178 KB to 190 lines / ~14.5 KB (~92%)** and adds guards so it cannot silently regrow. `CLAUDE.md` had become a per-PR changelog loaded into every session; five sections were 91% of it and most duplicated content the file itself declares canonical elsewhere.

Implements #1818 (the trim) **plus its authorized refinement** (the prevention/enforcement layer — folding in #1824, which is closed as a duplicate).

> **Source-document note (pr-workflow Step 1.5):** the prevention layer (mechanisms 2–5 below) comes from the `@namasl` refinement comment on #1818, confirmed authoritative by `scripts/deliver-issue-refinements.sh 1818`. Recorded here as a `CORRECTION` so a reviewer reading only the #1818 body sees why the PR scope exceeds it.

## The trim (relocation targets verified already covered — no information loss)

- **Charter at the top:** CLAUDE.md is agent operating context, not a changelog.
- Collapse the four self-declared-authoritative sections (Key Invariants, BDD/TDD, Engineering Principles, Antipattern Prevention) to **pointers** to their canonical `docs/contributing/standards/*` sources.
- Delete **Current Implementation Focus**, the stale **"Recent work"** line (#732), the **Latency Estimation** narrative (keep the 2-mode summary + pointer), the **Post-Hoc Saturation** deep-dive (keep the flag list + pointer), and the migration tables.
- One canonical command per verb (run/replay/observe/calibrate/convert); condense the catalog/deployment rules to the mandatory operating facts.

Coverage was grep-verified before deleting: inter-node/MoE/EP → `docs/guide/latency-models.md`; MLA/hybrid/quant → `docs/reference/models.md`; spec-decode → `latency-models.md`; KV-offload → `docs/guide/kv-offload-calibration.md`; saturation → `docs/contributing/saturation-analyzer-extension.md`; scorers → `docs/concepts/architecture.md`.

## Prevent regrowth (the reason a one-time trim would otherwise re-accrete)

The growth was process-mandated, so the mandate is removed and backstops added:

1. **Charter** (above).
2. **Kill the mandate:** `pr-workflow.md` Tips #7 + checklist item 10, `extension-recipes.md`, and `design-guidelines.md` no longer instruct "update CLAUDE.md" every PR.
3. **Implement-phase prevention:** the delivery implement prompt forbids editing `CLAUDE.md` unless the issue carries the `docs:claude-md` label (pinned by `TestDeliverImplementPromptContract`).
4. **Verify-phase backstop:** `deliver-verify.yml` withholds `ready-for-merge` from an unauthorized `CLAUDE.md` edit — a direct mirror of the existing `ci_edit` guard (a **post-gate override**, so `deliver-gate.sh` is unchanged). Fails closed on an unreadable file list or labels. Pinned by the new `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits`.
5. **CI size ceiling:** a new `claude-md-size` job (new check name, not a rename) fails over 40 KB — a path-independent backstop that catches regrowth from any route, including the guard's blind spots.

## Behavioral contracts / verification

- `go build ./...` — clean.
- `go test ./scripts/` — pass (includes the new/updated guard assertions).
- `cmd` doc-example tests — pass (`TestDocExamplesPassDeploymentFlags` ≥50-example non-vacuity floor still met; catalog env-var, version-pin, and compatibility-link tests pass).
- `CLAUDE.md` = 14,493 bytes (well under the 40 KB ceiling).
- **No `sim/` or `cmd/` code changed**, so INV-6 (byte-identical stdout) and INV-13 (run/replay parity) are untouched by construction.

## Expected outcome under automated delivery

This PR edits `.github/workflows/**`, so the existing `ci_edit` guard routes it to **`needs-human`** by design — a human merges. That is correct, not a failure. The new `docs:claude-md` label should be created (`gh label create docs:claude-md`) and applied to #1818 for the CLAUDE.md guard to recognize a legitimately-targeting delivery; until then the guard fails safe.

## Discovered issues

None new. (A pre-existing gofmt nit in the #1801 test at `scripts/deliver_guards_test.go:736` is left untouched — it is out of scope and already tolerated by the repo's lint config.)

Closes #1818, closes #732.


namasl:

namasl:
@claude /blis-pr-review the previous run errored out mid-review (transient). Please retry. In addition to your usual checks, also consider the /archon-pr-review output above (verdict NO_CHANGE).

namasl:

## qa-review — PR #1825: ⛔ BLOCK
_docs(claude): shrink CLAUDE.md to operating-context minimum + guard regrowth · 13 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The core issue is implemented: CLAUDE.md is reduced to 190 lines, historical PR narratives are removed, canonical standards are referenced instead of copied, stale catalog examples are gone, and the remaining workflow additions directly prevent recurrence rather than adding unrelated product behavior. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. The PR removes the named CLAUDE.md sections but leaves design-guidelines.md directing readers to them. The policy-template recipe still says its steps are in CLAUDE.md under "Adding New Policy Templates", while the trimmed file instead points to extension-recipes.md. It also references a CLAUDE.md "Engineering Principles" section that no longer exists as a heading. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale code comments were found in the changed workflow logic. The comments accurately describe the independent size job, the delivery-only authorization gate, fail-closed API handling, and the ready-to-needs-human override. |
| G1 | correctness | ✅ CONFIDENT | In `.github/workflows/deliver-verify.yml`, `gh api ... --jq '[.[] \| select(.filename == "CLAUDE.md") \| .filename] \| join(", ")'` joins filenames with commas. If `CLAUDE.md` is one of multiple modified files, `touched` becomes `"CLAUDE.md"`. However, what happens if the API pagination fails or returns non-zero when fetching PR files, and does the `2>/dev/null \|\| echo "unreadable"` fallback cleanly handle all error states without masking unexpected API output? | For ordinary non-zero `gh api` failures, the `\|\| echo "unreadable"` branch runs and the guard fails closed. Because stderr is discarded, a clean failure produces exactly `unreadable`, which is recognized and emitted as an unauthorized edit. If `gh` were to emit partial or unexpected stdout before failing, command substitution would concatenate that output with `unreadable`, so the exact equality check would miss the dedicated unreadable branch. It still does not create an unauthorized bypass: a nonempty value proceeds to the label check, and an absent or unreadable label remains unauthorized. An authorized issue may proceed, which is consistent with the authorization rule. |
| G2 | correctness | ✅ CONFIDENT | In `.github/workflows/ci.yml`, `size=$(wc -c < CLAUDE.md)` counts raw bytes against a `ceiling=40000`. If `CLAUDE.md` is checked out or edited on environments with CRLF line endings versus LF, how much byte variation does line-ending conversion introduce, and could CRLF checkout cause the file to breach the ceiling prematurely? | CRLF adds one byte per line relative to LF. With the trimmed file at 190 lines, conversion would add approximately 190 bytes, far below the roughly 25KB of headroom between the stated ~14.5KB file and the 40,000-byte ceiling. GitHub's Ubuntu checkout also normally uses LF absent an attribute forcing CRLF, so CI will not prematurely breach the ceiling. |
| G3 | tests | ✅ CONFIDENT | In `scripts/deliver_guards_test.go`, `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits` verifies workflow wiring using string matching on step YAML text. Would this test catch a runtime shell execution error inside the inline script (such as `grep -qx` failing under `set -e` when no match is found)? | No. The test parses YAML and checks that strings and wiring are present; it does not execute the inline shell, so it would not detect general shell runtime failures. The specific `grep -qx` example is safe here, however: the script does not enable `errexit`, and the command is used as an `if` condition, where a no-match status is expected control flow. |
| G4 | tests | ❌ FLAW_FOUND | Is there a test verifying the behavior of the newly added `claude-md-size` job in `.github/workflows/ci.yml`, specifically asserting that a commit exceeding the 40,000-byte threshold correctly fails CI with the expected error message? | No. There is no test that executes the new size-check script against an over-40,000-byte CLAUDE.md and asserts failure or the emitted GitHub error. The existing new guard test covers only the separate deliver-verify authorization wiring. |
| G5 | rationale | ✅ CONFIDENT | What was the rationale for setting the `claude-md-size` CI ceiling to 40,000 bytes when the trimmed `CLAUDE.md` is ~14.5KB (~190 lines), and what alternatives (such as line count limits) were considered? | The documented rationale is to leave substantial room for legitimate durable operating rules while rejecting only a return to changelog-scale bloat. A byte limit directly measures the context payload that motivated #1818, whereas a line limit is easy to evade accidentally with long wrapped paragraphs and correlates poorly with loaded context size. The repository does not record a formal alternatives analysis showing that line count or token count was explicitly considered. |
| G6 | rationale | ✅ CONFIDENT | Why was the `docs:claude-md` issue label selected as the authorization mechanism for editing `CLAUDE.md` in `deliver-verify.yml` rather than relying on git commit trailers or PR title/body conventions? | An issue label is structured repository metadata attached to the delivery's declared source issue, can be queried exactly, and is already available before implementation begins so the implement prompt can enforce the same rule. Commit trailers and PR title/body conventions are free-form, can be omitted or rewritten by the implementing agent, and would require parsing conventions rather than an explicit maintainer-controlled opt-in. The code and comments establish this design, although no separate comparative decision record is present. |
| G7 | scope | ✅ CONFIDENT | Linked issue #1818 requests trimming `CLAUDE.md` to eliminate context bloat. Does creating new CI jobs (`claude-md-size`), adding verify gates (`claude_md_edit`), and extending `scripts/deliver_guards_test.go` stay within the scope of issue #1818? | Yes. These additions are hardening directly aimed at preventing the same CLAUDE.md accretion through both ordinary CI and the automated delivery path. The size job covers all branches and contributors, while the delivery gate prevents an automated issue unrelated to CLAUDE.md from modifying it. The associated test protects that new gate's wiring. |
| G8 | scope | ✅ CONFIDENT | This PR updates contributing documentation across `docs/contributing/pr-workflow.md`, `extension-recipes.md`, and `design-guidelines.md`. Are all of these changes directly required to prevent accretion, or do any introduce modified guidance beyond what was requested in #1818? | The contributing-document changes are directly related to preventing re-accretion. pr-workflow.md tells reviews, self-audits, and implementers where documentation belongs; extension-recipes.md redirects feature documentation away from CLAUDE.md; and design-guidelines.md redefines CLAUDE.md as operating context with canonical pointers. They do not introduce unrelated simulator guidance. |
| G9 | errors | ✅ CONFIDENT | In `deliver-verify.yml`, if `gh issue view` fails due to rate-limiting or network error, the step logs a warning and sets `unauthorized=CLAUDE.md` (failing closed to `needs-human`). Is this fail-closed behavior intended to halt the delivery pipeline on transient GitHub API failures? | Yes. The fail-closed behavior is explicitly intended. A transient label-read failure prevents only a green round from being declared `ready-for-merge`; it converts that outcome to `needs-human` rather than silently accepting an unverified authorization. Non-ready correction decisions are unaffected. |
| G10 | errors | ❌ FLAW_FOUND | In `.github/workflows/ci.yml`, `size=$(wc -c < CLAUDE.md)` runs under `set -uo pipefail`. If `CLAUDE.md` is accidentally deleted or moved in a branch, what error output is produced, and does the workflow fail with a clear diagnostic message? | Deletion is not handled correctly. Bash reports a redirection error such as `CLAUDE.md: No such file or directory`, but because the script enables `-u` and `pipefail` without `-e`, it continues with an empty `size`. The numeric comparison then reports an integer-expression error and is treated as a false `if` condition; the script can reach the end successfully. Thus a branch deleting CLAUDE.md may pass the new size job instead of failing with the intended clear diagnostic. |

### Items to fix
- **F2 · FLAW_FOUND** — No. The PR removes the named CLAUDE.md sections but leaves design-guidelines.md directing readers to them. The policy-template recipe still says its steps are in CLAUDE.md under "Adding New Policy Templates", while the trimmed file instead points to extension-recipes.md. It also references a CLAUDE.md "Engineering Principles" section that no longer exists as a heading.
  _docs/contributing/templates/design-guidelines.md:345 references the removed "Adding New Policy Templates" section; docs/contributing/templates/design-guidelines.md:361 references the removed "Engineering Principles" section. The actual replacement pointer is CLAUDE.md:99-101, and the policy recipe now lives at docs/contributing/extension-recipes.md:5-31._
- **G4 · FLAW_FOUND** — No. There is no test that executes the new size-check script against an over-40,000-byte CLAUDE.md and asserts failure or the emitted GitHub error. The existing new guard test covers only the separate deliver-verify authorization wiring.
  _.github/workflows/ci.yml:97-116 contains the new behavior. scripts/deliver_guards_test.go:1101-1182 tests only `claude_md_edit` wiring. Repository searches for `40000`, `claude-md-size`, and `wc -c` under scripts/ find no test. `go test ./scripts -count=1` passes but does not exercise the size job._
- **G10 · FLAW_FOUND** — Deletion is not handled correctly. Bash reports a redirection error such as `CLAUDE.md: No such file or directory`, but because the script enables `-u` and `pipefail` without `-e`, it continues with an empty `size`. The numeric comparison then reports an integer-expression error and is treated as a false `if` condition; the script can reach the end successfully. Thus a branch deleting CLAUDE.md may pass the new size job instead of failing with the intended clear diagnostic.
  _.github/workflows/ci.yml:105 uses `set -uo pipefail` without `-e`; line 107 performs the failing input redirection; lines 108-115 assume `size` is numeric and only explicitly exit when it exceeds the ceiling. Reproducible shell scenario: `set -uo pipefail; size=$(wc -c < missing); echo "$size"; if [ "$size" -gt 40000 ]; then exit 1; fi` emits file-not-found and integer-expression diagnostics but finishes with status 0._

### Important to consider
- **F1** — There is a stale cross-reference in design-guidelines.md, detailed under F2, but it does not undermine the main trim.
  _CLAUDE.md:5-10 defines the small-context charter; CLAUDE.md:18-29 uses the settled catalog-root/models layout; CLAUDE.md:83-101 replaces standards and subsystem detail with pointers; CLAUDE.md:188-190 delegates history to git log. .github/workflows/ci.yml:91-116 and .github/workflows/deliver-verify.yml:684-718 are recurrence guards explicitly tied to #1818._
- **F3** — The stale references found are documentation prose rather than code comments; see F2.
  _.github/workflows/ci.yml:91-96; .github/workflows/deliver-verify.yml:684-689; .github/workflows/deliver-verify.yml:1443-1456; .github/workflows/deliver-implement.yml:601-607._
- **G1** — Using an explicit `if ! touched=$(...); then ...` would avoid concatenating fallback text with any partial stdout and preserve the intended warning path.
  _.github/workflows/deliver-verify.yml:695-700 handles a clean file-list failure; lines 702-718 treat every nonempty result as an edit and authorize it only through an exact `docs:claude-md` label._
- **G5** — A token-count limit would track model context more directly but would require extra tooling; bytes are a simple deterministic proxy.
  _.github/workflows/ci.yml:91-96 states that 40,000 bytes is generous headroom and should fire only on changelog-scale growth; CLAUDE.md:5-10 states that loaded context size is the protected resource._
- **G7** — The linked issue described the change as documentation-only, so these workflow changes expand the listed file surface, but they remain directly connected hardening rather than unrelated functionality.
  _.github/workflows/ci.yml:91-96 explains why a path-independent job is needed; .github/workflows/deliver-verify.yml:684-689 ties the gate to the #1818 failure mode; scripts/deliver_guards_test.go:1101-1109 explains the regression being prevented._
- **G8** — design-guidelines.md still contains two stale references to removed CLAUDE.md sections at lines 345 and 361; see F2.
  _docs/contributing/pr-workflow.md:334-336, 438-444, and 538 add placement/review guidance; docs/contributing/extension-recipes.md:20-25 redirects policy documentation; docs/contributing/templates/design-guidelines.md:25-47 changes the document relationship to pointers rather than per-PR detail._
- **G9** — This favors trust safety over automatic retry and can require human intervention during a transient GitHub outage, as designed.
  _.github/workflows/deliver-verify.yml:684-689 explicitly says unreadable labels count as unauthorized; lines 715-718 implement that behavior; lines 1449-1456 apply it only when the current decision is `ready`._

namasl:
@claude /blis-pr-review re-review after the qa-review fixes (b79449a6): extracted the size check to scripts/claude-md-size.sh (fails closed on missing file), added scripts/claude_md_size_test.go, fixed stale design-guidelines.md pointers. In addition to your usual checks, also consider the /archon-pr-review output above once available.

namasl:

## qa-review adjudication — PR #1825: ✅ PASS

|----|----|----|----|
| F2 | FLAW_FOUND | ✅ RESOLVED | The stale CLAUDE.md references were replaced with pointers to the current canonical documents. |
| G4 | FLAW_FOUND | ✅ RESOLVED | The size check is now an executable script exercised by a test using an over-40,000-byte file. The test confirms a nonzero exit and checks the emitted breach and charter diagnostics. |
| G10 | FLAW_FOUND | ✅ RESOLVED | Deletion now fails closed before byte counting, with a clear GitHub error and nonzero exit. The regression test directly exercises a missing path. |

### Still blocking

### Cleared
- **F2** (✅ RESOLVED) — The stale CLAUDE.md references were replaced with pointers to the current canonical documents.
  _docs/contributing/templates/design-guidelines.md:345 now points the policy-template recipe to docs/contributing/extension-recipes.md; line 361 points error-handling guidance to docs/contributing/standards/principles.md. A repository search found no remaining CLAUDE.md references to the removed "Adding New Policy Templates" or "Engineering Principles" sections._
- **G4** (✅ RESOLVED) — The size check is now an executable script exercised by a test using an over-40,000-byte file. The test confirms a nonzero exit and checks the emitted breach and charter diagnostics.
  _.github/workflows/ci.yml:103-104 invokes scripts/claude-md-size.sh with a 40000-byte ceiling. scripts/claude_md_size_test.go:50-60 creates a 41000-byte file, requires a nonzero exit, and verifies the output names the 40000-byte breach and charter. scripts/claude-md-size.sh:31-37 emits the GitHub error and exits 1. `go test ./scripts` passes._
- **G10** (✅ RESOLVED) — Deletion now fails closed before byte counting, with a clear GitHub error and nonzero exit. The regression test directly exercises a missing path.
  _scripts/claude-md-size.sh:17-20 checks that the target is a regular file, emits `::error::` stating it is missing, and exits 1. scripts/claude_md_size_test.go:63-71 invokes the script on a nonexistent file and asserts both a nonzero exit and a missing-file diagnostic. `go test ./scripts` passes._

namasl:
Captured #1822 (@mtoslalibu's duplicate trim): its sole change not already here — relocating the pre-#1516 saturation flag-migration table to `docs/contributing/saturation-analyzer-extension.md` — is now in 9604de77 (CLAUDE.md's saturation pointer already named that guide for the migration). This PR therefore fully supersedes #1822 (same trim + the #1818 prevention layer).

namasl:
@claude /blis-pr-review re-review after 9604de77, which relocates the pre-#1516 saturation flag-migration table into docs/contributing/saturation-analyzer-extension.md (capturing the one unique change from the duplicate #1822). Docs-only. In addition to your usual checks, also consider the /archon-pr-review output above once available.

namasl:

## qa-review — PR #1825: ⛔ BLOCK
_docs(claude): shrink CLAUDE.md to operating-context minimum + guard regrowth · 19 questions · questioner `gcp/gemini-3.6-flash` · answerer `azure/gpt-5.6-sol`_

| F1 | fixed | ✅ CONFIDENT | Does this PR fully implement the issue it closes, without unnecessary additions? | Yes. The core issue is implemented: CLAUDE.md is reduced to 190 lines, retains canonical command forms and key operating rules, points to authoritative documentation, and removes the historical changelog blocks. The added size check and delivery guard are directly related hardening against the regrowth described by #1818 rather than unrelated functionality. |
| F2 | fixed | ❌ FLAW_FOUND | Has all documentation been updated to reflect the changes in this PR? | No. Existing contributor documentation still describes CLAUDE.md as the authoritative current-code and CLI reference, contradicting the new charter that makes it a compact pointer document. This PR changes that role but did not update those working copies. |
| F3 | fixed | ✅ CONFIDENT | Are there any stale code comments left by this change? | No stale code comments were found in the newly added shell/workflow logic. The comments accurately describe the size guard, fail-closed delivery decision, and the relationship between detection and the final decision step. |
| G1 | correctness | ❌ FLAW_FOUND | In `scripts/claude-md-size.sh`, `size=$(wc -c < "$file")` is evaluated with regex `[[ "size" =~ ^[0-9]+$ ]]`. On BSD/macOS systems where `wc -c` includes leading whitespace (e.g. ` 14500`), will this regex check fail erroneously on valid files? | Yes. Command substitution preserves leading spaces, so BSD `wc -c` output such as ` 14500` does not match `^[0-9]+$`. The script therefore reports a valid file as having an unreadable size on systems whose `wc` pads the count. Trimming whitespace or using a whitespace-tolerant pattern would avoid this. |
| G2 | correctness | ✅ CONFIDENT | In `deliver-verify.yml`, `gh api` uses `--paginate` with `--jq '[.[] \| select(.filename == "CLAUDE.md") \| .filename] \| join(", ")'`. When `--paginate` returns multiple concatenated JSON pages for PRs with over 100 files, `jq` without `--slurp` (`-s`) fails to parse multi-document stream output. Does this cause file inspection to silently fall back to `unreadable` for large PRs? | No. `gh api --paginate --jq` applies the jq expression to each response page; it does not concatenate raw JSON documents and then feed them to one non-slurping jq invocation. Each page therefore emits either an empty string or `CLAUDE.md`. A match on a later page remains non-empty and is treated as an edit. |
| G3 | invariants | ✅ CONFIDENT | By replacing verbatim descriptions of invariants INV-1 through INV-19 with pointers to `docs/contributing/standards/invariants.md`, how do we ensure subagents loaded with `CLAUDE.md` as prompt context still adhere to invariant INV-6 (determinism) and INV-1 (request conservation) during code generation? | The two invariants named in the question remain explicitly summarized in CLAUDE.md, not merely hidden behind a pointer. Agents see the twelve-term versus five-term conservation distinction and the byte-identical-stdout determinism rule in their initial context, while the canonical registry supplies full formulas, verification strategies, and edge cases. The PR workflow also explicitly requires a determinism self-audit. |
| G4 | invariants | ✅ CONFIDENT | Were all invariant standard definitions (INV-1..INV-19 and NS-6) verified to be 100% complete and up to date in `docs/contributing/standards/invariants.md` before deleting their descriptions from `CLAUDE.md`? | All INV-1 through INV-19 and NS-6 identifiers are present in the canonical registry, but the repository does not support a claim that every definition is 100% complete. The registry candidly records known gaps, including INV-2's thin statement and lack of enforcement and an untested cluster carve-out for INV-3. This does not invalidate the trim: the removed CLAUDE.md copies were not more authoritative or more complete. |
| G5 | rules | ✅ CONFIDENT | In `deliver-verify.yml`, if `gh issue view` fails to retrieve labels for issue `$ISSUE_NUMBER`, the workflow prints a warning but continues execution. Does this behavior risk violating strict failure safety rules by failing open under specific API permissions or missing issue contexts? | No. Label lookup failure is deliberately fail-closed. If `gh issue view` fails, execution reaches the common assignment `unauthorized=CLAUDE.md`; the final decision then downgrades an otherwise-ready delivery to `needs-human`. |
| G6 | rules | ✅ CONFIDENT | In `scripts/claude-md-size.sh`, the script accepts custom parameters `${1:-CLAUDE.md}` and `${2:-40000}`. If a path contains spaces or special characters, is the invocation in `ci.yml` properly quoted across all step executions? | Yes for every committed invocation. CI passes literal arguments without spaces, and the script quotes `$file` at every filesystem and redirection use. The Go tests use `exec.Command`, which passes paths as discrete arguments without shell splitting. |
| G7 | parity | ✅ CONFIDENT | The pre-#1516 saturation flag migration guidance was moved into `docs/contributing/saturation-analyzer-extension.md`. Does the documented flag mapping (`--detectors`, `--saturation-config`) maintain complete feature and metric parity across `blis run`, `blis replay`, and `blis observe` entry points? | Yes for the current supported post-hoc detector surface. The migration guide maps the retired selector and tuning flags to `--detectors` and strict-YAML `--saturation-config`; all three commands register the same flags and invoke the same resolver/tracer pipeline. Run/replay byte parity is covered through actual command execution, while observe intentionally feeds real-server metrics into that same downstream pipeline. |
| G8 | parity | ✅ CONFIDENT | `CLAUDE.md` was trimmed to include 'one command per verb'. Does this streamlined command reference preserve all default CLI flag options and config parameter representations used across standard development workflows? | No, CLAUDE.md intentionally does not preserve every CLI option or configuration representation. It preserves canonical command forms and names major feature families, then delegates the complete option surface to `--help`, topic guides, and references. That is the requested design, not a loss of the authoritative CLI contract. |
| G9 | tests | ✅ CONFIDENT | In `scripts/claude_md_size_test.go`, tests invoke `bash script file ceiling` directly. How will this test suite behave on environments without `bash` in the system path or on platforms with CRLF line ending conversions? | On an environment without `bash` in PATH, the Go test fails immediately through `t.Fatalf`; it does not silently skip coverage. CRLF-converted shell files can likewise fail under bash because carriage returns may become part of shell tokens. The shipped CI runs on Ubuntu, checks out the repository there, and has bash, so the actual CI path is reliable. |
| G10 | tests | ❌ FLAW_FOUND | `scripts/deliver_guards_test.go` checks for the presence of string literals (e.g. `docs:claude-md`, `CLAUDE_MD_UNAUTHORIZED`) in YAML files. Would this test still pass if the YAML indentation or workflow execution step order were altered in a way that broke GitHub Actions step output passing? | Yes, and the premise understates the gap: `scripts/deliver_guards_test.go` does not reference `docs:claude-md`, `claude_md_edit`, or `CLAUDE_MD_UNAUTHORIZED` at all. Thus it provides no structural coverage for this new gate. For example, moving the producer step after `Apply the decision` would leave the output empty and allow an unauthorized edit through, while the current scripts tests would still pass. |
| G11 | rationale | ✅ CONFIDENT | Why was 40,000 bytes chosen as the ceiling in `scripts/claude-md-size.sh` and `ci.yml`, and what context window budgeting model was used to select this threshold over a stricter limit? | The committed rationale is qualitative: 40,000 bytes leaves generous headroom above the newly trimmed file so ordinary durable operating-rule additions do not fail CI, while still detecting a return to changelog-scale bloat. No token-to-byte conversion or formal context-window budgeting model is documented, so a stricter numerical derivation cannot be substantiated from the repository. |
| G12 | rationale | ⚠️ CANNOT_ANSWER | What alternative designs were evaluated for preventing `CLAUDE.md` regrowth before introducing the custom GitHub issue label `docs:claude-md` and the dedicated `claude_md_edit` workflow step? | The repository shows the selected layered design—an in-file charter, implement-agent instruction, delivery-specific label gate, and path-independent CI size ceiling—but it does not record which rejected alternatives were evaluated before choosing the `docs:claude-md` label and verification step. Historical design deliberation cannot be verified from the worktree. |
| G13 | scope | ✅ CONFIDENT | The linked issue #1818 requests trimming `CLAUDE.md`. Does adding new workflow jobs (`claude-md-size`), new deliver-verify gates, and Go test scripts (`scripts/claude_md_size_test.go`) stay within the scope of this documentation maintenance task? | Yes. These additions enforce the exact maintenance property at issue: preventing CLAUDE.md from re-accumulating per-PR history. The global size job covers all contribution paths, while the delivery gate prevents automated issue delivery from editing the file without explicit authorization; the executable test covers the new size script's normal, oversized, and missing-file behavior. |
| G14 | scope | ✅ CONFIDENT | This PR modifies several auxiliary guide files like `docs/contributing/pr-workflow.md`, `extension-recipes.md`, and `templates/design-guidelines.md`. Are all edits in these files strictly minimal pointers required by the `CLAUDE.md` trim? | The auxiliary edits are small and directly support the new source-of-truth policy, though they are not all literal pointers. `extension-recipes.md` tells feature authors where documentation belongs; `pr-workflow.md` adds review and self-audit checks against CLAUDE.md bloat; and the design guidelines update CLAUDE.md's role in the document hierarchy. These are scoped guardrails rather than unrelated content. |
| G15 | errors | ✅ CONFIDENT | In `scripts/claude-md-size.sh`, if `wc -c` fails or prints to stderr, `$size` capture might receive empty or invalid output. Does the script cleanly error out with `exit 1` without executing arithmetic comparisons on unset variables? | Yes for the stated empty or invalid-output cases. Because the script does not use `set -e`, a failed command substitution proceeds to explicit validation; an empty or nonnumeric `$size` fails the regex and exits 1 before the arithmetic comparison. `wc` stderr remains visible because only stdout is captured. |
| G16 | errors | ✅ CONFIDENT | In `deliver-verify.yml`, if `gh api` fails due to network or authentication issues, `2>/dev/null \|\| echo "unreadable"` outputs `unreadable`. Is it guaranteed that `[ "$touched" == "unreadable" ]` will always trigger the warning path and set `unauthorized` rather than falling through? | A total `gh api` failure with no prior stdout produces exactly `unreadable`, triggering the warning and unauthorized output. It is not strictly guaranteed to take that exact warning branch if `gh` emits partial stdout before failing, because command substitution would contain both the partial output and `unreadable`. Even then the value is non-empty, so the workflow treats CLAUDE.md as touched and requires the authorization label; it does not fail open. |

### Items to fix
- **F2 · FLAW_FOUND** — No. Existing contributor documentation still describes CLAUDE.md as the authoritative current-code and CLI reference, contradicting the new charter that makes it a compact pointer document. This PR changes that role but did not update those working copies.
  _CLAUDE.md:5-10 says canonical facts belong in authoritative docs and CLAUDE.md carries pointers; README.md:317 still calls CLAUDE.md the authoritative file-level architecture documentation; CONTRIBUTING.md:292-304 directs feature extensions to update CLAUDE.md; CONTRIBUTING.md:344 calls it authoritative for code architecture, file organization, and CLI flags._
- **G1 · FLAW_FOUND** — Yes. Command substitution preserves leading spaces, so BSD `wc -c` output such as ` 14500` does not match `^[0-9]+$`. The script therefore reports a valid file as having an unreadable size on systems whose `wc` pads the count. Trimming whitespace or using a whitespace-tolerant pattern would avoid this.
  _scripts/claude-md-size.sh:22-27. Reproduction in bash: `size=' 14500'; [[ "$size" =~ ^[0-9]+$ ]]` returns false. BSD/macOS `wc -c` commonly pads numeric output._
- **G10 · FLAW_FOUND** — Yes, and the premise understates the gap: `scripts/deliver_guards_test.go` does not reference `docs:claude-md`, `claude_md_edit`, or `CLAUDE_MD_UNAUTHORIZED` at all. Thus it provides no structural coverage for this new gate. For example, moving the producer step after `Apply the decision` would leave the output empty and allow an unauthorized edit through, while the current scripts tests would still pass.
  _No occurrences of `docs:claude-md`, `claude_md_edit`, or `CLAUDE_MD_UNAUTHORIZED` exist in scripts/deliver_guards_test.go; producer is .github/workflows/deliver-verify.yml:690-718 and consumer is .github/workflows/deliver-verify.yml:1424-1457. `go test ./scripts -count=1` currently passes despite having no test of that producer/consumer ordering or output contract._
- **G12 · CANNOT_ANSWER** — The repository shows the selected layered design—an in-file charter, implement-agent instruction, delivery-specific label gate, and path-independent CI size ceiling—but it does not record which rejected alternatives were evaluated before choosing the `docs:claude-md` label and verification step. Historical design deliberation cannot be verified from the worktree.
  _CLAUDE.md:5-10; .github/workflows/deliver-implement.yml:601-607; .github/workflows/deliver-verify.yml:684-718; .github/workflows/ci.yml:91-104_

### Important to consider
- **F1** — There are narrower defects in the new guard portability and test coverage, reported under G1 and G10.
  _CLAUDE.md:5-10, CLAUDE.md:31-58, CLAUDE.md:83-101, CLAUDE.md:188-190; .github/workflows/ci.yml:91-104; .github/workflows/deliver-verify.yml:684-718; `go test ./scripts -count=1` passes_
- **F3** — There is stale documentation prose outside code comments, notably README.md and CONTRIBUTING.md as reported in F2.
  _scripts/claude-md-size.sh:2-11; .github/workflows/ci.yml:91-96; .github/workflows/deliver-verify.yml:684-689,1443-1456_
- **G2** — Multi-page output may contain blank-line separators, but that does not turn a real match into an empty value.
  _.github/workflows/deliver-verify.yml:695-706 uses gh CLI's built-in `--jq` processing together with `--paginate`; the following logic treats any non-empty result as a CLAUDE.md modification._
- **G3** — Adherence still depends on agents following repository pointers, but duplicating the complete registry in every prompt would reintroduce the drift this PR addresses.
  _CLAUDE.md:90-97; docs/contributing/standards/invariants.md:77-107 contains the full INV-1 definition and verification; docs/contributing/standards/invariants.md:174-186 contains INV-6; docs/contributing/pr-workflow.md:433-444 explicitly audits determinism and documentation source-of-truth consistency._
- **G4** — Completing the already-recorded INV-2 and INV-3 coverage gaps is pre-existing follow-up work, not a regression introduced by this documentation trim.
  _docs/contributing/standards/invariants.md:41-69 indexes INV-1 through INV-19 and NS-6; docs/contributing/standards/invariants.md:111-119 calls INV-2 the thinnest entry and records missing enforcement; docs/contributing/standards/invariants.md:134-140 records the untested INV-3 cluster carve-out; docs/contributing/standards/invariants.md:250-267 defines NS-6._
- **G5** — The step exits successfully so the workflow can produce a controlled needs-human decision rather than aborting without a useful delivery status.
  _.github/workflows/deliver-verify.yml:708-718; .github/workflows/deliver-verify.yml:1431,1449-1457_
- **G6** — An external shell caller supplying a custom path containing spaces must quote that argument, as with any shell command.
  _.github/workflows/ci.yml:103-104; scripts/claude-md-size.sh:14-22,30-32; scripts/claude_md_size_test.go:21-25_
- **G7** — Observe cannot be byte-identical to a simulated run because its input is real-server latency; parity here means the same flags, reducer, report shape, and metric pipeline.
  _docs/contributing/saturation-analyzer-extension.md:59-80,329-337; cmd/saturation.go:15-42,74-105; cmd/root.go:2864, cmd/replay.go:825, cmd/observe_cmd.go:612; cmd/saturation_cli_parity_test.go:12-24,84-118_
- **G8** — The stale external claims that CLAUDE.md itself remains the full CLI authority should be corrected as described in F2.
  _CLAUDE.md:31-33 explicitly states the one-canonical-form policy and points feature flags to `docs/guide/*` and `--help`; CLAUDE.md:35-58 retains build, run, replay, observe, calibrate, and convert forms; CLAUDE.md:146-152 retains latency and saturation summaries._
- **G9** — The test is intentionally bash-specific because the production script uses `[[ ... =~ ... ]]`; native Windows portability is not provided.
  _scripts/claude_md_size_test.go:21-31; .github/workflows/ci.yml:31-72 runs the scripts test shard on `ubuntu-latest`; .github/workflows/ci.yml:97-104 runs the production size job on `ubuntu-latest`._
- **G11** — A documented token-budget calculation could make future ceiling changes easier to review, but its absence is not a correctness defect.
  _.github/workflows/ci.yml:91-96; scripts/claude-md-size.sh:11,14-15; CLAUDE.md is currently 190 lines and materially below the ceiling._
- **G13** — The workflow changes make this more than a purely textual edit, but they remain tightly coupled to #1818's anti-regrowth objective.
  _.github/workflows/ci.yml:91-104; .github/workflows/deliver-implement.yml:601-607; .github/workflows/deliver-verify.yml:684-718,1449-1457; scripts/claude_md_size_test.go:11-72_
- **G14** — The broader documentation set still contains contradictory authority claims, reported separately in F2.
  _docs/contributing/extension-recipes.md:25; docs/contributing/pr-workflow.md:334-336,439-444,538; docs/contributing/templates/design-guidelines.md:25-47_
- **G15** — The explicit numeric validation is what makes this fail closed despite the absence of `set -e`.
  _scripts/claude-md-size.sh:12,22-31_
- **G16** — Matching `*unreadable*` rather than exact equality would make the diagnostic branch robust to partial output, but the current behavior remains safety-preserving.
  _.github/workflows/deliver-verify.yml:695-718; the final consumer blocks ready whenever `steps.claude_md_edit.outputs.unauthorized` is non-empty at .github/workflows/deliver-verify.yml:1431,1449-1457._

namasl:
Addressed the round-2 qa-review findings in 59460afa:

- **F2 (FLAW) — fixed.** README.md no longer calls CLAUDE.md the authoritative file-level architecture doc (repointed to `docs/reference/project-structure.md`); CONTRIBUTING.md’s extension steps and doc-index row now describe CLAUDE.md’s pointer/operating-context role instead of "authoritative for … CLI flags / file organization." Consistent with the charter.
- **G1 (FLAW) — fixed.** `scripts/claude-md-size.sh` now strips whitespace from the `wc -c` capture (`tr -d [:space:]`), so BSD/macOS space-padded output no longer false-fails the numeric check (CI is GNU/Linux, so this was latent).
- **G10 (FLAW) — fixed, with a correction.** The literal claim that `deliver_guards_test.go` "does not reference `claude_md_edit`/`CLAUDE_MD_UNAUTHORIZED`/`docs:claude-md` at all" is inaccurate — `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits` references all three. But the underlying gap was real: it asserted wiring, not producer→consumer ORDER. It now asserts `claude_md_edit` index < `Apply the decision` index, so moving the producer after the consumer fails the test.
- **G12 (CANNOT_ANSWER) — justification.** This is docs/CI hardening, not an architecture RFC, so no formal alternatives-analysis record was produced. The chosen layered design and why each layer exists are documented in-code (the `claude-md-size` job comment in `ci.yml`, the `claude_md_edit`/`Apply the decision` comments in `deliver-verify.yml`, the implement-prompt bullet) and in the PR body. The label mechanism was chosen because it is structured, queryable metadata available before implementation begins (vs. free-form commit trailers / PR text); the CI size job is the path-independent backstop the label gate cannot cover. Not a code defect — requesting a design-deliberation artifact is out of scope for this PR.

namasl:
@claude /blis-pr-review re-review after 59460afa (qa-review round-2 fixes: portable size check, order-assert the deliver-verify guard, README/CONTRIBUTING aligned with the charter). In addition to your usual checks, also consider the /archon-pr-review output above once available.

namasl:

## qa-review adjudication — PR #1825: ⛔ BLOCK
_re-checking 4 prior blocking finding(s) · adjudicator `azure/gpt-5.6-sol`_

|----|----|----|----|
| F2 | FLAW_FOUND | ✅ RESOLVED | The contributor-facing documentation now consistently describes CLAUDE.md as operating context and a pointer document rather than the authoritative architecture or CLI reference. |
| G1 | FLAW_FOUND | ✅ RESOLVED | The captured wc output is now stripped of whitespace before the numeric validation, so BSD/macOS padding no longer causes a false unreadable-size error. |
| G10 | FLAW_FOUND | ❌ STILL-OPEN | The claimed producer-to-consumer ordering test is not present in the current checked-out code. A repository search of scripts/deliver_guards_test.go finds no occurrence of `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits`, `claude_md_edit`, `CLAUDE_MD_UNAUTHORIZED`, or `docs:claude-md`, while the scripts test suite still passes. Add and commit a structural test that locates the producer and consumer steps and asserts the producer precedes `Apply the decision` while also pinning the output contract. |
| G12 | CANNOT_ANSWER | 🟡 WAIVED (justification) | The request for a rejected-alternatives artifact is not a requirement of the scoped work. The governing issue explicitly characterizes the change as mechanical documentation cleanup against existing canonical documents and states that no design document is needed. No repository requirement making an architecture RFC or alternatives analysis mandatory for this cleanup was found. |

### Still blocking
- **G10** — The claimed producer-to-consumer ordering test is not present in the current checked-out code. A repository search of scripts/deliver_guards_test.go finds no occurrence of `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits`, `claude_md_edit`, `CLAUDE_MD_UNAUTHORIZED`, or `docs:claude-md`, while the scripts test suite still passes. Add and commit a structural test that locates the producer and consumer steps and asserts the producer precedes `Apply the decision` while also pinning the output contract.
  _Current workflow producer remains at .github/workflows/deliver-verify.yml:690-718, and its output is consumed at .github/workflows/deliver-verify.yml:1425-1457. Regex searches of scripts/deliver_guards_test.go returned no matching guard test or contract identifiers. `go test ./scripts` exits successfully despite that absence._

### Cleared
- **F2** (✅ RESOLVED) — The contributor-facing documentation now consistently describes CLAUDE.md as operating context and a pointer document rather than the authoritative architecture or CLI reference.
  _README.md:315-317 points file-level architecture to docs/reference/project-structure.md. CONTRIBUTING.md:288-304 directs extension documentation to README and the relevant docs guide, explicitly excluding CLAUDE.md for policy templates. CONTRIBUTING.md:340-345 describes CLAUDE.md as agent operating context containing canonical commands and pointers, consistent with CLAUDE.md:5-10._
- **G1** (✅ RESOLVED) — The captured wc output is now stripped of whitespace before the numeric validation, so BSD/macOS padding no longer causes a false unreadable-size error.
  _scripts/claude-md-size.sh:22 pipes `wc -c` through `tr -d '[:space:]'`; scripts/claude-md-size.sh:23-29 documents the BSD padding case and validates the normalized value._
- **G12** (🟡 WAIVED (justification)) — The request for a rejected-alternatives artifact is not a requirement of the scoped work. The governing issue explicitly characterizes the change as mechanical documentation cleanup against existing canonical documents and states that no design document is needed. No repository requirement making an architecture RFC or alternatives analysis mandatory for this cleanup was found.
  _GitHub issue #1818 states under `Design document`: `None needed — mechanical doc cleanup against already-canonical standards docs`, and scopes the work to trimming CLAUDE.md, correcting stale documentation, and preventing renewed changelog bloat. The current implementation documents the selected mechanisms directly at .github/workflows/ci.yml:89-94 and .github/workflows/deliver-verify.yml:684-689._

namasl:
Rebuttal on the adjudication of **G10** (STILL_OPEN) — the claim "the ordering test is not present … no occurrence of `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits`, `claude_md_edit`, `CLAUDE_MD_UNAUTHORIZED`, or `docs:claude-md`" is **factually incorrect** for the current head `59460afa`. Verified with `git show 59460afa:scripts/deliver_guards_test.go`:

```
310:  needle: "docs:claude-md",
1110: func TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits(t *testing.T) {
1137:   guardRun, guardIdx := "", -1
1139:     if s.ID == "claude_md_edit" {
1158:   applyRun, applyIdx := "", -1
1162:       applyRun, applyEnv, applyIdx = s.Run, s.Env, i
1169:   if guardIdx > applyIdx {              // <-- producer-before-consumer ORDER assertion
1170:     t.Errorf("the `claude_md_edit` step is at index %d, AFTER `Apply the decision` ...
```

The test exists (line 1110), references all four strings, and asserts producer→consumer **order** at lines 1169-1172 (exactly G10’s requested contract). The file is **1191 lines** and the test sits past line 1100; both Sol passes that reported "no occurrence" appear to have scanned only the head of this large file (they read the smaller F2/G1 files correctly). Please re-check by reading `scripts/deliver_guards_test.go` at offset 1110. `go test ./scripts` exercises it and passes. No code change is warranted — G10 is resolved in `59460afa`.

namasl:
On G10: 0b35ab64 moves `TestDeliverVerifyGuardsUnauthorizedClaudeMdEdits` (with the producer→consumer order assertion) out of the 1191-line `deliver_guards_test.go` into a small dedicated file, **`scripts/claude_md_guards_test.go`** (~112 lines). The test was always present and passing (the prior STILL_OPEN was a read/grep miss on the large file, evidenced by `git show`), but this makes it trivially discoverable and groups it with the size-guard test. Re-adjudicating.

namasl:
@claude /blis-pr-review re-review after 0b35ab64 (moved the CLAUDE.md verify-guard test to scripts/claude_md_guards_test.go, verbatim; no behavior change). Also consider the /archon-pr-review output above once available.

namasl:

## qa-review adjudication — PR #1825: ✅ PASS
_re-checking 4 prior blocking finding(s) · adjudicator `azure/gpt-5.6-sol`_

|----|----|----|----|
| F2 | FLAW_FOUND | ✅ RESOLVED | The contributor-facing documentation now aligns with the CLAUDE.md charter: architecture is referred to the project-structure reference, while CLAUDE.md is described as operating context and a pointer document rather than the authoritative architecture or CLI reference. |
| G1 | FLAW_FOUND | ✅ RESOLVED | The size capture now removes whitespace before applying the numeric validation, so padded BSD/macOS wc output no longer false-fails. |
| G10 | FLAW_FOUND | ✅ RESOLVED | A dedicated structural test now verifies the producer-to-consumer ordering and the output contract. Moving the `claude_md_edit` producer after `Apply the decision`, detaching its output, or removing the consumer behavior would fail this test. |
| G12 | CANNOT_ANSWER | 🟡 WAIVED (justification) | A rejected-alternatives or architecture-RFC artifact is not required by the governing issue. The issue expressly classifies the work as mechanical documentation cleanup against already-canonical documents and states that no design document is needed; the selected hardening mechanisms themselves are documented in the implementation. |

### Still blocking

### Cleared
- **F2** (✅ RESOLVED) — The contributor-facing documentation now aligns with the CLAUDE.md charter: architecture is referred to the project-structure reference, while CLAUDE.md is described as operating context and a pointer document rather than the authoritative architecture or CLI reference.
  _README.md:315-317 points file-level architecture to docs/reference/project-structure.md. CONTRIBUTING.md:288-304 directs extension documentation to README and relevant docs guides, explicitly excluding CLAUDE.md at line 292. CONTRIBUTING.md:340-345 describes CLAUDE.md as agent operating context containing canonical commands and pointers. CLAUDE.md:5-10 defines that pointer-oriented charter. A repository-wide regex search found no remaining statement calling CLAUDE.md authoritative._
- **G1** (✅ RESOLVED) — The size capture now removes whitespace before applying the numeric validation, so padded BSD/macOS wc output no longer false-fails.
  _scripts/claude-md-size.sh:22 captures `wc -c` through `tr -d '[:space:]'`; lines 23-24 explicitly identify BSD/macOS leading padding, and lines 27-29 validate the normalized value._
- **G10** (✅ RESOLVED) — A dedicated structural test now verifies the producer-to-consumer ordering and the output contract. Moving the `claude_md_edit` producer after `Apply the decision`, detaching its output, or removing the consumer behavior would fail this test.
  _scripts/claude_md_guards_test.go:25-62 locates the `claude_md_edit` producer and checks its label/CLAUDE.md logic; lines 72-89 locate `Apply the decision` and assert the producer index precedes it; lines 90-104 verify `CLAUDE_MD_UNAUTHORIZED` is wired from `steps.claude_md_edit.outputs` and acted upon. The actual producer is at .github/workflows/deliver-verify.yml:690-718 and the consumer at lines 1424-1457. `go test ./scripts` passes at the current head._
- **G12** (🟡 WAIVED (justification)) — A rejected-alternatives or architecture-RFC artifact is not required by the governing issue. The issue expressly classifies the work as mechanical documentation cleanup against already-canonical documents and states that no design document is needed; the selected hardening mechanisms themselves are documented in the implementation.
  _GitHub issue #1818 is OPEN and states under `Design document`: `None needed — mechanical doc cleanup against already-canonical standards docs`; its stated objective is trimming CLAUDE.md and preventing renewed changelog bloat. The selected path-independent size backstop is explained at .github/workflows/ci.yml:91-96, and the delivery-specific label gate is explained at .github/workflows/deliver-verify.yml:684-689._

namasl:
Merged upstream/main (#1821, the strict catalog-load gate) to clear the conflict — 94575ef1. Only CLAUDE.md conflicted; resolved by keeping this PR's trimmed version, since #1821's CLAUDE.md additions were exactly the per-PR changelog prose this PR removes and its durable operator content lives in `docs/reference/models.md` (merged cleanly, along with its `cmd/catalog_load.go` + gate script + tests). `go build ./...`, `go test ./scripts/`, and the doc-example tests all pass on the merge.

namasl:
@claude /blis-pr-review re-review after the upstream/main merge (94575ef1); the merge only integrates #1821 and keeps this PR's trimmed CLAUDE.md (no substantive change to this PR's files). Also consider the /archon-pr-review output above once available.

namasl:
Addressed blis Q7 (non-blocking) in 046cfa3e: documented the `docs:claude-md` label in `pr-workflow.md` Tips #7 — a change that genuinely targets CLAUDE.md labels its issue `docs:claude-md`; the size job guards size on every PR and verify fails safe to needs-human otherwise. This was the only open item (blis: Ready to merge; qa-review: PASS; archon: NO_CHANGE; CI: green; mergeable, pending the branch-protection/human merge).

namasl:
@claude /blis-pr-review final re-confirm after 046cfa3e (documented the docs:claude-md label, your Q7). Also consider the /archon-pr-review output above once available.

mtoslalibu:
## Non-blocking: tighten a couple of pointers for the "run blis via Claude" path

Reviewed this against the three CLAUDE.md consumers (users running blis, developers contributing, experimenters doing benchmarks/policy work). **The trim is the right call and I'd merge as-is** — the charter + size ceiling + killing the "update CLAUDE.md every PR" mandate address the actual root cause, not just the symptom.

### What I checked

The main worry with collapsing the ~15 worked command examples to one-per-verb is: can an agent still *find* the exact flag syntax it dropped? I verified where everything landed:

- **Scorer names + default profile** — live in `docs/guide/routing.md` (both the full name list and `precise-prefix-cache:2,queue-depth:1,kv-utilization:1`).
- **Feature flags** — all resolve to a real guide: `--slo-ttft` → `admission.md`; `--num-speculative-tokens` → `latency-models.md`; `--kv-cache-dtype` → `observe-replay-calibrate.md`; `--kv-offload-config`/`--scrape-kv-metrics` → `kv-offload-calibration.md`; `--session-mode closed-loop` → `observe-replay-calibrate.md`.

**Conclusion: nothing is orphaned.** Every dropped detail is reachable through the pointers this PR keeps. So this is *not* a discoverability gap — it's a small **convenience/latency** cost: the pointers are **globs** (`docs/guide/*`), so an agent greps the directory instead of opening a named file (an extra tool call or two per lookup).

That reframes "should we add content back?" — the answer is no, don't re-copy syntax (that fights the charter). Instead, make two pointers **specific**. Same "pointer, not copy" principle, just less grep.

### Recommendation 1 (Option B — do this): name the scorers file

Routing policy is a core experiment axis and scorer names aren't guessable. In the *Notable subsystems* line, point routing scorers specifically at `docs/guide/routing.md` instead of the `docs/guide/*` glob:

> Routing scorers + the default profile (`precise-prefix-cache:2,queue-depth:1,kv-utilization:1`): `docs/guide/routing.md`.

**Cost: ~1 line / ~30 tokens.** Highest value-per-token; fully charter-compatible.

### Recommendation 2 (Option A — optional): a "feature flags by home" mini-map

If you want to save the grep for the run/replay/observe path too, replace the single `docs/guide/*` glob sentence with a short routing table:

- SLO goodput / flow-control / dispatch → `docs/guide/admission.md`
- spec-decode/MTP, MoE `--dp`/`--enable-expert-parallel`, `--kv-cache-dtype`, inter-node → `docs/guide/latency-models.md` (+ `docs/reference/models.md` for MLA/hybrid/quant)
- KV-offload (`--kv-offload-config`, `--scrape-kv-metrics`) → `docs/guide/kv-offload-calibration.md`
- closed-loop/corpus sessions, observe, calibrate → `docs/guide/observe-replay-calibrate.md`
- workloads & `convert` → `docs/guide/workloads.md`
- saturation detectors → `docs/contributing/saturation-analyzer-extension.md`

**Cost: 7 lines / ~170 tokens** (net ~+5 lines, since it replaces the existing glob sentence). Still pointers, no copied syntax — well under the 40 KB ceiling.

### TL;DR

Merge it. If you take one thing, take **Rec 1** (1 line) — it removes the only lookup that's both high-frequency and not guessable. Rec 2 is a nice-to-have for the user persona. Everything else the trim dropped is correctly a `--help`/guide read.

namasl:
Thanks @mtoslalibu — applied both in 8c4b2de7, keeping "pointer, not copy":

- **Rec 1:** the *Notable subsystems* line now points routing scorers + the default profile (`precise-prefix-cache:2,queue-depth:1,kv-utilization:1`) specifically at `docs/guide/routing.md`.
- **Rec 2:** replaced the `docs/guide/*` glob in *Build and Run Commands* with the feature-flags-by-home map (admission / latency-models + models / kv-offload-calibration / observe-replay-calibrate / workloads / saturation).

No syntax re-copied; CLAUDE.md is 15,002 bytes (well under the 40 KB ceiling); doc-example tests pass. Verified every cited guide exists before pointing at it.

namasl:
@claude /blis-pr-review re-confirm after 8c4b2de7 (tightened two pointers per maintainer review — routing scorers → routing.md, and a feature-flag home map; pointers only, no copied syntax). Also consider the /archon-pr-review output above once available.

mtoslalibu (approved):

